PR #4, opened 2022-12-05
Add SIMD-0003: Dynamic Base fees

aeyakovenko:
#### Problem

Network doesn't have a mechanism to target a healthy load, resulting in two potential performance degradation attack vectors.  At 100% heterogenous hardware doesn't have uniform performance resulting in general performance degradation with downstream effects on block producers.  Without a dynamic base fee, a faulty majority can stuff blocks with 100% load making verification for non validating nodes unnecessarily expensive.

#### Solution

Allow the base fee to be dynamically modified based on average load with a target of 50% full blocks.

### DIFF
--- a/proposals/0003-dynamic-compute-unit-base-fee.txt
+++ b/proposals/0003-dynamic-compute-unit-base-fee.txt
@@ -58,12 +58,6 @@ Motivation for making votes pay the same exact base fee as other
 transactions is to prevent a faulty majority from stuffing the
 ledger with maximum votes.
 
-Maximum block limit is defined as part of the solana validator
-client.  Currently it is at 48 million compute units.  The block
-target is 50% of the maximum, which is 24 million compute units.
-This target may be updated with a minor version mainnet release
-followed by a feature activation.
-
 ### Reward split between leader and burn
 
 With this proposal, the base fee should be 100% burned, and the
--- a/proposals/0003-dynamic-compute-unit-base-fee.txt
+++ b/proposals/0003-dynamic-compute-unit-base-fee.txt
@@ -72,3 +72,11 @@ for this split is that block verification is something that every
 participant has to pay for, even unstaked nodes.  Prioritizing the
 top paying transactions out of 100gbps worth of spam is something
 that leaders have to acutely pay for.
+
+### Maximum expected cost for users
+
+When users sign a tx, they are implicitly authorizing the network
+to charge up to the maximum fee possible. If all the blocks were
+full for the entire duration of a recent blockhash, the maximum fee
+growth is roughly 600% at a 12.5% rate and a max blockhash age of
+120 - (1.125)^(120/8).
--- a/proposals/0003-dynamic-compute-unit-base-fee.txt
+++ b/proposals/0003-dynamic-compute-unit-base-fee.txt
@@ -1,0 +1,59 @@
+---
+simd: '0003'
+title: Dynamic Compute Unit Base Fee
+authors:
+  - Anatoly Yakovenko (Solana Labs)
+category: Sealevel Virtual Machine
+type: Standards Track
+status: Draft
+created: 2022-12-05
+---
+
+## Summary
+
+Original implementation of Solana had a dynamic governor that
+targeted blocks to be at 50% load. The governor was signature based,
+and was designed without compute units in mind.  As SVM matured
+compute units were added to track load instead of signatures, and
+priority fees were added for prioritizing transaction scheduling.
+The dynamic signature based governor was removed, and right now the
+network has no mechanism to try to target a specific load.
+
+The motivation for a 50% target load is two fold.  One is that at
+100% the heterogeneous hardware on the network doesn't perform
+uniformly.  This causes downstream issues with leaders delaying and
+missing their blocks, and overall performance degradation. Secondly,
+stuffing the ledger with full blocks without a fee is an attack on
+the verifiability of the ledger by 3rd parties. For example, a faulty
+majority of the validators start creating continuous full blocks
+while paying minimal base fees and zero priority fees.
+
+## Specification
+
+### Compute Unit Target
+
+Maximum block limit is defined as part of the solana validator
+client.  Currently it is at 48 million compute units.  The block
+target is 50% of the maximum, which is 24 million compute units.
+This target may be updated with a minor version mainnet release
+followed by a feature activation.
+
+### Governor
+
+If the average compute units for the last 8 blocks > 50%, base fee
+goes up by 12.5%.  If the average compute units for the last 8
+blocks < 50%, base fee drops by 50%.
+
+Lowest possible fee = 1000 lamports / 200_000 compute units.
+
+There is no maximum fee, see [###Votes].
+
+### Votes
+
+Votes are not exempt. Validators should start censoring to maintain
+load < 50% that includes votes and non votes after fees reach some
+critical point.
+
+Motivation for making votes pay the same exact base fee as other
+transactions is to prevent a faulty majority from stuffing the
+ledger with maximum votes.
--- a/proposals/0003-dynamic-compute-unit-base-fee.txt
+++ b/proposals/0003-dynamic-compute-unit-base-fee.txt
@@ -119,3 +119,30 @@ the nonce with a different high priority tx.
 
 This maybe updated with a feature activation, or by median stake
 weighted value signaled by validators.
+
+### Alternative Solution For Vote base fees
+
+Increasing vote fees may problematic for validators.  Alternative
+solution would be to reserve X% of the block for votes. For example:
+
+* 33% CUs are reserved for votes
+* 66% is the base fee threshold
+
+Validators always pay the lowest base fee for votes, but the maximum
+number of validators that are scheduled in the leader scheduler is
+
+* ((MAX_COMPUTE_LIMIT) * 1/3)/(VOTE_CU)
+
+Presently that would be (48m/3)/(5_000) = 3,200 voting validators.
+
+In this example, only the top 3,200 voting validators by stake
+weight would be added to the leader scheduler, and would be considered
+as active w.r.t. voting. Validators that do not have enough stake
+to be in the top 3,200 would be considered inactive and wouldn't
+be scheduled in the leader schedule and wouldn't earn any rewards
+from voting, and wouldn't be considered as part of the quorum in
+consensus.
+
+In this example, base fees would double when load > 66%, and base
+fees would double only for non-vote transactions.  Votes would
+continue paying the lowest base fee.
--- a/proposals/0003-dynamic-compute-unit-base-fee.txt
+++ b/proposals/0003-dynamic-compute-unit-base-fee.txt
@@ -30,6 +30,17 @@ while paying minimal base fees and zero priority fees.
 
 ## Specification
 
+### Transaction Fees on solana
+
+There are two kinds of fees on solana, base fees, and priority fees.
+This proposal is focused on base fees only.  Base fees are automatic
+and are set by the network. Prior to this proposal the base fees
+were fixed, 0.000005 sol per signature.
+
+After this proposal has been implemented the base fees are set by
+the network based on current load, and the fee is charged per Compute
+Unit requested by the transaction.
+
 ### Compute Unit Target
 
 Maximum block limit is defined as part of the solana validator
--- a/proposals/0003-dynamic-compute-unit-base-fee.txt
+++ b/proposals/0003-dynamic-compute-unit-base-fee.txt
@@ -124,7 +124,7 @@ weighted value signaled by validators.
 
 ### Alternative Solution For Vote base fees
 
-Increasing vote fees may problematic for validators.  Alternative
+Increasing vote fees may be problematic for validators.  Alternative
 solution would be to reserve X% of the block for votes. For example:
 
 * 33% CUs are reserved for votes
--- a/proposals/0003-dynamic-compute-unit-base-fee.txt
+++ b/proposals/0003-dynamic-compute-unit-base-fee.txt
@@ -11,6 +11,14 @@ created: 2022-12-05
 
 ## Summary
 
+This proposal changes the base fee from being based on signatures
+per transaction to requested compute units per transaction, as well
+as adds a mechanism to dynamically adjust the base fee based on
+desired target load for the network, and changes the economics of
+fees on the network such that base fee is 100% burned and priority
+fees 100% go to the leader. This proposal doesn't impact storage
+fees.
+
 Original implementation of Solana had a dynamic governor that
 targeted blocks to be at 50% load. The governor was signature based,
 and was designed without compute units in mind.  As SVM matured
@@ -41,7 +49,11 @@ After this proposal has been implemented the base fees are set by
 the network based on current load, and the fee is charged per Compute
 Unit requested by the transaction.
 
-### Compute Unit Target
+There are two ways the validators control the base fee on the
+network.  Validators can set the Maximum Compute Unit Limit, or the
+Lowest Base fee.
+
+### Maximum Compute Unit Limit
 
 Maximum block limit is defined as part of the solana validator
 client.  Currently it is at 48 million compute units.  The block
@@ -53,15 +65,20 @@ If the average load in compute units for the last 8 blocks > 50%,
 base fee goes up by 12.5%.  If the average load in compute units
 for the last 8 blocks < 50%, base fee drops by 12.5%.
 
+Note, the 12.5% and 8 slot limits are an approximation of eip 1559's
+targets per 12.5 second ethereum block.
+
+### Lowest Base Fee
+
 Lowest possible fee = 1000 lamports / 200_000 compute units.
 
 There is no maximum fee, see [###Votes].
 
 ### Votes
 
-Votes are not exempt. Validators should start censoring to maintain
-load < 50% that includes votes and non votes after fees reach some
-critical point.
+Votes are not exempt. Validators should start dropping non essential
+transactions to maintain load < 50% that includes votes and non
+votes after fees reach some critical point.
 
 Motivation for making votes pay the same exact base fee as other
 transactions is to prevent a faulty majority from stuffing the
--- a/proposals/0003-dynamic-compute-unit-base-fee.txt
+++ b/proposals/0003-dynamic-compute-unit-base-fee.txt
@@ -53,7 +53,7 @@ followed by a feature activation.
 
 If the average compute units for the last 8 blocks > 50%, base fee
 goes up by 12.5%.  If the average compute units for the last 8
-blocks < 50%, base fee drops by 50%.
+blocks < 50%, base fee drops by 12.5%.
 
 Lowest possible fee = 1000 lamports / 200_000 compute units.
 
--- a/proposals/0003-dynamic-compute-unit-base-fee.txt
+++ b/proposals/0003-dynamic-compute-unit-base-fee.txt
@@ -145,6 +145,6 @@ be scheduled in the leader schedule and wouldn't earn any rewards
 from voting, and wouldn't be considered as part of the quorum in
 consensus.
 
-In this example, base fees would double when load > 66%, and base
-fees would double only for non-vote transactions.  Votes would
-continue paying the lowest base fee.
+In this example, base fees would start increasing when load > 66%,
+and base fees would double only for non-vote transactions.  Votes
+would continue paying the lowest base fee.
--- a/proposals/0003-dynamic-compute-unit-base-fee.txt
+++ b/proposals/0003-dynamic-compute-unit-base-fee.txt
@@ -87,4 +87,8 @@ ledger with maximum votes.
 ### Reward split between leader and burn
 
 With this proposal, the base fee should be 100% burned, and the
-priority fee should 100% go to the block producer.
+priority fee should 100% go to the block producer. The motivation
+for this split is that block verification is something that every
+participant has to pay for, even unstaked nodes.  Prioritizing the
+top paying transactions out of 100gbps worth of spam is something
+that leaders have to acutely pay for.
--- a/proposals/0003-dynamic-compute-unit-base-fee.txt
+++ b/proposals/0003-dynamic-compute-unit-base-fee.txt
@@ -83,9 +83,13 @@ that leaders have to acutely pay for.
 When users sign a tx, they are implicitly authorizing the network
 to charge up to the maximum fee possible. If all the blocks were
 full for the entire duration of a recent blockhash, the maximum fee
-growth is roughly 600% at a 12.5% rate and a max blockhash age of
-120 - (1.125)^(120/8).
+growth is roughly 82x at a 12.5% rate and a max blockhash age of
+300. (1.125)^(300/8).
 
 To reduce the exposure, users can sign an older blockhash. One with
 a max lifetime of 30 slots would have a maximum fee increase of
 55%.
+
+Durable nonce transactions should limit their exposure via a maximum
+balance in the fee payer or cancel the transaction by invalidating
+the nonce with a different high priority tx.
--- a/proposals/0003-dynamic-compute-unit-base-fee.txt
+++ b/proposals/0003-dynamic-compute-unit-base-fee.txt
@@ -109,6 +109,8 @@ Durable nonce transactions should limit their exposure via a maximum
 balance in the fee payer or cancel the transaction by invalidating
 the nonce with a different high priority tx.
 
+Burn economics may need to be considered in a separate proposal.
+
 ### Constants
 
 * lowest fee per compute unit: 1000 lamports / 200_000 CUs
--- a/proposals/0003-dynamic-compute-unit-base-fee.txt
+++ b/proposals/0003-dynamic-compute-unit-base-fee.txt
@@ -14,10 +14,7 @@ created: 2022-12-05
 This proposal changes the base fee from being based on signatures
 per transaction to requested compute units per transaction, as well
 as adds a mechanism to dynamically adjust the base fee based on
-desired target load for the network, and changes the economics of
-fees on the network such that base fee is 100% burned and priority
-fees 100% go to the leader. This proposal doesn't impact storage
-fees.
+desired target load for the network.
 
 Original implementation of Solana had a dynamic governor that
 targeted blocks to be at 50% load. The governor was signature based,
@@ -84,7 +81,12 @@ Motivation for making votes pay the same exact base fee as other
 transactions is to prevent a faulty majority from stuffing the
 ledger with maximum votes.
 
-### Reward split between leader and burn
+### Fee economics for validators
+
+Changes to the current fee model should come in a separate proposal,
+the following outlines two possible designs.
+
+#### Burn base and congestion fees only
 
 The original solana design had a 50% burn and 50% validator reward
 for all fees. With priority fees, this split is suboptimal.  A user
@@ -101,11 +103,32 @@ With this proposal, the base fee, including the congestion fee,
 should be 100% burned, and the priority fee should 100% go to the
 block producer. The motivation for this split is that block
 verification is something that every participant has to pay for,
-even unstaked nodes.  Prioritizing the top paying transactions out
+even unstaked nodes. Prioritizing the top paying transactions out
 of 100gbps worth of spam is something that leaders have to acutely
 pay for. By rewarding 100% of the priority fees to the block producer,
 the incentive to use a side channel for prioritization fees is
 greatly reduced.
+
+#### Keep the 50/50 burn
+
+Generally, burning the user specified priorioty fee creates an
+incentive for the user to pay the validator for priority via a
+sidechannel.  If in the future, the solana protocol implements
+execution time preference for transactions using the same priority
+fee mechanism, it will not be possible to bypassed these fees without
+risk via sidechannel.
+
+General outline for execution time preference:
+
+1. Each block is segmented into 100ms slices. Currently that would
+be 16 PoH ticks.
+
+2. During replay, for each slice, the transactions are reordered
+by priority before execution.
+
+With multiple concurrent leaders, users best outcome for priority
+execution is to send the transaction to the closest leader and to
+pay the highest fee possible.
 
 ### Maximum expected cost for users
 
--- a/proposals/0003-dynamic-compute-unit-base-fee.txt
+++ b/proposals/0003-dynamic-compute-unit-base-fee.txt
@@ -86,12 +86,26 @@ ledger with maximum votes.
 
 ### Reward split between leader and burn
 
-With this proposal, the base fee should be 100% burned, and the
-priority fee should 100% go to the block producer. The motivation
-for this split is that block verification is something that every
-participant has to pay for, even unstaked nodes.  Prioritizing the
-top paying transactions out of 100gbps worth of spam is something
-that leaders have to acutely pay for.
+The original solana design had a 50% burn and 50% validator reward
+for all fees. With priority fees, this split is suboptimal.  A user
+and validator can prioritize via a side channel and avoid the burn
+for the priority portion of the fees all together. For example,
+instead of adding a 0.1 sol priority fee to the tx, user sends 0.75
+sol to the validator directly. The validator earns 25% more and the
+user spends 25% less, by bypassing the default priority fees.
+
+It's necessary to burn the congestion fees, otherwise a faulty
+majority is incentivized to keep the congestion fees high.
+
+With this proposal, the base fee, including the congestion fee,
+should be 100% burned, and the priority fee should 100% go to the
+block producer. The motivation for this split is that block
+verification is something that every participant has to pay for,
+even unstaked nodes.  Prioritizing the top paying transactions out
+of 100gbps worth of spam is something that leaders have to acutely
+pay for. By rewarding 100% of the priority fees to the block producer,
+the incentive to use a side channel for prioritization fees is
+greatly reduced.
 
 ### Maximum expected cost for users
 
--- a/proposals/0003-dynamic-compute-unit-base-fee.txt
+++ b/proposals/0003-dynamic-compute-unit-base-fee.txt
@@ -64,26 +64,6 @@ target is 50% of the maximum, which is 24 million compute units.
 This target may be updated with a minor version mainnet release
 followed by a feature activation.
 
-### Governor
-
-If the average compute units for the last 8 blocks > 50%, base fee
-goes up by 12.5%.  If the average compute units for the last 8
-blocks < 50%, base fee drops by 50%.
-
-Lowest possible fee = 1000 lamports / 200_000 compute units.
-
-There is no maximum fee, see [###Votes].
-
-### Votes
-
-Votes are not exempt. Validators should start censoring to maintain
-load < 50% that includes votes and non votes after fees reach some
-critical point.
-
-Motivation for making votes pay the same exact base fee as other
-transactions is to prevent a faulty majority from stuffing the
-ledger with maximum votes.
-
 ### Reward split between leader and burn
 
 With this proposal, the base fee should be 100% burned, and the
--- a/proposals/0003-dynamic-compute-unit-base-fee.txt
+++ b/proposals/0003-dynamic-compute-unit-base-fee.txt
@@ -57,3 +57,34 @@ critical point.
 Motivation for making votes pay the same exact base fee as other
 transactions is to prevent a faulty majority from stuffing the
 ledger with maximum votes.
+
+Maximum block limit is defined as part of the solana validator
+client.  Currently it is at 48 million compute units.  The block
+target is 50% of the maximum, which is 24 million compute units.
+This target may be updated with a minor version mainnet release
+followed by a feature activation.
+
+### Governor
+
+If the average compute units for the last 8 blocks > 50%, base fee
+goes up by 12.5%.  If the average compute units for the last 8
+blocks < 50%, base fee drops by 50%.
+
+Lowest possible fee = 1000 lamports / 200_000 compute units.
+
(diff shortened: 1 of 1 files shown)


proposals/0003-dynamic-compute-unit-base-fee.txt
```diff
@@ -0,0 +1,90 @@
...
+
+### Votes
+
+Votes are not exempt. Validators should start censoring to maintain
+load < 50% that includes votes and non votes after fees reach some
+critical point.
+
+Motivation for making votes pay the same exact base fee as other
+transactions is to prevent a faulty majority from stuffing the
+ledger with maximum votes.
```
barnjamin, line 85: Governor and votes sections appear duped?

overclock-validator:
Does it make the most sense to tie these increases to the base fee per compute units? I.e. we start with a fee of 0.0000005 sol per 1000 compute units. Blocks going over half the dedicated non-vote compute per block would lead to the next n number of blocks having their fee per compute unit adjusted up some predictable amount.

Validators could adjust the base fee through governance means at regular intervals to semi-adjust to changes in sol/fiat price (fiat being what hardware and bandwidth are priced according to). Sui has a system where fees are adjusted by the validator set regularly.

Also, this whole system might be made easier if non vote and vote compute are separated out a bit more? cc @AshwinSekar @taozhu-chicago @apfitzge

y2kappa:
A few (clarifying) questions/comments:
1. Why are leaders delaying and missing their blocks? Is this intentional, or because they can't catch up with verifying blocks from a more performant previous leader? Or is it because, in order to verify a block the block needs to wait and be verified by less performant nodes? Anyway, the idea here is that the network is as fast as its slowest nodes?
> Secondly, stuffing the ledger with full blocks without a fee is an attack on the verifiability of the ledger by 3rd parties. For example, a faulty majority of the validators start creating continuous full blocks while paying minimal base fees and zero priority fees. 
2. Don't understand the above section? Why can't 3rd parties verify the blocks, how is that an attack? Because fast validators filling the blocks with bogus txns won't allow slower 3rd parties to catch up and therefore not participate in consensus? Can you give an example of an attack?
3. A bit of context necessary: who pays for the votes, is it the other nodes? If votes are not exempt, doesn't that also prohibit other nodes from participating, kind of achieving the same thing? Is the logic that if you start paying a higher fee, as a voting node, when it's your turn you'll propose smaller blocks to decrease the fee? But doesn't that encourage even further the faulty majority to keep doing it to kick out the good nodes which are not willing to pay the higher fee?

proposals/0003-dynamic-compute-unit-base-fee.txt
```diff
@@ -0,0 +1,90 @@
...
+
+Maximum block limit is defined as part of the solana validator
```
taozhu-chicago, line 61: dedup this block with Ln35?

taozhu-chicago:
> Anyway, the idea here is that the network is as fast as its slowest nodes?

Don't think this is the case. Network performance should be close to the average performance of majority of nodes - excluding the slowest and fastest nodes.

taozhu-chicago:
> Does it make the most sense to tie these increases to the base fee per compute units?

Mmm, it does sound more fair to tie the fee to CU, so 1M cu tx would pay more compare to 1K cu tx when load is high, perhaps we can utilize `fee structurer`. Tho base fee as an adder is much simpler. 

>  this whole system might be made easier if non vote and vote compute are separated out a bit more?

Doesn't the proposal want the same base fee being applied to vote and non-vote in same way?

buffalu:
does this impact storage costs in anyway? or just compute?

y2kappa:
Got it. 

It looks like a mechanism to penalize some malicious behaviour / incentivize good behaviour, where good behaviour is not going above 50% capacity for longer than 8 blocks. 

Doesn't this also discourage higher performance? It seems like the goal to have uniform performance and nothing incentivises block producers to produce fuller blocks, shouldn't this be the goal also / optimized for? How and when will this target be raised to increase network capacity?

aeyakovenko:
> Doesn't this also discourage higher performance? It seems like the goal to have uniform performance and nothing incentivises block producers to produce fuller blocks, even if they can. Shouldn't this be the goal also / optimized for? How and when will this target be raised to increase network capacity? How will we get past 4k tps?

tps != compute limit, since transactions like transfers can be very cheap.  overall the validators need to agree to raise the system level compute limit above 48m CUs.

proposals/0003-dynamic-compute-unit-base-fee.txt
```diff
@@ -0,0 +1,91 @@
...
+load < 50% that includes votes and non votes after fees reach some
+critical point.
+
+Motivation for making votes pay the same exact base fee as other
+transactions is to prevent a faulty majority from stuffing the
+ledger with maximum votes.
+
+### Reward split between leader and burn
+
+With this proposal, the base fee should be 100% burned, and the
```
t-nelson, line 74: IMO this is a separate proposal for economics change.  no sense complicating one issue with another.  whether there should be any dependency between this proposal and the economics one, i'll leave up to you.

aeyakovenko:
> Does it make the most sense to tie these increases to the base fee per compute units? I.e. we start with a fee of 0.0000005 sol per 1000 compute units. Blocks going over half the dedicated non-vote compute per block would lead to the next n number of blocks having their fee per compute unit adjusted up some predictable amount.

Thats what the proposal does.  Base fee per CU is increased.  I think validators should signal a vote on total CUs per block and lowest base fee allowed that can be updated once per epoch.

proposals/0003-dynamic-compute-unit-base-fee.txt
```diff
@@ -0,0 +1,91 @@
...
+while paying minimal base fees and zero priority fees.
+
+## Specification
+
+### Transaction Fees on solana
+
+There are two kinds of fees on solana, base fees, and priority fees.
+This proposal is focused on base fees only.  Base fees are automatic
+and are set by the network. Prior to this proposal the base fees
+were fixed, 0.000005 sol per signature.
```
t-nelson, line 38: this isn't technically (but on MB, may have been practically) true.  the governor is definitely active and has been triggered during testnet bench tps runs.

proposals/0003-dynamic-compute-unit-base-fee.txt
```diff
@@ -0,0 +1,91 @@
...
+After this proposal has been implemented the base fees are set by
+the network based on current load, and the fee is charged per Compute
+Unit requested by the transaction.
+
+### Compute Unit Target
+
+Maximum block limit is defined as part of the solana validator
+client.  Currently it is at 48 million compute units.  The block
+target is 50% of the maximum, which is 24 million compute units.
+This target may be updated with a minor version mainnet release
```
t-nelson, line 49: could potentially vote on the threshold in the future.  similar to block timestamp oracle, but probably different statistics since pure stake weight is likely the wrong thing here.

proposals/0003-dynamic-compute-unit-base-fee.txt
```diff
@@ -0,0 +1,91 @@
...
+followed by a feature activation.
+
+### Governor
+
+If the average compute units for the last 8 blocks > 50%, base fee
+goes up by 12.5%.  If the average compute units for the last 8
+blocks < 50%, base fee drops by 50%.
```
t-nelson: these are back of envelope numbers, or is there some rigor behind them?  maybe replace with variables for the time being?

t-nelson: "compute units" should be "load" here right?  since we're speaking in percentages

proposals/0003-dynamic-compute-unit-base-fee.txt
```diff
@@ -0,0 +1,91 @@
...
+
+Lowest possible fee = 1000 lamports / 200_000 compute units.
+
+There is no maximum fee, see [###Votes].
+
+### Votes
+
+Votes are not exempt. Validators should start censoring to maintain
```
t-nelson, line 64: by "censoring" here, we mean "preferring vote transactions," right?

proposals/0003-dynamic-compute-unit-base-fee.txt
```diff
@@ -0,0 +1,91 @@
...
+load < 50% that includes votes and non votes after fees reach some
+critical point.
```
t-nelson, line 66: could make this a sliding scale based on tower distance.  basically a "progressive vote-only mode"

proposals/0003-dynamic-compute-unit-base-fee.txt
```diff
@@ -0,0 +1,91 @@
...
+top paying transactions out of 100gbps worth of spam is something
+that leaders have to acutely pay for.
+
+### Maximum expected cost for users
+
+When users sign a tx, they are implicitly authorizing the network
+to charge up to the maximum fee possible. If all the blocks were
+full for the entire duration of a recent blockhash, the maximum fee
+growth is roughly 600% at a 12.5% rate and a max blockhash age of
+120 - (1.125)^(120/8).
```
t-nelson, line 87: should be [`150 slots`](https://github.com/solana-labs/solana/blob/1b49c52d9896172f2ccc18a896b9c7c7b784cfbd/sdk/program/src/clock.rs#L96-L99) here.  `120` is max hash age in seconds (300slots * .4sec)

proposals/0003-dynamic-compute-unit-base-fee.txt
```diff
@@ -0,0 +1,91 @@
...
+
+To reduce the exposure, users can sign an older blockhash. One with
```
t-nelson, line 89: i'm not sure we actually have the rpc infra to do this reliably today.  shouldn't be too difficult to add though.

remember that the fee calculator has been decommissioned in favor of simulation-based fee calculations

aeyakovenko: yea, it would need to do both.

proposals/0003-dynamic-compute-unit-base-fee.txt
aeyakovenko: I think if max age is 300, and leaders drop at 150, then worst case from a faulty leader would be 300 afaik.

proposals/0003-dynamic-compute-unit-base-fee.txt
aeyakovenko: yes, limiting the CUs to votes and higher priority txs in the 24m cu limit.

proposals/0003-dynamic-compute-unit-base-fee.txt
aeyakovenko: ethereum 1559 does a 12.5% per 1 eth block which is roughly 8 slots.

proposals/0003-dynamic-compute-unit-base-fee.txt
t-nelson: mind updating for clarity?

proposals/0003-dynamic-compute-unit-base-fee.txt
t-nelson: maybe add a reference to eip1559 as the basis?

proposals/0003-dynamic-compute-unit-base-fee.txt
t-nelson: elaborating the supporting rpc changes would be valuable

zfedoran:
> When users sign a tx, they are implicitly authorizing the network to charge up to the maximum fee possible. If all the blocks were full for the entire duration of a recent blockhash, the maximum fee growth is roughly 82x at a 12.5% rate and a max blockhash age of 300. (1.125)^(300/8).

It would be nice if there was a way to limit the exposure outside of using an older blockhash with a lower fee. However, I can't think of a way that doesn't lend itself to abuse.

overclock-validator:
> > When users sign a tx, they are implicitly authorizing the network to charge up to the maximum fee possible. If all the blocks were full for the entire duration of a recent blockhash, the maximum fee growth is roughly 82x at a 12.5% rate and a max blockhash age of 300. (1.125)^(300/8).
> 
> It would be nice if there was a way to limit the exposure outside of using an older blockhash with a lower fee. However, I can't think of a way that doesn't lend itself to abuse.

What would be better I think, is to charge based on the fee at any referenced blockhash which is balanced by having a "stale fee" for transactions. Essentially, you can send a transaction with a blockhash as old as possible to get a cheaper transaction in, but you also increase the risk of having your transaction not included while the blockhash is still good. You have a second window after the normal blockhash life where the bulk of transactions are ignored and just the fee payer is charged -- low compute and highly parallelizable. If transactions are more at risk of actually getting charged instead of just dropped when expired, you'll have fewer people willing to spam loads of transactions. Having a slow escalating fee schedule doesn't help much with the initial flood of spam due to the slow ramp in fees but this reduces the incentive greatly. https://github.com/solana-labs/solana/issues/25211

proposals/0003-dynamic-compute-unit-base-fee.txt
overclock-validator: > > Does it make the most sense to tie these increases to the base fee per compute units? I.e. we start with a fee of 0.0000005 sol per 1000 compute units. Blocks going over half the dedicated non-vote compute per block would lead to the next n number of blocks having their fee per compute unit adjusted up some predictable amount.
> 
> Thats what the proposal does. Base fee per CU is increased. I think validators should signal a vote on total CUs per block and lowest base fee allowed that can be updated once per epoch.

Yeah this sounds like a cool system, every epoch is pretty frequent but I imagine validators can just have things coast and then rally together a majority vote when there's enough motivation to change things.

0xxbt:
Long or not?

Thanks,

nicovrg:
just curious, why did you choose 8 blocks and 12.5%?

proposals/0003-dynamic-compute-unit-base-fee.txt
nicovrg: had the same question and agree they could be replace by variables to avoid setting them in stone (most people may assume these number result from logic/debate/test and not question them)

proposals/0003-dynamic-compute-unit-base-fee.txt
aeyakovenko: maybe, I think they are related.  the motivation for the fee design is intertwined with resisting faulty majority spam.  Economics are basically how the network can protect itself from a faulty majority spamming all the users.

proposals/0003-dynamic-compute-unit-base-fee.txt
aeyakovenko: governor has been disabled on mb afaik

aeyakovenko:
> It would be nice if there was a way to limit the exposure outside of using an older blockhash with a lower fee. However, I can't think of a way that doesn't lend itself to abuse.

@zfedoran yea, maybe if this approach isn't sufficient the transaction can add a maximum fee later.  But id rather not add transaction fields when we don't need to.

aeyakovenko:
@overclock-validator 

there is a cost to maintain the older transactions in memory at each validator and shuffle them around to each leader.  imho, the 150 slot limit is way too long.  in practice if a tx doesn't make it into a block in 30 seconds the user should be given the chance to change their mind or bump the priority fee.

proposals/0003-dynamic-compute-unit-base-fee.txt
```diff
@@ -0,0 +1,104 @@
...
+### Transaction Fees on solana
+
+There are two kinds of fees on solana, base fees, and priority fees.
+This proposal is focused on base fees only.  Base fees are automatic
+and are set by the network. Prior to this proposal the base fees
+were fixed, 0.000005 sol per signature.
+
+After this proposal has been implemented the base fees are set by
+the network based on current load, and the fee is charged per Compute
+Unit requested by the transaction.
```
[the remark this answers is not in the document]
aeyakovenko: Yes, the goal is to remove the per signature fee and just charge CU costs.

overclock-validator:
> @overclock-validator
> 
> there is a cost to maintain the older transactions in memory at each validator and shuffle them around to each leader. imho, the 150 slot limit is way too long. in practice if a tx doesn't make it into a block in 30 seconds the user should be given the chance to change their mind or bump the priority fee.

Makes sense. In my mind the stale fee would just be charging the base fee for the amount of compute requested. If the user guessed the wrong priority, they wouldnt be charged for the priority, just the nearly non existent base fee of 1000 lamports / 200_000 CUs. I don't think normal users would notice... only the people who decide to send 100k transactions at the same time. 

On that note, bankless leaders will base block packing on compute estimations by the validator on some form of simulation and compute request by users? If it requires accurate compute requests by users then the requests will need to be more fine grained and fees that are meaningful will incentivize more accurate compute requests.

overclock-validator:
"Votes are not exempt. Validators should start dropping non essential
transactions to maintain load < 50% that includes votes and non
votes after fees reach some critical point.

Motivation for making votes pay the same exact base fee as other
transactions is to prevent a faulty majority from stuffing the
ledger with maximum votes."

"Votes are not exempt. Validators should start dropping non essential
transactions to maintain load < 50% that includes votes and non
votes after fees reach some critical point.

"Motivation for making votes pay the same exact base fee as other
transactions is to prevent a faulty majority from stuffing the
ledger with maximum votes."

I think the better design here is to have the EIP-1559 mechanism but have it applied to votes and non-votes completely separately and validators can decide on the proper compute limits and resources to dedicate to them (i.e. number of vote threads) instead of intermixing them? It's a lot easier to manage validator economics if voting is isolated like this and I think you might have a lot of problems and lack of predictability if voting fees are affected by non vote demand. 

Burning all base fees is also a very large change to validator economics (negative imo). Right now a change like this would result in block rewards almost completely disappearing due to priority fees still being very low (current compute amounts per block are around 10mil). It'll make it even harder for new validators who now can't even compete on commission to stand out. There's also no incentive for validators to include other validator's votes/voting fees anymore if they're all burned and this increases the incentive to delay the votes of other validators in the future vote latency based credit system, where validators are even more dependent on inflation rewards. I think the 100% base fee burn might add some other weird incentives but I need to ponder more.

aeyakovenko:
@overclock-validator this change would also reduce the vote fees, since the current base load is < 50% of the block.  the fees for votes will trend down towards the bottom.  


> I think the better design here is to have the EIP-1559 mechanism but have it applied to votes and non-votes completely separately 

wdym by this?  1559 has a base fee that is 100% burned and a dynamic governor that tries to target 50% load.  this is the same as what's in this proposal.

aeyakovenko:
@overclock-validator 
> Makes sense. In my mind the stale fee would just be charging the base fee for the amount of compute requested. If the user guessed the wrong priority, 

the problem is that there is no mempool in solana.  So where would the leader get stale transactions from?  Every forwarder should be prioritizing their resources for recent ones and would dump stale txs from memory.

overclock-validator:
> @overclock-validator this change would also reduce the vote fees, since the current base load is < 50% of the block. the fees for votes will trend down towards the bottom.
> 
> > I think the better design here is to have the EIP-1559 mechanism but have it applied to votes and non-votes completely separately
> 
> wdym by this? 1559 has a base fee that is 100% burned and a dynamic governor that tries to target 50% load. this is the same as what's in this proposal.
> 
> Also, pls take a look at
> 
> https://github.com/solana-foundation/solana-improvement-documents/pull/4/files#diff-bfb6ad0993b3f6dad04c8220c1d91e9d45b7b9f8d1563d4a40ff3323deed6b2bR125-R150

With the timeliness vote credit changes there is an incentive to delay the votes of other validators, but this is balanced out by the fact that you would lose block rewards if you push votes out of your leader slots (you would do this to have an apy advantage). If you burn all votes or perhaps make them inconsequential to block rewards by making their cost too low, then it becomes non or less important for the leader to include votes from other validators in their blocks. This is not really a strategy that smaller validators with fewer leader slots can use, but larger validators can manipulate the vote credit system more easily. 

By separate I mean that there should be a separate compute limit for non votes and votes and that non votes hitting the 50% load shouldn't impact the price of votes. Voting costs going up helps validators above the break even point but hurts smaller validators so this could even be gameable perhaps. Voting fees are currently the major source of block rewards for validators so the incentive is there to raise the base fees if it can be gamed (I didn't think through this that deeply though so maybe it's unrealistic). 

On another note, I feel like much of Solana traffic is typically at a very steady rate, is there a worry about certain things just needing to turn off if the steady state goes above 50% compute and very quickly ramps up in cost? I guess one thing that can be tuned by validators is the account compute limits --- decreasing those limits increases the strength of the local fee marketness of the system.

overclock-validator:
> @overclock-validator
> 
> > Makes sense. In my mind the stale fee would just be charging the base fee for the amount of compute requested. If the user guessed the wrong priority,
> 
> the problem is that there is no mempool in solana. So where would the leader get stale transactions from? Every forwarder should be prioritizing their resources for recent ones and would dump stale txs from memory.

In the "compostability" proposal a leader who just finished their slot would forward off unprocessed transactions to other validators to store and keep track of (perhaps the many validators who dont have many leader slots). But yeah the system is still pretty vague... the idea would be to have the fees from any stale transactions go into a pool that gets split among validators at the end of an epoch. Having a pool would encourage a cooperative system and gives small validators a job when they're more idle (our validator is about 20% cpu usage at steady state). 

Overall though, it discourages a transaction flood better than an eip1559 mechanism. If you send 5000 unique 500k cu transactions that have no hope of getting processed, then there is actually the possibility of getting charged for them once they become stale (the transaction's fee payer is just charged and the rest of transaction is ignored). This possibility of getting charged would likely disincentivize much of the spamming in the first place. It also encourages more accurate priority estimation. If you underbid by a lot, then it's more likely that your underbidded transaction doesnt get processed, but the stale transaction will (the fee reward to cu ratio is better with the stale transaction).

aeyakovenko:
> @overclock-validator 
> Maybe we want to increase the block compute limit to 96million for example and then the 1/3 compute reserved just ends up being a bit wasteful because economically we can't support nearly that many validators (due to stake distribution for example).

why not lower the base fee?  the vote fees are the vast majority of the costs in running a validator. the economics are tied to that.  if the CU capacity goes up, the lowest base fee should be lowered as well imho.

aeyakovenko:
> @overclock-validator 
> In the "compostability" proposal a leader who just finished their slot would forward off unprocessed transactions to other validators to store and keep track of (perhaps the many validators who dont have many leader slots).

This is worth a separate discussion, but basically it comes down to the previous leader can't take up 100% of the bandwidth of the next leader, and has to prioritize the highest paying TXs first, which is going to put the stale ones at a disadvantage.  and storing these stale ones takes resources that everyone else is contending with.

michaelh-laine:
> > @overclock-validator
> > Maybe we want to increase the block compute limit to 96million for example and then the 1/3 compute reserved just ends up being a bit wasteful because economically we can't support nearly that many validators (due to stake distribution for example).
> 
> why not lower the base fee? the vote fees are the vast majority of the costs in running a validator. the economics are tied to that. if the CU capacity goes up, the lowest base fee should be lowered as well imho.

lowering the base fee reduces block rewards significantly, which increases the break-even stake required, so while it might mean less SOL burned per day it means more days until a validator can achieve break even (and potentially fewer who can). It's asymmetric of course as non-votes also contribute to block rewards, but it will have an impact.

As for the economics of burning base vs 100% of prio going to leader we'd need some modelling on how various levels of priority fees would impact the overall validator economics, as well as a better understanding of the thinking behind this change, I feel that is insufficiently explained in the proposal. What does burning the base fee achieve vs just maintaining a 50% burn across all fees? The latter still incentivizes validators to pick the highest paying transactions first.

overclock-validator:
> > @overclock-validator
> > Maybe we want to increase the block compute limit to 96million for example and then the 1/3 compute reserved just ends up being a bit wasteful because economically we can't support nearly that many validators (due to stake distribution for example).
> 
> why not lower the base fee? the vote fees are the vast majority of the costs in running a validator. the economics are tied to that. if the CU capacity goes up, the lowest base fee should be lowered as well imho.

You may lower the base costs but you also shrink the size of the carrot on the stick because overall block rewards decrease :/. Competing on commission is very hard and I imagine you'll still see 0% commission validators even in an environment with even fewer block rewards. Obviously we dont want to hurt UX with high fees, but having a well balanced level of validator rewards will attract better validators and/or teams who want to use the validator as a revenue stream for a business (outside of MEV). I think it's hard to know how long it will take priority fees (and jito mev rewards) to match current block rewards. Priority fees might shrink too when the baseline fee is increased through the EIP1559 mechanism. Having the batch auctions within a slot and/or tuning account compute limits might increase the importance of priority fees though.

Anyway, I feel like there must be some cheap fee level where UX is still solid. Most current users wouldn't notice a 10x fee increase at the moment. I guess the teams that constantly need to send transactions would much moreso though (maybe why that other SIMD proposal where frequent transaction sending gets you some discount rate could make sense). 

"@overclock-validator, "voting threads" are just there to pack blocks. it has nothing to do with actual Compute Limits in the block. the reason there are dedicated voting threads is to prevent starvation when transactions are scheduled. You still need to define the limits in the block itself." 
-My understanding here was that Tao had found that 12mil compute is a rough approximation of what can be done by a thread in 400ms, and at the time 2 non-vote threads + 2 vote threads = 48mil total. A lot of measurements are in terms of compute to microsecond in here: https://github.com/solana-labs/solana/blob/master/runtime/src/block_cost_limits.rs

Regarding "compostability". I'll just create a separate proposal for that and maybe it can serve as inspiration for a different fee mechanism one day. The "stale transactions" only charge the signer so their cu is much less than the original transaction with less contention, but pay for the cu requested by the original transaction -- so overall the fee to cu ratio increases for the stale transaction. You wouldn't want stale fees taking up a lot of time/space though... Cherryworm thought these were perfect for batch processing though if you have one signer spamming a bunch. Ideally it's not a common thing anyway though and just serves a preventative purpose. But yeah... will take this to another proposal.

bji:
I'm officially piling onto the wagon of "don't burn 100% of base fees".  I would prefer to burn 50% of all tx fees (base, CU, priority) like Laine proposed.

> At 100% heterogenous hardware doesn't have uniform performance resulting in general performance degradation with downstream effects on block producers. Without a dynamic base fee, a faulty majority can stuff blocks with 100% load making verification for non validating nodes unnecessarily expensive.

I am not convinced of the necessity of taking action here.

Validators are already financially encouraged to produce blocks that can be replayed and voted on by the majority of the cluster, because producing blocks that can't be replayed by a significant number of validators increases the chances of being forked (because the next leader has to be able to replay your block in time to chain off of it rather than skipping back to a prior leader whos block *could* be replayed in time).

Validators are also financially encouraged to produce the largest block (most block rewards) that can meet the above criteria.

So I don't see a need to try to target a particular performance threshold.  Nodes are incentivized to find what the highest performance that can be achieved reliably is, and use that.

Furthermore, nodes are incentivized to stay within spitting distance of the performance required to reliably replay nodes, because falling behind means potentially missing votes and thus vote rewards.  And when/if timely vote credits happen, this effect is magnified since any delay in voting results in a real negative impact on vote credit earnings.

There are probably some things that validators could do to more intelligently and efficiently choose block sizes, and I will mention them briefly even though they're not relevant to the fee structure of this proposal per se:

1. Track avg CU of confirmed blocks and avg slot duration of confirmation of a block.  Use this information to target CU size to include in produced block.
2. Track blocks which were seen but never confirmed.  A difference in CU between confirmed and never-confirmed blocks could give some insight into dangerous CU levels (e.g. if blocks with CU no more than N usually are confirmed, and blocks with CU 1.1 * N are rarely confirmed, then 1.1*N is probably beyond the reasonable upper limit of blocks ize.

bji:
> They can, but a faulty majority can redline the cost.

A majority that can increase overall performance by increasing CU and tx counts consistently enough that their blocks are confirmed isn't "faulty" in my opinion -- it's pushing the performance envelope and demanding that slower nodes improve to keep up.

I think that's what we should want to happen.

bartenbach:
I'm casting my vote against burning 100% of base fees.

My reasons have already been voiced by others, so I won't reiterate the same things that have already been said.

bji:
I think that there should be five transaction fee categories:

1. Base fee: its purpose is to compensate validators for fixed costs of processing tx that are not a function of the size of the tx.
2. Byte fee: its purpose is to compensate validators for costs of processing tx that are a function of the size of the tx (reading bytes off the network, deserializing, buffering, etc)
3. CU fee: its purpose is to compensate validators for the costs of actually running the instructions of the tx, and assumes that compute units used is a reasonable proxy for these costs
4. Congestion fee: its purpose is to force higher fees to be paid to reduce demand for block space during times of high load
5. Priority fee: its purpose is to allow end users to pay more to have a higher chance of the tx being processed faster (or possibly even processed at all)

(1) and (2) probably can be rolled into a single fixed base fee since the byte size of transactions likely only marginally affects the fixed costs of processing them.  However I think ideally they'd be separate because aside from complexity of implementation, it would more accurately compensate validators for actual resources used if the two fees were considered separately.

All fees except congestion fees should go 50% to the validator, and 50% burned.  Congestion fees should be 100% burned (this is to prevent validators from increasing their fee take by artificially increasing congestion).

It should be obvious that CU fees are declared by the user as "maximum CU being paid for", and that since validators have to schedule assuming all CU will be used, the fee has to be charged as if all CU *were* used (i.e. no refunds for unused CU).  And it should also be obvious that the moment that tx processing would exceed paid-for CUs, the tx is aborted (but of course, CU fee still charged).

Finally, validators should prioritize tx based on revenue from the tx (i.e. (0.5 * (base_fees + CU fees + priority fee))) divided by (base_fees + CU fees), because (base_fees + CU fees) represents the expected cost of processing the tx.  Maximizing for the sum of (revenue / costs) across all tx gives the validator maximum possible revenue.

overclock-validator:
Congestion fees being fully burned makes sense to me. It'll be in the best interest of validators to keep block cu limits higher then. They can prevent any one market from increasing congestion costs too much by decreasing the account compute limits --- increasing the need to use priority fees for a particular hotspot/market..

aeyakovenko:
@bji 1-4 are rolled up into CU fees.  the transaction base fee = TX CUs * base rate

>Congestion fees should be based on the ratio of "CU of most recent N blocks" to "CU of most recent K blocks", where N is a number like 8, and N is a number like "number of blocks in an epoch".

I don't understand this part.  The MAX CU limit is a global constant.  congestion is based of what percentage of the MAX is utilized.  it doesn't matter what the historical average was.

overclock-validator:
> @bji 1-4 are rolled up into CU fees. the transaction base fee = TX CUs * base rate
> 
> > Congestion fees should be based on the ratio of "CU of most recent N blocks" to "CU of most recent K blocks", where N is a number like 8, and N is a number like "number of blocks in an epoch".
> 
> I don't understand this part. The MAX CU limit is a global constant. congestion is based of what percentage of the MAX is utilized. it doesn't matter what the historical average was.
> 
> > All fees except congestion fees should go 50% to the validator, and 50% burned. Congestion fees should be 100% burned (this is to prevent validators from increasing their fee take by artificially increasing congestion).
> 
> why not priority fees going to the validator? the current leader is doing the majority of the work of transaction scheduling. At a gut level it feels like they should capture most of the rewards there.
> 
> I think it's worth separating out the economics of the burn splits from the dynamic fees to deal with congestion from the proposal.

What is the current reason to burn 50% of fees btw? The main advantage I see for burning 50% of priority fees is that it may help with an investment narrative which validators would stand to benefit from if it helped. I can't really think of another reason to add priority burn really.... I know validators could try and abuse account compute limits (if that is under governance as well which I think would be good) by shrinking them and creating more artificial scarcity to increase priority fee usage. If that can be abused it seems good if non-validators could somehow be compensated.  Maybe my reasoning is dumb here.

aeyakovenko:
@overclock-validator 

if 0 was burned then majority can spam the network for free and increasing the cost of verification on all users including the unstaked ones.  

if 100% was burned, then there is no incentive to include any txs.

overclock-validator:
Looks like we are hitting block limits occasionally and account limits more often. Thanks to @taozhu-chicago for coming through with the new dashboard. 
full
costaccount

aeyakovenko:
priority fees are ~= 50% of the base fees currently.  I think burning base fees and letting leaders capture the priority fees wouldn't change the economics too much for validators currently. 

Screen Shot 2022-12-16 at 11 12 40 AM

the current 50/50% is not the worst option, but it is sub optional. Mempools like Jito can offer a prioritization scheme via a side channel that avoids the burn. A user then has the option of paying the priority fee of 0.1 sol or paying jito's validator directly 0.75 sol. The jito validator will earn 0.05 sol from the standard 50/50 priority fee, but 0.75 sol from the jito side channel. The jito path allows the user to split the burn between themselves.

overclock-validator:
> priority fees are ~= 50% of the base fees currently. I think burning base fees and letting leaders capture the priority fees wouldn't change the economics too much for validators currently.
> 
> Screen Shot 2022-12-16 at 11 12 40 AM
> 
> the current 50/50 burn is not the worst option, but it is sub optional. Mempools like Jito can offer a prioritization scheme via a side channel that avoids the burn. A user then has the option of paying the priority fee of 0.1 sol or paying jito's validator directly 0.75 sol. The jito validator will earn 0.05 sol from the standard 50/50 priority fee, but 0.75 sol from the jito side channel. The jito path allows the user and validator to split the burn between themselves.

Why not just keep base fees 50% but have the congestion fees 100% like @bji mentioned? Base fees 50% to validators incentivizes adding more compute like Zan mentioned. Validators also dont get 100% of Jito fees --- 8% commission take for validators is the social norm as of now. 

Would votes be treated differently here? On a side note, vote fees are a bit of a weird ponzi scheme where much of the network's validators are bearing the costs there and it shouldn't just be considered a pure network revenue source.

bji:
> @bji 1-4 are rolled up into CU fees. the transaction base fee = TX CUs * base rate

Yes and I'm saying that rolling them up is a mistake.  The ideal would be if every tx paid exactly what it cost the network to process it.  While we can never reach that idea, we can do better the more fee categories we can identify and make the price of a tx match as accurately as possible the cost of processing it.  Of course too many categories gets too confusing, so there is a limit.  Combining base + byte fees is a reasonable compromise I think; but combining all 4 fee types into one is I think too imprecise.

> 
> > Congestion fees should be based on the ratio of "CU of most recent N blocks" to "CU of most recent K blocks", where N is a number like 8, and N is a number like "number of blocks in an epoch".
> 
> I don't understand this part. The MAX CU limit is a global constant. congestion is based of what percentage of the MAX is utilized. it doesn't matter what the historical average was.

Yeah I'm not sure that part of my comment makes any sense at all.  I will revisit and revise as necessary. 

> > All fees except congestion fees should go 50% to the validator, and 50% burned. Congestion fees should be 100% burned (this is to prevent validators from increasing their fee take by artificially increasing congestion).
> 
> why not priority fees going to the validator? the current leader is doing the majority of the work of transaction scheduling. At a gut level it feels like they should capture most of the rewards there.

Sorry priority fees also should go 100% to validator.  Maybe I wasn't clear about that.

> I think it's worth separating out the economics of the burn splits from the dynamic fees to deal with congestion from the proposal.

In a scenario where 100% base fees are burned, all tx will need to include a priority fee otherwise the validator will not process them, assuming validators refuse to work for free.

So burning 100% of base fees is really just a play for more deflationary fee burn.  The 50% of fees that used to go to the validator and are now burned under 100% base fee burn, are replaced by a new floor on priority fees.

I'm not averse to more deflationary fee burn.

I guess that I've now landed on being neutral on the 100% base fee burn vs. 50% base fee burn issue.  It's just a question of how much deflationary burn is desired, and I really don't have an opinion either way on that.

bji:
> priority fees are ~= 50% of the base fees currently. I think burning base fees and letting leaders capture the priority fees wouldn't change the economics too much for validators currently.

I agree with this but requiring priority fees is going to make a whole bunch of existing code obsolete.  Any transaction author (wallet, defi app, other app, whatever) that doesn't include priority fees is going to suddenly become unable to submit tx that land.  That will be very disruptive.

How about this:  make base fee burn % something that is either controlled by governance, or increases from 50% to 100% on a schedule (say 1% per epoch for the next 50 epochs).  This will be less disruptive as it will allow the shift from base fee to priority fee to happen gradually.

proposals/0003-dynamic-compute-unit-base-fee.txt
```diff
@@ -0,0 +1,164 @@
+---
+simd: '0003'
```
mryalamanchi, line 2: Update the SIMD number to `0004` here as per the SIMD-0001 numbering [spec](proposals/0003-dynamic-compute-unit-base-fee.txt).

Also, please update the filename to `0004-dynamic-compute-unit-base-fee.md`

overclock-validator:
Block space is really starting to fill up now (see below). 

For choosing base fees, how about a system where validators adjust the base fee per cu (to track fiat/sol prices which hardware and bandwidth are denominated in). The way this could work is that the validator set can have a few options each epoch --- to adjust base fee per cu up by 5%, 0%, or decrease by 5% -- something like that. You'd need 80% quorum or so to have it adjusted (?). If it doesn't reach quorum for an epoch then it just stays the same. 
burn

IntokuSatori:
Any movement on this? I don't have much to add other than I think it's too important of a change to die on the vine without a resolution of the conversations above even if the proposal is ultimately not approved.

jacobcreech:
Hey @aeyakovenko. As per SIMD-1, this SIMD is now stagnant and to be closed. There's a lot of good discussion here, but is this even being still considered?

cc @taozhu-chicago

taozhu-chicago:
There are continued effort in the areas of this topic, Similar to SIMD-0017, it might benefit all if we go back to smaller group discussion. Would you agree @aeyakovenko ?

0xSol:
This SIMD has been inactive for 6 months. The status changed to Stagnant per [current SIMD process](https://github.com/0xSol/solana-improvement-documents/blob/main/proposals/0001-simd-process.md).  Close it for now, but feel free to reopen with new updates if still relevant, or create a new SIMD as appropriate.

PR #6, opened 2022-12-09
Increase base fees for infrequent users

betaconger:
Adding new SIMD 006 for community consideration


jacobcreech:
Hello @betaconger. I appreciate the enthusiasm for creating a new SIMD, but this one is lacking some key parts to help drive a discussion. Most notably a detailed design.

Could you please change the PR to draft until a design in added?

Also nit: newlines after every 80 characters to help with reviewing.

betaconger:
Hi all, I've made many changes to this SIMD, including an effort at a more detailed design and adherence to the template.md.  Can anyone please advise if this is ok to move from draft to pull request?  I recognize that I am not on Solana Foundation payroll, so if this needs to stay as a draft indefinitely, I understand that.  I also recognize that this may simply be an idea that no one is interested in pursuing.  However, if I can spur some helpful conversation and contribute in the community in the proper way, then I'll keep trying to improve this SIMD!

zfedoran:
Hi @betaconger,

I have taken a brief look at your proposal, admittedly not the whole thing. My initial feeling is that the issues raised by this proposal could potentially be indirectly modulated by the changes proposed in [SIMD-0003](https://github.com/solana-foundation/solana-improvement-documents/pull/4). If you have data to show that this is not the case, it would be good to add to the proposal.

Specifically, during a times of higher throughput the base fees would increase as a result of SIMD-0003, my hypothesis is that this would nudge participants to seek out times that are more cost effective; by definition lower throughput.

betaconger:
> Hi @betaconger,
> 
> I have taken a brief look at your proposal, admittedly not the whole thing. My initial feeling is that the issues raised by this proposal could potentially be indirectly modulated by the changes proposed in [SIMD-0003](https://github.com/solana-foundation/solana-improvement-documents/pull/4). If you have data to show that this is not the case, it would be good to add to the proposal.
> 
> Specifically, during a times of higher throughput the base fees would increase as a result of SIMD-0003, my hypothesis is that this would nudge participants to seek out times that are more cost effective; by definition lower throughput.

Hi @zfedoran,

Thanks for your comment.  It is definitely worth a high-level explanation about how this SIMD differs from https://github.com/solana-foundation/solana-improvement-documents/pull/4.  The main differences are as follows:

-This pull request seeks to raise the **ceiling** of fees that must be paid by a fee payer in order to be included in a block.  It seeks to increase fee revenue by specifically increasing fees on low-frequency users during **periods of low usage**.  The main motivation here is that many small validators operate at a net loss during time of low price and low usage.  Secondarily, we can observe that many fee-payers are willing to pay more for block space than the network is attempting to charge them during periods of low usage.  This is observed by examining fees paid to other blockchains by casual users, as well as by the progressive fee structures of many CEXs.  Nothing about this proposal is intended to incentivize fee-payers to seek lower times of usage, it simply requires that they pay higher fees/tx for infrequent usage.

-SIMD-003 by contrast seeks to raise the **floor** price of fees that must be paid by a fee payer in order to be included in a block during **period of high usage**.  The main motivation here is as you describe, to incentivize users who do not require immediate block space to seek times that are more cost effective.

Ideally, these two SIMDs would work together.  The following simple charts illustrate the expected fee a fee-payer must pay for block inclusion during low, moderate, and high usage using a combination of these SIMDs:

image

IntokuSatori:
> This pull request seeks to raise the ceiling of fees that must be paid by a fee payer in order to be included in a block. It seeks to increase fee revenue by specifically increasing fees on low-frequency users during periods of low usage.

I'm confused by this wording: "ceiling" would imply a max fee limit. This seems to say you're increasing the **minimum fee**, which implies a floor fee. As I read it, it seems this proposal sets the floor fee at times of low-usage, and SIMD-0003 sets the floor fee at times of high usage. 

Is this an incorrect interpretation? In my reading and by the Y-axis of the graph, it appears this and SIMD-0003 are setting the floor fee, just this proposal targets times of low usage and SIMD-0003 targets times of high usage.

betaconger:
@IntokuSatori This is a valid criticism.  I've updated the wording of my comment to reflect your thoughts.  I was hesitant to use the term "floor" for my comment, as I've called it the "Required Minimum Fee" in the SIMD, but it is really just a higher floor in times of low usage for infrequent users.

Your interpretation is correct.

aeyakovenko:
This would require the state to keep track of additional memory for each fee payer - the last slot used.  Which is pretty expensive. U64 for every account, or making account meta variable sized, which impacts perf.  

Also, I am not sure why there is any benefit from excluding less frequent users if they are willing to pay a higher priority fee.

bji:
I feel that this is an interesting proposal in abstract but I do not personally feel convinced that infrequent users should pay higher fees.  Also I think that any extra complexities added to how fees are computed needs to be very heavily scrutinized and it's hard to justify adding such complexity to increase fees for infrequent users.

jacobcreech:
Hello. This SIMD has largely become [stagnant](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0001-simd-process.md#stagnant) as part of SIMD-1. Being that it is still in the idea stage, feel free to discuss on core-technology on the [Solana-tech discord](solana.com/discord). If the SIMD gets fleshed out a bit more and has a chance of reaching consensus, feel free to open a new PR in the future.

PR #10, opened 2022-12-14
Add SIMD-10: Diet clients

aeyakovenko:
#### Problem

Users that do not run full node need a more secure way of trusting solana blockchain confirmations.

#### Solution

Propose a protocol for sampling validators such that if a minority of validators aren't faulty, they can notify the user that the majority confirmed an invalid state transition or a double spend.

### DIFF
--- a/proposals/0010-diet-clients.txt
+++ b/proposals/0010-diet-clients.txt
@@ -116,3 +116,65 @@ the staked validators.
 If the validator signed an invalid ProgramStateHash, this signed
 attestation can be used to slash them after the EpochHash has been
 computed by the network.
+
+### Future work: Minimized fraud proof
+
+Majority fraud is detectable when the majority has signed an invalid
+state transition. Currently, the above proposal requires a client
+to validate the entire chain starting from a previously trusted
+snapshot up to the invalid block.
+
+#### Overview
+
+1. Track all the inputs into state transitions
+
+Each input should have a merkle path to the original snapshot, or
+if it has been modified, to an output from a previous transaction.
+
+2. Merkelize all the txs and their inputs and outputs for each block
+
+       StateHash
+     h1       h2
+
+  tx1  tx2  tx3 tx4
+
+tx:
+inputs: path to snapshot | path to previous block output | path to intra-block tx output
+result: success or failure code
+outputs: hash of the result if result is successful, 0 otherwise
+
+The tree should contain an ordered sequence of txs. Each tx is its
+own tree that contains all the input paths to the accounts used for
+execution and all the outputs. Each input can either be an
+output from a snapshot root, or an output from a previous block,
+or from a transaction in the same block.
+
+3. User receives a path for their confirmed TX along with the
+majority of votes signing the root.  An honest node will generate
+a path for the same TX when processing the same block.  Either
+inputs will be different, or the outputs.
+
+4. If the input is different, user will need to recursively request
+a confirmation for the transaction that generated the input.
+
+5. If the output is different, user will need to run the program
+locally and confirm the result.
+
+Eventually user will be able to confirm the invalid state transition
+locally. The amount resources necessary to execute the confirmation
+should be on the order of executing 1 transaction, which generally
+fits in a mobile browser sandbox.
+
+#### Storage
+
+Upon storing the output of a transaction, AccountsDB should also
+record the merkle path of that output to the transaction. At the
+end of each block, each transaction in ledger order is merklelized
+into the StateHash.
+
+#### Retrieval
+
+When retrieving the inputs for execution, each input should have
+been tagged with a path to the transaction that generated it. That
+path, along the path to the root of the block that generated the
+transaction can be retrieved concurrently during replay.
--- a/proposals/0010-diet-clients.txt
+++ b/proposals/0010-diet-clients.txt
@@ -141,9 +141,10 @@ if it has been modified, to an output from a previous transaction.
   tx1  tx2  tx3 tx4
 
 tx:
-inputs: path to snapshot | path to previous block output | path to intra-block tx output
-result: success or failure code
-outputs: hash of the result if result is successful, 0 otherwise
+    inputs: all read account paths: path to snapshot | path to previous
+            block output | path to intra-block tx output
+    result: success or failure code
+    outputs: all write accounts if result is successful, 0 otherwise
 
 The tree should contain an ordered sequence of txs. Each tx is its
 own tree that contains all the input paths to the accounts used for
--- a/proposals/0010-diet-clients.txt
+++ b/proposals/0010-diet-clients.txt
@@ -25,13 +25,13 @@ state transition or a double confirmation attack on the network.
 contains 2/3+ votes for slot N, and a path from the transaction to
 the bankhash signed by all the votes.
 
-2. Concurrently, user requests at random votes from N different
-nodes. Votes maybe for later slots, and would contain a path to the
-requested slot's bankhash.
+2. Concurrently with 1, user requests votes for slot N from M random
+validators.  Votes returned may be for later slots, and would contain
+a merkle path to the requested slot's bankhash.
 
 3. If all the responses are the same, the user is done. 
 
-4. If X% or more of staked nodes responds that they do not have block,
+4. If X% or more of staked nodes respond that they do not have block,
 the user starts sampling nodes at random for missing shreds and
 submitting the shreds to the validators with the missing block.
 
@@ -43,7 +43,8 @@ submitting the shreds to the validators with the missing block.
 
 5. If X% or more of the staked nodes respond with different
 bankhash, a fault has occurred on the network. User shouldn't
-trust the confirmation.
+trust the confirmation. User should then submit the conflicting
+votes to M random validators.
 
 ### Confirmation Proof
 
--- a/proposals/0010-diet-clients.txt
+++ b/proposals/0010-diet-clients.txt
@@ -27,7 +27,10 @@ the bankhash signed by all the votes.
 
 2. Concurrently with 1, user requests votes for slot N from M random
 validators.  Votes returned may be for later slots, and would contain
-a merkle path to the requested slot's bankhash.
+a merkle path to the requested slot's bankhash. M should be large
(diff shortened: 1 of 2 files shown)


mryalamanchi (commented):
@jacobcreech please update the file to .md rather than .txt

anoushk1234:
A few clarifying questions/comments:
- The protocol interaction could go two ways:
	- The honest minority reports missing shreds / invalid blocks to the diet client which then verifies that
	- The diet client just observes the validator consensus after a transaction and if the superminority disagrees then the diet client samples the random shreds
- Do we sample for missing shreds or random shreds because here two things can happen:
	- Shreds are missing as in withheld by the majority which means that there is most certainly a problem because the honest supermajority wouldn't hide data if it were honest. Yes packet loss could cause missing shreds but DA sampling would check that anyway because we only need a small sample.
	- Diet client can request x amount of random shreds from the supermajority and then check if the proof in each shred computes to the root.
- For the implantation when the validator has to send random shreds to the client we thought of two ways we could go about it:
	- We can recompute shreds from the block in bank because it is deterministic
	- We could store a random set of the shreds for all blocks for the latest epoch and send them on request but this is probably a lot of extra work for no real benefit
	So I'm thinking the first is better

proposals/0010-diet-clients.txt
```diff
@@ -0,0 +1,183 @@
...
+
+### Overview
+
+1. User observes a confirmation for a TX at slot N. Confirmation
+contains 2/3+ votes for slot N, and a path from the transaction to
+the bankhash signed by all the votes.
+
+2. Concurrently with 1, user requests votes for slot N from M random
+validators.  Votes returned may be for later slots, and would contain
+a merkle path to the requested slot's bankhash.
```
valerini, line 30: How much smaller do you envision M to be compared to 2/3*N? Is it to be decided, or do you have some rough estimates?

valerini, line 30: If the slot's bankhash is the same across the requests, the correctly-verifiable merkle paths will also be the same (since it is computationally impossible to create two different merkle-paths leading to the same root-hash). Perhaps to reduce redundancy, you might NOT want to ask all of M validators to send merkle paths, since one path would be enough.

proposals/0010-diet-clients.txt
```diff
@@ -0,0 +1,183 @@
...
+
+3. If all the responses are the same, the user is done. 
+
+4. If X% or more of staked nodes respond that they do not have block,
+the user starts sampling nodes at random for missing shreds and
+submitting the shreds to the validators with the missing block.
```
valerini, line 36: Is the threshold X here yet to be determined?
Is it assumed here, that the block is confirmed (got 2/3+ votes), but still a substantial portion of validators (X < 1/3) does not have it? I think it's a really good idea to make users help the validators at obtaining the missing shreds!

aeyakovenko: Threshold is configured by the client.  It's a tradeoff on liveness vs safety.  If the client sets it to 1% it's more likely that 1% of the network happens to go down right in that moment.

proposals/0010-diet-clients.txt
aeyakovenko: 5% of stake is enough imho

proposals/0010-diet-clients.txt
```diff
@@ -0,0 +1,183 @@
...
+output from a snapshot root, or an output from a previous block,
+or from a transaction in the same block.
+
+3. User receives a path for their confirmed TX along with the
+majority of votes signing the root.  An honest node will generate
+a path for the same TX when processing the same block.  Either
+inputs will be different, or the outputs.
+
+4. If the input is different, user will need to recursively request
+a confirmation for the transaction that generated the input.
```
aeyakovenko, line 161: or a validator can do this for the user.

proposals/0010-diet-clients.txt
```diff
@@ -0,0 +1,183 @@
...
+    a. This process may need to run recursively until the fork block
+    is repaired.
+
+    b. If shreds cannot be recovered, a fault has occurred. User
+    shouldn't trust the confirmation.
+
+5. If X% or more of the staked nodes respond with different
+bankhash, a fault has occurred on the network. User shouldn't
+trust the confirmation. User should then submit the conflicting
+votes to M random validators.
```
valerini, line 47: If the user observed a confirmation (2/3+ of votes) and is not trusting it, does the user in this case conclude that there are more than 1/3+ byzantine nodes?

proposals/0010-diet-clients.txt
```diff
@@ -0,0 +1,183 @@
...
+
+This tree is necessary for step 4 in the protocol. If enough nodes
+respond that they do not have the data, the client should try to
+repair the data for them. This prevents the faulty majority from
+withholding the data from the rest of the network, so the rest can
+identify an invalid state transition or a double confirmation.
+
+Turbine, solana's block propagation protocol, transmits blocks in
+small chunks called shreds. Shreds are transmitted in batches.
+Each batch includes N data shreds and M erasure codes.
```
valerini, line 65: I guess, these N and M are different from the above, where N is a slot number and M is the number of random validators. May be you could use small n and m here to differentiate.

proposals/0010-diet-clients.txt
```diff
@@ -0,0 +1,183 @@
...
+    inputs: all read account paths: path to snapshot | path to previous
+            block output | path to intra-block tx output
+    result: success or failure code
+    outputs: all write accounts if result is successful, 0 otherwise
+
+The tree should contain an ordered sequence of txs. Each tx is its
+own tree that contains all the input paths to the accounts used for
+execution and all the outputs. Each input can either be an
+output from a snapshot root, or an output from a previous block,
+or from a transaction in the same block.
```
valerini, line 153: Should the user also be checking some non-inclusion proofs, to make sure no transaction has modified the input in-between? For example, if the user is given the input that is an output from a snapshot root, should they confirm there were no other transactions modifying the input between the snapshot and the current transaction?

aeyakovenko: @valerini, An honest validator would have computed a different input using a more recent value and path for that TX, so the result would have been different.  The honest validator can then return the more recent path and the majority signed header as the proof.

proposals/0010-diet-clients.txt
```diff
@@ -0,0 +1,205 @@
...
+Each batch includes J data shreds and K erasure codes.
+
+Upon executing the block, the validator should compute the merkle
+root of all the shreds, including the erasure shreds. The merkle
+root should be added to the BankHash, which is signed by validators
+when voting.
+
+When sampling for shreds, the user knows that the shreds are for
+the proposed confirmed block because each shred has a merkle path
+to the bankhash that was signed by all the validators.
```
anoushk1234, line 87: Don't the validators sign the merkle root generated by merkelizing all the shreds https://github.com/solana-labs/solana/blob/9db25655f72dc3cd4362b9ef077f09054573ccee/ledger/src/shred/merkle.rs#L1119
which is diff from the bank hash https://github.com/solana-labs/solana/blob/283a2b15402867baeb0735acda1cdc8632afd400/runtime/src/bank.rs#L6714-L6719

blasrodri:
This looks quite interesting. I’d like to understand why not building a protocol in which non full nodes can validate state tránsitos like Light Clients do. This is truly useful for trustless bridges (using IBC for example).

anoushk1234:
> This looks quite interesting. I’d like to understand why not building a protocol in which non full nodes can validate state tránsitos like Light Clients do. This is truly useful for trustless bridges (using IBC for example).

Diet Clients are non full nodes. Validating state transition is on the roadmap and part of the future work.

proposals/0010-diet-clients.txt
aeyakovenko: Yea, this would need to change.  We need the validators to sign the merkle root as well.

hydrogenbond007:
yo @anoushk1234 would be cool to share about tinydancer here just to keep the issue uptodate

anoushk1234:
> yo @anoushk1234 would be cool to share about tinydancer here just to keep the issue uptodate

yep we're implementing this with @tinydancer-io, once we release the whitepaper and v0 code we will put an update here, thanks @hydrogenbond007

hydrogenbond007 (approved):
Tinydancer LFG!

proposals/0010-diet-clients.txt
```diff
@@ -0,0 +1,205 @@
...
+## Summary
+
+Users that don't run a full node need secure access to the solana
+blockchain. The goal of this proposal is to define a protocol for
+sampling the validators such that if at least a small minority of
+validators are not faulty the user can be notified of an invalid
+state transition or a double confirmation attack on the network.
+
+## Specification
+
```
hydrogenbond007, line 21: Might wanna update this

anoushk1234:
All active development for the diet client can be followed at [tinydancer-io/half-baked-client](https://github.com/tinydancer-io/half-baked-client) and we have release the POC for v0

anoushk1234:
Hey @aeyakovenko had a small doubt about receipts being included in bank hash 
https://github.com/solana-labs/solana/blob/9281ab7d9778b23775d8de2415c48eb34cd92975/entry/src/entry.rs#L247

Screenshot 2023-05-26 at 11 19 56 AM

anoushk1234:
Also the SPV proposal which was accepted describes this but not sure if it was fully implemented
SPV Block Merkle

jacobcreech:
Hey @aeyakovenko, this [SIMD is considered stagnant](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0001-simd-process.md#stagnant) at this point with the majority of recent comments pointing to TinyDancer team's recent work.

Is this still something you'd want to push for acceptance or shall we close in favor of Tinydancer's future SIMDs?

0xSol:
This SIMD has been inactive for 6 months. The status changed to Stagnant per [current SIMD process](https://github.com/0xSol/solana-improvement-documents/blob/main/proposals/0001-simd-process.md).  Close it for now, but feel free to reopen with new updates if still relevant, or create a new SIMD as needed.

PR #14, opened 2022-12-17
Added some more details

codingshot:
This was inspired by EIP + NEP

https://github.com/ethereum/EIPs
https://github.com/near/NEPs


jacobcreech:
Hello @codingshot. I appreciate the contribution to the readme, but I don't think we're following exactly the EIP or NEP path. For example, we're not including token API standards in SIMDs as they are not relevant to the underlying protocol.

We'll update separately to define the different path we're taking.

codingshot:
Looking forward to it and watching this come to fruition

PR #32, opened 2023-01-30
The XXXX-template.md file is not in the proposals directory, so the

bji:
documentation should not refer to it as being there.


jacobcreech (approved):
Thanks for the update!

PR #33, opened 2023-01-30
Added timely vote credits proposal.

bji:


lurais:
hello , will the leader record the vote time of the validators? What if the leader is not honest and skip some validator's vote or   not record the validator's vote in time? @bji

bji:
> hello , will the leader record the vote time of the validators? What if the leader is not honest and skip some validator's vote or not record the validator's vote in time? @bji

The leader could already do that now (refuse to accept votes).  However before timely vote credits, the validator whose vote was ignored can just re-send it and assuming that some future leader accepts it, full credit would be earned.

With timely vote credits, it is indeed possible to penalize a voter by not accepting their vote because even if it does get accepted by a future leader, it would now have higher latency and earn fewer credits.

However, the incentives are not aligned for validators to skip votes like this.  A validator who refuses to accept a vote will not earn the transaction fee for that vote.  So a validator would reduce their own income when deferring votes in this way.  And the effect that a validator can have on another validator's overall earnings is fairly minute.

So technically, yes, a validator could skip an "enemy" validator's votes to attempt to reduce that enemy validator's credits.  However, this has a cost to the skipping validator (lost income) and has only a marginal effect on the credits earned by the skipped validator, as a single vote being reduced in credits earned is very minor.

brianlong:
There is some low-hanging fruit to pick before taking this on. If we get close to #ZeroForks, bad behavior (vote lagging) will not be a factor. This is a big change, and the ideal solution might be different after fixing other things.

I'm not down-voting, but I would prioritize the vote-sending & replay patches above this. The chain might look different after they are complete.

bji:
> There is some low-hanging fruit to pick before taking this on. If we get close to #ZeroForks, bad behavior (vote lagging) will not be a factor. This is a big change, and the ideal solution might be different after fixing other things.
> 
> I'm not down-voting, but I would prioritize the vote-sending & replay patches above this. The chain might look different after they are complete.

Hopefully we can live the dream and do more than one thing at a time.  We'll see.

proposals/0033-timely-vote-credits.md
```diff
@@ -0,0 +1,389 @@
...
+uses the Clock sysvar to identify the slot in which that instruction has
+landed.  For any newly voted on slot within that Vote or VoteStateUpdate
+transaction, the validator will record the vote latency of that slot as
+(voted_in_slot - voted_on_slot).
+
+These vote latencies are stored in u8 field of the Lockout data structure
+named *latency*.
+
+The credits awarded for a rooted slot are calculated using the latency value
+stored in the latency field the slot, using a formula that awards latencies
```
AshwinSekar, line 148:
```suggestion
for the corresponding vote, using a formula that awards latencies
```

proposals/0033-timely-vote-credits.md
```diff
@@ -0,0 +1,389 @@
...
+of 1 - 2 slots 8 credits, with a 1 credit reduction for each vote latency
+after 2.  Rooted slots will always be awarded a minimum credit of 1 (never
+0) so that very old votes, possibly necessary in times of network stress,
+are not discouraged.
+
+The implementation of the above is fairly simple for Vote transactions,
```
AshwinSekar, line 154: The remaining part of this section is not needed as `rooted slot` is well defined for both Vote and VoteStateUpdate instructions and was designed as part of the initial specs/proposal. Just makes it more complex re-explaining the rooting mechanism already present in the vote program.

proposals/0033-timely-vote-credits.md
```diff
@@ -0,0 +1,389 @@
...
+this).
+
+The existing VoteStateUpdate processing code already ensures that (2) new
+slots included in the new VoteState are only slots after slots that have
+already been voted on in the old VoteState (i.e. can't inject new slots in
+the middle of slots already voted on).
+
+## Impact
+
+1. Any code which deserialized vote account state will require a recompile
```
AshwinSekar, line 199: I think instead for this bullet point you should outline the feature flag for upgrading the new vote state format (and implications on any rent balance) as well as the non impact on the tower.

proposals/0033-timely-vote-credits.md
```diff
@@ -0,0 +1,389 @@
...
+   or other tooling will need to adapt to the additonal data being
+   displayed.
+
+## Security Considerations
+
+The are no security considerations; however there are changes to incentive
+structure for voting which may result in differing vote behavior becoming
+more optimal with regards to earning credits.
+
+To maximize vote credits, a validator must vote on all slots that end up
```
AshwinSekar, line 218: This paragraph and the 2 under it would be better in the motivation section, not really security considerations.

proposals/0033-timely-vote-credits.md
```diff
@@ -0,0 +1,389 @@
...
+way, perfect vote credits would be earned.  However, this optimal voting
+strategy, if employed cluster-wide, would halt the cluster as no validators
+would vote, all of them waiting for others to vote first.
+
+It's possible to get very close to this optimal strategy by waiting for some
+number of confirmations less than 31, with the number of incidents of
+picking the wrong fork increasing as the waiting time is decreased.  The
+current "vote laggers" have done this, with varying degrees of skill.
+
+After the timely vote credits change, the dynamics will change; it will no
```
AshwinSekar, line 238: These three paragraphs would be better in the impact section, the impact on the new optimal voting strategy.

proposals/0033-timely-vote-credits.md
```diff
@@ -0,0 +1,389 @@
...
+
+   Once this feature is enabled, vote accounts will earn credits based on
+   their vote latencies.  The maximum possible credits earnings per epoch
+   will increase from 432,000 to 3,456,000.  Each validator will continue to
+   earn vote rewards based on their fraction of total vote credits earned
+   across the cluster, and validators which earn more credits through
+   reduced vote latencies will earn a larger percentage of the total
+   rewards.
+
+## A Note About the Tower
```
AshwinSekar, line 351: I think this section can be folded into the previous, talking about the backwards compatibility of the tower.

proposals/0033-timely-vote-credits.md
AshwinSekar: Actually I see you talked about it down below, maybe just a note here that the on chain vote state format will change and see the `Backwards Compatibility` section for the upgrade procedure.

proposals/0033-timely-vote-credits.md
bji: Also it allows me to refer to existing implementation to give assurance that the melding of the new implementation into the old won't cause problems, and to describe why that is true.  But again I will remove it.

proposals/0033-timely-vote-credits.md
bji: I put that discussion under Security Considerations because are people who believe (and have said so in public discussion) that voting incentives can impact security; so here I lay out the case for why the new voting incentives do not have security implications.

proposals/0033-timely-vote-credits.md
bji: Again the "Security Considerations" section was meant to specifically address issues that have already been brought up several times in validator meetings to discuss this feature; having identified that as a point that people generally have concern about, I thought it best to address that topic very clearly here.

proposals/0033-timely-vote-credits.md
bji: It can be, but why?  It's meant to provide documentation about a specific and complex and confusing aspect to the implementation.

AshwinSekar (dismissed):
lgtm

AshwinSekar:
@jacobcreech looking to get this merged, can you advise on the next steps?

jacobcreech:
> @jacobcreech looking to get this merged, can you advise on the next steps?

Let me reach out to the other client teams for some additional review!

carllin (dismissed):

jacobcreech:
Going to give it until Friday on merging if no comments are made cc @ripatel-jump

bji:
> Going to give it until Friday on merging if no comments are made cc @ripatel-jump

Has that Friday passed yet?

bji:
> Going to give it until Friday on merging if no comments are made cc @ripatel-jump

Has that Friday passed yet?

jacobcreech:
> Has that Friday passed yet?

Yes, but I was really trying to get secondary review. We may have to implement a `Last Call` stage in the SIMD process like EIP has to get acceptance better. I'll bring up the to review SIMDs at the beginning of this week's core community call.

bji:
> Going to give it until Friday on merging if no comments are made cc @ripatel-jump

Has that Friday passed yet?

proposals/0033-timely-vote-credits.md
```diff
@@ -0,0 +1,342 @@
...
+The implementation of the above is fairly simple for Vote transactions,
+because Vote transactions include slot numbers voted on, and that list of
+slots is already pruned of already-voted-on slots; so each newly voted-on slot
+that survives this pruning can have its latency calculated simply.
+
+
+## Impact
+
+1. Any code which deserialized vote account state will require a recompile
+   with the newest Solana SDK in order to be able ot deserialize the new vote
```
apfitzge, line 151: typo:

```suggestion
   with the newest Solana SDK in order to be able to deserialize the new vote
```

proposals/0033-timely-vote-credits.md
```diff
@@ -0,0 +1,342 @@
...
+   latency field, which will result in the minimum vote credits awarded for
+   each vote for that vote account.  This provides economic incentive to
+   validators to sufficiently fund their vote accounts to allow the updated
+   VoteStateVersion to be serialized into them.
+
+   To summarize the issues surrounding the Vote account updates:
+
+   - A vote account will only be updated to the new version after the feature
+     is enabled, and only if the vote account includes enough SOL to satisfy
+     the new rent exempt minimum.  - If a vote account is not updated, then it
```
apfitzge, line 282: is this intended to be 2 bullet points?

proposals/0033-timely-vote-credits.md
```diff
@@ -0,0 +1,342 @@
...
+proposed there ("4 16 1", see later in this document for a description of what
+these numbers mean) has been long since superceded by different values ("2 8
+1").  This is because the overall performance of the Solana mainnet cluster
+has matured to a degree where "tighter timings" are tolerable.
+
+## New Terminology
+
+**Vote Latency** is defined as the difference, in slots, between the slot
+being voted on within a vote transaction, and the slot in which the vote
+transaction itself lands.  The minimum possible vote latency for any voted-on
```
apfitzge, line 60: > and the slot in which the vote transaction itself lands

I have some concern about this w.r.t. the proposed grace period. Currently, the leader scheduler gives leaders 4 slots at a time to be leader. The leader chooses which vote transactions get packed into those blocks.

As a malicious operator, I now have some incentive to not pack competing validator's votes; by delaying them, it reduces their rewards, and makes my validator appear more attractive to potential stakers.

I think with the current system, there's no real incentive to do this because it doesn't matter if I don't pack them, the next leader will and they'll get the same reward. And now I've only lost out on the transaction fees myself. 

I wonder if the grace period should be at least 4 slots to avoid this? Even so, I can imagine delaying processing votes until my 4th slot for the same reason while still earning the fees.

Haven't though through it entirely, but a potential way to address this is by making vote credits based on slot latency mod 4 (subject to change if leader schedule does).

proposals/0033-timely-vote-credits.md
bji: Will fix, got messed up in recent auto-fill

proposals/0033-timely-vote-credits.md
bji: Will fix, thanks.

proposals/0033-timely-vote-credits.md
bji: In doing so validators will be taking a risk of loss of block rewards: deferring any transactions until a later block will miss out on block rewards if those later blocks are skipped.  In addition, deferring votes to later blocks increases the chances that those blocks will be skipped as votes which could have landed earlier to "strengthen" the fork will land later instead, giving greater chance for a competing fork to win.

Both of these risks are anti-incentive for the validator to intentionally defer votes.

The effect that a validator can have on its "competitors" by deferring votes is very minor: a validator only gets approximately its stake weight in slots per epoch.  I don't think this has ever risen much above 3% since the early days of mainnet-beta, and anything above 2% is pretty rare.

Practically speaking this means that a validator could at worst reduce about 2% of a competitor's votes by 2 slots, or assuming an average of 7.5 credits per slot, a reduction from 7.5 to 5.5 credits for 2% of the target validators slots.  This is a reduction in 0.5% of credits -- and at the cost of greater risk of lost block rewards to the offending validator.

This issue was discussed in other forums previously and the conclusion was that it's not an issue large enough to counterbalance the benefits of the proposal.

But I'd be happy to hear your further thoughts on this.

ripatel-fd (dismissed):

proposals/0033-timely-vote-credits.md
AshwinSekar: > In addition, deferring votes to later blocks increases the chances that those blocks will be skipped as votes which could have landed earlier to "strengthen" the fork will land later instead, giving greater chance for a competing fork to win.

Adding to this: 

Votes for blocks on different forks cannot land (they will fail as invalid tx in vote program), so a leader can only censor votes for their own fork - which as zan mentioned hurts their chance of being selected in the heaviest fork calculation. I imagine a competing validator would have a sizable stake so this disincentives such behavior. 

At best a leader could censor tiny validators which would have a minimal reward for loss of tx fees. Based on the numbers zan cited this seems unlikely to be beneficial. 

I think if any such adversarial block packing strategy emerges successfully the best way to counteract it would be to tweak the ratio of block rewards/vote tx fees/vote credits

proposals/0033-timely-vote-credits.md
AshwinSekar: SIMDs are client agnostic, any references to the Solana Labs implementation are better suited for the PR in that repo

AshwinSekar: SIMDs are client agnostic, any references to the Solana Labs implementation are better suited for the PR in that repo

AshwinSekar (changes requested):

bji:
I removed the Note about the Tower.

AshwinSekar (approved):

AshwinSekar:
@jacobcreech looks like we have approval from firedancer as well, could we get this merged?

jacobcreech (approved):
Looks like we got general consensus on this. Thank you @bji for championing it!

diman-io:
Original on Discord https://discord.com/channels/428295358100013066/489504501749907468/1205857493692186684

Suggests using block height differences, not slot differences, to calculate voting lag, using sysvar. The idea is that during different network activities (calm times with linear forks vs. stormy times with many parallel forks and slot misses), the current method of slot difference makes intentional lagging less penalized during stormy periods. This hasn't been practically tested.

To clarify, I'm not advocating to pause or delay the release of the current implementation.

bji:
> Original on Discord https://discord.com/channels/428295358100013066/489504501749907468/1205857493692186684
> 
> Suggests using block height differences, not slot differences, to calculate voting lag, using sysvar. The idea is that during different network activities (calm times with linear forks vs. stormy times with many parallel forks and slot misses), the current method of slot difference makes intentional lagging less penalized during stormy periods. This hasn't been practically tested.
> 
> To clarify, I'm not advocating to pause or delay the release of the current implementation.

OK to bring my contribution from Discord into here:

If there's a fork, then validators who don't vote on it will necessarily be voting with some latency.  Let's say slot A was a slot on the fork, then some fork B, C, D, E was created, and some validators voted on that, but the fork that ends up getting finalized is F chained off of A.  OK so validators who didn't go down B, C, D, E will be voting on A in F at the earliest.  That's a slot latency of 5.

Now let's take the lagging validator who votes on A in G instead.  That's a slot latency of 6.

Without TVC, both voters earn the same, so the slower voter isn't penalized.

With TVC, the F voters earn 4 credits and the G voters earn 3 credits.

This is a reduction in credits earned in this scenario for the G voters of 25%.

Now let's say instead that TVC used block distance instead of slot distance when determining latency.

If this were the case, the latency for the F voters would be 1 and the latency of the G voters would be 2.

These would both be below the grace period threshold so both would earn 8 credits.  So no disincentive would be applied to the slower voter at all.

OK let's look at even slower voters.  Those that vote on A in J (and assuming that the fork ended up as: A - F - G - H - I - J).

Without TVC, again no penalty for the J voters.

With slot-based TVC, the F voters still earn 4, but the J voters (with slot latency 9) would earn 1.  That's a 75% reduction in credits for the slower J voters.

But with block-based TVC, the F voters would earn 8, but the J voters with block-latency 5 would earn 5 credits.  That's a reduction in credits of only 31.25%,  So half the penalty with block distance as would occur with slot distance. 
 
In both cases, TVC with block distance is less disincentivizing to slower voting than TVC with slot distance.

jasonslagter (commented):
[]()

proposals/0033-timely-vote-credits.md
nguyenvanthanh190892-web: Ok

nguyenvanthanh190892-web: Thanhbeu92

nguyenvanthanh190892-web: BqLmcBEGdEkV3e7tR7s8Yn1GkpSu1EH93W8jvfwC5GxG

proposals/0033-timely-vote-credits.md
nguyenvanthanh190892-web: Ok

PR #45, opened 2023-03-29
SIMD: (APExB) Asynchronous Program Execution and Broadcast

aeyakovenko:
### Problem

Transmitting blocks of user transactions and executing those transactions blocks fork choice. Only a single leader can propose user transactions at the moment, which means that client latency is going to be on average half way around the world.

### Solution

Propose a design to split out user transactions into separate user blocks that are transmitted concurrently with blocks of consensus votes.  This naturally follows into asynchronous execution of the user blocks.


proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,148 @@
...
+* UserBlock - a block full of none vote transactions.
+
+* UserBlockEntry - a hash of the block that was proposed
+
+* UserBlockSlot - a slot measuring half a normal slot in ticks.
+
+* Leader - the current leader for the slot that will propose a PoH
+ledger full of Votes and UserBlockEntry
+
+* Builder - a node that is scheduled to propose a block with non
```
buffalu, line 47: proposer?

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,148 @@
...
+number of builders can be scheudled per slot.  Builder are
+scheudled at 2x the rate of leaders.  So for every leader slot,
+there are N builders producing 2 UserBlocks.
+
+While a leader is scheduled, they receive and encode votes as normal.
+Any well formed UserBlocks that were received from the previous
+UserBlockSlot are added to the leaders PoH ledger as UserBlockEntry.
+
+The N concurrent Builder have 200ms slots to create blocks out
+of user transactions. These are transmitted to the network.  The
```
buffalu, line 66: only transmitted to the current proposer? or all nodes?

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,148 @@
...
+leader receives and decodes them and generates a UserBlockEntry,
+and adds it to PoH as soon as the leaders PoH has passed the
```
buffalu, line 68: how do they determine which one to add? is this in the world of bankless leader (leader isn't executing transactions, just doing poh recording + shred signing?)

buffalu: what if multiple UserBlockEntry contain same tx?

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,148 @@
...
+UserBlockSlot.
+
+### Fork Choice
+
+If a validator doesn't have the builder's UserBlock, the validator
```
buffalu, line 73: im confused where they repair from here if they're the only one getting the UserBlock. unless its propagated to the entire cluster somehow

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,148 @@
...
+doesn't vote on the proposed block and tries to repair the UserBlock.
+That fork is ignored for fork choice.
+
+Otherwise the validator evaluates the forks as usual. 
+
+### UserBlock execution
+
+Each UserBlock is assumed to have been created simultaneously
```
buffalu, line 81: a little confused in this paragraph; is the builder hashing their block with PoH? if the leader processes userblocks at the 200ms mark, why does the userblock need poh? can the userblock just be submitted with a poh in the past to get in the block earlier, which means all userblock submitters will not advance poh to game the system?

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,148 @@
...
+## Impact
+
+Multiple nodes can operate as Builder on the network concurrently.
+So clients can pick the nearest one, and the bandwidth to schedule
+and prioritize transactions is doubled.  There needs to be a design
+for BundleTransactions that allow the bundler to prioritize a batch
+of transactions for execution by paying a priority fee for all of
+them, and executing the whole batch together.
+
+Priority fees now also imply execution priority.
```
buffalu, line 123: LFG

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,148 @@
...
+been optimistically confirmed. Client can also wait for a status
+code weighted by stake weight, as each validator's execution catches
+up to the slot and includes a BankHash confirming the state of the
+execution.
+
+It may be sufficient to wait for 1/3+ validators executing the state
+transition instead of the full 2/3+, because if 1/3+ are incorrect
+the network will halt anyways.
+
+## Impact
```
buffalu, line 114: looks like we might want to include a bundle primitive in here too to avoid splitting up bundles that get reordered/not guaranteed to execute atomically. can imagine bundle like:
[user_high_priority_tx, arb_low_priority_tx]

buffalu:
thoughts:
- this is probably my favorite proposal so far.
- i like this proposal much more for the global centralization issues around solana. 
- the fact that state is **always changing** on solana means colo means **a lot**.  its good for throughput, but really encourages colo so you can ingest shreds -> replay -> geyser notif -> send mev tx as fast as possible.
- assuming this proposal suggests that state is paused for ~x ms (200ms in the proposal, could prob crank down even more to like 50ms), this would mean that someone could be semi-competitive doing mev from across the globe (latency will never not matter).

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,148 @@
...
+
+While a leader is scheduled, they receive and encode votes as normal.
+Any well formed UserBlocks that were received from the previous
+UserBlockSlot are added to the leaders PoH ledger as UserBlockEntry.
+
+The N concurrent Builder have 200ms slots to create blocks out
+of user transactions. These are transmitted to the network.  The
+leader receives and decodes them and generates a UserBlockEntry,
+and adds it to PoH as soon as the leaders PoH has passed the
+UserBlockSlot.
```
buffalu, line 69: this is also nice bc you can spend more time scheduling ahead of time and massively parallelize the execution of these transactions.
- sort by priority
- DAG
- batch
- execute

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,148 @@
...
+doesn't depend on program execution, and delaying voting creates
+forks and also delays the next block producer from starting.
+
+## Alternatives Considered
+
+APEX design that was proposed in https://github.com/solana-labs/solana/pull/24127
+
+## New Terminology
+
+* UserBlock - a block full of none vote transactions.
```
buffalu, line 38: to keep terminology, does RecordBatch or EntryBatch make more sense? this is essentially what the current validator sends to poh but in batch form? Vec<Vec<SanitizedTransaction>>

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,148 @@
...
+doesn't vote on the proposed block and tries to repair the UserBlock.
+That fork is ignored for fork choice.
+
+Otherwise the validator evaluates the forks as usual. 
+
+### UserBlock execution
+
+Each UserBlock is assumed to have been created simultaneously
+during its scheduled slot.  For each UserBlock, the transactions
+are ordered by priority fee before execution. If two transactions
```
eugene-chen, line 83: is this within a single UserBlock, or within the stage where multiple UserBlocks are merged?

proposals/0023-asynchronous-program-execution-and-broadcast.md
aeyakovenko: Isn't builder a better name? Proposer is too generic, since the leader also proposes a block

proposals/0023-asynchronous-program-execution-and-broadcast.md
aeyakovenko: UserBlocks are transmitted over turbine to the entire network

proposals/0023-asynchronous-program-execution-and-broadcast.md
aeyakovenko: Leader adds the UserBlockEntry to PoH as soon as it is received.  Since UserBlocks are transmitted over turbine, if a leader sees it, it's very likely the rest of the network has seen it as well.

proposals/0023-asynchronous-program-execution-and-broadcast.md
aeyakovenko: UserBlocks are sent over turbine. So repair and everything else works as is.

proposals/0023-asynchronous-program-execution-and-broadcast.md
aeyakovenko: UserBlocks don't need PoH, but follow the same ledger format with entries.  Fixed it in the update.

proposals/0023-asynchronous-program-execution-and-broadcast.md
aeyakovenko: Added the BundleTransaction

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,212 @@
...
+
+## Motivation
+
+1. Single leader for user transactions is a bottleneck.  Leaders
+are in a single spot in the world, while clients are all over the
+world. Latency is therefore based on how close a client is to a
+leader.  Leaders also have to spend a ton of resources on prioritization
+of transactions.
+
+2. Executing programs before voting is a bottleneck. Fork choice
```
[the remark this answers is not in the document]
aeyakovenko: Yes. Nodes need to be able to compute a snapshot once an epoch.

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,212 @@
...
+* UserBlock - a block full of none vote transactions.
+
+* UserBlockEntry - a hash of the block that was proposed
+
+* UserBlockSlot - a slot measuring half a normal slot in ticks.
+
+* Leader - the current leader for the slot that will propose a PoH
+ledger full of Votes and UserBlockEntry
+
+* Builder - a node that is scheduled to propose a block with non
```
segfaultdoc, line 47:
```suggestion
* Builder - a node that is scheduled to build a block with non
```

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,148 @@
...
+## Security Considerations
+
+* Validators may include incorrect BankHash values which may not
+be detected instantly. Validators may skip evaluating their own
+BankHash and copy results from other validators.
+
+* Network halts if it cannot compute a epoch snapshot hash once an
+epoch because it is overloaded with user transactions.
+
+* Leader could censor the builder. But then they would miss out
```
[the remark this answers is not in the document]
aeyakovenko: No, leader can only inject votes and UserBlocks from the scheduled Builders

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,148 @@
...
+Leaders are scheduled to propose blocks as they are currently by
+the LeaderSchedule.
+
+Builder's are scheduled along side leaders to propose blocks. N
+number of builders can be scheudled per slot.  Builder are
+scheudled at 2x the rate of leaders.  So for every leader slot,
+there are N builders producing 2 UserBlocks.
+
+While a leader is scheduled, they receive and encode votes as normal.
+Any well formed UserBlocks that were received from the previous
```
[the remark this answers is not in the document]
aeyakovenko: No, it's all or nothing.  UserBlocks are transmitted to the entire cluster over turbine by the Builder. So the only thing the leader can do is drop entire UserBlocks or reorder them.

proposals/0023-asynchronous-program-execution-and-broadcast.md
aeyakovenko: Within the stage when they are merged.

bji:
Might want to temper your enthusiasm.  This will make arbitrage *alot* more difficult since unless you can execute transactions faster than your competitors, you will be at a disadvantage.  They will know the "current state" but you will only know "state at some time in the past".  Trying to write arbitrage tx against old state seems like a really good way to waste priority fees.

bji:
> thoughts:
> 
>     * this is probably my favorite proposal so far.
> 
>     * i like this proposal much more for the global centralization issues around solana.
> 
>     * the fact that state is **always changing** on solana (shreds always being broadcasted) means colo means **a lot**.  its good for throughput, but really encourages colo so you can ingest shreds -> replay -> geyser notif -> send mev tx as fast as possible. its also hard AF to simulate things correctly in our block engine w/o significant RPC and replay modifications and will be even harder for others that don't have access to that :)
> 
>     * assuming this proposal suggests that state is paused for ~x ms (200ms in the proposal, could prob crank down even more to like 50ms), this would mean that someone could be semi-competitive doing mev from across the globe (latency will never not matter).

Might want to temper your enthusiasm. This will make arbitrage alot more difficult since unless you can execute transactions faster than your competitors, you will be at a disadvantage. They will know the "current state" but you will only know "state at some time in the past". Trying to write arbitrage tx against old state seems like a really good way to waste priority fees.

buffalu:
> This will make arbitrage alot more difficult since unless you can execute transactions faster than your competitors, you will be at a disadvantage

this will make arbitrage easier to be competitive, not harder.

proposals/0023-asynchronous-program-execution-and-broadcast.md
buffalu: a proposer proposes the block to the network that others vote on. in this case i think they'd be proposing the ordering?

proposals/0023-asynchronous-program-execution-and-broadcast.md
buffalu: ah ofc, need to fit in MTU-sized packets

proposals/0023-asynchronous-program-execution-and-broadcast.md
buffalu: i thought the leader was doing the ordering by poh?

proposals/0023-asynchronous-program-execution-and-broadcast.md
aeyakovenko: That is correct.  Before voting on a fork each validator has to have all the data, including all the data form all the user blocks in that fork.  Otherwise there is no way to guarantee that everyone can execute because the data could be withheld

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Alternatives Considered
+
+APEX design that was proposed in https://github.com/solana-labs/solana/pull/24127
+
+## New Terminology
+
+* UserBlock - a block full of none vote transactions.
+
+* UserBlockEntry - a hash of the block that was proposed
```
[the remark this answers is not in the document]
aeyakovenko: This is the entry for the UserBlock that goes into the Leaders PoH

proposals/0023-asynchronous-program-execution-and-broadcast.md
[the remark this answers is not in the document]
aeyakovenko: Validators can reject any leader blocks that contain other entries.  

TBD if a signature is necessary, because shreds are signed and validators will verify that the shreds are signed by the builder.

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,212 @@
...
+## Impact
+
+Multiple nodes can operate as Builder on the network concurrently.
+So clients can pick the nearest one, and the bandwidth to schedule
+and prioritize transactions is doubled.  There needs to be a design
+for BundleTransactions that allow the bundler to prioritize a batch
+of transactions for execution by paying a priority fee for all of
+them, and executing the whole batch together.
+
+Priority fees now also imply execution priority.
```
lheeger-jump, line 176: Can you elaborate? Why is that the case now?

lheeger-jump (commented):
It would be really awesome to get a state machine diagram for how clients are supposed to behave under this new scheme.

proposals/0023-asynchronous-program-execution-and-broadcast.md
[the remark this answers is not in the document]
aeyakovenko: that is also verified via turbine, but I'll add a note.

proposals/0023-asynchronous-program-execution-and-broadcast.md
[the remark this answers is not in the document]
aeyakovenko: Nodes can't fall behind that much because the overall CU limits are set for synchronous execution.  But with the option of async execution it is much easier to catch up.  Raw ledger processing without dealing with the network is 20-30x times faster.

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,212 @@
...
+doesn't vote on the proposed block and tries to repair the UserBlock.
+That fork is ignored for fork choice.
+
+Otherwise the validator evaluates the forks as usual. 
+
+### UserBlock execution
+
+Each UserBlock is assumed to have been created simultaneously during
+the UserBlockSlot that it was encoded by the leader. For each
+UserBlock, the transactions are ordered by priority fee before
```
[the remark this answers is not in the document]
aeyakovenko: I am not sure I agree.  The UserBlock will be limited in the amount of CUs it can take up, so the order of execution can be based on priority fee per CU.  CUs will take into account the cost of locks and reads and writes.  That is outside the scope of this design

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,212 @@
...
+for BundleTransactions that allow the bundler to prioritize a batch
+of transactions for execution by paying a priority fee for all of
+them, and executing the whole batch together.
+
+Priority fees now also imply execution priority.
+
+Fork choice and voting is not blocked by user block propagation or
+execution of user transactions.
+
+Builders are going to capture all the MEV.
```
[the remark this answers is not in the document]
aeyakovenko: How can the leader ignore a BundleTransaction?  It's part of the UserBlock, so the entire block must be accepted, including the BundleTransaction.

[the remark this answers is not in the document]
aeyakovenko: added some clarifications

bji:
> > This will make arbitrage alot more difficult since unless you can execute transactions faster than your competitors, you will be at a disadvantage
> 
> this will make arbitrage easier to be competitive, not harder, bc its less about latency and more about how much you're willing to pay

But latency is also "how much you're willing to pay".

proposals/0023-asynchronous-program-execution-and-broadcast.md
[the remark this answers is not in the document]
aeyakovenko: I think a signature is necessary, added the spec of it in the UserBlock format section

proposals/0023-asynchronous-program-execution-and-broadcast.md
aeyakovenko: See the UserBlock execution section.

aeyakovenko:
@lheeger-jump 

By clients do you mean wallets?  How do they decide if something has been confirmed?

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,280 @@
...
+blocks of non vote transactions. N number of builders can be scheudled
+concurrently to build blocks.
+
+While a leader is scheduled, they receive and encode votes as normal.
+Any well formed UserBlocks that were received from the previous or
+current UserBlockSlot are added to the leaders PoH ledger as
+UserBlockEntry.
+
+The N concurrent Builder create blocks out of user transactions.
+These are transmitted to the cluster via turbine.  The leader
```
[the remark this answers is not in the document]
aeyakovenko: Yep, if the leader observes the UserBlocks is very likely that the supermajority of the cluster has as well.

proposals/0023-asynchronous-program-execution-and-broadcast.md
aeyakovenko: yes, everything is converging.  I think the main difference is that solana has turbine and will provision validators to saturate the bandwidth available to them

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,280 @@
...
+
+The current leader schedule rotates each leader every 4 slots, but
+builders could be rotated at a different schedule then the leaders
+and rotation could be staggered.
+
+### UserBlock Compute Limits
+
+If the overall compute capacity for user transactions per leader
+block is 48m CU, and cluster is configured with 2 builders, then
+each UserBlock can use no more then 48m/4 or 12m CU.
```
apfitzge, line 109: This is a reasonable place to start imo. However, I have concerns that this could have unintentional side effects that lead to smaller blocks.

Validators are **not** spread evenly across the world, nor are users. There is significant clustering in North America and Europe.

As a hypothetical, let's say we have 2 builders 1 in Amsterdam and the other in Wellington NZ (roughly opposite sides of the world). The majority of users would simply target their nearest builder which will likely be Amsterdam. Amsterdam block gets packed and leaves people out while the Wellington block sits (relatively) idle.

It's certainly not a liveness issue, but I wonder if there is something that could be done to reduce this potential impact.

-------------------------------------------
**NOT WELL THOUGHT OUT SPITBALL IDEA**

Could we have more builders than can possibly fit into the `UserBlockSlot`, and leader takes the most heavily packed `UserBlockEntries`. This might be more complicated since then the leader has to scan these entries before selecting it.

Example above: 3 builders (AMS, WEL, +TOR).
AMS, TOR get packed because they are geographically (reduced latency) closer to users, and WEL gets excluded by the leader.


ugh, thinking about this more...maybe not a good idea. There's no way for non-leader to verify the leader chose the most packed block. Though the leader would (hopefully) be incentivized to choose the most packed due to fee-collection?

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,280 @@
...
+Leader could censor or delay the builder.
+
+Under heavy forking, validators will skip executing all the
+UserBlocks on minor forks.
+
+## Economic Considerations
+
+Builders should be shuffled and scheduled according to stake weight.
+
+TBD, deciding how should builders and leaders split the fees from
```
apfitzge, line 271: Something to consider here is how to handle duplicate transactions.

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,280 @@
...
+they have been hashed into the UserBlockEntry.
+
+### UserBlock format
+
+UserBlock follows the standard solana ledger format, except it
+MUST not contain any PoH ticks or votes.
+
+Invalid transactions are skipped, including duplicates, or those
+with invalid signatures, or completely malformed, and the rest of
+the block is processed.
```
apfitzge, line 143: A bit confused on this part - why should we begin to allow a builder to include transactions without proper signatures?

Shouldn't the builder's be running sigverify and basic checks?

apfitzge (commented):
Some initial comments/concerns. Nice write up!

proposals/0023-asynchronous-program-execution-and-broadcast.md
hydrogenbond007: sorry might be dumb but don't we already have dedup features?

apfitzge: No problem my previous message was not clear 😄 

We have deduping that prevents txs from appearing in multiple different slots. With this proposal, we have multiple builders who could potentially include the same transaction in their `UserBlock` since they can't know ahead of time what the other will include. We need to ignore one of them so we don't process the transaction twice, which is straight-forward to do.

My previous comment was mainly highlighting an edge-case for the economics of this deduplication. They both included a valid transaction in their `UserBlock`, so how do we handle fees?

Potentially a few options:
1) builder proportion of fees for tx split evenly between the two builders
2) first builder's block to reach leader

1 seems a better option, but it's also going to entirely depend how we split fees between the leader and builders in general.

hydrogenbond007 (approved):
great write up and would like to suggest some convo on something like [Whisk ](https://ethresear.ch/t/whisk-a-practical-shuffle-based-ssle-protocol-for-ethereum/11763) which will help protect the selected leaders from dos attacks. 
and once slashing is started, having leaders ddosed could become a concern
(sorry if the idea seems dumb still learning)

proposals/0023-asynchronous-program-execution-and-broadcast.md
nikhayes: Could the transaction being sent contain information that specifying which leader to process it? That would avoid duplicates and it's easier to filter the block for transactions referencing one of the particular leaders to know which fee rewards it should receive.

proposals/0023-asynchronous-program-execution-and-broadcast.md
aeyakovenko: Because if it is asynchronous, nodes may detect the invalid signature much later, well after they vote.  The only thing validators need to do to vote on forks is have the UserBlock data

proposals/0023-asynchronous-program-execution-and-broadcast.md
aeyakovenko: Yea, I think this is a v1 optimization.

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,280 @@
...
+
+Each UserBlock is assumed to have been created simultaneously during
+the UserBlockSlot that it was encoded by the leader.
+
+For each UserBlock in the PoH section spanning the UesrBlockSlot,
+the transactions are ordered by priority fee before execution.
+
+If two transactions from two different blocks have the same priority,
+they are ordered by which UserBlock appears first in the leaders
+PoH.
```
ilya-bobyr, line 125: Does this effectively creates something similar to the Ethereum mempool, but on chain?
As people are going to publish their transactions, yet the final order is determined only at the end of the `UserBlockSlot`.
Someone observing transactions sent to Builders can reliably front run any of those transactions, by including a transaction with a higher fee, as long as the current UserBlockSlot is not over yet.
And they can do so using any of the available Builders.
I am not necessarily criticizing - more like trying to make sure my understanding of the proposed mechanics is correct.

Or is the assumption that Builders do not share the transaction information until the end of the UserBlockSlot?

proposals/0023-asynchronous-program-execution-and-broadcast.md
ilya-bobyr: There is still an assumption that Builders validate transactions, making an effort to propose only UserBlocks with something that is valid, right?
As Builders are incentives for including transactions, they do not want to include garbage, and they could also be slashed for including transactions with invalid signatures or invalid in some other obvious way.
But, as pointed out, this should not stop the validator from including the UserBlock hash, even if it contains invalid transactions.

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,280 @@
...
+
+If the leader's block is attached to the heaviest fork,
+the validator can start execution of the UserBlocks optimistically.
+Otherwise the validator should only execute UserBlocks on the
+heaviest fork.
+
+Duplicate transactions in the UserBlock are skipped over without
+any state changes. They must still be retained in the ledger because
+they have been hashed into the UserBlockEntry.
```
ilya-bobyr, line 134: Wouldn't this mean that the most optimal behavior for the users would be to send every transaction for inclusion to every Builder?
If duplicates are removed for free, users lose nothing.  Yet, as the UserBlock hash order matters for transaction ordering in case of a conflict, it is beneficial to have all possible hashes assigned to your transaction.
It also removes the need of determining which Builder has the best RTT for the given user.

If it is the case, the downside is that it may effectively reduce the effective slot capacity, if too many transactions would turn out to be just duplicates in the final state.

proposals/0023-asynchronous-program-execution-and-broadcast.md
7layermagik: Maybe being able to specify the leader is also a feature though? Having your transaction processed by a certain leader might be undesirable for mev reasons?

ilya-bobyr:
IIUC, changes described in this proposal make leaders vote on slots without seeing the transaction actual content.  At the same time, Builders also do not seem to be doing anything useful with the transactions, except of forwarding them.
Sorry if this is something that has been discussed in the past, I did not read the previous SIMDs.
But I wonder what would happen if the next step is taken in the same direction.
And Builders do not see the transactions that they pack, either.

If users submit the fee and just the transaction hash to the Builders in UserBlockSlot N-1, and the actual transactions are submitted (and forwarded) during UserBlockSlot N.

For every transaction hash submitted in UserBlockSlot N-1, a matching transaction needs to be present in UserBlockSlot N, for the transaction to be included in the final state.
It would delay the moment when a complete state can be computed from the end of UserBlockSlot N-1 to some point during UserBlockSlot N, when the transaction itself is present.
But from the user standpoint, the latency is the same as in this proposal.
When a hash is included, the transaction is optimistically included.
They can also share the hash with a counterparty, if they want to show the transaction is pending before the end of UserBlockSlot N-1.

aeyakovenko:
@ilya-bobyr leaders and validators must see the transactions before voting.  They just don't need to complete execution before voting.  But data must be available locally.

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,280 @@
...
+status: Draft
+created: 2024-04-29
+feature: (fill in with feature tracking issues once accepted)
+---
+
+## Summary
+
+This feature changes how the ledger is broadcast and executed. It
+separates proposing blocks full of user transactions from blocks
+with votes.  It allows for N concurrent builders of user transaction
```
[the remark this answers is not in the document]
aeyakovenko: Configured by the cluster.

proposals/0023-asynchronous-program-execution-and-broadcast.md
aeyakovenko: Wdym processed by a certain leader? Either the UserBlcok is included or it's not. The big question is if it can be included in any slot, or only the scheduled slot.

apfitzge: I believe what is meant is a mechanism to make sure a transaction is processed by a specific **builder**

> Having your transaction processed by a certain leader might be undesirable for mev reasons so there could be reason for specificity?

If there's a certain builder I know does some MEV that is undesirable for my transaction, then I would request my tx is processed by the other.

proposals/0023-asynchronous-program-execution-and-broadcast.md
apfitzge: I have this concern as well, and it seems to be that @7layermagik may as well: https://github.com/solana-foundation/solana-improvement-documents/pull/45#discussion_r1156540615

ilya-bobyr:
> @ilya-bobyr leaders and validators must see the transactions before voting. They just don't need to complete execution before voting. But data must be available locally.

Are you saying that sharing the whole transaction details with a delay of 1 UserBlockSlot may also delay the voting process?
In the "Fork Choice" section, it says:
> the validator evaluates the forks as usual.

"As usual" means including execution, no?

I'm still relatively new to the Solana protocol.  But I was under the impression, that transaction execution is the main validation performed by the validators.
Voting without execution seems like a reduction in safety.

As described in this proposal, IIUC, transactions for the last UserBlockSlot in slot N might also not be fully available for all the validators until some later point during slot N+1 processing.  As the Builder only needs to share the hash with the leader, for the block to be produced.
In case when Builders are sharing transactions with a 1 UserBlockSlot delay, the situation seems similar.  Though I agree, the delay might be longer.  Increasing the number of UserBlockSlots per slot would reduce this delay.

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,280 @@
...
+only reference transactions prior to it in the UserBlock. TBD on
+format.
+
+## Detailed Design
+
+### Overview
+Leaders are scheduled to build blocks as they are currently by
+the LeaderSchedule.
+
+Builder's are scheduled along side leaders to build UserBlocks -
```
IntokuSatori, line 69: In this vision is the Builder set mutually exclusive with the Leader set? Are these Builders envisioned to be different types of nodes, or just normal validators? If normal validators, how will a Builder be prevented from being scheduled as a Leader in the same slot? (Assuming the sets aren't mutually exclusive)

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,280 @@
...
+This feature changes how the ledger is broadcast and executed. It
+separates proposing blocks full of user transactions from blocks
+with votes.  It allows for N concurrent builders of user transaction
+blocks, and it allows for asynchronous execution of user blocks.
+
+## Motivation
+
+1. Single leader for user transactions is a bottleneck.  Leaders
+are in a single spot in the world, while clients are all over the
+world. Latency is therefore based on how close a client is to a
```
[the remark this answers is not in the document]
apfitzge: I think the idea is it reduces latency to nodes choosing which transactions make it into the block, not total confirmation time.

Let's re-use the locations in your example, and suppose I see a really great opportunity in some market.
Currently, you have to send to LA which will delay you 180+ms at least. At that point, it's more likely you missed out regardless of your priority fee because the leader will start packing their block as soon as possible so they can collect the most fees.

There's actually 2 separate aspects in this proposal that seem like they might help you get the opportunity:
1. Priority Re-ordering (could be done indepedent of 2)
2. Multiple Builders (seems dependent on 1)

Let's say we **just** have priority re-ordering without multiple leaders. This means if you pay a higher priority fee than anyone else going for the opportunity, you would get it...if limits were not already hit AND you made it in time. 180ms + delay is a significant fraction of the 400ms block-time, and it's possible (maybe likely) your tx doesn't make it into the leader's queue for consideration early enough to get included.

Having multiple builders increases your chances that one is closer to you make it to the closest one in time for them to include you in their user block and before limits are hit.

Not sure what @aeyakovenko intended here though - **because if this `UserBlock` isn't transmitted early enough then the builder misses the leader's slot and you still don't get included.** Maybe this only helps with the not hitting limits aspect, but you'd still have to be very early.

proposals/0023-asynchronous-program-execution-and-broadcast.md
ilya-bobyr: I'm not the original author, but the way I understand it is:
1. Yes, Builders are selected from the pool of validators.
2. The same node is never both a Leader and a Builder.
   It seems to be just a technical detail of the selection function.

Currently, there is a function that determines the leader schedule based on the stake balances for the epoch of all the participating nodes.
This function can be arbitrary, as long as everyone can deterministically compute it and produce the same schedule.
So, it can be augmented to select 9 nodes per leader block instead of 1.
The first node will be the leader, producing 4 consecutive blocks.
And the rest 8 nodes will be split into 4 groups of 2 nodes, working as Builders in each of the 4 blocks for this leader.

And there could probably be certain details that I am overlooking.
But, in any case, I think, all these details became important only if some other questions are settled and overall the approach is decided to be the right one to address the latency issue.

proposals/0023-asynchronous-program-execution-and-broadcast.md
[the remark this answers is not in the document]
7layermagik: What is the advantage of having multiple block builders separated from a single leader vs having multiple leaders who are also responsible for buiiding blocks?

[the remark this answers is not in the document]
ilya-bobyr: What do you mean by "multiple leaders"?
Is there a proposal describing this kind of setup?

Currently, a leader is tasked with producing the next block in the chain.
Ignoring forks, every block in the chain contains a single child block.
How would multiple parallel leaders produce a single subsequent block to be added to the chain?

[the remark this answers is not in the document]
7layermagik: I think it's actually in the original whitepaper in section 4.4 https://github.com/solana-labs/whitepaper/blob/master/solana-whitepaper-en.pdf

Most of my understanding of how it could look like over the past year or so was informed from twitter discussions :) Basically, you'd need the bankless leader setup first, and from there you would go with an architecture similar to that described in the whitepaper.

aeyakovenko:
@ilya-bobyr 
Nodes evaluate forks by evaluating PoH entries to select the heaviest fork.  That doesn't require executing any UserBlocks. 

Sharing the UserBlock hash with the leader won't help, the rest of the network won't have it and will ignore that fork until they have the data for the UserBlocks.

proposals/0023-asynchronous-program-execution-and-broadcast.md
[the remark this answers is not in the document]
aeyakovenko: The multi leader approach will likely create more forks.  It's an attributable fault for a leader to produce two blocks for the same slot, but with multiple leaders if there is a turbine failure, the network has a similar partition but no one is at fault.

aeyakovenko:
@ilya-bobyr leader also received the UserBlocks through turbine.  It requires a large percentage of colluding stake to create a partition in turbine, so if the leader receives the UserBlock it's guaranteed to have been propagated to the rest of the network as well.

proposals/0023-asynchronous-program-execution-and-broadcast.md
7layermagik: Hmm yeah I guess u could have something in the transaction that specifies which leader it can be included by. If they want it to be processed by both leaders it costs double

proposals/0023-asynchronous-program-execution-and-broadcast.md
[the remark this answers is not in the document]
7layermagik: Ahh I see. Makes sense.

I had assumed there was some benefit with having multiple leaders in the past.... I guess with single leaders what can we do to improve ddos resistance? I worry that there might be a high incentive in the future for competitors to try and "kill" solana. Maybe "sentry nodes" in front of leaders (?)

proposals/0023-asynchronous-program-execution-and-broadcast.md
mschneider: 200ms is very short, I would assume that they just need to stream transactions as they arrive

proposals/0023-asynchronous-program-execution-and-broadcast.md
mschneider: Another reason for probably very small blocks: the builder only knows the compute limits requested not the compute units consumed.

proposals/0023-asynchronous-program-execution-and-broadcast.md
hydrogenbond007: Sorry again a dumb idea
Is there any way to encrypt the txns beforehand and reveal the transaction details only after they are included in the block to prevent bad reordering? Homomorphic encryption could allow operations to be run on the encrypted builder blocks

mschneider: > I believe what is meant is a mechanism to make sure a transaction is processed by a specific **builder**
> 
> > Having your transaction processed by a certain leader might be undesirable for mev reasons so there could be reason for specificity?
> 
> If there's a certain builder I know does some MEV that is undesirable for my transaction, then I would request my tx is processed by the other.

if the builder selection was based on the first few bits of the signature:

* builders could easily filter duplicate transactions submitted to the wrong builder
* power users could choose builders by bit-flipping data in their message
* users who don't care get properly load shedded

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -89,26 +90,63 @@ If a validator doesn't have the builder's UserBlock, the validator
...
+
+#### Persistent UserBlock slots
+
+Persistent UserBlock slots for the epoch are auctioned of to the top
+N bidders who burn the most lamports.
+
+In the first half of the epoch each builder deposits the lamports
+they are planning on burning, in the second half the builders may
+withdraw excess lamports.  The top N builders are assigned the slots
+in a dutch auction according to their remaining bids. If there are
```
apfitzge, line 125: Can you clarify the auction process here? I would have assumed an auction process where we take the top `N` bids with time-priority, not a dutch auction.

aeyakovenko: There are N slots to auction, so a Dutch auction makes the most sense.  Once bid could get 70% of the block and the other could get 30%

apfitzge: That doesn't sound like a Dutch auction as I understand them, so maybe I'm just getting confused by the terminology.

It'd help to give an example in the document to clarify.

apfitzge: Welp, since I'm the one who was confused I looked it up: https://www.investopedia.com/terms/d/dutchauction.asp

Seems to be an overloaded term, I was thinking of it as the first way they describe; i.e. a descending price auction, first bid wins.

I'm guessing that you mean the style they describe in the "Understanding Dutch Auctions for Initial Public Offerings (IPOs)" section:
> Once all the bids are submitted, the allotted placement is assigned to the bidders from the highest bids down, until all the allotted shares are assigned. However, the price that each bidder pays is based on the lowest price of all the allotted bidders, or essentially the last successful bid.

aeyakovenko: The auction is run for the next epoch. So it takes a full epoch to finish.

aeyakovenko: What's the most fair auction if the attacker can always place the last bid?

apfitzge: > What's the most fair auction if the attacker can always place the last bid?

What do you mean by attacker in this case? A node trying to become a persistent builder by gaming the auction?

If we're worried about nodes waiting to submit bids until they see everyone elses, we could run a sealed-bid auction.

aeyakovenko: Yea. Sealed bids might be better.  There is little harm in cancelled bids since the unused capacity would be scheduled randomly

aeyakovenko: Alternative to an auction would be extending staking to pick persistent builders.  That choice might be necessary if slashing for duplicate UserBlocks is necessary

aeyakovenko: Slashing duplicate UserBlocks shouldn't ever be necessary, since the leader would have picked 1 of the proposed blocks and the network would either accept or reject the leaders fork.  Everyone with inconsistent data can dump it and repair the right version.

proposals/0023-asynchronous-program-execution-and-broadcast.md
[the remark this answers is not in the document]
aeyakovenko: priority fees and base fees go to the leader

proposals/0023-asynchronous-program-execution-and-broadcast.md
aeyakovenko: or charge the fee on both if it is included in multiple user blocks by different leaders.  spammer would need to pre fund a bunch of fee payers, which wouldn't be free.

proposals/0023-asynchronous-program-execution-and-broadcast.md
segfaultdoc: i would advocate for a second-price sealed-bid auction, one where winner pays the bid of the second place bidder

segfaultdoc: > Alternative to an auction would be extending staking to pick persistent builders. That choice might be necessary if slashing for duplicate UserBlocks is necessary

how do you prove duplicate user blocks?

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,371 @@
...
+will be able to pick the closest one.
+
+The downside is that resource management becomes more complex. Each
+builder has 1/N compute and shred capacity and users have no idea
+which one is saturated when sending their transaction.  It's likely
+that priority fee floor will be different at each UserBlock builder.
+
+#### Persistent UserBlock builders
+
+Persistent UserBlock slots for the epoch are auctioned of to the top
```
segfaultdoc, line 120: how do you prevent the same N builders from winning every auction. imagine a whale builder bidding at a small loss every epoch pricing everyone out, only to earn all the mev for a volatile epoch

aeyakovenko: Someone else can outbid them. As long as the auction process isn't censored it's fine.  That's kind of the point of persistent builders.

proposals/0023-asynchronous-program-execution-and-broadcast.md
aeyakovenko: Same as you prove duplicate blocks.  Last shred is different

proposals/0023-asynchronous-program-execution-and-broadcast.md
hydrogenbond007: This is kinda the same with flashbots right now with 3 block builders making [70% of the](https://www.relayscan.io/) blocks because validators are looking for blocks with higher gas fees and optimized [uncle rate](https://blog.ethereum.org/2016/10/31/uncle-rate-transaction-fee-analysis)  and don't care much where the block is coming from.

proposals/0023-asynchronous-program-execution-and-broadcast.md
apfitzge: I don't think just charging fees for inclusion by multiple builders is sufficient.
That probably opens users up to attacks because builders are transmitting over turbine as they build.
Someone listening on turbine could just re-transmit processed transactions to other builders w/ hope of inclusion, and screwing over users on getting double-fee charged.

We can probably do some combination of hash/bit-based builder routing (like @mschneider suggested elsewhere) so signed message is only includable by certain builders.

This approach seems to check most of the boxes I think we want:
1. Users can select a specific builder by fiddling with bits
2. Users can send to multiple builders, but must pay to do so.

If users want to send to multiple builders, they'd want to have some sort of gate on the tx to prevent it from actually executing twice; seems reasonable this could just be done with a sequencer.

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,371 @@
...
+partially filled can be used by future blocks up at most N * (block
+or account limit). This would allow the network to handle bursts
+of demand and on average maintain the desired load.
+
+## Multi leader spam
+
+Spammers would be motivated to send the same message to every
+leader.
+
+1. builders censor fee payers by most significant bit of the fee
```
apfitzge, line 339: Am I missing something, shouldn't this be filtering by first `N` bits?

If we only look at first bit, then we'd get 2 sets of distinct fee-payers, rather than `N` (number of concurrent builders).

mschneider: if we use a power of 2 to limit the number of builders we can use ln(N) bits, as errror is always 0. Else using ln(N) + 4 is probably safer from a load shedding perspective error < 6.5%

mschneider: manipulating the fee payer is really annoying, for signature, I can just add a noop-instruction and rehash. why would we use fee payer and not signature?

proposals/0023-asynchronous-program-execution-and-broadcast.md
```diff
@@ -0,0 +1,371 @@
...
+While a leader is scheduled, they receive and encode votes as normal.
+Any well formed UserBlocks that were received from the previous or
+current UserBlockSlot are added to the leaders PoH ledger as
+UserBlockEntry.
+
+The N concurrent builders create blocks out of user transactions.
+These are transmitted to the cluster via turbine concurrently with
+the scheduled leader and other builders.  The leader receives and
+decodes them and generates a UserBlockEntry, and adds it to PoH as
+soon as the leaders PoH has started the UserBlockSlot.
```
mschneider, line 83: technically turbine is defined as never to forward to the leader. this change needs a bit of clarification, i guess shreds are never forwarded to their builder and the leader becomes the turbine root unless builder = leader?

aeyakovenko: Leader can add the UserBlockEntry on the next block

proposals/0023-asynchronous-program-execution-and-broadcast.md
aeyakovenko: @apfitzge yea, I meant N bits

aeyakovenko: @mschneider  for well behaving users, I think its more predictable to use a fee payer that is closest to the leader.  so latency sensitive users can rotate fee payers automatically

0xSol:
This SIMD has been inactive for 6 months. The status changed to Stagnant per [current SIMD process](https://github.com/0xSol/solana-improvement-documents/blob/main/proposals/0001-simd-process.md).  Close it for now, but feel free to reopen with new updates if still relevant, or create a new SIMD as needed.

PR #48, opened 2023-05-15
SIMD-0048: Native Program for secp256r1 signature verification

iceomatic:
## General Goal: 

Add support for secp256r1 signature verification, so we can enable the usage of Passkeys, WebAuthn and other account abstractions

## Notes
Would love to get the ball rolling on this :) Have talked briefly to @mvines & @samkim-crypto about it, and would greatly appreciate any additional inputs/ideas


proposals/0048-syscall-for-secp256r1-sigverify.md
```diff
@@ -0,0 +1,86 @@
...
+General Documentation:
+
+[WebAuthn](https://webauthn.io/)
+
+[Passkeys](https://fidoalliance.org/passkeys/)
+
+## Alternatives Considered
+We have discussed the following alternatives:
+
+1.) Realising signature verification with a native program/instruction similar to `KeccakSecp256k11111111111111111111111111111` instead of a syscall. As the signature is known/generated by the user upfront for a tx, an implementation as a native program would also work.
```
mvines, line 48: @lichtso - from the perspective of ABIv2, a native program/instruction would be strongly preferred over another syscall right? Or does it not really matter?

iceomatic: Would there be any benefit / significant man hours saved if I went with the native program approach?

mvines: I think we probably should hold here until @Lichtso and the runtime team decides what we're going to do in general about the "Precompile" instructions that currently exist for runtime-v2. We've had some Discord discussions about that over the last day or so.

Generally I feel like adding a syscall isn't the best approach if we decide to retain the notion of Precompiles in runtime-v2, since that limits our future ability to optimize the performance of this signature verification should it gain widespread usage. Others might argue that would be a premature optimization though..

iceomatic: Gotcha, okay 👍

iceomatic: Just want to ping in here again :) @Lichtso would you have anything against an implementation as a native program / do you see any up or downsides for either option?

iceomatic:
Would like to grab some more input from other contributors, so if you have any comments or ideas, please don't hesitate :D

samkim-crypto:
Yes, I think the secp256r1 verification have potential to have widespread usage, so we should do what we can to leave any potential for optimization, which would mean the native program route.

I think @Lichtso's opinion should weigh heavily here though, so it might be good to wait until he has some input.

Lichtso:
Sorry for the late reply. So we have three options:
1. Precompile: Preferred method for potential optimizations, but prevents being used from CPI
2. Syscall: Also fine, will be later be moved into a built-in (together with other syscalls) in program-runtime-v2
3. Built-in: A bit too much management overhead to have a standalone built-in with only one method, so would advise against it

mvines:
I suggest we proceed with the Precompile route for now. These are signatures that are always created off-chain so there's not really a use-case for a program to be able to provide the input (eg, Syscall approach)

iceomatic:
Rewrote the SIMD to reflect our consensus on the route of making it a precompile/native program 👍

mvines (dismissed):
This proposal looks good to me, I don't see any concerns from the Labs validator client side. Would love to see a bit of feedback from somebody else though.

If I recall correctly, @sakridge implemented the secp256k1 program so just tagging him in the event he wishes to review and comment

iceomatic:
> This proposal looks good to me, I don't see any concerns from the Labs validator client side. Would love to see a bit of feedback from somebody else though.
> 
> If I recall correctly, @sakridge implemented the secp256k1 program so just tagging him in the event he wishes to review and comment

Great! Thx 👍

sakridge (approved):
looks reasonable to me

proposals/0048-native-program-for-secp256r1-sigverify.md
```diff
@@ -0,0 +1,149 @@
...
+phrases is too cumbersome. It would be ideal to have an authetication form that
+relies on a more familiar second factor, such as a users mobile device.
+
+Passkeys & WebAuthn are a standardized implementation of this. They enable users
+to save keypairs associated to different services natively on the secure
+element of their mobile device. To authenticate with those services, the user
+uses their biometrics to sign a message with the stored private key.
+
+And although this is meant to enable password-less logins in web2, it makes for
+an excellent canditate as a second factor of on-chain authentication.
```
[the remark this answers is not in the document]
iceomatic: wups, fixed

mvines (approved):

antoineherzog:
Amazing! Passkeys and multisig are the future of self-custody ❤️

mvines:
Enough time has passed with approvals and no additional comments, merging

iceomatic:
Here's some clarification around the spec:

**Curve Parameters:** NIST P-256/secp256r1/prime256v1 curve parameters are outlined in the [SEC2](https://www.secg.org/SEC2-Ver-1.0.pdf) document in section 2.7.2

**Point Encoding:** Point encoding is well outlined in [SEC1](https://www.secg.org/sec1-v2.pdf#page=49) in Section 2.3.3 as either uncompressed or compressed SEC1 encoded points.

**Signing & Verification Functions:** Functions for signing messages and message verification are outlined in [SEC1](https://www.secg.org/sec1-v2.pdf#page=49) in Section 4.1. Additionally the updated algorithm in [RFC6979](https://www.rfc-editor.org/rfc/rfc6979) specifically outlines a the Deterministic ECDSA procedure. Test Vectors for P-256 are also present in [RFC6979](https://www.rfc-editor.org/rfc/rfc6979#appendix-A.2.5) in section A.2.5.

**p256 rust crate:** The ```p256``` [crate](https://docs.rs/p256/latest/p256/) crate is an implementation of the NIST-P256 curve and is part of the widely used and well maintained```rust-crypto``` library. It's tests have parity with the vectors found in [RFC6979](https://www.rfc-editor.org/rfc/rfc6979#appendix-A.2.5)

As a NIST certified curve, NIST-P256 and its algorithms are well documented and tested. Apart from the performance optimizations found in secp256k1, it relies on the same arithmetic and only differs in curve parameters.  An implementation in C for instance can be found [here](https://github.com/ConsenSys/besu-native-ec). 

A suitable testing suite to ensure general safety could be the [Wycheproof Project](https://github.com/google/wycheproof#project-wycheproof)

iceomatic:
Yes will do. I see your point regarding precise computational determinism. Obviously any mismatch there would be catastrophic. Trying to think of the best way to move forward.

Spitballing:
Has fd created a library for field arithmetic on secp256k1 yet? Might it be possible to use that, with adjusted curve parameters and without the optimisations around k1, and then instead of having a rust implementation, rely on a pre- compiled C version in both the Labs client, as well as fd?

jbrower95:
Hey folks. Driving resolution here would be awesome. Passkey signers powered by secp256r1, per the spec above, empower experiences like (https://defifortheworld.com), which are both safe and portable in ways that most current web SDKs can't achieve.

Where can I find the latest discussion?

4player90 (commented):
.

4player90 (commented):
.

anton584oh (commented):
👌

PR #53, opened 2023-05-31
SIMD-0053: Block Propagation Using QUIC Protocol

behzadnouri:
Use QUIC protocol (instead of current UDP implementation) to propagate blocks
across cluster.


mvines:
@ripatel-jump - could you please take a look at this? def want FD input on the direction here

proposals/0000-quic-turbine.md
```diff
@@ -0,0 +1,95 @@
...
+Once the cluster has upgraded to the new
+[`ContactInfo`](https://github.com/solana-labs/solana/blob/2fc1dc1bf/gossip/src/contact_info.rs#L68-L85)
+nodes may explicitly specify a socket address for TVU QUIC connections.
+Until then, similar to QUIC migration for TPU, we will use the port at
+[`QUIC_PORT_OFFSET`](https://github.com/solana-labs/solana/blob/2fc1dc1bf/sdk/src/quic.rs#L4)
+from the TVU socket for QUIC connections.
+
+For the initial implementation in the Solana Labs client we will use the same
+constructs as TPU QUIC implementation so the specs are the same. See [TPU/QUIC
+Protocol v1](https://github.com/solana-foundation/specs/blob/42f2058b7/p2p/tpu.md#tpuquic-protocol-v1).
```
mschneider, line 61: stake weighted stream / connection limit might become an issue with this use-case

lurais:
Do you have some example? Such as some one send different shreds to different node? And how does he generate the different shreds? Will the shreds be droped by others?

proposals/0000-quic-turbine.md
behzadnouri: Turbine random shuffle is also stake weighted; so I would expect some positive correlation between stake and who I receive shreds from.
It might be the case that the current cache-eviction policy for TPU connections wouldn't work well for turbine, in which case we can plug in a different eviction policy; but I guess that will still need to favor higher staked nodes somehow.

proposals/0000-quic-turbine.md
```diff
@@ -0,0 +1,95 @@
...
+* Implement [ping/pong
+  protocol](https://github.com/solana-labs/solana/blob/2fc1dc1bf/gossip/src/ping_pong.rs)
+  between nodes to verify IP addresses before retransmitting any shreds to the
+  recipient.
+* Each node also signs and attaches its signature to each shred it is
+  retransmitting.
+
+This will cost 64 bytes of each shred payload to embed retransmitter's
+signature and an extra signature verification overhead for each shred at the
+recipient node.
```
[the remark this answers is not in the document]
behzadnouri: What do you mean? why does merkle vs legacy matter here?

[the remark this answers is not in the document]
behzadnouri: > reduce signature verification rate for leader signatures. The same mechanism could be repurposed for retransmitter signatures

Which of the below is the mechanism you are referring to?
1. all shreds within the same erasure batch having the same signature?
2. or, signed messages being shorter with Merkle shreds? (only 32 bytes merkle root vs the whole payload before).

Obviously (1) doesn't apply to the retransmitter signature.

proposals/0000-quic-turbine.md
```diff
@@ -0,0 +1,95 @@
...
+
+There are ways to improve above but it gradually approaches to re-inventing
+QUIC which defeats the point of not using QUIC to begin with.
```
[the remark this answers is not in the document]
behzadnouri: I believe the only functionalities we need are:
1. Prevent IP address spoofing and any traffic amplification for unverified addresses.
2. Obtain and authenticate sender's pubkey during connection establishment.
   * i.e. verify that the sender is indeed in possession of the private key.
4. Ability to rate-limit senders by their pubkey.

(anything else?)

Any other QUIC feature which does not put above 3 at risk can be left disabled in favor of simplicity and achieving better throughput.

So please feel free to suggest any features which we should leave out if you see them as an overhead and you believe it has no impact on above 3 features.

For example I totally agree that shreds do not need to be encrypted, but if we leave encryption entirely out, wouldn't that make it possible for someone to claim a pubkey that they don't own the private key of?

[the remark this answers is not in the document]
behzadnouri: > > I believe the only functionalities we need are:
> > 
> > 1. Prevent IP address spoofing and any traffic amplification for unverified addresses.
> > 2. Obtain and authenticate sender's pubkey during connection establishment.
> >    
> >    * i.e. verify that the sender is indeed in possession of the private key.
> > 3. Ability to rate-limit senders by their pubkey.
> > 
> > (anything else?)
> 
> Yes, I think a few more:
> ...

Thank you for adding these. But what I meant was required functionalities we need from turbine's perspective in theory regardless of which protocol we use. For example if we implement a custom protocol on top of UDP just for turbine, what minimum set of features would be necessary.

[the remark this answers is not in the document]
behzadnouri: That is totally fair.
I added your comments to the doc.
Please feel free to edit or add.

jacobcreech:
@behzadnouri @ripatel-fd looks like there's some level of consensus between ya'll on this SIMD. Correct me if I'm wrong though - not sure if there's enough information to get to acceptance.

linuskendall:
One of the limitations of a QUIC based approach is that there are currently afaict no mature open source gateways to manage filtering or DDOS mitigation on QUIC? The UDP protocol is quite a lot easier to work with, even if Solana's specific implementation is custom so the current commercial solutions can't exactly handle it well. This in effect means that we are trading something easy to build custom filtering around for something that is easier to buy a commercial solution for (but not really open source solutions last I checked),

behzadnouri:
> One of the limitations of a QUIC based approach is that there are currently afaict no mature open source gateways to manage filtering or DDOS mitigation on QUIC?

What did you guys do when TPU migrated to QUIC?
or are you thinking of a filtering/ddos-mitigation specific for turbine?

> Also anything that's meant to process this traffic would need to have access to the validator identity? 

I believe so but I am not sure if that is enough. The QUIC spec says:
https://www.rfc-editor.org/rfc/rfc9000.html#section-21.1.1.4

    The entire handshake is cryptographically protected, with the Initial
    packets being encrypted with per-version keys and the Handshake and later
    packets being encrypted with keys derived from the TLS key exchange.
    Further, parameter negotiation is folded into the TLS transcript and thus
    provides the same integrity guarantees as ordinary TLS negotiation.


> Would there be a point in splitting up the validator voting identity and the network identity used for signing QUIC? In terms of operational management, the less trusted keys we can use for middlewares would be better.

We already only use the validator identity for the networking purposes. To my knowledge, there is no use of voting identity anywhere. So unless I did not understand your question, this is already split out.

linuskendall:
> > One of the limitations of a QUIC based approach is that there are currently afaict no mature open source gateways to manage filtering or DDOS mitigation on QUIC?
> 
> What did you guys do when TPU migrated to QUIC? or are you thinking of a filtering/ddos-mitigation specific for turbine?

We had to spend a lot of time to work out good approaches for QUIC, and TBH they still aren’t very well-developed as yet. As far as I see it there’s limited proxying support for QUIC. Had we still had UDP TPU we could have already had much better firewalling support in place as well as do edge TPU firwealling. Instead we end up with Jito relayer + firewall rules that kind of support 1 RTT ratelimiting.

In the case of TPU I think there’s a decent argument for the benefits of QUIC slightly outweighing some of the costs. 

Looking at what you are outlining above as the features we need from QUIC for block propagation I really don’t understand how the costs of QUIC are outweighed by the benefits. 

> 
> > Also anything that's meant to process this traffic would need to have access to the validator identity?
> 
> I believe so but I am not sure if that is enough. The QUIC spec says: https://www.rfc-editor.org/rfc/rfc9000.html#section-21.1.1.4

> 
> ```
> The entire handshake is cryptographically protected, with the Initial
> packets being encrypted with per-version keys and the Handshake and later
> packets being encrypted with keys derived from the TLS key exchange.
> Further, parameter negotiation is folded into the TLS transcript and thus
> provides the same integrity guarantees as ordinary TLS negotiation.
> ```
> 

Yeah, so now if I want to deploy a gateway to protect the validators ports that ratelimits or controls shred forwarding I *need* to deploy the validators key on to that node. Alternatively,  I need to have full nodes on the network that act gateways which adds additional strain on the network.


>  > Would there be a point in splitting up the validator voting identity and the network identity used for signing QUIC? In terms of operational management, the less trusted keys we can use for middlewares would be better.
> 
> We already only use the validator identity for the networking purposes. To my knowledge, there is no use of voting identity anywhere. So unless I did not understand your question, this is already split out.

The node identity receive rewards. Therefore it has value beyond the networking purposes. It also is encoded in the vote account, meaning that rotating this identity in case of compromise can be complicated (requires that the person in control of the vote account does the updating of the identity).

This means that the node identity still is a “secret” and “valuable” identity used for submitting votes. It therefore cannot be used more freely for networking purposes or deployed on gateways without a fair bit of security concerns.

linuskendall:
When it comes to the benefits this is giving:


*Prevent IP address spoofing and any traffic amplification for unverified addresses.*

If the amplifying node (e.g. agave-validator) already de-dupes shreds, how would traffic amplification happen? There is only forwarding of valid shreds and presumably not forwarding of dupe shreds (or at least that would be easy to implement). If we receive multiple copies of the shreds, and this is a concern we can already de-dupe raw-UDP shreds very easily on the network layer. We do not need ip address authentication for this. 

So I don’t see the need for QUIC when it comes to traffic amplification. QUIC also makes it harder to avoid DDoS since we need to implement QUIC aware DDoS capabilities if we want to retain the benefit of the authenticated / anti-IP spoofing. It would be better to just work on network layer DDoS mitigation that does not need to be QUIC aware. QUIC aware proxying would need to have access to the validator’s records of pubkey identities and be able to handle some decryption of the QUIC streams. 


**Obtain and authenticate sender's pubkey during connection establishment.
i.e. verify that the sender is indeed in possession of the private key.**

Why? What if we receive shreds from an unknown source? aren’t shreds already authenticated in that we can validate if they are produced by the correct leader? Why also authenticate the “forwarder” of the shreds when we can authenticate the source of the shreds?

**Ability to rate-limit senders by their pubkey** 
Rate limiting by IP wouldn’t really cause an issue even if the IP was spoofed. With UDP based Turbine we can use checksums to ratelimit duplicate UDP packets earlier on the network infra. Why do we need pubkey aware ratelimting for shreds? This requires a gateway or the solana validator to be in charge of ratelimiting in userspace/application layer. With UDP ratelimiting and duplicate packet detection we could very easily ratelimit on the network layer. Unlike TPU spoofing is not an issue.

proposals/0053-quic-turbine.md
```diff
@@ -0,0 +1,120 @@
...
+## Motivation
+
+* The Solana protocol currently uses an datagram-based block distribution
+  mechanism (Turbine/UDP) using out-of-band peer discovery.
+* We find that Turbine/UDP lacks important features to support reliable block
+  data distribution over WAN, namely:
+  * Turbine/UDP does not specify congestion control mechanism and thus lacks
+    the ability to dynamically adapt to packet corruption/loss (currently
+    mitigated via forward error correction with a hardcoded ratio).
+  * Turbine/UDP lacks data authentication and a mechanism to reject incoming
```
linuskendall, line 27: why is this required when the data is already authenticated? invalid shreds can be detected?

proposals/0053-quic-turbine.md
```diff
@@ -0,0 +1,120 @@
...
+    block data streams.
+  * This makes Turbine/UDP recipients susceptible to packet floods.
+  * Commercial DDoS protection solutions are considered ineffective for custom
```
linuskendall, line 30: can you show me a generally good available QUIC DDoS mitigation solution?

proposals/0053-quic-turbine.md
```diff
@@ -0,0 +1,120 @@
...
+    UDP protocols like Turbine/UDP.
+* QUIC is an Internet standard defining a protocol for a UDP-based multiplexed,
+  authenticated, and encrypted transport.
+  * QUIC is already in use in other parts of the Solana protocol.
+  * QUIC implements many of the features required that Turbine/UDP currently
+    lacks.
+  * We expect wider specialization of network infrastructure towards QUIC from
```
linuskendall, line 37: For effective management of QUIC protocol this would require border gateways with ability to decrypt the traffic, which means that without a separate network identity the protocol/node identity would need to be used.

proposals/0053-quic-turbine.md
```diff
@@ -0,0 +1,120 @@
...
+## Security Considerations
+
+We expect migrating from UDP to QUIC will provide useful features in improving
+security and QoS of the block propagation protocol.
+
+## Drawbacks
+
+Obviously QUIC will add extra overheads compared to a bare UDP implementation.
+We expect these overheads will be mitigated by using a large enough
+connection-cache to reuse the same connections and minimize handshake costs.
```
linuskendall, line 114: Seeing the issues of workign with the QUINN library for TPU I think the potential risks and overhead with QUIC in this context and the pitfalls of selecting the correct QUIC parameters is considerable.

QUIC with its connection oriented nature and it’s choice of receive windows leads to a linkage between bandwidth and latency, something that we have already seen happen in the QUIC TPU case (our estimate is that in the current implementation you have 20-40x the bandwidth if you are located close to the leader - i.e. NYC/AMS vs far from the leader TYO).

This leads to a geographic centralisation as we would be pretty much forced to move our turbine receivers/ read nodes to locations that have a lower latency to potential leaders. 

UDP Turbine does not suffer from this problem, since bandwidth and latency can be disconnected to a greater degree in a connection less design.

behzadnouri:
I will reach out to @linuskendall directly to discuss their comments.

As for this SIMD, we already have an implementation of Turbine over QUIC already released but it is not activated yet because we want to explore more lightweight and efficient protocols. I will close this pull request for now.

PR #57, opened 2023-06-13
SIMDXX: Events

mschneider:


proposals/0057-events.md
```diff
@@ -0,0 +1,192 @@
...
+A lot of developers would prefer a reliable and universally compatible solution
+so recently Anchor has added a new feature called cpi_events. This is a
+convention on how to call your own program via CPI and encode a binary string
+as instruction data. This can be indexed via the RPC Transaction response, as
+innerInstruction data is always persisted. It is more reliable than sol_log_
+but it effectively reduces runtime limits like maximum CPI call depth for the
+actual application. It also causeses unnecessary overhead in replay as a new
+execution context neeeds to be created, for something as trivial as a log
+message. There is a potential security risk in the this system in that external
+programs could spoof anchor events (and the program emitting them) with CPI
```
jarry-xiao, line 56: The spoofing risk was mitigated by adding in a PDA event authority that must sign the self CPI

proposals/0057-events.md
```diff
@@ -0,0 +1,192 @@
...
+plaintext debug message, hence become very difficult to parse, which is why
+many developers avoid structured event logging via anchor and just use
+formatted strings.
+
+A lot of developers would prefer a reliable and universally compatible solution
+so recently Anchor has added a new feature called cpi_events. This is a
+convention on how to call your own program via CPI and encode a binary string
+as instruction data. This can be indexed via the RPC Transaction response, as
+innerInstruction data is always persisted. It is more reliable than sol_log_
+but it effectively reduces runtime limits like maximum CPI call depth for the
```
jarry-xiao, line 52: Unclear if this is true for the average cases. If my program already CPIs to another program then the depth doesn't actually change.

proposals/0057-events.md
```diff
@@ -0,0 +1,192 @@
...
+actual application. It also causeses unnecessary overhead in replay as a new
+execution context neeeds to be created, for something as trivial as a log
```
jarry-xiao, line 54: This is an extremely valid criticism, but this should get resolved in Runtime V2, where VM's are only spun up once per transactions as opposed to once per instruction

jarry-xiao:
I fully support the development of more robust indexing solutions + persistent event data, but it's hard for me to see a strict benefit in the near term future of native events over CPI events (which is admittedly a hack). From a resource allocation standpoint, this feels like large engineering lift + ecosystem buy-in lift for insufficient gain IMO. No opposition if a native solution gets prioritized and implemented with 5x improvement to existing devex.

proposals/0057-events.md
mschneider: We definitely need events in spl token and system program to deprecate the existing rpc features. Afaik everyone is calling those at some point when cpi depth is critical. We already ran into the limit and stopped encouraging CPI into mango because of it.

jarry-xiao: What's wrong with the current approach of just directly parsing instruction data for those programs?

mschneider: Won’t work for token 2022

jarry-xiao: I'm not too familiar with how Token22 differs from the original. Transfer/mint indexing doesn't break with the extensions right? 

Does `set_return_data` work for Token22? I know that syscall has some annoying quirks associated with it

jarry-xiao: Also, runtime V2 should in theory increase CPI stack depth. Probably should increase it an arbitrary amount if there's literally no more overhead

mschneider: this assumption is wrong, there will always be overhead, it's an execution context, there's isolation guarantees. yes it currently is incredible overhead, docker container level. the v2 will be roughly equivalent of a fork syscall. but that's still pretty bad for a log message.

proposals/0057-events.md
mschneider: just to confirm, does this require to add another account per program that wants to emit logs to the instruction?

jarry-xiao: Yep, but it’s only 34 bytes. (0 data, only pubkey + flags)

proposals/0057-events.md
jarry-xiao: Was exaggerating, but what I meant was that runtime v2 should get rid of much of the nasty stuff that makes CPI super inefficient (new VM, copy all of the accounts). CPI no longer being depth bounded makes this viable IMO. If there was still a cap on depth then I agree that native events legitimately adds something compelling/differentiating

proposals/0057-events.md
```diff
@@ -0,0 +1,192 @@
...
+    context: {
+        eventIndexInBlock: number,
+        slot: number
+    }
+) => void)
+```
+
+DataSizeFilter and MemcmpFilter can be used to filter for specific event types
+using their discriminator. AccountMetaFilter allows to restrict on events
+emitted from transactions reading or writing from/to specific POA & EOA.
```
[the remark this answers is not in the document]
mschneider: program owned account and externally owned account

proposals/0057-events.md
```diff
@@ -0,0 +1,261 @@
...
+those are technically not part of the SIMD specification process, it's
+important to publish a recommended spec for 3rd party providers so that Solana
+Dapps don't run into vendor lock-in issues.
+
+### RPC-Client
+
+Clients should be able subscribe to events emitted with various filters:
+
+```
+rpc.onEvents(
```
mvines, line 123: There should be some way for the client to know the slot height that their subscription starts at. This'll allow them to backfill any missed events prior to their subscription using some other method.

proposals/0057-events.md
```diff
@@ -0,0 +1,261 @@
...
+
+### Runtime 
+
+In order to log an event, the SVM runtime must implement a new syscall
+`sol_emit_`. The runtime must charge 100 + ceil(48 * len / 1024) CU per
+invocation.
+
+```
+#include <sol/emit.h>
+void sol_emit_(/* r1 */ uint64_t len,
```
mvines, line 108: How is the "Discriminator" expressed? 

Also it seems better that the program specify the addresses it wants to associate with the event no? This would be part of the price, and allow the indexing backends to only persist the addresses that actually matter instead of requiring every address in the transaction.

proposals/0057-events.md
```diff
@@ -0,0 +1,261 @@
...
+POA - Program Owned Account
+
+EOA - Externally Owned Account
+
+## Detailed Design
+
+### Runtime 
+
+In order to log an event, the SVM runtime must implement a new syscall
+`sol_emit_`. The runtime must charge 100 + ceil(48 * len / 1024) CU per
```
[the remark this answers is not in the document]
mschneider: i tried to find an answer in my notes, but it's too long ago, will need to figure out why CPI has a 50MB / 200k CU conversion

proposals/0057-events.md
```diff
@@ -0,0 +1,261 @@
...
+definitions / account state.
+
+## New Terminology
+
+Discriminator - a unique n byte sequence often a 8 byte hash of a method
+signature or data type, that should be used to tag the type of a data structure
+and encoded always in the very first bytes of an event, account or instruction
+for the sake of efficient parsing.
+
+POA - Program Owned Account
```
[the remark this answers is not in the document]
mschneider: i can scrub the difference from the proposal, it's not essential

proposals/0057-events.md
mschneider: discriminator: not at all, i thought we can get by without supporting it explicitly in the standard. clients can decide to have an index on the program id + first n (1..32) bytes  for performance reasons, but that's implementation detail.

indexing backends anyways have a gsfa index, an additional discriminator -> signature index can be used for merged index queries.

proposals/0057-events.md
```diff
@@ -0,0 +1,261 @@
...
+### Runtime 
+
+In order to log an event, the SVM runtime must implement a new syscall
+`sol_emit_`. The runtime must charge 100 + ceil(48 * len / 1024) CU per
+invocation.
+
+```
+#include <sol/emit.h>
+void sol_emit_(/* r1 */ uint64_t len,
+               /* r2 */ const char *message);
```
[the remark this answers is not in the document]
mschneider: why? you can cpi with 5k instruction data right now for 1020 CU

proposals/0057-events.md
mschneider: it should work well to wait for the first event an then backfill (the event includes the slot)
in the mean-time the user can manually query getSlot after opening the subscription and backfill from there

mvines: waiting for the first event might not work well for events that don't happen frequently though right?

proposals/0057-events.md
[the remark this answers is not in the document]
mvines: sorry, 5k was an example. i just think we need some reasonable event payload limit defined

proposals/0057-events.md
mvines: sorry, what's a gsfa?

> clients can decide to have an index on the program id + first n (1..32) bytes for performance reasons, but that's implementation detail.

i think leaving this as an implementation detail means we could then expect to see different behaviours across rpc endpoints, but that's the problem with `sol_log()` that this proposal is attempting to solve though

proposals/0057-events.md
mschneider: yeah but slot updates itself are frequent enough imo

proposals/0057-events.md
mschneider: gsfa = getSignaturesForAddress

imo this is very application specific, i would expect the program IDL to declare to the indexing provider which parts of the event (for each event type) need indexing and hence which queries can have improved latency. the difference here is that the data query API is defined which is important for compatibility, so same query bytes -> same result bytes. latency is still dependent on rpc provider, it's not something you could enforce via spec.

proposals/0057-events.md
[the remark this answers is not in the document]
mschneider: sure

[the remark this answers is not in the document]
jumpsiegel: The events should have a sequence number so that I can "subscribe" to an event stream and restart at a specific sequence number.  This would match up with the event behavior for aptos, sui, and (via a hack or two) algorand.

proposals/0057-events.md
theDefiBat: Also would be nice to add this example of cpi call depth problem which is currently max: 4 

So if the anchor cpi event is added to the token transfer instruction 
This scenario might break when composing::
Mango-> Jupiter -> Orca/ Phoenix-> Token Program -> self-cpi-event

proposals/0057-events.md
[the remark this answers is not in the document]
mschneider: the events on stream and when queried via rpc call include a sequence number, just to clarify, would it be enough to be able to send a request to backfill missing events (once you notice discontinuity?) i find stream semantic for historic events hard to reason about. curious to see the api's you mentioned, could you drop me the links @jumpsiegel ?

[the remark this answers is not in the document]
mschneider: @ripatel-fd brought up in the discussion today that we should consider the cost of including a hash of all events in consensus. this would allow light-clients to verify the authenticity of events received from third party rpc providers. 

bank deltas are currently hashed via sha256, the sha256 syscall charges 85 + 0.5 CU / byte. Given that sha256 is kinda overkill for what we are trying to achieve a faster hash function would be adequate, e.g. blake3. Albeit blake3 being roughly 2x faster to compute (see [blake3-avx2 vs sha256-ni](https://github.com/BLAKE3-team/BLAKE3/issues/207)) the runtime currently charges the same amount for blake3 hashing.

Happy to go either way from here, but 100 + 0.25 * len CU would be a good approximation in this case. CPI for event logging would be only "cheaper" when using  >4k event payloads in this case, most events are <512b in length.
Limit per 200k CU: 800kB
Limit per 48M CU: 200MB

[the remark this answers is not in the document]
jumpsiegel: https://aptos.dev/concepts/events

https://docs.sui.io/build/event_api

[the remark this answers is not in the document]
mschneider: @jumpsiegel i added a way to query events directly and included the sequence number for that purpose. due to the way solana randomizes block entry execution there is some weirdness around how events can change their order inside the response, I clarified that. Please let me know if this is sufficient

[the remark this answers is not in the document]
mschneider: @mvines please let me know, if this issue is resolved

proposals/0057-events.md
mschneider: > that this proposal is attempting to solve though

just want to comment again on this:

proposal solves provider specific truncation and lack of query api schema, basically enforces data consistency across providers.

it does **not** guarantee reliable query performance, similar to how SQL spec can't enforce query performance.

proposals/0057-events.md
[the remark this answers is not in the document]
jumpsiegel: I must admit, I am confused by this statement.  Does this mean different validators might have events with different sequence numbers?  ie, the sequence number is non-deterministic and not part of consensus..   The effect of this would be that if I change event providers (validators) while going through a load balancer, the sequence numbers would change and I would find myself scrambling to figure out what I have missed?

proposals/0057-events.md
```diff
@@ -0,0 +1,313 @@
...
+    }
+): Promise<TransactionSignature[]>
+```
+
+In addition clients should be able to query events directly and index them
+reliably working around the randomized execution of entries inside a slot.
+The response is sorted by the following priorities:
+
+1. slots ascending
+2. signature order for transactions inside a block
```
mvines, line 217: Rather than trying to order by signature here, there is a single natural ordering of how transactions are placed in entries and then in a block. This order is common across the cluster (`solana-ledger-tool` outputs it). During replay, the entry transactions are intentionally shuffled randomly but if that shuffling didn't exist then the entire cluster would replay entries the same way

mschneider: Oh wasn't aware, if we can keep that alive across replay, we should be able to have even more easy indexing. Will adapt.

proposals/0057-events.md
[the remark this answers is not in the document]
mschneider: Yes, that's the current behaviour of gSFA. @mvines mentioned it's possible to clean up, so i'll update the proposal wrt. i think it's a huge improvement to change this part.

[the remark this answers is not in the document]
mschneider: in general sequence numbers can still change due to block re-orgs. i updated the proposal text, would appreciate a final review and ideally thumbs-up, so we can get started on implementation :)

proposals/0057-events.md
theDefiBat: cc : @askibin

alessandrod:
> I would rather bring these fixes to the existing log mechanism instead of adding a new log facility with a different name.

Just a drive by comment (sorry!) since someone asked about this. If we reused the existing log facility, how would RPC be able to tell whether something is an event or a dumb log? Would it filter/match all logs? That seems terrible perf wise

steveluscher:
Previously: https://forum.solana.com/t/srfc-00004-native-events-program/32. cc/ @ngundotra

steveluscher:
@steviez, do you have some time to think on the ways in which event data could be queried and filtered upon (both the ones stated in this SIMD and any speculative ones you can think of) with respect to the implications about how it might be stored and indexed in blockstore? The heading `RPC-Client` is the section to look at.

proposals/0057-events.md
```diff
@@ -0,0 +1,341 @@
...
+### Runtime 
+
+In order to log an event, the SVM runtime must implement a new syscall
+`sol_emit_`. The runtime must charge 100 + 0.25 * len CU per invocation to
+account for the potential cost of including a blake3 hash of all log messages
+in consensus.
+
+```
+#include <sol/emit.h>
+void sol_emit_(/* r1 */ const char *message,
```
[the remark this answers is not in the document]
mschneider: change of CU cost bc size restriction is no longer a client side decison. sol_log_ truncates for programs that log a lot of data

proposals/0057-events.md
```diff
@@ -0,0 +1,341 @@
...
+    data: [string, "base64" | "base64+zstd"],
+    programId: string,
+    outerInstructionIndex:  number,
+    innerInstructionIndex?: number,
+}
+```
+
+
+### Geyser client
+
```
apfitzge, line 286: we really should not be adding **new** interfaces in geyser that are also broken for non-rust languages.
we eventually need to fix the old ones, but imo we shouldn't add new ones that are also broken - please make the new `Event` structs ffi-safe.

proposals/0057-events.md
```diff
@@ -0,0 +1,341 @@
...
+authenticity of events received from third party rpc providers.
+
+Bank deltas are currently hashed via sha256, the sha256 syscall charges
+85 + 0.5 CU / byte. blake3 a roughly 2x faster to compute hash, would be
+sufficient as well, hence 100 + 0.25 * len CU would be a good approximation
+in this case. CPI for event logging would be only "cheaper" when using >4k
+event payloads but most events are <512b in length. This cost enforces
+theoretical limits to the amount of logs per transaction & block:
+Limit per 200k CU: 800kB
+Limit per 48M CU: 200MB
```
apfitzge, line 126: It seems odd to me that this limit, which is per block, is higher than the amount of state-growth we allow per block.
That to me implies that the CU cost is too low? why would we let events emit faster than our state growth if infrastructure has to store both?

proposals/0057-events.md
```diff
@@ -0,0 +1,341 @@
...
+for the sake of efficient parsing.
+
+## Detailed Design
+
+### Runtime 
+
+In order to log an event, the SVM runtime must implement a new syscall
+`sol_emit_`. The runtime must charge 100 + 0.25 * len CU per invocation to
+account for the potential cost of including a blake3 hash of all log messages
+in consensus.
```
apfitzge, line 101: I don't see a section specifying how this would be included in consensus, what's the plan here?

proposals/0057-events.md
mschneider: Happy to go with double the cost per byte, if that simplifies the process moving forward.

proposals/0057-events.md
```diff
@@ -0,0 +1,308 @@
...
+important to publish a recommended spec for 3rd party providers so that Solana
+Dapps don't run into vendor lock-in issues.
+
+The cost for this operation is derived from a possible future inclusion of a
+hash of all events in consensus. So future light-clients can verify the
+authenticity of events received from third party rpc providers.
+
+Bank deltas are currently hashed via sha256, the sha256 syscall charges
+85 + 0.5 CU / byte. blake3 a roughly 2x faster to compute hash, would be
+sufficient as well, hence 100 + 0.25 * len CU would be a good approximation
```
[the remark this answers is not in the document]
mschneider: Ack.

proposals/0057-events.md
mschneider: There is no concrete plan yet. I wanted to leave that open for a future SIMD, as the roadmap towards light clients is not clear yet.

proposals/0057-events.md
mschneider: had a look at this PR https://github.com/anza-xyz/agave/pull/2125 and updated the events struct, lmk if that works for you

apfitzge:
With this being outside of consensus, I'm not sure how this is different than logs wrt forcing operators to save the data.

For RPCs, sure they'll likely need to save it to serve users. But for normal validator operators, couldn't they just drop the logs after execution, or even patch logic to not emit events at all? Even RPCs, if they didn't want to offer this service, could drop them.

I guess what I'm missing is how this is different than logs where we truncate, since I'm not seeing anything that prevents me from truncating. With my current reading, this effectively seems like a specific way to do logging.

steveluscher:
Because we have an opportunity to do _better_ than logs here, I wanted to ask if we can offer guarantees that an event hasn't been modified.

Let's imagine that I was building some event-driven system that listened for events on the network (listing/oracle/order/etc) and took some action involving a non-trivial amount of value. I would want to know that the event was:

* logged by the entity I think it was logged by, and
* unmodified

Regarding the existing userspace implementation of events atop self-CPI, @jarry-xiao mentioned that the ‘spoofing risk was mitigated by adding in a PDA event authority that must sign the self CPI.’ That's good. For the next point though – for the same reason that RPCs are free to truncate logs, I don't think there's anything to stop a malicious RPC from modifying the _content_ of the log in a way that the consumer can't detect.

I don't have any ideas of how to do this. Normally I'd produce an [HMAC](https://en.wikipedia.org/wiki/HMAC) and include it in the content of the event data itself, but at the point where the program is emitting the event, there's nothing secret available to include in the HMAC, and at the point where the signers are signing the transaction they don't know what event data will be emitted.

Anyone have any thoughts?

mschneider:
@ripatel-fd had a read-through SIMD-64, please let me know if I correctly summarized it, i also have a couple questions:
- transaction receipt is a 32 byte hash that commits transaction execution status (success/failure) and transaction message
- given all transaction receipts and their PoH order for a given slot, anyone can verify the correctness of that information by collecting the 32 byte commitment tree root of all validators or a reasonable subset
- it does not specify how the commitments are shared, just mentions it would not be ideal to add it to block building. did you plan any extensions to voting or gossip?
- inclusion proofs are not part of the proposal, were these formalized since then?

Benhawkins18:
Stale

PR #5, opened 2022-12-08
Bankless leader proposal

taozhu-chicago:
Refreshed and moved bankless leader proposal to SIMD process

PR #61, opened 2023-07-05
SIMD-0061: Fair congestion control with intra-block exponential local base fee

ryoqun:
just starting. :)

this is my pet idea, been sitting in my mind for awhile, considering recent talk at discord's #ecnomics, seems it's prime time to engage in this battle field. :)

this is much serious rambling than this (https://discord.com/channels/428295358100013066/987445907031199844/1080717702672429107): 

image

### related (read: competing) proposals.

dynamic base fees
https://github.com/solana-foundation/solana-improvement-documents/pull/4

program rebatable account write fees:
https://github.com/solana-foundation/solana-improvement-documents/pull/16

asynchronous program execution:
https://github.com/solana-foundation/solana-improvement-documents/pull/45

increase prioritization fee:
https://github.com/solana-foundation/solana-improvement-documents/pull/50

bankless
https://github.com/solana-foundation/solana-improvement-documents/pull/5

dynamic base fee2?
https://github.com/solana-foundation/solana-improvement-documents/pull/19

### DIFF
--- a/proposals/0061-fair-congestion-control.md
+++ b/proposals/0061-fair-congestion-control.md
@@ -80,10 +80,11 @@ crowded subset of transactions waiting for block inclusion, while allowing
 other transactions to be processed for block inclusion.
 
 In this way, under-capitalized usage demand like spl-token transfers among
-wallets will still find its way for timely block inclusion, accomplishing this
-mechanism's advertised _fairness_ in the sense of each isolated local fee
-markets even at the time of unrelated financial market's volatility, which
-entices well-capitalized usage demand in DeFi activities on chain. 
+wallets will still find its way for _timely_ block inclusion, accomplishing
+this mechanism's advertised _fairness_ in the sense of blockspace and
+transaction fee isolation for each independent local fee markets even at the
+time of unrelated financial market's volatility, which entices well-capitalized
+usage demand in DeFi activities on chain. 
 
 This rate-limiting gets enforced only when the cluster deemed to be congested.
 Also, those increased `cu_price`s will be decreased exponentially as soon as
--- a/proposals/0061-fair-congestion-control.md
+++ b/proposals/0061-fair-congestion-control.md
@@ -68,7 +68,7 @@ address base fee:
 reserve <=> reward
 requested <=> required
 
-## High-Level Description with Example
+## High-Level Description
 
 This proposal tries to localize congestions by means of increasing minimum
 required `cu_price`s for each write-locked addresses, which is newly introduced
@@ -90,10 +90,11 @@ Also, those increased `cu_price`s will be decreased exponentially as soon as
 its address-specific CUs remain to be unchanged due to the temporal stalemate
 of relevant transaction processing.
 
-On top of the direct appreciation of fairness in congestion control for heavily
-write-locked addresses, this proposal also obsoletes both the existing
-block-wide CU limit and the account-write CU limit to overcome their inherent unfair ness
-
+On top of the direct appreciation of fairness, this proposal also obsoletes
+both the existing block-wide CU limit and the account-write CU limit to
+overcome their inherent unfairness and problems. Towards that end, rather
+drastic technical and economical changes are needed for this proposal as
+described below.
 
 ## Detailed Design
 
--- a/proposals/0061-fair-congestion-control.md
+++ b/proposals/0061-fair-congestion-control.md
@@ -86,10 +86,13 @@ transaction fee isolation for each independent local fee markets even at the
 time of unrelated financial market's volatility, which entices well-capitalized
 usage demand in DeFi activities on chain. 
 
-This rate-limiting gets enforced only when the cluster deemed to be congested.
-Also, those increased `cu_price`s will be decreased exponentially as soon as
-its address-specific CUs remain to be unchanged due to the temporal stalemate
-of relevant transaction processing.
+This rate-limiting gets enforced only when the cluster deemed to be congested,
+meaning no idling block space when not congested. Also, those increased
+`base_cu_price`s will be decreased exponentially as soon as its
+address-specific CUs remain to be unchanged due to the temporal stalemate of
+relevant transaction processing. Collectively, each addresses get equal amount
+of opportunity to execute transactions in round-robin fashion, assuming no
+priority-fee premium is paid to interrupt the orderly processing.
 
 On top of the direct appreciation of fairness, this proposal also obsoletes
 both the existing block-wide CU limit and the account-write CU limit to
--- a/proposals/0061-fair-congestion-control.md
+++ b/proposals/0061-fair-congestion-control.md
@@ -120,7 +120,7 @@ processing model is transactions are linearized by the order of appearance in le
 
 ### Congestion Detection
 
-thread count
+thread count or runnable/scheduled tx queue len?
 
 
 ### Nonconflicting transaction group
--- a/proposals/0061-fair-congestion-control.md
+++ b/proposals/0061-fair-congestion-control.md
@@ -34,40 +34,6 @@ obvious base fee manipulation with such short intervals.
   auction can't guarantee the scheduling deadline of casual payment txes (which
   needs 99.99% sub-second confirmation at very minimum).
 
-## Alternatives Considered
-
-Related proposals:
-
-(TODO: add any relation of this to them)
-
-dynamic base fees
-https://github.com/solana-foundation/solana-improvement-documents/pull/4
-
-program rebatable account write fees:
-https://github.com/solana-foundation/solana-improvement-documents/pull/16
-
-asynchronous program execution:
-https://github.com/solana-foundation/solana-improvement-documents/pull/45
-
-increase prioritization fee:
-https://github.com/solana-foundation/solana-improvement-documents/pull/50
-
-bankless
-https://github.com/solana-foundation/solana-improvement-documents/pull/5
-
-## New Terminology
-
-Is there any new terminology introduced with this proposal?
-
-casual tx:
-fairness:
-block fullness in terms of number of actively-execution threads
-dark/filler tx: 
-tx base fee:
-address base fee:
-reserve <=> reward
-requested <=> required
-
 ## Overview of the Design
 
 This proposal tries to localize congestions by means of increasing minimum
@@ -196,3 +162,38 @@ Why should we not do this?
 
 Does the feature introduce any breaking changes? All incompatibilities and
 consequences should be listed.
+
+## Alternatives Considered
+
+Related proposals:
+
+(TODO: add any relation of this to them)
+
+dynamic base fees
+https://github.com/solana-foundation/solana-improvement-documents/pull/4
+
+program rebatable account write fees:
+https://github.com/solana-foundation/solana-improvement-documents/pull/16
+
+asynchronous program execution:
+https://github.com/solana-foundation/solana-improvement-documents/pull/45
+
+increase prioritization fee:
+https://github.com/solana-foundation/solana-improvement-documents/pull/50
+
+bankless
+https://github.com/solana-foundation/solana-improvement-documents/pull/5
+
+## New Terminology
+
+Is there any new terminology introduced with this proposal?
+
+casual tx:
+fairness:
+block fullness in terms of number of actively-execution threads
+dark/filler tx: 
+tx base fee:
+address base fee:
+reserve <=> reward
+requested <=> required
+
--- a/proposals/0061-fair-congestion-control.md
+++ b/proposals/0061-fair-congestion-control.md
@@ -71,13 +71,13 @@ requested <=> required
 ## Overview of the Design
 
 This proposal tries to localize congestions by means of increasing minimum
-required `cu_price`s for each write-locked addresses, which is newly introduced
-by this proposal. This increase will be done exponentially measured by the CU
-consumed by each addresses at the moment. This means a transaction must cost
-the sum of `requested_cu * base_cu_price` for all of its write-locked addresses
-at least. This results in selectively pricing out crowded subset of
-transactions waiting for block inclusion, while allowing other transactions to
-be processed for block inclusion.
+required `cu_price`s for each write-locked addresses (= `base_cu_price`), which
+is newly introduced by this proposal. This increase will be done exponentially
+measured by the CU consumed by each addresses at the moment. This means a
+transaction must cost the sum of `requested_cu * base_cu_price` for all of its
+write-locked addresses at least. This results in selectively pricing out
+crowded subset of transactions waiting for block inclusion, while allowing
+other transactions to be processed for block inclusion.
 
 In this way, under-capitalized usage demand like spl-token transfers among
 wallets will still find its way for timely block inclusion, accomplishing this
--- a/proposals/0061-fair-congestion-control.md
+++ b/proposals/0061-fair-congestion-control.md
@@ -78,21 +78,55 @@ high-achieving objective, rather drastic technical and economical changes are
 needed for the introduction of this consensus-level congestion control as
 described below in detail.
 
-## Example
+## Detailed Design
 
-For simplicity, assume all transactions request 100k CU and always succeed after consuming 100k CU.
+### Introductory example
 
-Available transactions for block inclusion:
+Due to this proposal's inherent complexities, the following very-simplified
+example is presented at first before diving into each design topic:
 
-- `Tx1_AB@9` (read: transaction numbered as `1`, which write-locks address `A` and `B` with `cu_price == 9`)
+Assume all transactions request 100k CU and always succeed after consuming 100k
+CU. Also assume the cluster is always congested.
+
+Available transactions for block inclusion (= `bufferd transactions`):
+
+- `Tx1_AB@9` (read: transaction numbered as `1`, which write-locks both address `A`
+  and address `B` with `cu_price == 9`)
 - `Tx2_A@8`
 - `Tx3_C@8`
 
 (1) 
 
-## Detailed Design
+#### update freq
+
+why block interval is bad?
+  this proposal's primary objective is for congestion control, not like EIP 1559's price discovery
+  so non-interactiveness is desired for little or no room of active auctioning
+  so, quick and acute is desired much like tcp's rwin.
 
 ### Incentive alignment
+
+
+
+
+
+
+### `base_cu_price` Calculation and block reward adjustment
+
+processing model is transactions are linearized by the order of appearance in ledger entries.
+
+### `reserved_fee` calculation and its accruing
+
+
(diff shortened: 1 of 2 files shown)


proposals/0061-fair-congestion-control-with-intra-block-exponential-local-base-fee.md
```diff
@@ -0,0 +1,127 @@
...
+fairness:
+block fullness in terms of number of actively-execution threads
+
+## Detailed Design
+
+(i jotted this down in 10min before going to bed! pardon for being so random
+writings...)
+
+to *determiniscally* define active thread count (`TC_a`), additionally record
+transaction termination events into poh stream.
```
apfitzge, line 72: How does this give us thread count? PoH recording is single-threaded.

mschneider: this opens up for a special type of MEV where the leader can artificially inflate the base fee, e.g. mango's validator could increase the base fee for zeta traders.

proposals/0061-fair-congestion-control-with-intra-block-exponential-local-base-fee.md
```diff
@@ -0,0 +1,127 @@
...
+will naturally starts ceiling up. That's unavoidable no matter what.
+
+## Impact
+
+How will the implemented proposal impacts dapp developers, validators, and core contributors?
+
+## Security Considerations
+
+What security implications/considerations come with implementing this feature?
+Are there any implementation-specific guidance or pitfalls?
```
mschneider, line 118: this creates opportunity for a new type of MEV where an actor purposefully schedules transactions in a way that creates larger base fees. mango's validator could try to increase the fees for zeta to gain monetary benefit through attracting traders with lower fees while using another exchange.

ryoqun: hey, thanks for jumping in.

the actual base fee calculation will be defined so that these arbitrary re-ordering of txes won't result in higher base fees, and larger total (base + priority) fees.

so, given 10 zeta honest transactions, if naive scheduling results in 1000 lamports are burnt, any reordering also results in 1000 lamports.

the base fee in this proposal is intended to throttle particularly hot address's tx throughput as the first objective.

also, tx specifies fixed base fee and fixed priority fee and it's 100% paid.

proposals/0061-fair-congestion-control-with-intra-block-exponential-local-base-fee.md
ryoqun: hey, thanks for jumping in.

like this?

serialized new ledger entries:
1. SanitizedTransactionA
2. TerminationMarkerForTransactionA
3. SanitizedTransactionB
4. SanitizedTransactionC
5. TerminationMarkerforTransactionB
6. TerminationMarkerForTranactiionC



```
 ..1: TC_a: 0
1..2: TC_a: 1
2..3: TC_a: 0
3..4: TC_a: 1
4..5: TC_a: 2
5..6: TC_a: 1
6.. : TC_a: 0
```

this proposal assumes a single threaded banking scheduler, not thread-local-MI.

so, banking threads newly start to record transaction terminations via the scheduler thread. and single threaded poh can be able to linearize these events without ambiguity and replay stage can reconstruct `TC_a` just given list of these ledger entries.

proposals/0061-fair-congestion-control-with-intra-block-exponential-local-base-fee.md
```diff
@@ -0,0 +1,127 @@
...
+`TC_stake_weighted`. (i.e. as if validator stuffed spam txes to capture the
+exessive part (2) of prirotiy fee)
+
+in this way, there's no meaning to spam blocks by leaders. at the same time,
+it's still incentivied to pack txes according to p.f. desceding order, because
+leaders and clients alike are want to increse of their single threaded tps)
+
+also requested fee is basis for fee cals, block fullness calc, not the actual
+cu.
+- to prevent bad behavior, rebate 50% of (requested CU - actual CU)?
```
apfitzge, line 100: This is in conflict of move toward asynchronous execution

proposals/0061-fair-congestion-control-with-intra-block-exponential-local-base-fee.md
apfitzge: It seems as a greedy leader I can lie about the termination in order to maximize parallelism, in order to raise fees. Seems like it'd be easy to just keep a cache of the terminations and just record them right at the end of the slot.

Even with conflicts, could just record & (remove from cache) cached tx-terminations that are in conflict.

proposals/0061-fair-congestion-control-with-intra-block-exponential-local-base-fee.md
apfitzge: That's not necessarily a problem; if we decided that this proposal/approach outweighs the benefits of asynchronous execution (I know you have thoughts on that!).

This was just a comment to raise awareness on the conflict with other outstanding SIMDs: #5, #45

proposals/0061-fair-congestion-control-with-intra-block-exponential-local-base-fee.md
mschneider: you can still re-order across slots

ryoqun: no worries. i haven't got time to write details, but i think this proposal can address this just like intra-block malicious reorderings.

as a teaser, a general idea is that base fee is reset across block boundaries unlike eip-1559. and raising base fees like that won't help because these hot addresses are forced to throttle executions according to new consensus logic. so, best strategy for leaders is spread out hot txes as much as possible across ledger ticks.

aeyakovenko:
@ryoqun why is it necessary?  Digital systems aren't heat engines, they can safely run at 100% capacity.  

The problem with automatic increases is that 1 single use case will increase the price of using the chain for everything.  If there is spare capacity, I would expect the capacity to filed up with opportunistic users that use the chain up to price $X per tx.  That doesn't mean they want to price out anyone else that arrives first at $X, they simply wan to use up spare capacity.  There is nothing wrong with that kind of user, and because of defi, there is basically always demand for any slack capacity.  If the floor price is forced to automatically increase until that user is priced out, that means that any chain that has defi is going to be more expensive then a dedicated chain for payments, or some other non defi use case, as defi will always consume slack capacity.

proposals/0061-fair-congestion-control-with-intra-block-exponential-local-base-fee.md
```diff
@@ -0,0 +1,127 @@
...
+
+also, derive stake-weighted average transaction execution thread count
+(`TC_stake_weighted`).
+
+so, full is defined as the duration when  `TC_a == TC_stake_weighted` (this is
+updated at ~10ms intervals).
+
+when not full, maximize throughput of each of any single threaded transaction
+executions. note that, this mode exponentially cools down any hot addresses if
+any.
```
ryoqun, line 82: here

ryoqun:
> @ryoqun why is it necessary? Digital systems aren't heat engines, they can safely run at 100% capacity.
> 
> The problem with automatic increases is that 1 single use case will increase the price of using the chain for everything. If there is spare capacity, I would expect the capacity to filed up with opportunistic users that use the chain up to price $X per tx. That doesn't mean they want to price out anyone else that arrives first at $X, they simply wan to use up spare capacity. There is nothing wrong with that kind of user, and because of defi, there is basically always demand for any slack capacity. If the floor price is forced to automatically increase until that user is priced out, that means that any chain that has defi is going to be more expensive then a dedicated chain for payments, or some other non defi use case, as defi will always consume slack capacity.
> 
> The challenge for us is to construct a fee market for multiple use cases where no use case would have a better price in a dedicated chain. By aggregating resources and sharing the cost of the validation, they should all actually enjoy a cheaper price assuming there is equal capacity. That means the state of 1 usecase can't impact global pricing.


hey, thanks for jumping in while this is still in very much draft.. as i noted here: https://github.com/solana-foundation/solana-improvement-documents/pull/61/files#r1263813462, slack capacity will be fully utilized and local base fee won't be increased by this proposal:

> when not full, maximize throughput of each of any single threaded transaction
> executions.

so, automatic increase comes into effect only if there's more than concurrently executable series of transactions than the given cluster's super-majority can timely replay with their multi-core hardware.

other than that, i strongly agree with your comment. this will be my best to achieve the challenge under these mentioned nuances.

> That doesn't mean they want to price out anyone else that arrives first at $X, they simply wan to use up spare capacity. There is nothing wrong with that kind of user, and because of defi, there is basically always demand for any slack capacity.

yeah, these kind of transactions will be categorized as _idling_ prioritized transactions. These transactions will only start to compete with each other when there are more than that aforementioned cluster's hw capacity.

proposals/0061-fair-congestion-control-with-intra-block-exponential-local-base-fee.md
ryoqun: quick counterargument for async exec. is unpredicatable latency for simple payments amid nft mint / volatile market. i need to grok the details, but i think merchant can't exclude the possibility of classic double-spend when some delay is occurring in async exec, maybe?

proposals/0061-fair-congestion-control-with-intra-block-exponential-local-base-fee.md
ryoqun: yeah, so, this proposal strongly encodes disincentivation of parallelism manipulation. as you pointed out  base fee is indeed raised, but throughput is forcibly limited. the rate limit is adjusted to that total transaction fee after manipution would never greater than scheduled normally.

in this proposal, cu & active thread are both available for determisnic replaying.

proposals/0061-fair-congestion-control-with-intra-block-exponential-local-base-fee.md
ryoqun: also, market makers don't like async exec as far as i had chat with subject-matter experts at #16 , right? :)

btw, if not, i can strongly push my scramble tx thing: https://github.com/solana-labs/solana/pull/23837

apfitzge: I don't see the discussion of async in #16, can you link? There's several hundred comments now, and after expanding all and searching I'm not seeing any "aync"

mschneider: i remember we discussed the following issue with async (but in other SIMD):
estimating CU is terribly hard for HFT, they often need to over-request because failing TX are way worse than slightly more expensive gas fees, unused CU accounting for the purpose of block packing requires a sync leader

ryoqun: > > also, market makers don't like async exec as far as i had chat with subject-matter experts at https://github.com/solana-foundation/solana-improvement-documents/pull/16 , right? :)
> 
> I don't see the discussion of async in https://github.com/solana-foundation/solana-improvement-documents/pull/16, can you link? There's several hundred comments now, and after expanding all and searching I'm not seeing any "aync"

there's no explicit mention about `async`. but, i generally sensed that praw is so deadly wanted because market maker want to refresh their orders periodically / reliably to begin with.

ryoqun: > estimating CU is terribly hard for HFT, they often need to over-request because failing TX are way worse than slightly more expensive gas fees

yeah, i want to improve the situation as well.... however, this is very difficult. used-cu based fee calc creates no incentive for users to estimate their cu correctly. on the hand hand, portion of fees of unused cu can't be rebated to users as well. That's because the rest of it would still go to the leaders partially, meaning leaders are incentivized to reorder txes to fail.

> unused CU accounting for the purpose of block packing requires a sync leader

... also unused cu can't be utilized for bankingstage scheduling (block packing). this also would create an incentive for leaders to make txes to fail by reordering for more packed block. however, unused CU accounting will be fully exploited for fastest replayingstage scheduling (block verifying).

btw, i prefer a sync leader for a different reason. :)

ryoqun:
fyi, i just pushed a simulation rust code. i need to write down proper wordz from it...

mschneider:
I’ll try to summarize what I understood, I must admit I’m not sure I can fully follow:
1. Validators vote on available core count, basically there’s another native oracle on chain.
2. Validators pack multiple threads of blocks as fast as they can execute them locally, there is no more limit to CU per block, CU per account
3. There is another mechanism at play which allows the validator to raise an intra block base fee when a a specific account is write locked while a lot of CU are executed

Would appreciate clarification

ryoqun:
@mschneider yeah, that's almost correct! from now on, i'll try to fill this proposal with proper words. other's high-level perception helps to guide the upcoming arrangement of the detailed design writings.. thanks!

> Validators vote on available core count, basically there’s another native oracle on chain.

at initial stage, this will be some hard-coded value like 4 (most conservative by being aligned with current default non-vote thread count) or 10 (should be safe with my replay scheduler), given current leader's hw.

dunno what do you mean by _vote_. but thread count will be automatically derived/updated at epoch boundaries. no human interaction is needed.
the sim is adaptive even in the current form. and, all constants in simulation is carefully defined to be derivable from past blocks.

proposals/0061-fair-congestion-control.md
```diff
@@ -0,0 +1,258 @@
...
+temporal stalemate of relevant transaction processing. Collectively, each
+addresses get equal amount of opportunity to execute transactions in
+round-robin fashion while contributing to the decrease of other
+`base_cu_price`s, assuming no exponential priority-fee premium is paid among
+users to interrupt the orderly processing with such a high economical
+justification.
+
+On top of the direct appreciation of aforementioned fairness, this proposal
+also obsoletes both the existing block-wide CU limit and the account-write CU
+limit to overcome their inherent unfairness and problems. Also, no global
```
taozhu-chicago, line 71: Will the dynamic `base-cu-price` fully safe-guard reply timeliness (eg <=400ms)? Is it still somewhat desirable to have a hard limit to be safe?

proposals/0061-fair-congestion-control/simulation/src/main.rs
```diff
@@ -0,0 +1,565 @@
...
+        if tx.addrs.is_empty() {
+            return Err(NoAddress);
+        }
+
+        let (is_congested, reset_counter) =
+            match updated_active_tx_count.cmp(&POLICY.congestion_threshold) {
+                Less if self.is_congested => (false, self.reset_counter + 1),
+                Equal | Greater if !self.is_congested => (true, self.reset_counter),
+                _ => (self.is_congested, self.reset_counter),
+            };
```
taozhu-chicago, line 140: Wondering the reason of determine congestion by number of txs instead of accumulated CUs?

proposals/0061-fair-congestion-control.md
```diff
@@ -0,0 +1,258 @@
...
+
+This is realized by means of some incentive tweaks to combat against the
+obvious base fee manipulation with such short intervals.
+
+## Motivation
+
+(still in half-finished...)
+
+- Write lock cu limit is bad (bot can lock out at the very first of block for
+  the entire duration of whole blocktime (400ms)
```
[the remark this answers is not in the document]
ryoqun: (thanks for peeking into this half-based proposal)

> In order to lock out an account for entire slot, wouldn't bot need to time they 10M cu txs are at the top of the queue when a block starts, by paying the highest priority fee for that account?

yes. i think this is problem per se. this creates uneven value of time only around block boundaries. ideally, there should be some measure to unlock the limit if economically desired by users.

> If SIMD https://github.com/solana-foundation/solana-improvement-documents/pull/50 is approved, bot would be economically disincentivized, wouldn't it?

i guess that's limited. with 10M cus, it's just auction by the increments of 10_000 lamports? I think it's still cheap, considering arb's profits could be unbounded.

proposals/0061-fair-congestion-control.md
```diff
@@ -0,0 +1,258 @@
...
+  needs 99.99% sub-second confirmation at very minimum).
+
+## Overview of the Design
+
+This proposal tries to localize congestions by means of increasing minimum
+required `cu_price`s for each write-locked addresses (= `base_cu_price`), which
+is newly introduced by this proposal. This increase will be calculated
+exponentially, measured by the CU consumed by each addresses at the moment.
+This means a transaction must cost the sum of `requested_cu * base_cu_price`
+for all of its write-locked addresses at least. This results in selectively
```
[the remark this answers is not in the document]
ryoqun: > is `base_cu_price` per writ-locked account

yes.

> will/should there be a `base-cu-price` for the block itself (in case the congestion is for block)?

no, there shouldn't be no such thing as written below in the proposal.

proposals/0061-fair-congestion-control.md
```diff
@@ -0,0 +1,258 @@
...
+pricing out crowded subset of transactions waiting for block inclusion, while
+allowing other transactions to be processed for block inclusion.
```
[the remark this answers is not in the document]
ryoqun: assume block is saturated like this: L highly lucrative trading pairs like sol/usdc, btc/usdc, eth/usdc and banking thread N. and `L > N` with 100 market makers quoting 10/s paying 1 cents for priority fee. 

so, normal spl transfers will cost at least 1 cents (assuming quoting and transfer consume around same amount of cu).

also, when market gets volatile for some economic news, normal spl transfers wil need to pay the more priority fee during that period, than normal days. 

I think these are bad.

proposals/0061-fair-congestion-control/simulation/src/main.rs
ryoqun: nice question. and really thanks for even taking look into the sim code. :)

accumulated cu could be a possibility. however, i think active number of running tx is more direct. and accumulated cu is hard to use to define congestion with very short feedback delay and could be skewed due to overpaid txes so far.

proposals/0061-fair-congestion-control.md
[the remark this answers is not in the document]
taozhu-chicago: >  it's still cheap.

it really depends on who you asking 😉

proposals/0061-fair-congestion-control/simulation/src/main.rs
taozhu-chicago: this is definitely an open question. Appreciate for putting forward your thoughts.

ryoqun:
Well, I noticed this idea is broken. closing and won't revisit this....

sybil attack is too easy. anyone can submit cu-heavy fake txes which locks unrelated addresses. so that their real tx's write address get cooled down...

PR #75, opened 2023-10-25
SIMD-0075: Secp256r1 Precompile (Supersedes SIMD-0048)

iceomatic:
I've updated the SIMD to include all the things discussed in conversation on github and the core-technology channel over the past few days.

This includes:
- More specification & detail on the implementation 
- New security considerations
- Citations to relevant documentation

It has become quite a bit more opinionated and therefore requires more discussion.


iceomatic:
Tagging participants of the former discussion around SIMD-0048 for visibility
@samkim-crypto, @Lichtso , @mvines, @ripatel-fd , @sakridge

iceomatic:
> I'd also like to point out https://github.com/guidovranken/cryptofuzz as a dynamic analysis security tool. This fuzzer has found many differences in behavior between secp256k1 ECDSA implementations and can probably be applied to secp256r1 too.

Noted. Will look into incorporating it into testing 🫡

proposals/0048-precompile-for-secp256r1-sigverify.md
```diff
@@ -0,0 +1,273 @@
...
+those services, 
+the user uses their biometrics to sign a message with the stored private key.
+
+While originally intended to solve for password-less authentication in Web2
+applications, WebAuthn and Passkeys also make an excellent
+candidate for on-chain 
+second-factor authentication. Beyond simply securing funds,
+there are also many 
+other potential beneficial abstractions that could make use of the simple UX
+they provide.
```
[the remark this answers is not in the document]
iceomatic: resolved 👍

proposals/0048-precompile-for-secp256r1-sigverify.md
```diff
@@ -0,0 +1,273 @@
...
+
+General Documentation:
+
+[WebAuthn](https://webauthn.io/)
+
+[Passkeys](https://fidoalliance.org/passkeys/)
+
+## Alternatives Considered
+
+We have discussed the following alternatives:
```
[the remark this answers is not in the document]
iceomatic: added 👍

proposals/0048-precompile-for-secp256r1-sigverify.md
```diff
@@ -0,0 +1,273 @@
...
+- P-256
+- P-384
+- P-521
+- ed25519
+
+P-256 is the only one supported by both Android & MacOS/iOS
+(MacOS/iOS being the more
+restrictive of the two), hence the goal being to implement
+secp256r1 signature
+verification
```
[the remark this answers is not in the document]
iceomatic: > **Note: P-256 / secp256r1 / prime256v1 are used interchangably in this document
> as they represent the same elliptic curve. The choice of nomenclature depends on
> what RFC or SEC document is being referenced.**

resolved 👍

proposals/0048-precompile-for-secp256r1-sigverify.md
```diff
@@ -0,0 +1,277 @@
+---
+simd: "0048"
```
jacobcreech, line 2: @iceomatic since SIMD-48 is not living, this would actually be another SIMD-75 that "supersedes" SIMD-48. I'd recommend updating the headers to display this.

iceomatic: Just out of curiosity, how come it's not considered "living"? The original was merged a while ago and we've been working on it since then.

jacobcreech: It's to avoid continually updating a single SIMD, or give the impression that "we can accept now and change everything later" leading to less rigorous review. The [living](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0001-simd-process.md#living) status is only for specific SIMDs that could update a decent amount, like simd-1 or simd-7.

iceomatic: Ah totally fair, makes sense. Will amend, thx 🫡

iceomatic: Resolved 👍

iceomatic:
I have re-written the `Implementation` part of the SIMD to include the pivot into using the `OpenSSL` crate instead of the `p256` crate. Ive also added detailed implementation details as to how sig verify will be accomplished using OpenSSL.

TL;DR: Its ~3x faster than `p256`, is already a dependency in the labs/anza client, and uses underying C code, rather than native rust. 

The idea here is to:

A.  make implementation comparison between labs/anza and Firedancer easier
B. Use an rely on a more reputable and well maintained crate rather than an "unknown" and somewhat new one
C. Have faster/more efficient sigverify 

Let me know what you think

proposals/0048-precompile-for-secp256r1-sigverify.md
```diff
@@ -12,6 +12,10 @@ created: 2023-05-14
 feature: (fill in with feature tracking issues once accepted)
 ---
 
+
+## [DEPRECATED IN FAVOR OF SIMD-0075]
```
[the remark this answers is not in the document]
CriesofCarrots: (Chiming in because I also have a `supersedes` simd in progress.) We can ask how foundation wants to handle, but it doesn't make sense to me to have a document marked as superseding something that doesn't exist. I think you should leave the original simd-0048. @jacobcreech , thoughts?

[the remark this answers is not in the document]
iceomatic: Good point. Let see what the verdict is 👍

proposals/0048-native-program-for-secp256r1-sigverify.md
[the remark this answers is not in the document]
iceomatic: Noted! Seems like most agree on this, will go ahead and restore 👍

proposals/0048-precompile-for-secp256r1-sigverify.md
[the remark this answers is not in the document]
iceomatic: Reverted the commit removing the SIMD and also marked the status as withdrawn
👍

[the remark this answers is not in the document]
jacobcreech: Yes, this is the right approach. We should supersede the previous SIMD, withdraw it as it was never actually implemented, but not remove it.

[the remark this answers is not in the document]
iceomatic: Makes sense 👍 Everything should be correct now ✅

proposals/0048-precompile-for-secp256r1-sigverify.md
```diff
@@ -0,0 +1,274 @@
...
+
+Would enable the on-chain usage of Passkeys and the WebAuthn Standard, and
+turn the vast majority of modern smartphones into native hardware wallets.
+
+By extension, this would also enable the creation of account abstractions and
+forms of Two-Factor Authentication around those keypairs.
+
+## Security Considerations
+
+As [Firedancer](https://github.com/firedancer-io/firedancer) is being developed
```
riptl, line 251: Please remove Firedancer and any mentions by name.
SIMDs should be a timeless document without any implementation specifics.
The compatibility risks are present to anyone who wants to develop a client. 
It would be great to list out potential risk areas (e.g. handling of invalid points). If further research is needed, that should be mentioned too.

riptl (changes requested):

proposals/0048-precompile-for-secp256r1-sigverify.md
iceomatic: Removed any mentions of firedancer 👍 
With regard to the risk areas, shall I add that as a subsection in the `Security Considerations` ?

proposals/0048-precompile-for-secp256r1-sigverify.md
riptl, line 1: Can we leave the original file name intact? IMO, there shouldn't be any changes needed to SIMD-0048 other than marking it as abandoned. @jacobcreech Wdyt?

t-nelson: agreed the superceded document should be retained minimally modified for historical purposes. for the 0015 -> 0118 [supercede](https://github.com/solana-foundation/solana-improvement-documents/pull/118), we've added `supercedes: '0015'` in 0118 and i've proposed adding `superceded-by: '0118'` to 0015.

proposals/0048-precompile-for-secp256r1-sigverify.md
```diff
@@ -0,0 +1,274 @@
...
+accepting a 20 octet Ethereum address and recovery id to recover an
+SEC1 encoded uncompressed point. This is due to the primary aim of the
+program not being to verify ECDSA signatures but to provide parity with
+`ecrecover` on EVM. Conversely, the `ed25519` program, which is
+primarily concerned with verifying ed25519 signatures, utilises the most
+common ed25519 convention of encoding x_p as a single 32 octet string.
+As the goals of the `secp256r1` program are more analogous to those of the
+`ed25519` program, we propose the SEC1 compressed point encoding to conform
+to the most widely-used standard in common ECDSA applications.
+
```
t-nelson, line 141: this section seems superfluous and distracts from the design details

proposals/0048-precompile-for-secp256r1-sigverify.md
```diff
@@ -0,0 +1,274 @@
...
+order of the curve. This introduces "s malleability", allowing an attacker
+to produce an alternative version of `s` without invalidating the signature.
+
+The pitfalls of this in authentication systems can be particularly perilous,
+opening up certain implementations to signature replay attacks over the same
+message by simply flipping the `s` value over the curve.
+
+As the primary goal of the `secp256r1` program is secure signature validation
+for authentication purposes, the precompile should mitigate these attacks
+by enforcing the usage of `lowS` values, in which `s <= n/2`.
```
t-nelson, line 170: assuming this is a restriction beyond what is described in the above standards/rfcs, this is the only detail that belongs in the design. the rest of this section can live in `security considerations`

proposals/0048-precompile-for-secp256r1-sigverify.md
```diff
@@ -0,0 +1,274 @@
...
+
+### Program
+
+ID: `Secp256r1SigVerify1111111111111111111111111`
+
+The program instruction will be composed of the following:
+
+- A first u8 as the count for the number of signatures to check
+- Single byte of padding
+- The following struct serialized, for each signature to verify
```
t-nelson, line 188: please specify this layout as a pseudo-coded struct. avoid assuming implementation language, be explicit about things like alignment, padding, signedness, endianness, bitwidth, (dis)allowed enum values, etc. idea is that someone can read this doc and implement it from scratch in whatever language they like with no other references

applies to the struct below as well

proposals/0048-precompile-for-secp256r1-sigverify.md
```diff
@@ -0,0 +1,274 @@
...
+
+The program logic will be constructed and built using a `verify`
+function, as outlined in
+[sdk/src/precompiles.rs](https://github.com/solana-labs/solana/blob/9ffbe2afd8ab5b972c4ad87d758866a3e1bb87fb/sdk/src/precompiles.rs).
+
+Apart from the signature verification, the remaining
+logic should be constructed analogously to the existing
+[ed25519](https://github.com/solana-labs/solana/blob/master/sdk/src/ed25519_instruction.rs)
+& [secp256k1](https://github.com/solana-labs/solana/blob/9ffbe2afd8ab5b972c4ad87d758866a3e1bb87fb/sdk/src/secp256k1_instruction.rs#L4)
+precompiles.
```
t-nelson, line 213: avoid relying on existing code to describe this feature (though this does suggest the need for some kind of "generic precompiles specification" simd)

proposals/0048-precompile-for-secp256r1-sigverify.md
```diff
@@ -0,0 +1,274 @@
...
+This crate is part of the `Rust Crypto` library and implements the NIST P-256
+curve as well as ECDSA. It conforms with the test vectors found in
+[RFC6979](https://datatracker.ietf.org/doc/html/rfc6979#appendix-A.2.5).
+
+- `p256::ecdsa::VerifyingKey::from_sec1_bytes()`
+- `p256::ecdsa::Signature::from_scalars()`
+- `p256::arithmetic::Scalar::is_high()`
+- `p256::ecdsa::VerifyingKey::verify()`
+
+Note: The crate is well maintained, but has never been externally audited.
```
t-nelson, line 227: these details are up to the implementer and don't belong in this document

proposals/0048-precompile-for-secp256r1-sigverify.md
```diff
@@ -0,0 +1,274 @@
...
+
+### Compute Cost / Efficiency
```
t-nelson, line 229: something else that begs its own simd :thinking: 

@tao-stones wdyt about a short simd for guidance on compute cost estimation?

t-nelson (commented):
this is a good start! please take a look at [rfc2119](https://datatracker.ietf.org/doc/html/rfc2119) and make a pass at this document to correct usage of these key words. there are several SHOULDs that need to be MUSTs

proposals/0048-precompile-for-secp256r1-sigverify.md
tao-stones: > something else that begs its own simd 🤔
> 
> @tao-stones wdyt about a short simd for guidance on compute cost estimation?

like this: https://github.com/solana-foundation/solana-improvement-documents/pull/121

iceomatic: Excellent, thx for this @tao-stones 👍

proposals/0048-precompile-for-secp256r1-sigverify.md
jacobcreech: Yes, let's keep it all intact and just add the `superceded-by` + update status to `withdrawn`

iceomatic:
All fixes/resolutions to open comments should be done. I'm unsure if the pseudo-coded structs and the processing logic are okay in the new format, so please leave feedback wrt to changes if you have any 👍

proposals/0048-native-program-for-secp256r1-sigverify.md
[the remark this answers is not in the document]
iceomatic: Unintentional, my bad. I merged main into my patch branch, but I guess it still took over changes. Will fix 👍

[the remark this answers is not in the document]
iceomatic: Resolved :)

0xSol:
Thanks @iceomatic for updating the SIMD and addressing fixes/resolutions in comments. 
@ripatel-fd, @sakridge, @t-nelson, could you please take another look, and comment / approve as appropriate? Thanks.

proposals/0075-precompile-for-secp256r1-sigverify.md
```diff
@@ -0,0 +1,298 @@
...
+Multiple signatures can be verified. If any of the signatures fail to verify,
+an error must be returned.
+
+If the instruction data is empty, the program must return an error.
+
+If `count == 0` and the length of the instruction data > 1, the program must
+return an error.
+
+If `count == 0` and the length of the instruction data is not at least
+`count * Secp256r1SignatureOffsets + 1`, the program must return an error.
```
ptaffet-jump, line 173: This should be 2 to include the padding byte, right? You should also make it clear that `Array` does not contain any padding between elements or any length prefixes.

```suggestion
If `count > 0` and the length of the instruction data is not at least
`count * sizeof(Secp256r1SignatureOffsets) + 2`, the program must return an error.
```

proposals/0075-precompile-for-secp256r1-sigverify.md
```diff
@@ -0,0 +1,298 @@
...
+}
+```
+
+Multiple signatures can be verified. If any of the signatures fail to verify,
+an error must be returned.
+
+If the instruction data is empty, the program must return an error.
+
+If `count == 0` and the length of the instruction data > 1, the program must
+return an error.
```
ptaffet-jump, line 170: I know existing precompiles have this check, but I think it's unnecessary.

proposals/0075-precompile-for-secp256r1-sigverify.md
```diff
@@ -0,0 +1,298 @@
...
+count = data[0]
+if count == 0 && length_of_data > 1 {
+  return Error
+}
+if length_of_data < (count * SERIALIZED_OFFSET_STRUCT_SIZE + 1) {
+  return Error
+}
+instructions = instruction_datas;
+for i in 0..count {
+    signature = instructions[signature_instruction_index].data[signature_offset..signature_offset+64]
```
ptaffet-jump, line 199: You should add checks for indices and offsets being outside the range to the pseudocode

proposals/0075-precompile-for-secp256r1-sigverify.md
```diff
@@ -0,0 +1,298 @@
...
+/// Secp256r1SignatureOffsets struct
+
+if length_of_data == 0 {
+  return Error
+}
+count = data[0]
+if count == 0 && length_of_data > 1 {
+  return Error
+}
+if length_of_data < (count * SERIALIZED_OFFSET_STRUCT_SIZE + 1) {
```
ptaffet-jump, line 194: Again, this should be a +2, right?

ptaffet-jump (commented):
Some clarifications for the pseudocode

proposals/0075-precompile-for-secp256r1-sigverify.md
iceomatic: So in the case of ```count == 0``` the program passes, irrespective of the instruction data?

proposals/0075-precompile-for-secp256r1-sigverify.md
iceomatic: absolutely correct, fixing 👍 

will also note the lack of length prefix and padding between elements, thanks for the suggestion 👍

proposals/0075-precompile-for-secp256r1-sigverify.md
iceomatic: again correct, fixing. thx 👍 
not sure how I didn't notice this...

proposals/0075-precompile-for-secp256r1-sigverify.md
ptaffet-jump: That would make sense to me, but I'd be interested in hearing from the folks who made this design decision for the other precompiles to make sure I'm not missing something.

iceomatic: Will leave this unresolved until we get some more input from the other contributors 🫡

iceomatic:
Thanks @ptaffet-jump for the input, all comments with the exception of the ```count == 0 && length_of_data``` behaviour should be resolved 👍

proposals/0075-precompile-for-secp256r1-sigverify.md
samkim-crypto: For this issue, I think it makes sense either way. This predates me but https://github.com/solana-labs/solana/pull/19300 has the justification for the count, which I think is reasonable.

samkim-crypto:
Looks good to me!

It is worth noting (I apologize if this has already been discussed before) that there is going to be an inconsistency in the way ecdsa malleability is handled with the existing secp256k1 [implementation](https://github.com/anza-xyz/agave/blob/master/sdk/src/secp256k1_instruction.rs#L149), which does not handle malleability.

sakridge (approved):

proposals/0075-precompile-for-secp256r1-sigverify.md
tlambertz: > That would make sense to me, but I'd be interested in hearing from the folks who made this design decision for the other precompiles to make sure I'm not missing something.

This is done as a safeguard for smart-contracts. There have been a number of contracts that wanted to verify a single signature, and thus omitted the check for count.
They just used instruction introspection to assert that the correct data is present in the call to secp verify.

An attacker could thus call secp verify with a count of 0, and pretend to have a valid signature for arbitrary data.

This is obviously a bug in the smart-contract using secp, but having this count=0 check made for an easy fix to prevent this issue from cropping up. It makes no sense to call with a zero count and non-zero data after all.

iceomatic: I guess as there's no document or SIMD detailing otherwise, I'd we just stay in line with [#19300](https://github.com/solana-labs/solana/pull/19300).

Thoughts @ptaffet-jump ?

iceomatic:
> Looks good to me!
> 
> It is worth noting (I apologize if this has already been discussed before) that there is going to be an inconsistency in the way ecdsa malleability is handled with the existing secp256k1 [implementation](https://github.com/anza-xyz/agave/blob/master/sdk/src/secp256k1_instruction.rs#L149), which does not handle malleability.

Yep. In the description of the SIMD we made our case as to why it would be advantageous to handle malleability, but if contributors disagree I can happily remove that check to put it in line with the other precompiles 
@samkim-crypto

iceomatic:
With the approval from Anza we're just missing an approval on the FD side now 👍

0x0ece:
I'd like to recommend the following 3 changes:

1. For the data structure, let's use the same as Ed25519.
https://github.com/anza-xyz/agave/blob/v1.18.15/sdk/src/ed25519_instruction.rs#L22-L30
This is because 1) Ed25519 is newer than Secp256k1, 2) it's best suited for JS impl, 3) since Webauthn supports both Ed25519 and r1, it'll make it easier to develop client libs.
Concretely, this means switching to 2-byte instruction ids.

2. If count==0, return Error.
This simplifies the checks and avoid unnecessary edge cases.
Independently, I'm going to propose a SIMD to "clean up" existing precompiles and adopt this same behavior.

3. Reuse the same `get_data_slice` as Ed25519.
(BTW, I personally prefer the approach outlined in this SIMD where the precompile gets data slices, and then the sigverify function is responsible for parsing data. Unlike existing precompiles that interleave fetching and parsing, that's very implementation specific.)

Other than these 3 minor changes, everything looks good to me and I can approve on the FD side.

iceomatic:
Adjusted the pseudocode and the data structs in accordance with @0x0ece 's comments and [SIMD-0152](https://github.com/solana-foundation/solana-improvement-documents/pull/152)

proposals/0075-precompile-for-secp256r1-sigverify.md
```diff
@@ -0,0 +1,382 @@
...
+
+/// SERIALIZED_PUBLIC_KEY_LENGTH and SERIALIZED_SIGNATURE_LENGTH represent the 
+/// length of the serialized public key and signature respectively
+
+function verify() {
+  if length_of_data == 0 {
+    return Error
+  }
+  num_signatures = data[0]
+  if num_signatures > 8 {
```
0x0ece, line 224: This is unnecessary as it's checked before running precompiles.

proposals/0075-precompile-for-secp256r1-sigverify.md
```diff
@@ -0,0 +1,382 @@
...
+      data_position += SERIALIZED_OFFSET_STRUCT_SIZE
+
+      signature = get_data_slice(all_tx_data,
+                                offsets.signature_instruction_index,
+                                offsets.signature_offset
+                                signature_length)
+      if !signature {
+        return Error
+      }
+      if signature_S == highS {
```
0x0ece, line 248: My personal preference is to not mix memory issues with logic issues, checking for highS can be done inside `sigverify`

proposals/0075-precompile-for-secp256r1-sigverify.md
```diff
@@ -0,0 +1,382 @@
...
+      // sigverify includes validating signature and public_key
+      result = sigverify(signature, public_key, message)
+      if result != Success {
+        return Error
+      }
+    }
+    return Success
+}
+
+fn get_data_slice(all_tx_data, instruction_index, offset, length) {
```
0x0ece, line 277: This should be reused across precompiles.

0x0ece (approved):
LGTM. I left some comments for whoever is going to develop this, but the spec looks good!

proposals/0075-precompile-for-secp256r1-sigverify.md
iceomatic: resolved 👍

proposals/0075-precompile-for-secp256r1-sigverify.md
iceomatic: resolved 👍

proposals/0075-precompile-for-secp256r1-sigverify.md
iceomatic: added a comment to make re-use and reference to SIMD-0152 explicit 👍

iceomatic:
Have approval on the FD side, re-requesting approval from Anza 🫡
@sakridge @samkim-crypto

samkim-crypto (approved):
Looks fine to me!

sakridge (approved):

iceomatic:
Sweet, we now have approvals from both Anza and FD 🙌
Kindly asking for a merge 🫶

@joncinque @jacobcreech

proposals/0075-precompile-for-secp256r1-sigverify.md
```diff
@@ -0,0 +1,381 @@
...
+
+## Summary
+
+Adding a precompile to support the verification of signatures generated on
+the secp256r1 curve. Analogous to the support for secp256k1 and ed25519
+signatures that already exists in form of the
+`KeccakSecp256k11111111111111111111111111111` and
+`Ed25519SigVerify111111111111111111111111111` precompiles.
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL
```
jacobcreech, line 24: Why is this a part of the SIMD? This is from the template as guidance on how to write the SIMD.

iceomatic: Trent wanted me to reference RFC 2119 and it mandates the following: 
`Authors who follow these guidelines
 should incorporate this phrase near the beginning of their document:`

iceomatic: Nudging this, would love to sort this out @jacobcreech 🤝

jacobcreech (approved):
Looks like we've had approvals from current core maintainers as well as ample time to make any additional comments. Thanks @iceomatic for championing this!

PR #79, opened 2023-10-26
Added allow_commission_decrease_at_any_time proposal.

bji:


jacobcreech:
Hey @bji, was this discussed at all in the Solana Tech discord? I don't see any discussion or coming to consensus that this change should be made into a SIMD.

CriesofCarrots:
@jacobcreech , the discussion happened on github: https://github.com/solana-labs/solana/pull/33847#issuecomment-1779973970

CriesofCarrots:
@jacobcreech , should this be SIMD 79 because it's github issue number 79? Is the numbering scheme articulated somewhere? Sorry if I missed it.

jacobcreech:
> @jacobcreech , should this be SIMD 79 because it's github issue number 79? Is the numbering scheme articulated somewhere? Sorry if I missed it.

Yes, it should be 79 according to https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0001-simd-process.md#draft

bji:
I don't know how I got 80.  I swear I read issue #80 somewhere, but I guess I made a mistake.  My apologies.  I will rename.

proposals/0080-allow_commission_decrease_at_any_time.md
```diff
@@ -0,0 +1,89 @@
...
+stake to a validator while still charging 0% commission.  This is not considered
+to be a technique harmful to stakers.
+
+## Detailed Design
+
+A feature will be added which will enable set-commission logic to first check
+whether the proposed commission change is a decrease, and if so, do not invoke
+the "only allow commission change in first half of epoch" rule.
+
+In order to facilitate testing, all of the functionality for handling the
```
AshwinSekar: lets avoid client specific details in SIMD and move this paragraph to the solana labs PR.

bji: Fixed in most recent version by describing the general functionality that vote programs must implement without specifically mentioning any code base.

proposals/0079-allow_commission_decrease_at_any_time.md
```diff
@@ -0,0 +1,83 @@
...
+commission to 0% to ensure that stake accounts are not disadvantaged by my
+actions. Not being allowed to do this in the second half of an epoch is a real
+problem for me.
+
+Allowing commission changes during the second half of epochs also enables
+"reverse rug pulls" where a validator has commission set to 100% during the
+majority of an epoch but reduces it to 0% across the epoch boundary.  It's not
+clear why some validator operators choose to do this, but it does not seem
+harmful, and there does appear to be motivation for some validators to do this
+as it's been done numerous times in the past.
```
joncinque, line 40: Well now this makes me worry about a potential security issue for making the change! I don't think there is a problem either, but I'll definitely think it over.

The only one I can come up with: when an entity provides validator-operation-as-a-service for whales, I believe they typically run 100% commission validators and pay out the inflation rewards through some separate channel. If they can "reverse rug", then they can artificially decrease the amount that goes back to their client.  Since they already have a separate channel for setting up this agreement though, I don't think it's a problem.

proposals/0079-allow_commission_decrease_at_any_time.md
```diff
@@ -0,0 +1,83 @@
...
+will diverge.
+
+## Motivation
+
+Some validator operators may need to decrease commission in order to satisfy
+their own operational criteria.  Speaking personally, when I have an operator
+error that results in reduced stake account rewards for the epoch, I reduce
+commission to 0% to ensure that stake accounts are not disadvantaged by my
+actions. Not being allowed to do this in the second half of an epoch is a real
+problem for me.
```
joncinque, line 33: This is extremely useful, thanks for providing the story. As a nit, please clarify who "I" is, so that the use case becomes clear to anyone reading the proposal.

proposals/0079-allow_commission_decrease_at_any_time.md
```diff
@@ -0,0 +1,83 @@
...
+category: Standard
+type: Core
+status: Draft
+created: 2023-10-26
+feature: TBD
+---
+
+## Summary
+
+The commission_updates_only_allowed_in_first_half_of_epoch feature disallows
```
joncinque, line 15: Please link the feature here https://github.com/solana-labs/solana/issues/29361

proposals/0079-allow_commission_decrease_at_any_time.md
```diff
@@ -0,0 +1,83 @@
...
+## New Terminology
+
+Reverse Rug Pull: Setting commission to 100% for most of an epoch, and to 0%
+only across epoch boundaries.  This has the effect of discouraging third party
+stake to a validator while still charging 0% commission.  This is not considered
+to be a technique harmful to stakers.
+
+## Detailed Design
+
+A feature will be added which, when enabled, should cause all node
```
joncinque, line 55: nit: this can be more forceful
```suggestion
A feature will be added which, when enabled, must cause all node
```

proposals/0079-allow_commission_decrease_at_any_time.md
```diff
@@ -0,0 +1,83 @@
...
+implementations' vote program's set-commission instruction handling to first
+check whether the proposed commission change is a decrease, and if so, do not
+invoke the "only allow commission change in first half of epoch" rule.
```
joncinque, line 58: How about relaxing it a tiny bit more and also allowing validators to set the same commission that is currently set? Ie. allowing any number `<=` current commission instead of just `<`

joncinque (commented):
Looks great to me overall! Mostly small comments

proposals/0079-allow_commission_decrease_at_any_time.md
bji: I will do this, but I don't understand the motivation.  Allowing the commission to be set to the same value is a no-op, and I don't know why allowing no-ops to succeed has value.  I think it would be more valuable for no-ops to fail as an indication to whoever submitted the tx that they probably didn't do what they *thought* they were trying to do.

proposals/0079-allow_commission_decrease_at_any_time.md
bji: I have altered the document to no longer contain first-person wording.  I describe the generalized operational technique now.

proposals/0079-allow_commission_decrease_at_any_time.md
bji: Agreed, there's no concern about that.

AshwinSekar (approved):
lgtm

proposals/0079-allow_commission_decrease_at_any_time.md
joncinque: In some cases that definitely makes sense, but I think your point addresses a separate issue than this SIMD, which is about restoring previous behavior in certain cases.

If we wanted to implement the fail-on-noop behavior, then my take would be to *always* do it. But that would probably be better addressed in a separate SIMD, which I am *not* advocating for right now

proposals/0079-allow_commission_decrease_at_any_time.md
joncinque: Oh sorry, I think we had a misunderstanding -- I meant to mention the feature for `commission_updates_only_allowed_in_first_half_of_epoch` here. The feature that you mention in line 10 should be the *new* feature added for your change, not the previous one.

I created a feature gate issue for it which you can use in line 10: https://github.com/solana-labs/solana/issues/33994

joncinque:
We just need someone from Firedancer to sign off too. I posted in the Discord firedancer channel about it.

proposals/0079-allow_commission_decrease_at_any_time.md
```diff
@@ -0,0 +1,88 @@
...
+## Alternatives Considered
+
+No alternatives were considered.
+
+## New Terminology
+
+Reverse Rug Pull: Setting commission to 100% for most of an epoch, and to 0%
+only across epoch boundaries.  This has the effect of discouraging third party
+stake to a validator while still charging 0% commission.  This is not
+considered to be a technique harmful to stakers.
```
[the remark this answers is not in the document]
bji: I didn't realize that the purpose of this section was to introduce "official" terminology (not even sure what "official" means here) - I thought it was just to introduce terms that are used or might be used in discussion of the feature that are not otherwise commonly known.  However, I will remove this section as it doesn't accomplish much in this SIMD (was just trying to provide some background).

proposals/0079-allow_commission_decrease_at_any_time.md
```diff
@@ -0,0 +1,88 @@
...
+this in the second half of an epoch is a problem because it would prevent that
+commission change until the next epoch, which will not allow this policy to
+take effect for stake accounts which were de-activating during that epoch.
+
+Allowing commission changes during the second half of epochs also enables
+"reverse rug pulls" where a validator has commission set to 100% during the
+majority of an epoch but reduces it to 0% across the epoch boundary.  It's not
+clear why some validator operators choose to do this, but it does not seem
+harmful, and there does appear to be motivation for some validators to do this
+as it's been done numerous times in the past.
```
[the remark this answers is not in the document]
bji: OK I'll remove.

proposals/0079-allow_commission_decrease_at_any_time.md
```diff
@@ -0,0 +1,88 @@
...
+some types of commission changes succeed only in the first half of epochs while
+others succeed always.
+
+## Backwards Compatibility
+
+This feature requires a feature gate because software which includes the
+implementation will allow certain set-commission transactions to succeed where
+software without the implementation would fail those transactions.  Thus all
+validators must be updated to the new functionality at an epoch boundary so
+that all validators agree on the result of executing those transactions.
```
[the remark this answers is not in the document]
bji: OK

proposals/0079-allow_commission_decrease_at_any_time.md
```diff
@@ -0,0 +1,88 @@
...
+check whether the proposed commission change is a decrease or no change, and if
+so, do not invoke the "only allow commission change in first half of epoch"
+rule.
+
+## Impact
+
+Validators will now be able to decrease commission at any time in the epoch,
+but only increase commission in the first half of epochs (because of the
+commission_updates_only_allowed_in_first_half_of_epoch feature already
+implemented).
```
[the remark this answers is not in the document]
bji: I think this is splitting hairs on the purpose of SIMD sections.  I'm going to leave this as-is.

proposals/0079-allow_commission_decrease_at_any_time.md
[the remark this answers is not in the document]
bji: So I'll include the note about the ledger, but I don't think I'll include the note about why it's fully backwards compatible for users of the stake program, since that isn't necessary in a section that is described (in the template) as " All incompatibilities and consequences should be listed."

proposals/0079-allow_commission_decrease_at_any_time.md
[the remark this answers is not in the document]
bji:
```
## Impact

How will the implemented proposal impacts dapp developers, validators, and core contributors?
```

The text included states how the change will impact validators.  There is no impact to the other categories outside of the truism that "any change could affect anyone".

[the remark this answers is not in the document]
jacobcreech: Do we know how many validators have the policy mentioned in the motivation section, or how common this problem is? Something like that would be nice to document as impact, because I agree with Richard that unless we understand that this change does seem "nice to have"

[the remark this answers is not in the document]
bji: Is one good reason not enough?

jumpsiegel (approved):

proposals/0079-allow_commission_decrease_at_any_time.md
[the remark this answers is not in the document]
bji: The original feature which prevented changes to the commission in the second half was an over-reach.  It had a goal ("prevent rug pulls") and it created collateral damage by implementing a more restrictive than necessary policy.

It doesn't seem like the best way to have technical discussions to have to provide justification for whether or not the reasons are "good enough".  What you think is important, someone else might not.  It's necessary to include importance in prioritization of effort, but when discussing technical merits of a proposal, I think justification just should be a minor factor, and having "one good reason" should be good enough.

In this case, I've already given a justification of why this feature is important *to me*.

I can imagine other scenarios where it might matter.  Will these sway anyone?  Or is a back and forth discussion about whether "enough reasons have been imagined yet" going to be a part of the process?

- Someone may accidentally set a high commission in the first half of epoch, and not realize it until the second half.  It would be better if they could correct their mistake than if they were prevented from correcting it.

- Someone may set a high commission in the first half of the epoch but change their mind and want to change it back down in the second half.  It would be better if they could make their change rather than being restricted without reason from doing so.

- Someone may receive a command from their government or other authority notifying them that if they accept commission, they are breaking the law and will go to jail.  If this occurs in the second half of an epoch, and they can't set commission to 0, they will have no way to prevent going to jail since their validator already earned vote credits and cannot be prevented otherwise from earning commission (YES this is super far-out, but since the threshold for someone feeling this feature is justified is arbitrary, I have no idea how far out to go here).

- Someone may run a sophisticated commission selection scheme that sets a 100% commission in the first half of the epoch, and then has their stakers vote on reducing it incrementally until they reach a compromise (there would be some other, not disclosed here, reason for stakers to vote for 'some commission' above 0 instead of just 0).  If commission cannot be set in the second half of epochs, this mechanism may not work properly (YES this is super far out; see the previous paragraph for why)

...

bji:
> Green light from Firedancer side, but considering this low priority.

Thank you; the implementation has already been provided to Solana Labs as a PR so from their perspective it should be an easy thing to take.  Once it's reached consensus within the cluster Firedancer would have to implement it; but I think it's a very easy implementation.

PR #82, opened 2023-11-02
SIMD-0082: Relax Transaction Constraints

apfitzge:
The first in a series of SIMDs aimed at the relaxation of constraints blocking us from (potentially) having asynchronous execution.
This proposal does not lock Solana into moving towards asynchronous execution.
This proposal stands on its own to simplify the protocol and give more flexibility in implementation of both block-production and block-validation.

### DIFF
--- a/proposals/0082-relax-transaction-constraints.md
+++ b/proposals/0082-relax-transaction-constraints.md
@@ -8,8 +8,6 @@ type: Core
 status: Draft
 created: 2023-10-30
 feature:
-supersedes:
-extends:
 ---
 
 ## Summary
--- a/proposals/0082-relax-transaction-constraints.md
+++ b/proposals/0082-relax-transaction-constraints.md
@@ -12,26 +12,37 @@ feature:
 
 ## Summary
 
-This proposal aims to relax some of the constraints on which individual
-transactions can be included in a valid block.
-The proposal does not relax the constraints required for a transaction to be
-executed.
+Transaction errors fall into two categories: protocol violating errors and
+runtime errors.
+Protocol violation errors are those that break some constraint of the protocol,
+and any blocks containing such transactions must be rejected by the network.
+Runtime errors are those that occur during the processing of a transaction, and
+generally these errors mean that state changes are limited to a few accounts,
+but the transaction itself may be included in a block.
+Any block containing a transaction that results in a runtime error is still
+valid.
+This proposal aims to change several protocol violation errors into runtime
+errors, in order to simplify the protocol, and give more flexibility to
+block-producer and block-validator implementations.
 
 ## Motivation
 
 The current protocol places many constraints on the structure and contents
-of blocks; if any constraints are broken, block-validators will mark the block
-as invalid.
+of blocks; if any constraints are broken, block-validators will mark the
+entire block as invalid.
 Many of these constraints are necessary, but some of them are not, and lead to
 additional complexity in the protocol.
 This proposal aims to relax some of the constraints at the individual
 transaction level, in order to simplify the protocol, and give more flexibility
 to block-producer and block-validator implementations.
 
-More specifically, this proposal aims to relax the constraints which require
-account state in order to determine the validity of a transaction's inclusion.
-The reason these constraints are targetted specifically, is that reliance on
-account state necessarily requires synchronous execution within the protocol.
+More specifically, this proposal changes several protocol violation errors into
+runtime errors.
+This means that transactions that would previously be dropped with an error,
+can now be included in a block, but will either not be executed at all, or
+will have limited state changes.
+The goal is to remove much of the reliance on account-state in order to
+validate a block.
 This proposal on its' own, will not enable asynchronous execution, but it will
 remove one of the barriers to asynchronous execution.
 
@@ -40,13 +51,15 @@ remove one of the barriers to asynchronous execution.
 1. Do nothing
     - This is the simplest option, as we could leave the protocol as is.
     However, this leaves the protocol more complex than it needs to be.
-2. Relax all but fee-paying constraint
-    - This was actually the initial proposal, but it was decided that it would
-    be better to relax as many constraints as possible, rather than just some.
-    Any reliance on account state, means the protocol requires synchronous
-    execution.
-    For this reason, the constraint on fee-paying was also decided to be
-    relaxed.
+2. Also relax fee-paying constraint
+    - This was considered, and included in the intially reviewed proposal.
+    However, this was decided to be moved to a separate follow-up proposal,
+    should this one be accepted.
+    There was significant disagreements on how exactly transactions that cannot
+    pay fees should be handled.
+    By keeping fee-paying as a constraint for the current proposal, it will
+    allow this proposal to be accepted more quickly, which will give
+    significant benefits to the network.
 3. Additionally, relax the address lookup table resolution constraint
     - This was considered, since it is a transaction-level constraint that is
     depdendent on account-state. However, due to entry-level and block-level
@@ -60,103 +73,188 @@ None
 
 ## Detailed Design
 
-Specifically, this proposal relaxes constraints that a transaction included in
-a block must:
+### Current Protocol-Violation Errors
 
-1. Have a fee-payer with enough funds to pay fees
-2. Have a valid nonce account, if specified
-3. Have program accounts that:
-   1. exist
-   2. are executable
-4. Have writable accounts that are:
-   1. not executable, unless owned by
-   `BPFLoaderUpgradeab1e11111111111111111111111`
-   2. if owned by `BPFLoaderUpgradeab1e11111111111111111111111`,
-   then `BPFLoaderUpgradeab1e11111111111111111111111` must be included
-   3. if owned by `Stake11111111111111111111111111111111111111`, then the
-   current slot must not be within the epoch stakes reward distribution period
-5. Have executable `BPFLoaderUpgradeab1e11111111111111111111111` owned accounts
-with `UpgradeableLoaderState::Program` state, and derived program data account
-that exist
-6. Have a call chain depth of 5 or less
-7. Have no builtin loader ownership chains
-(pending `4UDcAfQ6EcA6bdcadkeHpkarkhZGJ7Bpq7wTAiRMjkoi`)
-8. Have total loaded data size which does not exceed
-requested_loaded_accounts_data_size_limit
-(pending `DdLwVYuvDz26JohmgSbA7mjpJFgX5zP2dkp8qsF2C33V`)
+Prior to this change, the list of protocol violation errors at the individual
+transaction level are:
 
-The intent with relaxing these constraints is to minimize the amount of
-account-state which is required in order to validate a block. This gives more
-flexibility to when and how account-state is updated during both block
-production and validation.
+1. Transaction must be deserializable via `bincode` (or equivalent) into a structure:
 
-With these constraints removed, there is still one account-state constraint
-at the transaction level: address lookup table resolution.
-With this proposal, this constraint is intentionally not relaxed since it is
-necessary for validation of entry-level and block-level constraints.
-However, if/when those constraints are relaxed, this constraint should be
-relaxed as well.
+    ```rust
+    pub struct VersionedTransaction {
+        #[serde(with = "short_vec")]
+        pub signatures: Vec<Signature>,
+        pub message: VersionedMessage,
+    }
+    ```
 
-The relaxation of these constraints is only relaxed for transactions being
-included in a block. During block validation, if any of these constraints
-are broken, the entire block is marked invalid; with this proposal, the
-block is not marked invalid, but the transaction will not be executed.
-These constraints must still be satisfied in order for the transaction to be
-executed, and have an effect on account state.
-Any transaction included in the block, regardless of whether or not the
-transaction is executed, must be inserted into the status cache.
-Nonced transactions must advance the nonce account, even if the transaction,
-is not executed.
-However, there are some new considerations which must be taken into account,
-specifically as it relates to fees and block limits.
+    where `Signature` and `VersionedMessage` are defined in `solana-sdk`.
+2. Transaction serialized side must be less than or equal to 1232 bytes.
+3. Transaction signatures must be valid and in the same order as the static
+   account keys in the `VersionedMessage`.
+4. Transaction must have exactly the number of required signatures from the
+   `VersionedMessage` header.
+5. Transaction must not have more signatures than static account keys.
+6. The `VersionedMessage` must pass sanitization checks:
+    - The sum of the number of required signatures and the number of read-only
+      unsigned accounts must not be greater than the number of static account
+      keys.
+    - The number of readonly signed accounts must be less than the number of
+      required signatures.
+    - Each lookup table, if present, must be used to load at least one account.
+    - The total number of accounts, static or dynamic, must be less than 256.
+    - Each instruction's `program_id_index` must be less than the number of
+      static account keys.
+    - Each intruction's `program_id_index` must not be the payer index (0).
+    - All account indices in instructions must be less than the number of total
+      accounts.
+7. Transactions that use address look up tables must be resolvable:
+    - The address look up table account must exist.
+    - The address look up table account must be owned by the address look up
+      table program: `AddressLookupTab1e1111111111111111111111111`
+    - The address look up table account data must be deserializable into
+      `AddressLookupTable` as defined in `solana-sdk`.
+    - All account table indices specified in the transaction must be less than
+      the number of active addresses in the address look up table.
+8. Transactions containing pre-compile instructions must pass pre-compile
+   verification checks.
+9. The transaction must not load the same account more than once.
+10. The transaction must have fewer than 64 accounts.
+    - The limit is subject to change to 128 with the activation of
+      `9LZdXeKGeBV6hRLdxS1rHbHoEUsKqesCC2ZAPTPKJAbK`.
+11. The `recent_blockhash` of the transaction message must be valid:
+    - It must exist and not have an age greater than 150.
+    - OR the transaction must be a nonced transaction, and the nonce
+      account must exist and be valid for the given `recent_blockhash`.
+12. The transaction must not have already been processed.
+13. The transaction fee-payer account must:
+    - exist
+    - be owned by the the system program: `11111111111111111111111111111111`
+    - have more lamports than the fee
+    - have more lamports than the fee plus the minimum balance
+14. Any non-writable program accounts used in the transaction must exist,
+    regardless of if they are used in any instructions.
+15. The total loaded data size of the transaction must not exceed
+    `requested_loaded_accounts_data_size_limit`, or the default limit (64MB).
+16. Any account used as a program in a top-level instruction must:
+    - be the native loader: `NativeLoader1111111111111111111111111111111`
+    - OR
+      - exist
+      - be executable
+      - be owned by the native loader: `NativeLoader1111111111111111111111111111111`
+    - OR
+      - exist
+      - be executable
+      - the owner account be owned by the native loader: `NativeLoader1111111111111111111111111111111`
+      - the owner account must be executable
 
-### Fee-Paying
+### Proposed Protocol-Violation Errors
 
-Currently, if a transaction's fee-payer does not have enough funds to pay the
-fee, the transaction cannot be included in a block. With this proposal, it is
-possible for such a transaction to be included in the block, and there are
-four different possibilites for what happens to the fee-payer account:
+The proposal is to move some items of the above list from protocol violations to runtime errors.
+Specifically, the following constraints will remain as protocol violations:
 
-1. The fee-payer account does not exist (0 lamports)
-2. The fee-payer account does not have enough funds for the entire fee
-3. The fee-payer account has enough funds for the fee, but would no longer be
-rent-exempt
-4. The fee-payer account has enough funds for the fee, and would still be
-rent-exempt after paying the fee
+1. Transaction must be deserializable via `bincode` (or equivalent) into a structure:
 
-In case 1, the transaction should simply be ignored for execution, and have no
-effect on state.
-In case 2, the fee-paying account should be drained of all funds, with fees
-being paid to the block producer, and have no other effect on state.
-In case 3, the fee-paying account should be drained of all funds, with fees
-being paid to the block-producer, and the remainder of lamports being burnt,
-and have no other effect on state.
-In case 4, the fee-paying account should be charged the fee, and execution
-should proceed as normal if all other executability checks pass.
-Case 4 is the only case in which a transaction is possibly executed, and thus
-is the only case that has can have effect on account state beyond fees.
+    ```rust
+    pub struct VersionedTransaction {
+        #[serde(with = "short_vec")]
+        pub signatures: Vec<Signature>,
+        pub message: VersionedMessage,
+    }
+    ```
+
+    where `Signature` and `VersionedMessage` are defined in `solana-sdk`.
+2. Transaction serialized side must be less than or equal to 1232 bytes.
+3. Transaction signatures must be valid and in the same order as the static
+   account keys in the `VersionedMessage`.
+4. Transaction must have exactly the number of required signatures from the
+   `VersionedMessage` header.
+5. Transaction must not have more signatures than static account keys.
+6. The `VersionedMessage` must pass sanitization checks:
+    - The sum of the number of required signatures and the number of read-only
+      unsigned accounts must not be greater than the number of static account
+      keys.
+    - The number of readonly signed accounts must be less than the number of
+      required signatures.
+    - Each lookup table, if present, must be used to load at least one account.
+    - The total number of accounts, static or dynamic, must be less than 256.
+    - Each instruction's `program_id_index` must be less than the number of
+      static account keys.
+    - Each intruction's `program_id_index` must not be the payer index (0).
+    - All account indices in instructions must be less than the number of total
+      accounts.
+7. Transactions that use address look up tables must be resolvable:
+    - The address look up table account must exist.
+    - The address look up table account must be owned by the address look up
+      table program: `AddressLookupTab1e1111111111111111111111111`
+    - The address look up table account data must be deserializable into
+      `AddressLookupTable` as defined in `solana-sdk`.
+    - All account table indices specified in the transaction must be less than
+      the number of active addresses in the address look up table.
+8. The `recent_blockhash` of the transaction message must be valid:
+    - It must exist and not have an age greater than 150.
+    - OR the transaction must be a nonced transaction, and the nonce
+      account must exist and be valid for the given `recent_blockhash`.
+9. The transaction must not have already been processed.
+10. The transaction fee-payer account must:
+    - exist
+    - be owned by the the system program: `11111111111111111111111111111111`
+    - have more lamports than the fee
+    - have more lamports than the fee plus the minimum balance
+
+### Proposed New Runtime Errors
+
+The following constraints will be moved to become runtime errors.
+Transactions that break these constraints may be included in a block.
+The fee-payer account must be charged fees, nonce advanced, but the transaction
+will not be executed.
+
+1. Transactions containing pre-compile instructions must pass pre-compile
+   verification checks.
+2. The transaction must not load the same account more than once.
+3. The transaction must have fewer than 64 accounts.
+    - The limit is subject to change to 128 with the activation of
+      `9LZdXeKGeBV6hRLdxS1rHbHoEUsKqesCC2ZAPTPKJAbK`.
+4. Any non-writable program accounts used in the transaction must exist,
+   regardless of if they are used in any instructions.
+5. The total loaded data size of the transaction must not exceed
+   `requested_loaded_accounts_data_size_limit`, or the default limit (64MB).
+6. Any account used as a program in a top-level instruction must:
+    - be the native loader: `NativeLoader1111111111111111111111111111111`
+    - OR
+      - exist
+      - be executable
+      - be owned by the native loader: `NativeLoader1111111111111111111111111111111`
+    - OR
+      - exist
+      - be executable
+      - the owner account be owned by the native loader: `NativeLoader1111111111111111111111111111111`
+      - the owner account must be executable
 
 ### Block-Limits
 
-Pending the activation of `2ry7ygxiYURULZCrypHhveanvP5tzZ4toRwVp89oCNSj`,
-validators must validate a block is within block-limits.
-With this proposal, some transactions may not be executable, but will still
-count towards block-limits.
-Transactions have a cost towards block limits without needing to execute the
-transaction.
-The cost of a transaction consists of signature verification, write locks for
-accounts, builtin instructions, requested or default bpf compute units, and
-data size costs.
-With this proposal, regardless of whether a transaction can pay fees, or is
-executed, the total cost of the transaction will count towards block limits.
+This proposal allows for transactions which do not execute to be included in a
+block.
+This was not previously possible, so how transaction costs are applied to
+limits must be clarified.
+This proposal proposes that an included unexecuted transaction should have all
+non-execution transaction costs applied towards block-limits, as well as all
+the writable account limits.
 
-If these transactions did not count towards block-limits, the validation of
-block-limits would require the validator to check whether or not a transaction
-is executable, which negates the benefits of this proposal.
+### Rationale
 
-Additionally, if these transactions did not count towards block-limits, a
-malicious leader could produce a block with non-executable transactions and
-overload the network.
+The intent with relaxing these constraints is to reduce the amount of
+account-state which is required in order to validate a block.
+This gives more flexibility to when and how account-state is updated during
+both block production and validation.
+Particularly with the relaxation of the program account constraints,
+block-packing can be done without needing to load large program accounts
+for the initial decision to include or not.
+This is a major step towards asynchronous block production and validation.
+
+Any transaction included in the block, regardless of whether or not the
+transaction is executed, must be charged fees, inserted into the status cache,
+and the nonce account, if present, must be advanced.
 
 ## Impact
 
@@ -164,28 +262,19 @@ overload the network.
   included and will be charged fees.
   - Users must be more careful when constructing transactions to ensure they
     are executable if they don't want to waste fees
-- The validity of a block is no longer dependent on intra-block account-state
-updates. This is because the only account-state required for transaction
-validation is the ALT resolution, which is resolved at the beginning of a slot.
-  - Block-production could be done asynchronously
-  - Block-validation could be done without execution, but still relies on the
-  execution of previous blocks.
+- Block-production is simplified as it can be done without needing to load
+  large program accounts for the initial decision to include or not.
 
 ## Security Considerations
 
-- Removing fee-paying requirement could allow validator clients to produce
-  blocks that do not pay fees. If not checked by block-producer, this could
-  allow malicious users to abuse such producers/leaders.
+None
 
 ## Drawbacks
 
-- Any dynamic fees based on block utilization cannot/should not reward the
-block-producer directly. Otherwise, this incentivizes block-producers to fill
-blocks with non-fee paying transactions in order to raise fees.
-- Non-fee paying transactions included in a block will be recorded in long-term
-transaction history. Without any fees being paid, there is no way to reward
-long-term storage of these transactions. It is important to note, there is also
-currently no way to reward this storage.
+- Users must be more careful about what they sign, as they will be charged fees
+  for transactions that are included in a block, even if they are not executed.
+- This will likely break a lot of tooling, such as explorers, which may expect
+  all transactions to attempt execution.
 
 ## Backwards Compatibility
 
--- a/proposals/0082-relax-transaction-constraints.md
+++ b/proposals/0082-relax-transaction-constraints.md
@@ -62,7 +62,7 @@ remove one of the barriers to asynchronous execution.
     significant benefits to the network.
 3. Additionally, relax the address lookup table resolution constraint
     - This was considered, since it is a transaction-level constraint that is
-    depdendent on account-state. However, due to entry-level and block-level
+    dependent on account-state. However, due to entry-level and block-level
     constraints that rely on the address lookup table resolution, this
     constraint cannot easily be relaxed without also relaxing those
     constraints.
@@ -105,7 +105,7 @@ transaction level are:
     - The total number of accounts, static or dynamic, must be less than 256.
     - Each instruction's `program_id_index` must be less than the number of
       static account keys.
-    - Each intruction's `program_id_index` must not be the payer index (0).
+    - Each instruction's `program_id_index` must not be the payer index (0).
     - All account indices in instructions must be less than the number of total
       accounts.
 7. Transactions that use address look up tables must be resolvable:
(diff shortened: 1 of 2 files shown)


proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+   current slot must not be within the epoch stakes reward distribution period
+5. Have executable `BPFLoaderUpgradeab1e11111111111111111111111` owned accounts
+with `UpgradeableLoaderState::Program` state, and derived program data account
+that exist
+6. Have a call chain depth of 5 or less
+7. Have no builtin loader ownership chains
+(pending `4UDcAfQ6EcA6bdcadkeHpkarkhZGJ7Bpq7wTAiRMjkoi`)
+8. Have total loaded data size which does not exceed
+requested_loaded_accounts_data_size_limit
+(pending `DdLwVYuvDz26JohmgSbA7mjpJFgX5zP2dkp8qsF2C33V`)
```
apfitzge, line 84: Tried to be as specific as possible in the document itself, but these constraints (3-8) can be summarized as "transaction must be executable"

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+## New Terminology
+
+None
+
+## Detailed Design
+
+Specifically, this proposal relaxes constraints that a transaction included in
+a block must:
+
+1. Have a fee-payer with enough funds to pay fees
```
lheeger-jump, line 64: This list seems to denote the _current_ status of constraints on transactions. This is useful background, but we need to precisely specify the new constraints on transactions.

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+    constraint cannot easily be relaxed without also relaxing those
+    constraints.
+
+## New Terminology
+
+None
+
+## Detailed Design
+
+Specifically, this proposal relaxes constraints that a transaction included in
```
lheeger-jump, line 61: This list seems to mix "constraints on transactions which invalidate entries and blocks" and "constraints on transaction execution which invalidate the transaction". The list should be split up or the ones of the latter type should be removed.

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+   3. if owned by `Stake11111111111111111111111111111111111111`, then the
+   current slot must not be within the epoch stakes reward distribution period
+5. Have executable `BPFLoaderUpgradeab1e11111111111111111111111` owned accounts
+with `UpgradeableLoaderState::Program` state, and derived program data account
+that exist
+6. Have a call chain depth of 5 or less
+7. Have no builtin loader ownership chains
+(pending `4UDcAfQ6EcA6bdcadkeHpkarkhZGJ7Bpq7wTAiRMjkoi`)
+8. Have total loaded data size which does not exceed
+requested_loaded_accounts_data_size_limit
```
lheeger-jump, line 83: Can you please add links to the feature documents.

proposals/0082-relax-transaction-constraints.md
lheeger-jump: ... in a separate list.

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+
+Pending the activation of `2ry7ygxiYURULZCrypHhveanvP5tzZ4toRwVp89oCNSj`,
+validators must validate a block is within block-limits.
+With this proposal, some transactions may not be executable, but will still
+count towards block-limits.
+If these transactions did not count towards block-limits, the validation of
+block-limits would require the validator to check whether or not a transaction
+is executable, which negates the benefits of this proposal.
+
+Additionally, f these transactions did not count towards block-limits, a
```
lheeger-jump, line 136:
```suggestion
Additionally, if these transactions did not count towards block-limits, a
```

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+In case 1, the transaction should simply be ignored for execution, and have no
+effect on state.
+In case 2, the fee-paying account should be drained of all funds, but have no
+other effect on state.
+In case 3, the fee-paying account should be drained of all funds, with fees
+being paid to the block-producer, and the remainder of lamports being dropped.
+
+### Block-Limits
+
+Pending the activation of `2ry7ygxiYURULZCrypHhveanvP5tzZ4toRwVp89oCNSj`,
```
lheeger-jump, line 128: Please add a link.

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+validators must validate a block is within block-limits.
+With this proposal, some transactions may not be executable, but will still
+count towards block-limits.
+If these transactions did not count towards block-limits, the validation of
```
lheeger-jump, line 132: How do they specifically contribute to block limits.

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+1. The fee-payer account does not exist (0 lamports)
+2. The fee-payer account does not have enough funds for the entire fee
+3. The fee-payer account has enough funds for the fee, but would no longer be
+rent-exempt
+
+In case 1, the transaction should simply be ignored for execution, and have no
+effect on state.
+In case 2, the fee-paying account should be drained of all funds, but have no
+other effect on state.
+In case 3, the fee-paying account should be drained of all funds, with fees
```
lheeger-jump, line 123: That seems a bit harsh, no?

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+three different edge-cases to consider:
+
+1. The fee-payer account does not exist (0 lamports)
+2. The fee-payer account does not have enough funds for the entire fee
+3. The fee-payer account has enough funds for the fee, but would no longer be
+rent-exempt
+
+In case 1, the transaction should simply be ignored for execution, and have no
+effect on state.
+In case 2, the fee-paying account should be drained of all funds, but have no
```
lheeger-jump, line 121: probably a good exercise to also enumerate the normal case

lheeger-jump (changes requested):
I have done a first pass but will need to go more in depth later.

proposals/0082-relax-transaction-constraints.md
apfitzge: That's a very fair comment, as that is something is changing with this proposal. Currently, the list of constraints that invalidate the block is a superset of those which invalidate execution. That will no longer be the case, and I should make that more clear.

proposals/0082-relax-transaction-constraints.md
apfitzge: Bit confused on what you're suggesting would make the document more clear here; this proposal does not add any new constraints.
The list here is an incomplete list of the constraints on transactions, it is only the constraints that are being removed/relaxed.

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+other effect on state.
+In case 3, the fee-paying account should be drained of all funds, with fees
+being paid to the block-producer, and the remainder of lamports being dropped.
```
ryoqun, line 124: `dropped` => `burnt`?

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+necessary for validation of entry-level and block-level constraints.
+However, if/when those constraints are relaxed, this constraint should be
+relaxed as well.
+
+The relaxation of these constraints is only relaxed for transactions being
+included in a block. During block validation, if any of these constraints
+are broken, the entire block is marked invalid; with this proposal, the
+block is not marked invalid, but the transaction will not be executed.
+These constraints must still be satisfied in order for the transaction to be
+executed, and have an effect on state. However, there are some new considerations
```
ryoqun, line 103: maybe better off being explicit about side-effects on some on-chain states for these constraints-violating transactions. namely, the status cache to avoid repeated inclusion of same transactions.

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+  - Block-production could be done asynchronously
+  - Block-validation could be done without execution, but still relies on the
+  execution of previous blocks.
+
+## Security Considerations
+
+- Removing fee-paying requirement could allow validator clients to produce
+  blocks that do not pay fees. If not checked by block-producer, this could
+  allow malicious users to abuse such producers/leaders.
+
```
ryoqun, line 158: also, malicious leaders could create full of garbage blocks, which only contains bad transactions at no cost, burdening all the validations onto other validators, esp, heavy account state lookup.

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+With these constraints removed, there is still one account-state constraint
+at the transaction level: address lookup table resolution.
+With this proposal, this constraint is intentionally not relaxed since it is
+necessary for validation of entry-level and block-level constraints.
+However, if/when those constraints are relaxed, this constraint should be
+relaxed as well.
+
+The relaxation of these constraints is only relaxed for transactions being
+included in a block. During block validation, if any of these constraints
```
godmodegalactus, line 99: How does this affect transaction simulation? Currently, we get an error when we simulate transactions where the payer does not have enough balance. This will make these kinds of errors indetectable during transaction simulation.

proposals/0082-relax-transaction-constraints.md
godmodegalactus: SIMD should stand for itself and should not have any links, as far as i know.

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+fee, the transaction cannot be included in a block. With this proposal, it is
+possible for such a transaction to be included in the block, and there are
+three different edge-cases to consider:
+
+1. The fee-payer account does not exist (0 lamports)
+2. The fee-payer account does not have enough funds for the entire fee
+3. The fee-payer account has enough funds for the fee, but would no longer be
+rent-exempt
+
+In case 1, the transaction should simply be ignored for execution, and have no
```
godmodegalactus, line 119: Should also consider that base fee will eventually change and could be much greater than now.

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+effect on state.
+In case 2, the fee-paying account should be drained of all funds, but have no
+other effect on state.
+In case 3, the fee-paying account should be drained of all funds, with fees
+being paid to the block-producer, and the remainder of lamports being dropped.
+
```
godmodegalactus, line 125: Could burn lots of SOLs if you create big accounts and use the same accounts as a payer. Should be okay though if anyone wants to burn their SOLs through an obscure mechanism.

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+- Removing fee-paying requirement could allow validator clients to produce
+  blocks that do not pay fees. If not checked by block-producer, this could
+  allow malicious users to abuse such producers/leaders.
+
+## Drawbacks
+
+- Any dynamic fees based on block utilization cannot/should not reward the
+block-producer directly. Otherwise, this incentivizes block-producers to fill
+blocks with non-fee paying transactions in order to raise fees.
+- Non-fee paying transactions included in a block will be recorded in long-term
```
godmodegalactus, line 164: This could be a way to attack the cluster and use more space. A malicious leader could create 48M CU / (CU required to load payer account which currently stands at 0) number of transactions. This could be a massive number of transactions, depending on CUs charge that we decide to load the payer account. If each transaction was given the maximum size of 1280 bytes or eventually more if we include ZK transactions, then I can possibly create a huge block burdening the whole cluster's network resources because of the turbine.

I feel like the leader should be somehow punished/slashed for adding transactions where the payer does not have enough balance.

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+transaction level, in order to simplify the protocol, and give more flexibility
+to block-producer and block-validator implementations.
+
+More specifically, this proposal aims to relax the constraints which require
+account state in order to determine the validity of a transaction's inclusion.
+The reason these constraints are targetted specifically, is that reliance on
+account state necessarily requires synchronous execution within the protocol.
+This proposal on its' own, will not enable asynchronous execution, but it will
+remove one of the barriers to asynchronous execution.
+
```
taozhu-chicago, line 37: A bit confused by "aim" in Summary section and Motivations here. As "aim" seems to yield benefit to block producer ("...transaction can be included..."), which is not part of protocol; while Motivation talks about modifying protocol. 

Is it correct yo say this proposal has two parts: 
1. remove listed constraints/requirement of transaction for block producer to include them into block; this part solely in Producer implementation, does not modify protocol;
2. instead of invalidate entire block at earlist failure of transaction due to named constraints, validator only _not_ execute bad transactions but will continue validate the block. (bad transactions included in ledger). This part of protocol altering.

or maybe I am thinking on wrong track?

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+category: Standard
+type: Core
+status: Draft
+created: 2023-10-30
+feature:
+---
+
+## Summary
+
+This proposal aims to relax some of the constraints on which individual
```
taozhu-chicago, line 15: > to relax some of the constraints

to clarify, "relax" means "removal"?

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Block-Limits
+
+Pending the activation of `2ry7ygxiYURULZCrypHhveanvP5tzZ4toRwVp89oCNSj`,
+validators must validate a block is within block-limits.
+With this proposal, some transactions may not be executable, but will still
+count towards block-limits.
+If these transactions did not count towards block-limits, the validation of
+block-limits would require the validator to check whether or not a transaction
+is executable, which negates the benefits of this proposal.
```
taozhu-chicago, line 134: I think the case here is "producer builds a block of 48M cu, but validator only spent less than 48M to execute all transactions (as some of TXs are failed and not executed)", and this should be OK (except the DoS vector - that entire block is unexecutable txs, which can be addressed by making producer pay for these unexecutable txs)

proposals/0082-relax-transaction-constraints.md
jstarry: agreed with @ryoqun, this attack vector is the main thing that needs to be solved. Maybe this specific problem can be discussed in a separate SIMD since relaxing other transaction constraints is more straightforward.

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+
+Additionally, f these transactions did not count towards block-limits, a
+malicious leader could produce a block with non-executable transactions and
+overload the network.
+
+## Impact
+
+- Transactions that would previously be dropped with an error, can now be
+  included, and even charged fees.
```
jstarry, line 143:
```suggestion
- Transactions that would previously be dropped with an error without paying fees, can now be
  included and will be charged fees.
```

I think "and even charged fees" doesn't sound declarative enough. My understanding is that the crux of this SIMD is that before we dropped txs for a variety of reasons and didn't charge fees and after we will allow invalid txs to be in a block and will charge fees for them.

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+## Alternatives Considered
+
+1. Do nothing
+    - This is the simplest option, as we could leave the protocol as is.
+    However, this leaves the protocol more complex than it needs to be.
+2. Relax all but fee-paying constraint
+    - This was actually the initial proposal, but it was decided that it would
+    be better to relax as many constraints as possible, rather than just some.
+    Any reliance on account state, means the protocol requires synchronous
+    execution.
```
jstarry, line 47: To clarify, "it would be better to relax as many constraints as possible" just means that the fee-paying constraint is _also_ removed right? I think you could say that directly here.

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+3. Additionally, relax the address lookup table resolution constraint
+    - This was considered, since it is a transaction-level constraint that is
+    depdendent on account-state. However, due to entry-level and block-level
+    constraints that rely on the address lookup table resolution, this
+    constraint cannot easily be relaxed without also relaxing those
+    constraints.
```
jstarry, line 53: I don't really understand why there is still a constraint for address lookup tables. If the lookup table is invalid for some reason can't validators just charge the fee payer some fees?

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+
+None
+
+## Detailed Design
+
+Specifically, this proposal relaxes constraints that a transaction included in
+a block must:
+
+1. Have a fee-payer with enough funds to pay fees
+2. Have a valid nonce account, if specified
```
jstarry, line 65: Did you consider that both normal transactions and durable nonce transactions can only be processed once to be valid? Validators and bankless leaders can use the status cache to avoid including normal transactions more than once, but they need to check that the nonce is valid or not to avoid including durable nonce transactions more than once.

proposals/0082-relax-transaction-constraints.md
lheeger-jump: It at least needs to describe the feature.

proposals/0082-relax-transaction-constraints.md
lheeger-jump: The list describes what the constraints are now, but does not explicitly list what the relaxed constraints are. Consider that I or another FD dev needs to stare at this and implement the exact rules for what defines a valid transaction.

proposals/0082-relax-transaction-constraints.md
apfitzge: ALT resolution contributes to constraints at higher levels i.e. entry and block.

Entries need to have non-conflicting transactions (without adoption of #83).
Blocks are required to not exceed account write limits (currently 12M CUs).

In both these situations, if we allow ALTs to not resolve I think the behavior becomes very unclear. Do we only consider the static accounts? all resolvable accounts? no accounts?
Not clear to me any of those options is ideal or makes sense. I think we'd like to remove this constraint eventually, but I think it probably makes more sense to do after these complicating entry and block constraints are also removed.

proposals/0082-relax-transaction-constraints.md
apfitzge: Considered, but I need to rewrite sections of the proposal to be very explicit.

Status cache should still be used to avoid including transactions multiple times. Duplicate inclusion of transactions is still a violation of constraints and leads to the entire block being marked as invalid. This does not depend on account state, which is why I think this should be left in.

With respect to nonces, we want to avoid depdendence on account state, so cannot check the validity of the nonce account for the question of block validity. However, the nonce acount state must still be checked before execution. If the nonce account does not exist or is otherwise invalid, the transaction is not executed and fees are not charged to the fee-payer.
I view this as a similar situation to the fee-payer check; The block-producer is incentivized to check these states. They are free to do that in some non-strict method, but if done incorrectly, they lose out on rewards and the transaction has no/limited on state and have effectively burned their CUs.

There's an ongoing discussion of if fees should be paid by the block-producer should a transaction be unable to pay. So whether or not the producer gets charged fees in these cases is still an open question.

proposals/0082-relax-transaction-constraints.md
apfitzge: The constraints still exist in the sense we need to check "can this transaction be executed". They do not exist for the validity of the block.

Right now, those 2 different types of constraints are overlapping. I need to re-write this proposal with @lheeger-jump's suggestion on making it very explicit that these 2 things are different.

proposals/0082-relax-transaction-constraints.md
apfitzge: I think only your 2nd point is included in the proposal. Block-production changse are enabled by the changes in the proposal, but not required beyond potential changes to the fee-colection.

> instead of invalidate entire block at earlist failure of transaction due to named constraints, validator only not execute bad transactions but will continue validate the block. (bad transactions included in ledger). 

yes. Those bad transactions may end up having limited effect if we are still able to collect fees, or if we decide to make the block-producer cover those fees. They will not be executed though.

----

Proposal does not suggest or require the block-producer to make any changes, but allows for changes. For several of the constraints, the producer is still incentivized to perform the checks (fee-payer, nonce) because otherwise they will not receive rewards for those transactions. There's freedom in how that is done, and the block-producer could potentially do this in a non-strict fashion.

I'll cover this in the updated prop, but essentially block-validation and consensus really consists of 2 primary questions:

1. is this block valid?
2. what is the result of the block?

The changes here are intending to make answering this first question, "is this block valid?", a simpler process that does not require executing the block. This proposal separates these questions logically, but w/ current voting we still require the 2nd question, "what is the result of the block?", be answered synchronously.

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+
+1. The fee-payer account does not exist (0 lamports)
+2. The fee-payer account does not have enough funds for the entire fee
+3. The fee-payer account has enough funds for the fee, but would no longer be
+rent-exempt
+
+In case 1, the transaction should simply be ignored for execution, and have no
+effect on state.
+In case 2, the fee-paying account should be drained of all funds, but have no
+other effect on state.
```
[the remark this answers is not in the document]
apfitzge: > With this,

Might have missed a comment this is attached to. With what?

Case 2 would probably not occur if it was a malicious actor, but could happen if some non-strict block-producer made and assumption a fee-payer was good and was not.

proposals/0082-relax-transaction-constraints.md
apfitzge: I'm not sure what you mean. Transaction simulation is unrelated to block invalidation, and doesn't change with this proposal.
You'd still get the same transaction error which indicates it cannot pay for fees.

proposals/0082-relax-transaction-constraints.md
apfitzge: Currently transactions do have a minimum cost of 1070 CUs = 720 SigVerify + 300 WriteLock.
This corresponds to a maximum of 44776 txs per slot, assuming 48M CU limit.

Additionally, cost has 1 CU per 4 bytes of instruction data. So large ix data txs would cost more than this bare minimum of 1070 CUs as well.
It's still a large number of transactions of 34532 txs per slot.

apfitzge: It's worth noting that currently, all direct rewards are going to the block-producer anyway.
If that large number of transactions were valid fee-paying transactions, it would only be more work for other validators than if they could not pay fees. Either way, other nodes are not directly rewarded for the validation/execution.

There is still some discussion to be had around **indirect** rewards via the burning of SOL in those fees. I think that's potentially a valid reason we may want to force the block-producer into paying fees. Hoping to discuss this more

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,173 @@
...
+Additionally, f these transactions did not count towards block-limits, a
+malicious leader could produce a block with non-executable transactions and
+overload the network.
+
+## Impact
+
+- Transactions that would previously be dropped with an error, can now be
+  included, and even charged fees.
+  - Users must be more careful when constructing transactions to ensure they
+    are executable if they don't want to waste fees
```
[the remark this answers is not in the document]
apfitzge: Simulation doesn't change. It currently would tell you that your transaction had an error. That error doesn't change, it only changes what will happen to a block it may be included in.

For example, if a transaction had an invalid program account, simulation will still return `ProgramAccountNotFound` error.
If the user simulates, and then still submits that transaction, then they probably deserve to be charged, imo.

proposals/0082-relax-transaction-constraints.md
apfitzge: made a comment in this chain: https://github.com/solana-foundation/solana-improvement-documents/pull/82#discussion_r1393361600

I don't think there's a DOS that's creating more work than would be possible within current protocol.
I can currently create a block thats full of **valid** heavy transactions, and then tag in some invalidating transaction at the last tick. All the validators wasted their time executing all those txs, and there's no reward for them since the block is thrown out entirely.

This would change the vector, since they could have **all** txs that do not pay fees. But, that only makes the execution less intense, since they still have a cost that counts towards block limits.

proposals/0082-relax-transaction-constraints.md
apfitzge: Potentially this is too harsh. Alternatively we could just drain them to the rent-exempt level.
Not sure which seems fairer. If submitting a transaction that cannot pay fees, they **should** be punished imo.

apfitzge:
Pushed some clarifications. As mentioned, I need to re-write a few sections wrt being explicit on separation of "includability" and "executability" constraints.

I think the main open question is how to treat non fee-paying transactions. In my opinion, this should be answered and included as part of this SIMD.

1. Current proposal allows txs to go entirely unpaid
2. Could charge the block-producer unpaid fees

I'm not that oppposed to 2, although it might be more complicated. It opens some new questions around what to do if the producer also cannot pay for the fees.

proposals/0082-relax-transaction-constraints.md
diman-io: Imho, since the commission is paid, the transaction should be executed. Deleting the account is a different event.

proposals/0082-relax-transaction-constraints.md
jstarry: I see, thanks for explaining what the entry and block constraints are, I hadn't considered those. FWIW ALTs were designed to be apend only and the warmup time (currently one slot) is configurable by the runtime and can be increased in the future if it helps bankless leaders manage a cache of ALTs. So I think we could try to move forward with removing this constraint.

jstarry:
I think it's hard to imagine a fee-maximizing validator not maintaining a status cache, durable nonce cache, ALT cache, etc (are there more?). I lean towards enforcing all validators to maintain those caches in order to produce valid blocks. Is that also an option?

jstarry:
>  I think the main open question is how to treat non fee-paying transactions

Are these the only cases in which transactions don't pay fees?
1. Replayed transactions (normal or durable)
2. Fee payer lacks sufficient funds

Can you document them in the SIMD and explain why they can't be avoided and need to be addressed?

apfitzge:
> Are these the only cases in which transactions don't pay fees?
> 
>     1. Replayed transactions (normal or durable)
> 
>     2. Fee payer lacks sufficient funds

I'll document it, but to be clear. Normal transactions can still not be replayed. This SIMD does not propose changes to the status cache. If a block contains an already executed **normal** transaction, then the entire block will be invalid.

With nonced transactions, correct. They could be included multiple times, would not pay fees or be executed after the first one.

> I think it's hard to imagine a fee-maximizing validator not maintaining a status cache, durable nonce cache, ALT cache, etc (are there more?). I lean towards enforcing all validators to maintain those caches in order to produce valid blocks. Is that also an option? (EDIT: If we continue forward assuming that validators are incentivized to avoid non-fee paying transactions, can we explain how they could reasonably implement that? I think block space could be DOS'd otherwise)

Just to re-iterate, the SIMD doesn't propose changes to status-cache.

> Is that also an option?

Sure, rejecting wholely or part of the SIMD would leave those constraints in the protocol.

> validators are incentivized to avoid non-fee paying transactions, can we explain how they could reasonably implement that

There's no reason a validator could not continue to do strict fee-paying checks, as they do today, for block-production.

> I think block space could be DOS'd otherwise

Yeah, so this part of prevention is behind a currently inactive feature: https://github.com/solana-labs/solana/issues/29595

That feature will cause transactions to count towards block-limits with the **requested** CUs (+builtin costs), not **actually executed**.
WRT this proposal, transactions regardless of if they can pay fees or are executed, will count towards block-limits as if they could.
So I don't think there's any additional worry about block space DOS attacks, these non-fee paying transactions can only be cheaper to execute than if they could pay fees, but count towards block-limits at full cost.

jstarry:
Couldn't some malicious actor send a bunch of non-fee paying transactions then to saturate block-limits and prevent fee-paying users from using the chain? I think this should be explicitly prevented by the design of this SIMD if possible.

apfitzge:
> Couldn't some malicious actor send a bunch of non-fee paying transactions then to saturate block-limits and prevent fee-paying users from using the chain? I think this should be explicitly prevented by the design of this SIMD if possible.

A malicious user could send those transactions, but any reasonable producer will have some logic, albeit not necessarily *strict*, for filtering out transactions that cannot pay fees. Since if they pack those transactions they either get no rewards, or have to pay the fees themselves (hoping to discuss this tomorrow at community call).

I will explicitly mention this potential case in the SIMD.

proposals/0082-relax-transaction-constraints.md
godmodegalactus: Yeah I agree, block producers should pay a fee as they are using network resources, storage space to transfer invalid transactions. Malicious node can just start creating blocks with invalid transactions of 1280 bytes using 44.20096 MBs/ block in network and storage.

apfitzge:
Converting to a draft as I rework the proposal to fit the feedback received during the core community call.

apfitzge:
@lheeger-jump made the lists much more clear, tried to enumerate all relevant checks that affect whether or not a transaction can be accepted in a block.

Because many of the other changes are less controversial, I've opted to remove the fee-payer relaxation out of this SIMD and would do that as a follow-up proposal.
Acceptance of this SIMD will still require that all transactions can pay fees.

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,284 @@
...
+    However, this was decided to be moved to a separate follow-up proposal,
+    should this one be accepted.
+    There was significant disagreements on how exactly transactions that cannot
+    pay fees should be handled.
+    By keeping fee-paying as a constraint for the current proposal, it will
+    allow this proposal to be accepted more quickly, which will give
+    significant benefits to the network.
+3. Additionally, relax the address lookup table resolution constraint
+    - This was considered, since it is a transaction-level constraint that is
+    depdendent on account-state. However, due to entry-level and block-level
```
nazreen, line 65: small typo here @apfitzge 

"depdendent"

apfitzge: Thanks! Found a few other spelling, consistency, or grammar issues while fixing it.
Pushed the fixes.

apfitzge:
TVC eliminates the 1:1 vote-to-credit relationship, so those number would need to be tuned.
Not sure where you are getting 100k as the maximum number of invalid transactions? Malicious leader could pack much more than than in 4 slots.

Even without the complicating factors of relaxing fee-payer or signature constraints, which are contentious changes, this proposal seems to be idling waiting for reviews.
We can have follow-up or even separate SIMDs for those ideas, but it seems adding too many things to any SIMD will slow it down - particularly when those things are contentious.

Maybe I'm off - @jstarry lmk if I should add these back into this proposal.

aeyakovenko:
How can proposer exceed the shred limits @apfitzge?  

I don't think they can exceed 100k txs in 4 slots, but they can submit super large txs and reduce the penalty. But the penalty can just be tuned to be 10 points per invalid tx, or X points per tx byte. Either way. 

TVC changes don't really impact this since the best case is about the same.

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,284 @@
...
+    - OR
+      - exist
+      - be executable
+      - be owned by the native loader: `NativeLoader1111111111111111111111111111111`
+    - OR
+      - exist
+      - be executable
+      - the owner account be owned by the native loader: `NativeLoader1111111111111111111111111111111`
+      - the owner account must be executable
+
```
jumpsiegel, line 150: We need a complete test suite that tests every single one of these constraints (via txn execution) so that we can also test the behavior when the feature flag is flipped to make sure we relax all the correct constraints

apfitzge: Yes, any client should be testing against the protocol, and those tests much change if the protocol changes.

There should already (big should in our codebase...) be tests for all these cases, as they are currently protocol violations, that we must check for.
Tests can be modified to use feature-flag which would change the result of processing pipeline from rejected -> not executed.

apfitzge:
> How can proposer exceed the shred limits @apfitzge?
> 
> I don't think they can exceed 100k txs in 4 slots, but they can submit super large txs and reduce the penalty. But the penalty can just be tuned to be 10 points per invalid tx, or X points per tx byte. Either way.
> 
> TVC changes don't really impact this since the best case is about the same.



> How can proposer exceed the shred limits @apfitzge?
> 
> I don't think they can exceed 100k txs in 4 slots, but they can submit super large txs and reduce the penalty. But the penalty can just be tuned to be 10 points per invalid tx, or X points per tx byte. Either way.
> 
> TVC changes don't really impact this since the best case is about the same.

because that's not the shred limits, afaik shred limits would be 131072 txs per **slot**.

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,284 @@
...
+    - OR the transaction must be a nonced transaction, and the nonce
+      account must exist and be valid for the given `recent_blockhash`.
+12. The transaction must not have already been processed.
+13. The transaction fee-payer account must:
+    - exist
+    - be owned by the system program: `11111111111111111111111111111111`
+    - have more lamports than the fee
+    - have more lamports than the fee plus the minimum balance
+14. Any non-writable program accounts used in the transaction must exist,
+    regardless of if they are used in any instructions.
```
jstarry, line 136: This rule doesn't really make any sense and should be removed

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,284 @@
...
+        #[serde(with = "short_vec")]
+        pub signatures: Vec<Signature>,
+        pub message: VersionedMessage,
+    }
+    ```
+
+    where `Signature` and `VersionedMessage` are defined in `solana-sdk`.
+2. Transaction serialized side must be less than or equal to 1232 bytes.
+3. Transaction signatures must be valid and in the same order as the static
+   account keys in the `VersionedMessage`.
```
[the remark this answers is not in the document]
apfitzge: Yeah, but that was a separate change not part of this SIMD.

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,284 @@
...
+    - All account indices in instructions must be less than the number of total
+      accounts.
+7. Transactions that use address lookup tables must be resolvable:
+    - The address lookup table account must exist.
+    - The address lookup table account must be owned by the address lookup
+      table program: `AddressLookupTab1e1111111111111111111111111`
+    - The address lookup table account data must be deserializable into
+      `AddressLookupTable` as defined in `solana-sdk`.
+    - All account table indices specified in the transaction must be less than
+      the number of active addresses in the address lookup table.
```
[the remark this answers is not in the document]
apfitzge: Plan was for this to be in a separate follow-up SIMD since the choice here is very much not obvious imo.

How do we apply CUs to account-level cost if we cannot resolve ALTs?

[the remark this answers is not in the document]
jstarry: I'm fine with it being covered in a separate SIMD

proposals/0082-relax-transaction-constraints.md
apfitzge: Yeah think this comes from me translating code -> requirements, since we have no official spec.

Code logic has a check for this later on, but we identify programs earlier from the account-data...if we identified it as a program it has to exist.

jstarry: In order to identify an account as a program, it surely exists already. That's the part that doesn't make sense to me

proposals/0082-relax-transaction-constraints.md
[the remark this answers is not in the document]
jstarry: Ok maybe rather than having a single all encompassing SIMD for transaction constraints, we limit this one to account loading related constraints?

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,284 @@
...
+9. The transaction must not load the same account more than once.
+10. The transaction must have fewer than 64 accounts.
+    - The limit is subject to change to 128 with the activation of
+      `9LZdXeKGeBV6hRLdxS1rHbHoEUsKqesCC2ZAPTPKJAbK`.
+11. The `recent_blockhash` of the transaction message must be valid:
+    - It must exist and not have an age greater than 150.
+    - OR the transaction must be a nonced transaction, and the nonce
+      account must exist and be valid for the given `recent_blockhash`.
+12. The transaction must not have already been processed.
+13. The transaction fee-payer account must:
```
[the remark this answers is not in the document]
apfitzge: dc42394b40e4221fb3ec9bcce685c7c5b6bd2ec9

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,284 @@
...
+    ```rust
+    pub struct VersionedTransaction {
+        #[serde(with = "short_vec")]
+        pub signatures: Vec<Signature>,
+        pub message: VersionedMessage,
+    }
+    ```
+
+    where `Signature` and `VersionedMessage` are defined in `solana-sdk`.
+2. Transaction serialized side must be less than or equal to 1232 bytes.
```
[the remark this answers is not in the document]
apfitzge: 6d6200b22964baaa09f31f2bc076ff63ca7917a4

proposals/0082-relax-transaction-constraints.md
```diff
@@ -0,0 +1,284 @@
...
```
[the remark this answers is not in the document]
apfitzge: 6d6200b22964baaa09f31f2bc076ff63ca7917a4

proposals/0082-relax-transaction-constraints.md
apfitzge: 561b44f7947321f43537f77f02d55d568df550a2

jstarry (commented):
I think this SIMD can be trimmed down a lot and just focus on what I consider to be "transaction loading" constraints. We don't need to enumerate all of the protocol constraints in this document, so I think we should get rid of the "Current Protocol-Violation Errors" and the "Proposed Protocol-Violation Errors" sections which just leaves the "Proposed New Runtime Errors" section which is the only thing anyone will want to know when reading this document.

I know this SIMD started off as a general relax all the tx constraints proposal, but now that it's been scoped down, I think it should be renamed and focused to only relaxing account loading related transaction accounts. Which, in my opinion, includes all of the new runtime constraints you've listed with the exception of the precompile verification constrain which is already covered in https://github.com/solana-foundation/solana-improvement-documents/pull/159

apfitzge:
This SIMD has been split into a few smaller SIMDs (biggest one being https://github.com/solana-foundation/solana-improvement-documents/pull/191). Closing this issue since I will not be pushing for this to be merged. Discussions are not deleted and this PR is linked to from the split SIMD(s)

proposals/0082-relax-transaction-constraints.md
agca52: ＼(^_^)／

PR #85, opened 2023-11-05
SIMD-0085: Additional Fee Collector Constraints

jstarry:
Add SIMD for adding fee collector constraints

### DIFF
--- a/proposals/0085-additional-fee-collector-constraints.md
+++ b/proposals/0085-additional-fee-collector-constraints.md
@@ -24,8 +24,8 @@ constraint was already added for rent fee collection in
 
 ## Motivation
 
-1. Fee collection occurs outside of the transaction runtime because the Solana
-   protocol mandates that fees are collected at the end of each block. By
+1. Fee distribution occurs outside of the transaction runtime because the Solana
+   protocol mandates that fees are distributed to "fee collector" accounts at the end of each block. By
    restricting fee collector accounts to be system owned, the number of account
    modification edge cases as well as protocol complexity are both reduced.
 2. Prevent new rent-paying accounts from being created since rent collection is
--- a/proposals/0085-additional-fee-collector-constraints.md
+++ b/proposals/0085-additional-fee-collector-constraints.md
@@ -35,7 +35,7 @@ constraint was already added for rent fee collection in
 
 ### Elide the system-owned constraint
 
-Restricting fee collection accounts to be system-owned is perhaps overly
+Restricting fee collector accounts to be system-owned is perhaps overly
 restrictive and limits the amount of flexibility that validator operators have
 when managing sensitive accounts with funds. However, the risk of having more
 runtime edge cases is too high to allow any program-owned account to collect
@@ -66,7 +66,7 @@ these constraints, the fees MUST be burned by not distributing them to anyone.
 
 ## Impact
 
-New and existing validators must ensure that their fee collection account is
+New and existing validators must ensure that their fee collector account is
 rent-exempt and owned by the system program in order to receive fees. Since the
 Solana Labs validator implementation currently requires the fee collector
 account to be same account as the fee payer for vote transactions, this is
--- a/proposals/0085-additional-fee-collector-constraints.md
+++ b/proposals/0085-additional-fee-collector-constraints.md
@@ -47,14 +47,14 @@ loopholes.
 
 Rather than restricting fee collector accounts to be system-owned, a new type of
 "validator node" account could be introduced. Currently, in normal validator
-operations, the "fee collector" account is also used as the "node id" as well as
-the vote fee payer. Introducing a "validator node" account that is owned by a
-"validator node" program which allows configuring a "withdraw authority" and
-"vote fee payer" could help increase validator operation flexibility and
+operations, the fee collector account is also used as the node id as well as
+the vote fee payer. Introducing a validator node account that is owned by a
+validator node program which allows configuring a withdraw authority and
+vote fee payer could help increase validator operation flexibility and
 increase clarity in how validator keys are used in the protocol.
 
-This approach requires a migration of all "fee collector" accounts as well as
-the development of a new on-chain program to manage the new "validator node"
+This approach requires a migration of all fee collector accounts as well as
+the development of a new on-chain program to manage the new validator node
 accounts. It will be a big effort compared to the proposed constraints in this
 SIMD and should be discussed in a new SIMD if this approach is desired.
 Furthermore, durable nonce accounts already have a configurable authority field
--- a/proposals/0085-additional-fee-collector-constraints.md
+++ b/proposals/0085-additional-fee-collector-constraints.md
@@ -1,6 +1,6 @@
 ---
 simd: "0085"
-title: Additional Fee Collector Constraints
+title: Additional Fee-Collector Constraints
 authors:
   - Justin Starry
 category: Standard
@@ -12,22 +12,26 @@ feature: https://github.com/solana-labs/solana/issues/33888
 
 ## Summary
 
-Burns fees if they would otherwise be collected into an account that violates
-one of the following constraints:
+Every validator defines a node id which is also used as the validator's
+fee-collector account for collecting earned protocol fees. After implementing
+this proposal, validator implementations will burn fees if they would otherwise
+be distributed into a fee-collector account that violates one of the following
+constraints:
 
 1. Must be a system owned account
 2. Must be rent-exempt
 
-This applies to both transaction fees and rent fees. Note that the "rent-exempt"
-constraint was already added for rent fee collection in
+These constraints apply to both transaction fees and rent fees. Note that the
+rent-exempt constraint was already added for rent fee collection in
 [Feature 30151](https://github.com/solana-labs/solana/issues/30151).
 
 ## Motivation
 
 1. Fee distribution occurs outside of the transaction runtime because the Solana
-   protocol mandates that fees are distributed to "fee collector" accounts at the end of each block. By
-   restricting fee collector accounts to be system owned, the number of account
-   modification edge cases as well as protocol complexity are both reduced.
+   protocol mandates that fees are distributed to "fee-collector" accounts at the
+   end of each block. By restricting fee-collector accounts to be system owned, the
+   number of account modification edge cases as well as protocol complexity are
+   both reduced.
 2. Prevent new rent-paying accounts from being created since rent collection is
    planned to be disabled in SIMD-0084.
 
@@ -35,7 +39,7 @@ constraint was already added for rent fee collection in
 
 ### Elide the system-owned constraint
 
-Restricting fee collector accounts to be system-owned is perhaps overly
+Restricting fee-collector accounts to be system-owned is perhaps overly
 restrictive and limits the amount of flexibility that validator operators have
 when managing sensitive accounts with funds. However, the risk of having more
 runtime edge cases is too high to allow any program-owned account to collect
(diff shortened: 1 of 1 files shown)


proposals/0085-additional-fee-collector-constraints.md
```diff
@@ -0,0 +1,61 @@
...
+
+## New Terminology
+
+NA
+
+## Detailed Design
+
+At the end of a block, validators MUST NOT distribute fees to accounts that are
+not system owned and/or rent-exempt. Instead, they MUST burn the fees by not
+distributing them to anyone.
```
CriesofCarrots, line 43:
```suggestion
At the end of a block, validators MUST NOT distribute fees to accounts that are
rent-paying or owned by a non-system program. Instead, they MUST burn the fees by not
distributing them to anyone.
```
Nit. I thought this was a little easier to follow without one set of negatives.

CriesofCarrots (approved):
r+ nit and sign-off from other clients

proposals/0085-additional-fee-collector-constraints.md
```diff
@@ -0,0 +1,61 @@
...
+feature: https://github.com/solana-labs/solana/issues/33888
+---
+
+## Summary
+
+Burns fees if they would otherwise be collected into an account that violates
+one of the following constraints:
+
+1. Must be a system owned account
+2. Must be rent-exempt
```
lheeger-jump, line 19: SIMD-0084 aims to disable rent. Do you want to just

proposals/0085-additional-fee-collector-constraints.md
```diff
@@ -0,0 +1,61 @@
...
+
+This applies to both transaction fees and rent fees. Note that the "rent-exempt"
+constraint was already added for rent fee collection in
+[Feature 30151](https://github.com/solana-labs/solana/issues/30151).
+
+## Motivation
+
+1. Since fee collection occurs outside of the runtime, it's generally a good
```
lheeger-jump, line 27: How is it that fee collection occurs outside of the runtime? This seems like an implementation detail.

proposals/0085-additional-fee-collector-constraints.md
```diff
@@ -0,0 +1,61 @@
...
+   idea to reduce the number of account modification edge cases.
+2. Prevent new rent paying accounts from being created
```
lheeger-jump, line 29: This does not address the improvement this change aims to make.

lheeger-jump (changes requested):

proposals/0085-additional-fee-collector-constraints.md
jstarry: In order to disable rent collection, all accounts must be enforced to be rent exempt when created. Fee collection was discovered to be an exception because it modifies accounts when depositing collected fees and didn't enforce that the new balance is rent exempt. So this exception needs to be patched to disable rent.

proposals/0085-additional-fee-collector-constraints.md
jstarry: Fees are deposited at the end of the block and must be done consistently across clients to maintain consensus. By outside the runtime, I meant outside of normal transaction processing.

proposals/0085-additional-fee-collector-constraints.md
jstarry: Yes it does, can you elaborate more what could be clearer?

proposals/0085-additional-fee-collector-constraints.md
```diff
@@ -0,0 +1,61 @@
...
+created: 2023-11-05
+feature: https://github.com/solana-labs/solana/issues/33888
+---
+
+## Summary
+
+Burns fees if they would otherwise be collected into an account that violates
+one of the following constraints:
+
+1. Must be a system owned account
```
diman-io, line 18: Why system owned? This restricts the use of multisig wallet for this for example

diman-io: I just want to understand why non-system owned are bad if all other constraints are met?

diman-io: I read the comments on the PR too, but there are no explanations there either as to why the system owned restriction is necessary.

jstarry: Yeah that isn't ideal, the main reason is that it limits the number of edge cases that core protocol devs need to consider when thinking about the ways that accounts could change. You could use a nonce account as the fee collector still and set the authority to a multisig. It's kinda hacky and not ideal, though. What do you think?

diman-io: I’m really sorry, but I have to disagree. The problem of forgetting to make a check is more an issue of implementation rather than the model itself. If, in the implementation, for example, an account can self-validate against declarative rules before saving, then this problem would not exist. The account wouldn't be recorded, the block wouldn't be created, and nothing bad would happen. I understand that such an implementation is the most expensive, but it remains a problem of implementation. And the check here is no more than after a regular transfer (any account can receive lamports). In my experience, artificially imposed constraints have led to further constraints in the future (like the random example about multisig; a nonce account is not quite the same, the signature of the account receiving fees is not needed on the node).

To be honest, for a complete picture, in my opinion, the ability to create non-rent exempt accounts was a very cool feature. But in that case, as I understand, there was a real reason why they needed to be phased out.

In this case, I still don’t see any real reasons. Sorry.

proposals/0085-additional-fee-collector-constraints.md
lheeger-jump: Your previous comment resolves this. Can disregard

proposals/0085-additional-fee-collector-constraints.md
jstarry: I flipped this to a set of positives which I hope is even clearer, thanks for pointing this out

jstarry:
I added more context to the motivations of this SIMD and added some arguments against other alternative solutions. Can everyone take another look and either approve or suggest changes? Thanks!

proposals/0085-additional-fee-collector-constraints.md
diman-io: Thank you for the explanation, now I understand the difference in our thinking approaches. I need some time to organize my thoughts.

PS. Sorry for the silence, I was on a trip and my brain is fried after long flights.

buffalojoec: > My preferred solution is to accept that depositing fees should continue happening outside of any transactions and then constrain the fee collector account to limit the "nasty edge cases." If you take issue with limiting to system owned accounts, I think that having some concept of a fee collector program and fee collector accounts (which have configurable authorities) in the protocol would be ideal and flexible enough for any usecases. But I think they end up looking pretty similar to the system program + nonce accounts so I don't it's necessary to create such a program.

Would you suggest this hypothetical alternative would involve the non-transactional fee deposit to land in the "fee collector" program's account, and then the authority would have to withdraw?

I personally don't think it should be a deterrent to end up with something like "looks like" the system-nonce relationship, because with this suggestion, it sounds like we might get two things:
- separation of duties, albeit maybe just in nomenclature
- ability to withdraw with _your_ signature of choice, not specifically a system-owned keypair
  - Think, for example, if I wanted to use my on-chain program to invoke-signed and withdraw from my fee collector account. This could be perfectly valid and still shave away the nasty edge cases mentioned.

proposals/0085-additional-fee-collector-constraints.md
```diff
@@ -0,0 +1,90 @@
...
+"validator node" program which allows configuring a "withdraw authority" and
+"vote fee payer" could help increase validator operation flexibility and
+increase clarity in how validator keys are used in the protocol.
+
+This approach requires a migration of all "fee collector" accounts as well as
+the development of a new on-chain program to manage the new "validator node"
+accounts. It will be a big effort compared to the proposed constraints in this
+SIMD and should be discussed in a new SIMD if this approach is desired.
+Furthermore, durable nonce accounts already have a configurable authority field
+which can be used to manage fee collector account funds in a more flexible way.
```
buffalojoec, line 61: How nasty would it be to allow either one? System-owned or this special program? 😅

buffalojoec (commented):
Nice write-up @jstarry!

I think 2) is a given, but in my opinion 1) is worth a bit more discussion.

Sure you can totally accomplish a "multi-sig-esque" setup with System and nonce accounts, but that's not what they're designed for. I think it's at least worth exploring alternatives that may steer us away from overloading functionality/responsibility on one particular component.

proposals/0085-additional-fee-collector-constraints.md
buffalojoec: > Yeah I think that the hypothetical alternative of having a specific fee collector program account is still a possible path to take in the future. For now, limiting to system program seems appropriate to limit edge cases.

All right, I think a two-phase approach is appropriate. Once anyone gets on the nonce method, though, I don't really see a reason why anybody would start up any dialogue about an official program for fee collection 😅

buffalojoec (approved):

jstarry:
Any other thoughts? Can I get an approval to merge this in @CriesofCarrots @lheeger-jump?

proposals/0085-additional-fee-collector-constraints.md
diman-io: Hello,

I've given this a lot of thought. I've revisited it many times but still haven't made a final decision. However, I want to share my thoughts.

On one hand, I understand the issue you're trying to solve. If we think on the level of transaction abstractions, our thinking breaks down, and it seems right to do what you suggest. Because (for me) in this case, we are (adding these rewards) somewhere in the on_write of a transaction (in this case a block, but we can abstract a block as a big transaction), and the necessary checks for modified accounts should be in on_write, putting us in a dilemma (since we have to do everything we actually want to postpone and have done for us).
So, looking at the model as I've tried to explain (not sure if I've made it clear), I want to minimize risks and keep it as simple (reliable) as possible.
However, the paradox is that, thinking this way, I always concluded that thinking at the transaction abstraction level here is wrong, not that it should be done this way.
That is, awarding rewards is a thing in itself and should be limited exclusively by the protocol's rules regarding accounts.
And at this point, it seems illogical not to be able to apply all protocol rules. Yes, I realize that these rules may change, but in any case, there should be a place where they are all collected in one place in the code (i.e., there should be a function capable of doing this). And the number of places where this function is called seems to be quite small because it's the lowest level. Thus, any change in protocol rules that leads to changing such a function simply entails reviewing its use.
So, the real risks are not substantial.

Therefore, imposing restrictions on the owner still seems excessive to me. And although I understand its reasons, I cannot accept it wholeheartedly. But this can hardly be considered an argument.

Also, in my reflections, I came to the idea that there could be more than one reward (i.e., a validator might split rewards based on some principle, like automatically replenishing an account to pay for voting fees and putting the rest in the treasury), but while I'd really like such flexibility, here I'm more inclined to say no, because it could create an unaccounted CU expenditure (as a resource of the validator), i.e., a potential point for attack.

Furthermore, I am negative about creating a special program for accounting for rewards. I don't really believe in it, given that the number of clients is increasing. Introducing such a program in the future would require agreement by many and could be delayed indefinitely. And considering that nonce accounts can be managed by programs, it's likely to be rejected. So, changes to the current proposal are likely to be the last (as I see it).
Moreover, I thought that the development of such programs should be the business of validators. That is, any validator can develop their own program according to their needs (which may include factors like taxation, for example).
Simple steps could include, for instance, using a PDA for each epoch or, say, a calendar month (which is closer to tax issues).
But creating a universal program that satisfies everyone is unlikely to succeed. Or it would be something very simple, in which case why a new program? Why not just expand the vote program, adding an account to which to attribute rewards (with the requirement that its owner be the vote program).
But returning to the current discussion, having the ability to attribute rewards to any account in this case does not contradict all this. But managing the nonce account from an external program doesn't either. So this is an argument neither for nor against.

Something like that

diman-io: tldr I do not mind.

proposals/0085-additional-fee-collector-constraints.md
```diff
@@ -0,0 +1,90 @@
...
+
+This applies to both transaction fees and rent fees. Note that the "rent-exempt"
+constraint was already added for rent fee collection in
+[Feature 30151](https://github.com/solana-labs/solana/issues/30151).
+
+## Motivation
+
+1. Fee collection occurs outside of the transaction runtime because the Solana
+   protocol mandates that fees are collected at the end of each block. By
+   restricting fee collector accounts to be system owned, the number of account
```
CriesofCarrots, line 29: Hmm, while I realize we call the accounts `fee_collector` accounts in code (and it's defined as a term below), the word "collected" feels problematic here, because to someone who hasn't read that stuff yet, it implies reaping as well as distributing. (The reaping (debiting) explicitly happens during transaction processing not at the end of the block.) How about this instead?

```suggestion
1. Fee distribution occurs outside of the transaction runtime because the Solana
   protocol mandates that fees are distributed to "fee collector" accounts at the end of each block. By
   restricting fee collector accounts to be system owned, the number of account
```

proposals/0085-additional-fee-collector-constraints.md
CriesofCarrots:
```suggestion
Rather than restricting fee collector accounts to be system-owned, a new type of
"validator node" account could be introduced. Currently, in normal validator
operations, the fee collector account is also used as the node id as well as
the vote fee payer. Introducing a validator node account that is owned by a
validator node program which allows configuring a withdraw authority and
vote fee payer could help increase validator operation flexibility and
increase clarity in how validator keys are used in the protocol.

This approach requires a migration of all fee collector accounts as well as
the development of a new on-chain program to manage the new validator node
accounts. It will be a big effort compared to the proposed constraints in this
SIMD and should be discussed in a new SIMD if this approach is desired.
Furthermore, durable nonce accounts already have a configurable authority field
which can be used to manage fee collector account funds in a more flexible way.
```
Typically, quotes for new terminology are only used on first use.

CriesofCarrots: Also, side-note, I prefer hyphenating modifiers like "fee-collector (account)" and "validator-node (account)"

proposals/0085-additional-fee-collector-constraints.md
```diff
@@ -0,0 +1,90 @@
...
+the development of a new on-chain program to manage the new "validator node"
+accounts. It will be a big effort compared to the proposed constraints in this
+SIMD and should be discussed in a new SIMD if this approach is desired.
+Furthermore, durable nonce accounts already have a configurable authority field
+which can be used to manage fee collector account funds in a more flexible way.
+
+## New Terminology
+
+Fee Collector: The account that receives block and rent fees collected by
+validators. Also known as the "node id".
```
CriesofCarrots, line 66: The "Also known as" sentence is a bit odd. I think remove it. However, I think in the design section it is probably worth making explicit how the fee-collector account is chosen/identified and that this isn't changing from existing protocol behavior.

CriesofCarrots (commented):
Coming back to this with fresher eyes, I had additional thoughts about the writing/clarity. Nothing major, though; seems really close.

proposals/0085-additional-fee-collector-constraints.md
jstarry: Yeah that's much clearer, thanks!

proposals/0085-additional-fee-collector-constraints.md
jstarry: Thanks for your thoughts @diman-io, I'm happy we agree that nonce accounts are flexible enough for now.

CriesofCarrots (approved):
Thanks for the polish! Lgtm

lheeger-jump (approved):
I am approving. This seems sufficiently well specified and, although I can imagine scenarios where you would want the fee collector to be a non-system account, it complicates the fee collection process. We can, and likely will, relax this constraint in the future but right now we need simplicity and clarity of design.

jacobcreech:
Looks like we got consensus from multiple clients. Thank you @jstarry!

PR #87, opened 2023-11-14
SIMD-0087: Explicit Sign Extension in SBPFv2

lheeger-jump:
This proposal adds a new sign extension instruction to SBPFv2 and removes implicit sign extension in some instructions under SBPFv2.


lheeger-jump:
We will be convening with the runtime team to clarify some of the instructions to be included here and the content of the proposal prior to releasing it from draft.

lheeger-jump:
Closing in favor of instead specifying the existing SBPF opcodes, then will revisit this.

PR #88, opened 2023-11-18
SIMD-0088: Enable Core BPF Programs

buffalojoec:
This proposal defines the concept of "Core BPF Programs" - programs which are essential to cluster operations but exist as BPF programs, rather than native built-in programs.

The proposal describes, in a general sense, how native programs are migrated to Core BPF, and defines the requirements for proposing a native program migration to Core BPF.

### DIFF
--- a/proposals/0088-enable-core-bpf-programs.md
+++ b/proposals/0088-enable-core-bpf-programs.md
@@ -12,13 +12,13 @@ feature: (fill in with feature tracking issues once accepted)
 
 ## Summary
 
-This proposal outlines the process by which engineers will manage upgrades to
-enshrined Core BPF programs, as well as migrate existing native programs to core
-BPF.
+This proposal introduces the concept of Core BPF programs: programs which are
+essential to network operations. Currently, these exist as built-in programs
+known as "native" programs.
 
-This SIMD also introduces the concept of Core BPF programs: programs which are
-essential to network operations. Currently, these are embedded programs known as
-"native" programs.
+This SIM also details the process by which existing native programs shall be
+ported to Core BPF, as well as the process for introducing brand-new Core BPF
+programs.
 
 ## Motivation
 
@@ -27,46 +27,39 @@ developers will no longer have to keep up with program changes in their runtime
 implementations. Instead, the program can just be modified once. 
 
 For this reason, it makes sense to introduce the concept of Core BPF programs:
-BPF programs the network depends on that should be upgraded with a special
-process.
+BPF programs the network depends on that should be treated with special care.
 
 ## Alternatives Considered
 
-An alternative approach to managing upgrades to Core BPF programs could be to
-manually upgrade these programs behind a feature gate using the runtime. This
-process is more complex and involves circumventing the established program
-upgrade process.
-
-It's possible that multi-sig program upgrades could be combined with
-feature-gates, however this implementation would add additional layers of
-complexity to the upgrade process.
+The alternative to Core BPF programs is to keep these essential programs as
+native programs. This would mean each validator client implementation would have
+to build and maintain these built-in programs with their runtime
+implementations, including any future changes to these programs introduced via
+SIMDs or other fixes.
 
 ## New Terminology
 
-- `Core BPF Program`: An **upgradeable** BPF program relied on by any part of
-  the Solana stack, including (but not limited to) consensus, transaction
-  processing, voting, staking, and account creation.
+- `Core BPF Program`: A BPF program relied on by any part of the Solana stack,
+  including (but not limited to) consensus, transaction processing, voting,
+  staking, and account creation.
 
 ## Detailed Design
 
-Core BPF programs shall be upgradeable BPF programs deployed with
-`BPFLoaderUpgradeab1e11111111111111111111111`.
+Core BPF programs in many ways will be designed no differently than any other
+BPF program on Solana. However, some programs may require special privileges,
+which is beyond the scope of this SIMD.
 
-The upgrade authority shall be a multi-sig comprised of keyholders from each
-validator client. Right now, that list includes Solana Labs, Jito, and Jump.
-This list can be updated in the future to include newer clients like Sig.
+When an existing native program is being proposed to migrate to Core BPF, or
+when a new Core BPF program is being introduced, at least one SIMD shall be
+published outlining at least the following details:
 
-**Upgrading a core BPF program** shall consist of a coordinated effort amongst
-core contributors from all validator clients. Similar to the feature gate
-process, this upgrade should occur on testnet, then devnet, then mainnet-beta.
-
-Upgraded versions of core BPF programs shall always be compiled using a
-verifiable build process.
+- How any required special privileges will be granted to the program in its BPF
+  form
+- How this program's upgrades will be managed after it becomes Core BPF
 
 **Migrating a native program to core BPF** shall consist of deploying a modifed
-version of the native program to a new arbitrary address as an **upgradeable**
-BPF program and using a feature gate to move the modified program in place of
-the existing program.
+version of the native program to a new arbitrary address and using a feature
+gate to move the modified program in place of the existing program.
 
 The feature gate is required to circumvent normal transaction processing rules
 and replace the contents of one account with another directly at the runtime
@@ -92,16 +85,13 @@ With this change, validator clients would no longer be required to implement
 changes to essential programs. Instead these programs could be modified just
 once. This reduces some engineering overhead on validator teams.
 
-This also introduces a multi-sig process, which core contributors across all
-validator teams must, in a timely manner, participate in.
-
 ## Security Considerations
 
 This proposal establishes the concept of relying on BPF programs that are not
-built into the runtime for essential cluster operations. With these programs
-being upgradeable, there are some obvious security considerations around who can
-upgrade these programs and when. With a proper multi-sig process in place, these
-risks are mitigated.
+built into the runtime for essential cluster operations. Depending on how these
+programs are elected to be upgraded, there are some obvious security
+considerations around who can upgrade these programs and when. Any new Core BPF
+program should follow a fully-fledged SIMD process addressing these concerns.
 
 When it comes to migrating native programs to core BPF, this change introduces a
 serious security consideration surrounding the replacement of an essential
--- a/proposals/0088-enable-core-bpf-programs.md
+++ b/proposals/0088-enable-core-bpf-programs.md
@@ -20,15 +20,6 @@ This SIMD also introduces the concept of Core BPF programs: programs which are
 essential to network operations. Currently, these are embedded programs known as
 "native" programs.
 
-With an explicit process in place for upgrading Core BPF programs, engineers can
-safely manage changes to essential programs through feature gates. The feature
-gate process requries incremental activations on each cluster: testnet, then
-devnet, then mainnet-beta. This is extremely valuable for ensuring a change has
-been safely integrated and will not have negative effects on mainnet-beta.
-
-This same explicit process can be used to migrate existing native programs to
-their new Core BPF versions.
-
 ## Motivation
 
 BPF programs offer less complexity for other clients, such as Firedancer, since
@@ -41,16 +32,10 @@ process.
 
 ## Alternatives Considered
 
-An alternative approach to managing upgrades to Core BPF programs could be
-to make these core BPF programs _upgradeable_ and establish a multi-signature
-upgrade authority. This authority could be comprised of key-holders from each
-validator client: Solana Labs, Jito, Jump, and possibly Syndica (Sig).
-
-Gating upgrades behind a feature-gate has numerous benefits, including a long
-lead time to deployment and requiring validators to upgrade their software to
-signal their acceptance of the change. Allowing Core BPF programs to be upgraded
-by a multi-sig without feature-gating strips core contributors of these
-benefits.
+An alternative approach to managing upgrades to Core BPF programs could be to
+manually upgrade these programs behind a feature gate using the runtime. This
+process is more complex and involves circumventing the established program
+upgrade process.
 
 It's possible that multi-sig program upgrades could be combined with
 feature-gates, however this implementation would add additional layers of
@@ -58,32 +43,40 @@ complexity to the upgrade process.
 
 ## New Terminology
 
-- `Core BPF Program`: A **non-upgradeable** BPF program relied on by any part of
+- `Core BPF Program`: An **upgradeable** BPF program relied on by any part of
   the Solana stack, including (but not limited to) consensus, transaction
   processing, voting, staking, and account creation.
 
 ## Detailed Design
 
-In the context of this design, **source program** shall be the modified version
-of a core BPF program intending to be moved in place of the existing program,
-while **target program** is that existing program.
+Core BPF programs shall be upgradeable BPF programs deployed with
+`BPFLoaderUpgradeab1e11111111111111111111111`.
 
-Core BPF programs shall be non-upgradeable BPF programs deployed with
-`BPFLoader2111111111111111111111111111111111`.
+The upgrade authority shall be a multi-sig comprised of keyholders from each
+validator client. Right now, that list includes Solana Labs, Jito, and Jump.
+This list can be updated in the future to include newer clients like Sig.
 
-Core BPF programs must only be modified through feature gates.
+**Upgrading a core BPF program** shall consist of a coordinated effort amongst
+core contributors from all validator clients. Similar to the feature gate
+process, this upgrade should occur on testnet, then devnet, then mainnet-beta.
 
-**Upgrading a core BPF program** shall consist of deploying the modified program
-to a new arbitrary address as a **non-upgradeable** BPF program and using a
-feature gate to move the modified program in place of the existing program. The
-feature gate is required to circumvent normal transaction processing rules and
-replace the contents of one account with another directly at the runtime level.
+Upgraded versions of core BPF programs shall always be compiled using a
+verifiable build process.
 
-Note, since this deployed program will be deployed using the non-upgradeable BPF
-loader (`BPFLoader2111111111111111111111111111111111`), it will consist of only
-a program account, with no additional program _data_ account.
+**Migrating a native program to core BPF** shall consist of deploying a modifed
+version of the native program to a new arbitrary address as an **upgradeable**
+BPF program and using a feature gate to move the modified program in place of
+the existing program.
 
-This process must adhere to the following steps:
+The feature gate is required to circumvent normal transaction processing rules
+and replace the contents of one account with another directly at the runtime
+level.
+
+In the context of this design, **source program** shall be the modified BPF
+version of a native program intending to be moved in place of the existing
+native program, while **target program** is that existing program.
+
+The migration process must adhere to the following steps:
 
 1. Verifiably build the modified program.
 2. Generate a new keypair for the **source** program.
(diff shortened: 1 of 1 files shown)


proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,143 @@
...
+process.
+
+With an explicit process in place for upgrading Core BPF programs, engineers can
+safely manage changes to essential programs through feature gates. The feature
+gate process requries incremental activations on each cluster: testnet, then
+devnet, then mainnet-beta. This is extremely valuable for ensuring a change has
+been safely integrated and will not have negative effects on mainnet-beta.
+
+This same explicit process can be used to migrate existing native programs to
+their new Core BPF versions.
```
joncinque, line 46: This is going outside of the motivation and summarizing the approach. How about moving this to the summary?

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Alternatives Considered
+
+An alternative approach to managing upgrades to essential core programs could be
```
joncinque, line 50: nit: let's use the new terminology!
```suggestion
An alternative approach to managing upgrades to core BPF programs could be
```

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,143 @@
...
+to make these core BPF programs _upgradeable_ and establish a multi-signature
+upgrade authority. This authority could be comprised of key-holders from each
+validator client: Solana Labs, Jito, Jump, and possibly Syndica (Sig).
+
+However, this approach adds additional layers of complexity around upgrading
+these programs behind a feature-gate.
```
joncinque, line 56: Sorry, I don't think I understand this point. How does a feature gate come into play if you're using a multisig? 

From the way I understand it, a multisig is more dangerous because it means potential changes at any time, whereas a feature gate is preferred because it has such a long lead time to deployment, and requires validators to upgrade their software to signal their acceptance of the change.

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,143 @@
...
+  processing, voting, staking, and account creation.
+- `Target Program`: The core program intending to be changed by some
+  feature activation.
+- `Source Program`: The modified version of a core program intending to be moved
+  in place of the existing core program.
+
+## Detailed Design
+
+Core BPF programs shall be non-upgradeable BPF programs deployed with
+`BPFLoader2111111111111111111111111111111111`.
```
joncinque, line 71: Side note: we should check with the runtime team to see if it's preferred to have them as "final" upgradeable programs. I always forget which one is better for the runtime and avoids copying the program data into the vm

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,143 @@
...
+
+## New Terminology
+
+- `Core BPF Program`: A **non-upgradeable** BPF program relied on by any part of
+  the Solana stack, including (but not limited to) consensus, transaction
+  processing, voting, staking, and account creation.
+- `Target Program`: The core program intending to be changed by some
+  feature activation.
+- `Source Program`: The modified version of a core program intending to be moved
+  in place of the existing core program.
```
joncinque, line 66: This doesn't seem like new terminology, and more about defining terms for the design. Let's move these into the detailed design portion

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Detailed Design
+
+Core BPF programs shall be non-upgradeable BPF programs deployed with
+`BPFLoader2111111111111111111111111111111111`.
+
+Core BPF programs must only be modified through feature gates.
+
+**Upgrading a core BPF program** shall consist of deploying the modified program
+source code to a new arbitrary address as a **non-upgradeable** BPF program and
```
joncinque, line 76: nit: it's not the source code deployed
```suggestion
to a new arbitrary address as a **non-upgradeable** BPF program and
```

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,143 @@
...
+creating a feature gate to, using the runtime, move the modified program in
+place of the existing program.
```
joncinque, line 78: nit: It might be worth making it completely clear what the runtime needs to do, ie
```suggestion
using a feature gate to move the modified program in
place of the existing program. The feature gate is required to circumvent
normal transaction processing rules and replace the contents of one account
with another directly at the runtime level.
```

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,143 @@
...
+the exact steps outlined above with the modified BPF version of the program.
+This will effectively replace the program at the target address with its core
+BPF implementation.
+
+Some additional checks should be run when doing a migration instead of an
+upgrade, such as:
+
+- Validating the native program's owner is
+  `NativeLoader1111111111111111111111111111111`
+- Ensuring the core BPF program has the correct state
```
joncinque, line 107: Sorry, what does this mean?

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,143 @@
...
+   **source** (modified) program.
+6. Follow the existing process for activating features.
+
+An additional optional field could be added to feature gate issues for the
+**target program** being upgraded.
+
+**Migrating a native program** to a core BPF program shall consist of following
+the exact steps outlined above with the modified BPF version of the program.
+This will effectively replace the program at the target address with its core
+BPF implementation.
```
joncinque, line 100: Up to you, but I think there might be more to this since the runtime will also need to remove any special-casing for that address during transaction processing.

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,143 @@
...
+
+This proposed change would result in a new process for upgrading core programs.
+Although this process will still require feature gates as it does now, the
+change is handled completely differently.
+
+The act of deploying the modified program and using a runtime feature gate to
+move it into the proper account will be an entirely new way of upgrading
+programs that were previously native.
+
+Core contributors will have to become comfortable with this process.
```
joncinque, line 119: I think we can remove this line

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,143 @@
...
+
+If a core program is modified incorrectly, or somehow erased during migration,
+it could have immediate and fatal consequences for the network.
+
+## Backwards Compatibility
+
+This feature itself does not directly introduce any breaking changes.
+
+When this proposed process is used to migrate a native program to core BPF, that
+program will be completely backwards compatible.
```
joncinque, line 139: It might be nice to point out that the core BPF implementation must be more than backwards compatible, and provide exactly the same results as the current native program.

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,143 @@
...
+surrounding the replacement of a core program with the contents of another
+account. This is an extremely sensitive process that must be handled with
+maximum caution.
+
+If a core program is modified incorrectly, or somehow erased during migration,
+it could have immediate and fatal consequences for the network.
+
+## Backwards Compatibility
+
+This feature itself does not directly introduce any breaking changes.
```
joncinque, line 136: I'm not sure this is true, since the BPF program will exist at the address instead of the native program, which is a breaking change on the network.

joncinque (commented):
Looks really good for the most part! Mostly some small bits to clean up, then we can start getting more eyes on it

proposals/0088-enable-core-bpf-programs.md
buffalojoec: Oh, I was referring to the addition of the code that does the swap/upgrade. Without actually using it on a program, merely introducing it off the hot path, doesn't introduce any breaking changes.

proposals/0088-enable-core-bpf-programs.md
buffalojoec: I think it's worth including the additional steps in the SIMD. What does it look like to remove special casing?

proposals/0088-enable-core-bpf-programs.md
buffalojoec: I was originally moving toward some validation/checks that had to do with native program instruction/state, but I don't think it's going to work or worth adding.

This was a straggler from those ideas.

proposals/0088-enable-core-bpf-programs.md
buffalojoec: Asked the VM guys, will circle back with an update.

proposals/0088-enable-core-bpf-programs.md
joncinque: Oh I see, so you have the option of doing either upgrade path? That seems overly complicated to describe, and we'll get much more out of differentiating with a plain multisig

buffalojoec: Lmk if the updated description is too verbose.

proposals/0088-enable-core-bpf-programs.md
buffalojoec: Actually circling back, it appears neither has a significant VM-related advantage.

With that being said, I feel like, since v2 (non-upgradeable) is _not_ deprecated, it actually feels a little more official to lock these special Core BPF programs into the "final" loader - as written in the SIMD now. Basically, they're intentionally designed to not be upgraded by a keypair authority. What do you think?

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Backwards Compatibility
+
+This proposal itself does not directly introduce any breaking changes. The code
+introduced to migrate native programs to core BPF programs will exist off of the
+runtime's "hot path" until it's actually used for a migration.
+
+When a migration is conducted, the core BPF version will be more than backwards
+compatible. It must provide the exact same results as the native program it aims
+to replace.
```
CriesofCarrots, line 111: This makes me feel more strongly about removing the phrase "modified program" everywhere above.

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -49,40 +49,41 @@ Core BPF programs in many ways will be designed no differently than any other
...
+In the context of this design, **target prorgam** refers to an existing native
+program, while **source program** refers to the BPF version to be moved into the
+existing program account.
 
-In the context of this design, **source program** shall be the modified BPF
-version of a native program intending to be moved in place of the existing
-native program, while **target program** is that existing program.
+In the context of this design, **source program** shall be the BPF version of
+a native program intending to be moved in place of the existing native program,
+while **target program** is that existing program.
```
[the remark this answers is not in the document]
buffalojoec: ahh whoops!

proposals/0088-enable-core-bpf-programs.md
buffalojoec: Just for clarity, the concept of "which loader owns all formerly native Core BPF programs" has been removed from this SIMD. Now, one must define this specification in the SIMD that outlines that specific native program's migration to Core BPF.

proposals/0088-enable-core-bpf-programs.md
buffalojoec: Since this SIMD is now slimmed down and mentions that upon any migration, a SIMD outlining the migration and Core BPF specifications should be published, I think describing special casing removal isn't necessary here.

However, I've added it to the necessary details for such a SIMD as mentioned in this proposal.

buffalojoec:
@joncinque @CriesofCarrots I think this SIMD is good to go.

@lheeger-jump @ripatel-fd can you guys have a look please?

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,112 @@
...
+feature: (fill in with feature tracking issues once accepted)
+---
+
+## Summary
+
+This proposal introduces the concept of Core BPF programs: programs which are
+essential to network operations. Currently, these exist as built-in programs
+known as "native" programs.
+
+This SIMD also details the process by which existing native programs shall be
```
CriesofCarrots, line 19:
```suggestion
This SIMD details the process by which existing native programs can be
```

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,112 @@
...
+ported to Core BPF, as well as the process for introducing brand-new Core BPF
+programs.
+
+## Motivation
+
+BPF programs offer less complexity for other clients, such as Firedancer, since
```
CriesofCarrots, line 25:
```suggestion
BPF programs offer less complexity than native programs for other clients, such as Firedancer, since
```

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,112 @@
...
+or when a new Core BPF program is being introduced, at least one SIMD shall be
+published outlining at least the following details:
+
+- How any required special privileges will be granted to the program in its BPF
+  form
+- How this program's upgrades will be managed after it becomes Core BPF
+
+**Migrating a native program to core BPF** shall consist of deploying a
+BPF version of the native program to a new arbitrary address and, using a
+feature gate, move the BPF program, replacing the existing native program at
```
CriesofCarrots, line 62:
```suggestion
BPF version of the native program to a new arbitrary address and using a
feature gate to move the BPF program, replacing the existing native program at
```

CriesofCarrots (commented):
Just a couple language suggestions. Content lgtm!

CriesofCarrots (approved):

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,113 @@
...
+existing program account.
+
+The migration process must adhere to the following steps:
+
+1. Verifiably build the source BPF program.
+2. Generate a new keypair for the source program.
+3. Deploy the program to the source program address.
+4. Generate a new keypair for the feature gate.
+5. Create a new feature gate to replace the target program with the source BPF
+   program.
```
2501babe, line 77: do you have a mechanism in mind for how this replacement would take place? unless im missing something even with loader v4 you would need to run a transaction to replace the bytecode of target with source

2501babe:
just hopping in because ill be working on some of these ports, ty for proposing a process!

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,113 @@
...
+
+In the context of this design, **target program** refers to an existing native
+program, while **source program** refers to the BPF version to be moved into the
+existing program account.
+
+The migration process must adhere to the following steps:
+
+1. Verifiably build the source BPF program.
+2. Generate a new keypair for the source program.
+3. Deploy the program to the source program address.
```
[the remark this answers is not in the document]
buffalojoec: I'm thinking it _should_ work if you're just deploying and not invoking the program (right?).

In theory, if the core BPF program at say `11111111111111111111111111111111` has special privileges, then unless the executable lives at `11111111111111111111111111111111` it's not going to be able to use those privileges. So you deploy the BPF version of it to some arbitrary address, and if it was to be invoked at that arbitrary address it would fail to load. However, it shouldn't ever be invoked at that arbitrary address. It's only temporarily living there until the swap is complete.

Additionally, I think if any tweaks are necessary in the future as we figure out how to provide all of these core BPF programs with the necessary privileges, we can issue another proposal to modify the process or introduce a new one.

proposals/0088-enable-core-bpf-programs.md
buffalojoec: We would basically build out a revamped version of this function, and call it from a runtime feature gate on epoch rollover.

https://github.com/solana-labs/solana/blob/a2d2eb3ba80824f55c9cbd267ab881a285ca324a/runtime/src/bank.rs#L7985

joncinque (approved):

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,113 @@
...
+BPF program on Solana. However, some programs may require special privileges,
+which is beyond the scope of this SIMD.
+
+When an existing native program is being proposed to be migrated to Core BPF,
+or when a new Core BPF program is being introduced, at least one SIMD shall be
+published outlining at least the following details:
+
+- How any required special privileges will be granted to the program in its BPF
+  form
+- How this program's upgrades will be managed after it becomes Core BPF
```
lheeger-jump, line 59: This show be standardized

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,113 @@
...
+
+Core BPF programs in many ways will be designed no differently than any other
+BPF program on Solana. However, some programs may require special privileges,
+which is beyond the scope of this SIMD.
+
+When an existing native program is being proposed to be migrated to Core BPF,
+or when a new Core BPF program is being introduced, at least one SIMD shall be
+published outlining at least the following details:
+
+- How any required special privileges will be granted to the program in its BPF
```
lheeger-jump, line 57:
```suggestion
- How any required special privileges will be granted to the program in its BPF
- The interface for the program.
- A precise and complete specification of the behavior.
```

proposals/0088-enable-core-bpf-programs.md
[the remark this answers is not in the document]
lheeger-jump: More discussion is required from VM stakeholders about privileged execution before I can consider approval. I do not want to add more syscalls if they all have to be re-invented in PRv2.

lheeger-jump, line 74: Can't we just use the same address?

lheeger-jump (changes requested):
I also apologize for my very late review. We need more input from the VM team and the deployment procedure is unclear. Would this be an EOA that controls the account? I think changing the address would really hurt app devs and UX.

proposals/0088-enable-core-bpf-programs.md
Lichtso: I think we need to replace the program at the original address anyway, otherwise all its owned accounts will not be available to the replacement.

buffalojoec: To replace a native program with a BPF program, you'd have to deploy the BPF version to some arbitrary address, then move it into the account at the original address. That would result in the address being unchanged and the program being replaced with BPF.

See [my comment here](https://github.com/solana-foundation/solana-improvement-documents/pull/88#discussion_r1441929536).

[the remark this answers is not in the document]
buffalojoec: > More discussion is required from VM stakeholders about privileged execution before I can consider approval. I do not want to add more syscalls if they all have to be re-invented in PRv2.

Fair point. I guess the question is, does the number of syscalls block us from migration _any_ native program to BPF?

In other words, is it possible we could move this SIMD along to acknowledge that at least _some_ programs will be migrated, and then assess the impact of any necessary syscalls when the SIMDs outlining migration specifications are proposed for a program requiring privileges?

lheeger-jump: That is fine.

Lichtso:
Another slight change in behavior in which the on-chain version can not precisely match the original built-in is the compute meter and other resource limits. I think that should be mentioned in the document.

proposals/0088-enable-core-bpf-programs.md
buffalojoec: For example, we may choose to make some programs upgradeable - like Feature Gate - but others could be non-upgradeable - like System - therefore they'd have separate loaders and different ways of introducing changes, if need be.

proposals/0088-enable-core-bpf-programs.md
buffalojoec: Added.

buffalojoec:
> I also apologize for my very late review. We need more input from the VM team and the deployment procedure is unclear. Would this be an EOA that controls the account? I think changing the address would really hurt app devs and UX.

As mentioned in [this comment](https://github.com/solana-foundation/solana-improvement-documents/pull/88#discussion_r1443124524), the address would _not_ be changed. I've added some details on this into the SIMD.

buffalojoec:
> Another slight change in behavior in which the on-chain version can not precisely match the original built-in is the compute meter and other resource limits. I think that should be mentioned in the document.

Added.

proposals/0088-enable-core-bpf-programs.md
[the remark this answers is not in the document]
buffalojoec: Up-voting previous message by @ripatel-fd. If we can move this proposal along to introduce these programs and their migration path, then each program's migration SIMD - outlining the details of its migration to BPF - can tackle privileges.

If we learn that privileged programs will need a different migration path, we can publish another SIMD that adds the necessary changes to the migration process, and block any program migration SIMDs on that new proposal.

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,113 @@
...
+status: Draft
+created: 2023-11-07
+feature: (fill in with feature tracking issues once accepted)
+---
+
+## Summary
+
+This proposal introduces the concept of Core BPF programs: programs which are
+essential to network operations. Currently, these exist as built-in programs
+known as "native" programs.
```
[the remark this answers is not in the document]
buffalojoec: I think this is a good shout. Technically, many programs are implemented as native/built-in for performance or privilege reasons. It just so happens these enhanced programs are often relied on by the cluster.

The naming convention for these programs should encapsulate:
1. The program may be relied on by the cluster, but may not be
2. The program may have certain privileges, but may not
3. The program may be non-upgradeable and changed via runtime feature gates, but it may not

In my opinion:
1. "essential" (or similar) captures true, not false
2. "enshrined" (or similar) captures true, not false
3. "special" (or similar) captures true, not false

I feel like "core" is broad enough to encapsulate all branches here. I'm happy to update the proposal to relax some of the wording around "essential to cluster operations" in favor of something similar to what I've written here.

What do you think?

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -0,0 +1,128 @@
...
+
+This proposal itself does not directly introduce any breaking changes. The code
+introduced to migrate native programs to core BPF programs will exist off of the
+runtime's "hot path" until it's actually used for a migration.
+
+When a migration is conducted, the core BPF version will be more than backwards
+compatible. It must provide the exact same results as the native program it aims
+to replace.
+
+When a migration is conducted, the BPF version of a native program will be
```
pgarg66, line 116: Dup of line 112?

buffalojoec: Yes, sorry. Thanks!

pgarg66 (approved):

proposals/0088-enable-core-bpf-programs.md
buffalojoec: @lheeger-jump Curious what you think here. I'm leaning towards less standardization for the reasons listed above.

proposals/0088-enable-core-bpf-programs.md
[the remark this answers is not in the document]
buffalojoec: @ripatel-fd If we're good with "core" I think I'll resolve this one? Lmk, thanks.

Lichtso (approved):

jacobcreech:
Looks like we've got general consensus. I'll give a last call for reviews and merge on Thursday, January 25th if there are no further objections from those with level 2 access.

jacobcreech (approved):
LGTM. Thanks Joe for moving this along

Dakota48423 (approved):

Pietrek96 (approved):

Pietrek96 (approved):

t-nelson:
can the bot org block users that approve that aren't on the approver list?

Benhawkins18:
That could be arranged. Fixing the bot so it doesn't respond to those not on the approved list and making it ignore closed permitted pull requests

PR #90, opened 2023-11-21
fix custom ci for metadata

buffalojoec:
Aligns CI with the template by adding `supersedes` and `extends` to the optional metadata.


buffalojoec:
@jacobcreech I can't request review.

jacobcreech (approved):
Thanks!

PR #94, opened 2023-12-13
SIMD-0094: deprecate executable update for bpf loader

HaoranYi:
Add deprecate executable metadata update for bpf loader SIMD


Lichtso (approved):
This is already in effect on MNB

Lichtso (changes requested):
This SIMD now extends #162 and needs to be be rewritten to address the outstanding points in the beginning of the "Detailed Design" section of SIMD-0162.

HaoranYi:
> This SIMD now extends #162 and needs to be be rewritten to address the outstanding points in the beginning of the "Detailed Design" section of SIMD-0162.

Yeah. I pushed a commit to update this SIMD according to the changes in SIMD-0162.

proposals/0094-deprecate-executable-update.md
```diff
@@ -0,0 +1,67 @@
...
+
+This will affect the following scenarios.
+
+- program accounts hash after deployment. Since this change is guarded by a
+  feature, program accounts hash change won't break consensus.
+- any dapps that depend on `is_executable` to be true on the serialized
+  instruction accounts. Existing program deployed on chain will work fine.
+  However, if the program is redeployed, it may be broken. Before redeployment,
+  dapps developer will need to check and update the program if it depends on
+  `is_executable`.
```
buffalojoec, line 55: Possibly a big deal for dApp developers, cc @acheroncrypto

proposals/0094-deprecate-executable-update.md
```diff
@@ -0,0 +1,67 @@
...
+- CPI won't ignore changes made by caller to instruction accounts (Currently CPI
+  ignores changes made by the caller to instruction accounts which has the
+  `is_executable` flag set). For correctness, this change will be fine. It just
+  becomes a bit less efficient with more checks.
```
buffalojoec, line 59: I think this could use a little more fleshing out. Are you suggesting we accept this behavioral change, or add checks to handle changes to non-executable program accounts in CPI? Perhaps this deserves a paragraph in "Detailed Design".

HaoranYi: Yes, updated.

I am suggesting to accept the behavioral change. Since we are removing the shortcut and handle it in a more general way, the correctness is guaranteed, fwik.

proposals/0094-deprecate-executable-update.md
acheroncrypto: > Possibly a big deal for dApp developers, cc @acheroncrypto

It shouldn't affect Anchor as long as `AccountInfo.executable` doesn't change (see https://github.com/solana-labs/solana/issues/34425#issuecomment-1915250031).

acheroncrypto: The usage of `is_executable` and `executable` is a bit confusing. I think using something like `<TYPE>.is_executable` would help resolve this ambiguity.

HaoranYi: > It shouldn't affect Anchor as long as `AccountInfo.executable` doesn't change (see [solana-labs/solana#34425 (comment)](https://github.com/solana-labs/solana/issues/34425#issuecomment-1915250031)).

Yeah. updated.

HaoranYi: > It shouldn't affect Anchor as long as `AccountInfo.executable` doesn't change (see [solana-labs/solana#34425 (comment)](https://github.com/solana-labs/solana/issues/34425#issuecomment-1915250031)).

@acheroncrypto With this proposal, `AccountInfo.executable`  will change. In the original implementation that you mentioned above, `AccountInfo.executable` doesn't change because we emulate it at serialization time. But that turns out to be very expensive - 10X increase of serialization. Therefore,  in this proposal, we won't emulate executable at serialization time. Therefore, dapp can't depends on `AccountInfo.executable` directly.

buffalojoec (commented):
I think things look good from my side, but leaving to @Lichtso for approval. Thanks @HaoranYi!

PR #108, opened 2024-01-18
SIMD-108: Transaction size fee

joncinque:
The small size of transasctions on Solana can limit how much users can do on the network atomically, forcing apps to do unsafe things such as signing multiple transactions with the assumption that they can be executed serially.

Before we can increase the maximum transaction size, however, the network needs an economic backstop to make sure that validators are properly compensated in the case that large transactions negatively impact transaction throughput or bandwidth usage.

This proposal only covers the economic backstop by including a new fee based on transaction size. It includes the motivation for the fee along with proposed parameters for a safe initial iteration.

### DIFF
--- a/proposals/0108-transaction-size-fee.md
+++ b/proposals/0108-transaction-size-fee.md
@@ -1,0 +1,190 @@
+---
+simd: '0108'
+title: Transaction Size Fee
+authors:
+  - Jon Cinque (Solana Labs)
+category: Standard
+type: Core
+status: Draft
+created: 2024-01-18
+feature: (fill in with feature tracking issues once accepted)
+---
+
+## Summary
+
+Add a new transaction fee based on the serialized size of the transaction.
+
+## Motivation
+
+Currently, fees are based entirely on the time it takes to process the
+transaction in a node:
+
+* the fee per signature covers the sigverify time, before execution
+* priority fees are priced on compute units, which cover the virtual machine
+execution time
+
+A transaction takes up more than time resources, however. The network must also
+broadcast the transaction through Turbine so that all nodes can replay it, which
+consumes bandwidth.
+
+To see this in practice, let's imagine we have two transactions. They take the
+same amount of time to execute, meaning that they have the same number of
+signatures to verify, and they consume the same number of compute units.
+
+If one of those transactions has a larger payload, however, it will use up more
+bandwidth, and potentially reduce the overall transaction throughput of the network.
+
+To quantify the impact of transaction payload size, we ran tests and reported
+the results in [this issue](https://github.com/solana-labs/solana/issues/28063).
+We setup a small private test network, and used `solana-bench-tps` to send large
+numbers of transactions with useless random padding bytes, ranging from 0 to 2206.
+
+The test network removed all QUIC congestion control to better capture total
+throughput.
+
+The inner work in each transaction was exactly the same, performing a CPI into 
+the system program to transfer SOL. This way, we fixed the time dimension of
+transaction process and isolated the impact of transaction size.
+
+The base transaction size is 257 bytes, and the tool sent batches of 50k
+transactions. Here is the table of results, taken from
+[this comment](https://github.com/solana-labs/solana/issues/28063#issuecomment-1727151064):
+
+Without congestion control, there is slight decrease in transactions per second
+with more padding bytes. However, the overall variance is larger than any effect
+from padding bytes. So there may be some negligible effect based on transaction
+size.
+
+This test, however, was only on a small network, so we can't come to any
+conclusions about how larger transactions will impact a real network.
+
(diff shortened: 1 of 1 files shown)


aeyakovenko:
The system should handle more bandwidth in fewer larger txs. I am not sure the super linear component makes sense.

aeyakovenko:
Wouldn't just accounting for CUs used from the larger txs be enough? Then priority fees will handle the rest.

proposals/0108-transaction-size-fee.md
```diff
@@ -0,0 +1,190 @@
...
+
+The fee coefficients can be updated through a stake-weighted vote from the
+validator set, but the update process is out of scope for this document.
+
+### Validator Changes
+
+When calculating the fee, the transaction size fee must be calculated and deducted
+from the fee payer account as outlined earlier.
+
+As with other fees, half of the fee is burned, and half goes to the leader.
```
aeyakovenko, line 124: this breaks prioritization.  should leaders do priority * CUs + 1/2 * base + 1/2 * tx size / CUs as the priority function?

the fee doesn't make sense and is out of the blue tbh.  it should just be a CU modifier based on benchmarks.

aeyakovenko (changes requested):
fee doesn't make sense with the prioritization model, should be a CU modifier based on benchmarks

proposals/0108-transaction-size-fee.md
```diff
@@ -0,0 +1,190 @@
...
+Currently, fees are based entirely on the time it takes to process the
+transaction in a node:
+
+* the fee per signature covers the sigverify time, before execution
+* priority fees are priced on compute units, which cover the virtual machine
+execution time
+
+A transaction takes up more than time resources, however. The network must also
+broadcast the transaction through Turbine so that all nodes can replay it, which
+consumes bandwidth.
```
tao-stones, line 28: From the perspective of evaluating "resources consumed by transaction", transaction byte size correlates to network ingress/egress cost, and separately long term storage cost. 

I was thinking alone [the line ](https://gist.github.com/tao-stones/6fc77b477d1b946db98e8fe8e3e005ee) that transaction base fee (or its total static fee) accounts for all resources used (including CPUs, memory etc). The cost of bandwidth and ledger storage can all be folded into `CU`. 

(now I caught up with Discord discussion about CU :) I still in fav of pricing everything in CU - perhaps in a different name, that better implies "it takes this amount computing power to do xyz".

proposals/0108-transaction-size-fee.md
```diff
@@ -0,0 +1,190 @@
...
+
+We propose a solution with a few important features:
+
+* the new fee does not impact current transactions
+* the fee contains a non-linear component to dissuade huge transactions,
+especially while we see the impact of large transactions on a real cluster
+* the fee also includes a linear component for the future, if the effect on a
+real cluster is small
+
+Given `total_transaction_bytes` and the fee parameters
```
tao-stones, line 78: Assume `total_transaction_bytes` is cheaply available to validators after deserializing?

proposals/0108-transaction-size-fee.md
```diff
@@ -0,0 +1,190 @@
...
+`square_lamports_per_byte` and `linear_lamports_per_byte`, the fee is given as:
+
+```
+transction_bytes_over_mtu = total_transaction_bytes - 1232
+linear_fee = ceil(linear_lamports_per_byte * transaction_bytes_over_mtu)
+square_fee = ceil((square_lamports_per_byte * transaction_bytes_over_mtu) ^ 2)
+transaction_size_fee = linear_fee + square_fee
+```
```
tao-stones, line 86: +1 for being deterministic, so it can be calculated once then stored for later use

joncinque:
Thanks for your feedback and discussion everyone. It seems like it makes the most sense to fold this into the CU model since that's the only way to prioritize transactions at the moment. I'll go back to do some testing and propose a CU-based model for the cost. Closing this in the meantime.

aeyakovenko:
@joncinque i had a recent shower thought on this.  The model should both, add to CUs as best we can as an approximation, and also have a watermark which can't be exceeded.  So similar to the watermark for CUs per account

bw-solana:
Need to revive something like this.
Replay is network bound.
Today, a txs number of bytes is arguably more important than what's inside the tx.
Large tx does not necessarily imply high CU count (or deserialization of input data does not cost enough).

proposals/0108-transaction-size-fee.md
natalejoeraco1: Thanks

proposals/0108-transaction-size-fee.md
natalejoeraco1: export-blocks-1738806427744.csv

PR #116, opened 2024-01-28
Amend SIMD-15 with latest labs implementation details

jstarry:
The labs implementation diverged from the initial SIMD 15 proposal so this change amends that proposal to reflect the labs implementation.

### DIFF
--- a/proposals/0015-partitioned-epoch-reward-distribution.md
+++ b/proposals/0015-partitioned-epoch-reward-distribution.md
@@ -123,19 +123,81 @@ deterministic manner for the current epoch, while also randomizing the
 distribution across different epochs, the partitioning of all rewards will be
 done as follows.
 
-The reward results are sorted by Stake Account address, and randomly shuffled
-with a fast `rng` seeded by current epoch. The shuffled reward results are then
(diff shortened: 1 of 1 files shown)


proposals/0015-partitioned-epoch-reward-distribution.md
```diff
@@ -123,21 +123,83 @@ deterministic manner for the current epoch, while also randomizing the
...
+address into a u64 hash value. The reward distribution block index `I` can then
+be computed by applying the following formula to the hash:
+
+```
+I = (M * stake_address_hash) / 2^64
+```
+
+### Reward Distribution Snapshot State
+
+An additional field `epoch_rewards_status` will be added to serialized bank
```
HaoranYi, line 154: Maybe it is worth emphasizing that `epoch_rewards_status` are of type option. Thats means only snapshots, which are taken during reward period, store the fulllist of stake rewards. . For snapshots that are taken outside of reward period, we don't need to store the rewards.

HaoranYi (commented):
Overall looks good to me.

jstarry:
Closed in favor of https://github.com/solana-foundation/solana-improvement-documents/pull/118

PR #118, opened 2024-02-16
SIMD-0118: Partitioned Epoch Rewards, amend/extend design

CriesofCarrots:
This SIMD supersedes SIMD-0015 with some new design elements.
I have begun by copying in the original SIMD so that the changes can be seen more easily in subsequent commits.

https://github.com/solana-foundation/solana-improvement-documents/pull/116 may be useful as a reference, as it describes the existing Labs implementation.

### DIFF
--- a/proposals/0118-partitioned-epoch-reward-distribution.md
+++ b/proposals/0118-partitioned-epoch-reward-distribution.md
@@ -103,17 +103,17 @@ Reward calculation phase computes all the rewards that need to be distributed
 for the active stake accounts, and partitions the reward into a number of
 chunks for distribution in the next phase.
 
-Currently, on Solana Mainnet Beta with ~550K active stake accounts, it shows
-that epoch reward calculation takes around 10 seconds on average. This will
-make it impossible to perform rewards computation synchronous within one block.
+Currently, on Solana Mainnet Beta with ~550K active stake accounts, epoch reward
+calculation takes around 10 seconds on average. This will make it impossible to
+perform rewards computation synchronous within one block.
 
 However, there are quite a few promising optimizations that can cut down the
 reward computation time. An experiment for reward calculation optimization
 (https://github.com/solana-labs/solana/pull/31193) showed that we can cut the
 reward calculation time plus vote reward distribution time to around 1s. This
 makes synchronous reward computation and asynchronous reward distribution a
-feasible approach. We also believe that there is still more rooms for more
-optimization to further cut down the above timing.
+feasible approach. We also believe that there is room for more optimization to
+further cut down the above timing.
 
 Therefore, the following design is based on the above optimization. The
 reward calculation will be performed at the first block of the epoch. Once the
@@ -126,19 +126,30 @@ deterministic manner for the current epoch, while also randomizing the
 distribution across different epochs, the partitioning of all rewards will be
 done as follows.
 
-The reward results are sorted by Stake Account address, and randomly shuffled
-with a fast `rng` seeded by current epoch. The shuffled reward results are then
-divided into `M` chunks. This process will ensure that the reward distribution
-is deterministic for the current epoch, while also introducing a degree of
-randomness between epochs.
-
 To minimize the impact on block processing time during the reward distribution
-phase, a total of 4,096 accounts will be distributed per block. The total
-number of blocks needed to distributed rewards is given by the following
-formula to avoid using floating point value arithmetic:
+phase, a target of 4,096 stake rewards will be distributed per block. The total
+number of blocks `M` needed to distributed rewards is given by the following
+formula to round up to the nearest integer without using floating point value
+arithmetic:
 
 ```
 M = ((4096 - 1)+num_stake_accounts)/4096
+```
+
+To safeguard against the number of stake accounts growing dramatically and
+overflowing the number of blocks in an epoch, the number of blocks is capped at
+10% of the number of block in an epoch (currently 432,000).
+
+The [SipHash 1-3](https://www.aumasson.jp/siphash/siphash.pdf) pseudo-random
+function is used to hash stake account addresses efficiently and uniformly
+across the blocks in the reward distribution phase. The hashing function for an
+epoch is created by seeding a new `SipHasher` with the parent block's blockhash.
+This hashing function can then be used to hash each active stake account's
+address into a u64 hash value. The reward distribution block index `I` can then
+be computed by applying the following formula to the hash:
+
+```
+I = (M * stake_address_hash) / 2^64
 ```
 
 ### `EpochRewards` Sysvar Account
--- a/proposals/0118-partitioned-epoch-reward-distribution.md
+++ b/proposals/0118-partitioned-epoch-reward-distribution.md
@@ -117,7 +117,8 @@ M = ((4096 - 1)+num_stake_accounts)/4096
 
 To safeguard against the number of stake accounts growing dramatically and
 overflowing the number of blocks in an epoch, the number of blocks is capped at
-10% of the total slots in an epoch.
+10% of the total slots in an epoch. If the block cap is reached, accounts per
+partition is allowed to surpass the 4,096 target.
 
 The [SipHash 1-3](https://www.aumasson.jp/siphash/siphash.pdf) pseudo-random
 function is used to hash stake account addresses efficiently and uniformly
--- a/proposals/0118-partitioned-epoch-reward-distribution.md
+++ b/proposals/0118-partitioned-epoch-reward-distribution.md
@@ -153,23 +153,30 @@ Prefixes indidate the offset of each field, and padding bytes are always zero.
 #[repr(C)] // C representation, 8-byte alignment
 struct EpochRewards{
    // whether the rewards period (calculation and distribution) is active
-   /* 0x00 */ active: bool, // byte. false: 0x00, true: 0x01
+   // byte. false: 0x00, true: 0x01
+   /* 0x00 */ active: bool,
 
-   /* 0x08 */ distribution_starting_block_height: u64, // little-endian unsigned 64-bit integer
+   // little-endian unsigned 64-bit integer
+   /* 0x08 */ distribution_starting_block_height: u64,
    
-   /* 0x10 */ num_partitions: u64, // little-endian unsigned 64-bit integer
+   // little-endian unsigned 64-bit integer
+   /* 0x10 */ num_partitions: u64,
    
-   /* 0x18 */ parent_blockhash: Hash, // byte-array of length 32
+   // byte-array of length 32
+   /* 0x18 */ parent_blockhash: Hash,
 
    // total points calculated for the current epoch, where points equals the sum
    // of delegated stake * credits observed for all delegations
-   /* 0x38 */ total_points: u128, // little-endian unsigned 128-bit integer
+   // little-endian unsigned 128-bit integer
+   /* 0x38 */ total_points: u128, 
 
    // total rewards for the current epoch, in lamports
-   /* 0x48 */ total_rewards: u64, // little-endian unsigned 64-bit integer
+   // little-endian unsigned 64-bit integer
+   /* 0x48 */ total_rewards: u64,
 
    // distributed rewards for the current epoch, in lamports
-   /* 0x50 */ distributed_rewards: u64, // little-endian unsigned 64-bit integer
+   // little-endian unsigned 64-bit integer
+   /* 0x50 */ distributed_rewards: u64, 
 }
 ```
 
--- a/proposals/0118-partitioned-epoch-reward-distribution.md
+++ b/proposals/0118-partitioned-epoch-reward-distribution.md
@@ -147,28 +147,29 @@ include: the parameters needed to recalculate the reward partitions, the total
 rewards to be distributed, and the amount of rewards distributed so far.
 
 The layout of `EpochRewards` sysvar is shown in the following pseudo code.
+Prefixes indidate the offset of each field, and padding bytes are always zero.
 
 ```
-#[repr(C)] // C representation, ie. 8-byte alignment
+#[repr(C)] // C representation, 8-byte alignment
 struct EpochRewards{
    // whether the rewards period (calculation and distribution) is active
-   active: bool, // byte. false: 0x00, true: 0x01
+   /* 0x00 */ active: bool, // byte. false: 0x00, true: 0x01
 
-   distribution_starting_block_height: u64, // little-endian unsigned 64-bit integer
+   /* 0x08 */ distribution_starting_block_height: u64, // little-endian unsigned 64-bit integer
    
-   num_partitions: u64, // little-endian unsigned 64-bit integer
+   /* 0x10 */ num_partitions: u64, // little-endian unsigned 64-bit integer
    
-   parent_blockhash: Hash, // byte-array of length 32
+   /* 0x18 */ parent_blockhash: Hash, // byte-array of length 32
 
    // total points calculated for the current epoch, where points equals the sum
    // of delegated stake * credits observed for all delegations
-   total_points: u128, // little-endian unsigned 128-bit integer
+   /* 0x38 */ total_points: u128, // little-endian unsigned 128-bit integer
 
    // total rewards for the current epoch, in lamports
-   total_rewards: u64, // little-endian unsigned 64-bit integer
+   /* 0x48 */ total_rewards: u64, // little-endian unsigned 64-bit integer
 
    // distributed rewards for the current epoch, in lamports
-   distributed_rewards: u64, // little-endian unsigned 64-bit integer
+   /* 0x50 */ distributed_rewards: u64, // little-endian unsigned 64-bit integer
 }
 ```
 
--- a/proposals/0118-partitioned-epoch-reward-distribution.md
+++ b/proposals/0118-partitioned-epoch-reward-distribution.md
@@ -211,10 +211,10 @@ are in progress in order to prevent any changes interfering with reward
 distribution.
 
 Therefore, the Stake Program needs access to a syscall which reports whether the
-distribution phase is active. This new syscall `sol_get_epoch_rewards_status`
-should report the value of `EpochRewards::active`. All Stake Program
+distribution phase is active. This new syscall `sol_get_epoch_rewards_sysvar`
+should return the values of the `EpochRewards` sysvar. All Stake Program
 instructions that mutate stake data or debit stake balances must be disabled
-when `sol_get_epoch_rewards_status` is true.
+when `EpochRewards::active` is true.
 
 Any transaction that attempts to invoke such an instruction will fail with this
 new error code:
@@ -225,8 +225,9 @@ StakeError {
 }
 ```
 
-Other users can access the `sol_get_epoch_rewards_status` to determine the
-distribution-phase status from within the SVM.
+Other users can access the `active` field from the
+`sol_get_epoch_rewards_sysvar` syscall to determine the distribution-phase
+status from within the SVM.
 
 ## Impact
 
--- a/proposals/0118-partitioned-epoch-reward-distribution.md
+++ b/proposals/0118-partitioned-epoch-reward-distribution.md
@@ -206,24 +206,29 @@ partition indicated by comparing its current block height to
 comparing a partial sum of the rewards calculation (those partitions expected to
 have been distributed) with the `EpochRewards::distributed_rewards` field.
 
-### Restrict Stake Account Access
+### Restrict Stake Account Mutation
 
-To avoid the complexity and potential rewards in unexpected stake accounts,
-the stake program is disabled during the epoch reward distribution period.
+The stake accounts need to be preserved in credit-only state while distributions
+are in progress in order to prevent any changes interfering with reward
+distribution.
 
-Any transaction that invokes staking program during this period will fail with
-a new unique error code - `StakeProgramUnavailable`.
+Therefore, the Stake Program needs access to a syscall which reports whether the
+distribution phase is active. This new syscall `sol_get_epoch_rewards_status`
+should report the value of `EpochRewards::active`. All Stake Program
+instructions that mutate stake data or debit stake balances should be disabled
+when `sol_get_epoch_rewards_status` is true.
 
-That means all updates to stake accounts have to wait until the rewards
-distribution finishes.
+Any transaction that attempts to invoke such an instruction will fail with this
+new error code:
 
-Because different stake accounts are receiving the rewards at different blocks,
-on-chain programs, which depends on the rewards of stakes accounts during the
-reward period, may get into partial epoch reward state. To prevent this from
-happening, loading stake accounts from on-chain programs during reward period
-will be disabled. However, reading the stake account through RPC will still be
-available.
+```
+StakeError {
+   EpochRewardsDistributionActive = 15,
+}
+```
 
+Other users can access the `sol_get_epoch_rewards_status` to determine the
+distribution-phase status from within the SVM.
 
 ## Impact
 
--- a/proposals/0118-partitioned-epoch-reward-distribution.md
+++ b/proposals/0118-partitioned-epoch-reward-distribution.md
@@ -1,13 +1,16 @@
 ---
-simd: '0015'
+simd: '0118'
 title: Partitioned Epoch Rewards Distribution
 authors:
-  - Haoran Yi (Solana Labs)
+  - Haoran Yi (Anza)
+  - Justin Starry (Anza)
+  - Tyera Eulberg (Anza)
 category: Standard
 type: Core
 status: Draft
-created: 2023-03-02
+created: 2024-02-16
 feature: (fill in with feature tracking issues once accepted)
+supersedes: 0015
 ---
 
 ## Summary
--- a/proposals/0118-partitioned-epoch-reward-distribution.md
+++ b/proposals/0118-partitioned-epoch-reward-distribution.md
@@ -238,9 +238,9 @@ epoch rewarding phase.
 The first impact is that stake accounts will see their rewards being credited a
 few blocks later in the epoch than before.
 
-The second impact is that users will not be able to update their stakes during
-the epoch reward phases, and will have to wait until the end of the epoch
-reward period to make any changes.
+The second impact is that users will only be able to credit their stake accounts
+during the epoch reward phase. Any other updates will have to wait until the end
+of the phase.
 
 Nonetheless, the overall amount of time that the user must wait before
 receiving and updating their stake rewards should be roughly equivalent to what
@@ -258,8 +258,12 @@ beta today.
 While the proposed new approach does impact and modify various components of
 the validators, it does not alter the economics of the reward system.
 
-If all the changes are implemented correctly and tested fully, there are no
-security issues.
+Reward distribution relies on restricting any lamport debit or state changes for
+stake accounts until distribution is completed.
+
+The initial reward calculation and reward-distribution progress should be
+recoverable from snapshots produced during the reward distribution period to
+avoid consensus failure.
 
 
 ## Backwards Compatibility
--- a/proposals/0118-partitioned-epoch-reward-distribution.md
+++ b/proposals/0118-partitioned-epoch-reward-distribution.md
@@ -1,10 +1,7 @@
 ---
 simd: '0118'
 title: Partitioned Epoch Rewards Distribution
-authors:
-  - Haoran Yi (Anza)
-  - Justin Starry (Anza)
-  - Tyera Eulberg (Anza)
+authors: Haoran Yi (Anza), Justin Starry (Anza), Tyera Eulberg (Anza)
 category: Standard
 type: Core
 status: Draft
--- a/proposals/0118-partitioned-epoch-reward-distribution.md
+++ b/proposals/0118-partitioned-epoch-reward-distribution.md
@@ -10,7 +10,7 @@ type: Core
 status: Draft
 created: 2024-02-16
 feature: (fill in with feature tracking issues once accepted)
-supersedes: 0015
+supersedes: '0015'
 ---
 
 ## Summary
@@ -170,8 +170,8 @@ The layout of `EpochRewards` sysvar is shown in the following pseudo code.
 
 ```
 struct EpochRewards{
-   // whether rewards distribution is active
-   active: bool, // byte
+   // whether the rewards period (calculation and distribution) is active
+   active: bool, // byte. false: 0x00, true: 0x01
 
    distribution_starting_block_height: u64, // little-endian unsigned 64-bit integer
    
@@ -215,7 +215,7 @@ distribution.
 Therefore, the Stake Program needs access to a syscall which reports whether the
 distribution phase is active. This new syscall `sol_get_epoch_rewards_status`
 should report the value of `EpochRewards::active`. All Stake Program
-instructions that mutate stake data or debit stake balances should be disabled
+instructions that mutate stake data or debit stake balances must be disabled
 when `sol_get_epoch_rewards_status` is true.
 
 Any transaction that attempts to invoke such an instruction will fail with this
--- a/proposals/0118-partitioned-epoch-reward-distribution.md
+++ b/proposals/0118-partitioned-epoch-reward-distribution.md
@@ -147,7 +147,7 @@ include: the parameters needed to recalculate the reward partitions, the total
 rewards to be distributed, and the amount of rewards distributed so far.
 
 The layout of `EpochRewards` sysvar is shown in the following pseudo code.
-Prefixes indidate the offset of each field, and padding bytes are always zero.
+Prefixes indicate the offset of each field, and padding bytes are always zero.
 
 ```
 #[repr(C)] // C representation, 8-byte alignment
--- a/proposals/0118-partitioned-epoch-reward-distribution.md
+++ b/proposals/0118-partitioned-epoch-reward-distribution.md
@@ -94,9 +94,11 @@ distribution during rewarding phases.
 ### Rewards Calculation
 
 The reward calculation will be performed at the first block of the epoch, block
-height `X`. Once the full rewards are calculated, the rewards will be
-partitioned into distribution chunks stored in the bank, which will then be
-distributed during the `reward distribution` phase.
+height `X`. (Some documentation on how rewards are calculated is available
+[here](https://docs.solanalabs.com/consensus/stake-delegation-and-rewards#basics).)
+Once the full rewards are calculated, the rewards will be partitioned into
+distribution chunks stored in the bank, which will then be distributed during
+the `reward distribution` phase.
 
 To ensure that each block distributes a subset of the rewards in a
 deterministic manner for the current epoch, while also randomizing the
--- a/proposals/0118-partitioned-epoch-reward-distribution.md
+++ b/proposals/0118-partitioned-epoch-reward-distribution.md
@@ -208,15 +208,18 @@ current block height can be discarded.
 
 ### Restrict Stake Account Mutation
 
-The stake accounts need to be preserved in credit-only state while distributions
-are in progress in order to prevent any changes interfering with reward
-distribution.
+In order to ensure partition recalculations will be the same across the
+distribution period, the number and addresses of stake accounts, as well as
+stake-delegation amounts, must remain unchanged. There are existing use-cases
+for crediting stake accounts at epoch boundaries; so as to not disrupt these
+credits, but otherwise take the most conservative approach, account credits will
+be the only stake-account changes permitted during distribution.
 
-Therefore, the Stake Program needs access to a syscall which reports whether the
-distribution phase is active. This new syscall `sol_get_epoch_rewards_sysvar`
-should return the values of the `EpochRewards` sysvar. All Stake Program
-instructions that mutate stake data or debit stake balances must be disabled
-when `EpochRewards::active` is true.
+To limit permissible actions, the Stake Program needs access to a syscall which
+reports whether the distribution phase is active. This new syscall
+`sol_get_epoch_rewards_sysvar` should return the values of the `EpochRewards`
+sysvar. All Stake Program instructions that mutate stake data or debit stake
+balances must be disabled when `EpochRewards::active` is true.
 
 Any transaction that attempts to invoke such an instruction will fail with this
 new error code:
--- a/proposals/0118-partitioned-epoch-reward-distribution.md
+++ b/proposals/0118-partitioned-epoch-reward-distribution.md
@@ -152,54 +152,59 @@ be computed by applying the following formula to the hash:
 I = (M * stake_address_hash) / 2^64
 ```
 
-### `EpochRewards` Sysvar Account
+### Reward Distribution
 
-`EpochRewards` sysvar account records the total rewards and the amount of
-distributed rewards for the current epoch internally. And the account balance
-reflects the amount of pending rewards to distribute.
+The reward distribution phase happens after the reward computation phase, ie.
+during the second block in the epoch as per the current design. It lasts for `M`
+blocks (see previous section).
+
+#### `EpochRewards` Sysvar Account
+
+The `EpochRewards` sysvar account records whether the rewards distribution phase
+is in progress, as well as the details needed to resume distribution when
+starting from a snapshot during the reward distribution phase. These details
+include: the parameters needed to recalculate the reward partitions, the total
+rewards to be distributed, and the amount of rewards distributed so far.
 
 The layout of `EpochRewards` sysvar is shown in the following pseudo code.
 
 ```
-struct RewardRewards{
+struct EpochRewards{
+   // whether rewards distribution is active
+   active: bool, // byte
+
+   distribution_starting_block_height: u64, // little-endian unsigned 64-bit integer
+   
+   num_partitions: u64, // little-endian unsigned 64-bit integer
+   
+   parent_blockhash: Hash, // byte-array of length 32
+
    // total rewards for the current epoch, in lamports
-   total_rewards: u64,
+   total_rewards: u64, // little-endian unsigned 64-bit integer
 
-   // distributed rewards for  the current epoch, in lamports
-   distributed_rewards: u64,
-
-   // distribution of all staking rewards for the current
-   // epoch will be completed before this block height
-   distribution_complete_block_height: u64,
+   // distributed rewards for the current epoch, in lamports
+   distributed_rewards: u64, // little-endian unsigned 64-bit integer
 }
 ```
 
-The `EpochRewards` sysvar is created at the start of the first block of the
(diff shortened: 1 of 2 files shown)


CriesofCarrots:
@t-nelson (can't request as reviewer)

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -152,54 +152,59 @@ be computed by applying the following formula to the hash:
...
+rewards to be distributed, and the amount of rewards distributed so far.
 
 The layout of `EpochRewards` sysvar is shown in the following pseudo code.
 
 ```
-struct RewardRewards{
-   // total rewards for the current epoch, in lamports
-   total_rewards: u64,
+struct EpochRewards{
+   // whether rewards distribution is active
```
jstarry, line 173: Is it also true during rewards calculation?

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -206,24 +206,29 @@ partition indicated by comparing its current block height to
...
-the stake program is disabled during the epoch reward distribution period.
+The stake accounts need to be preserved in credit-only state while distributions
+are in progress in order to prevent any changes interfering with reward
+distribution.
 
-Any transaction that invokes staking program during this period will fail with
-a new unique error code - `StakeProgramUnavailable`.
+Therefore, the Stake Program needs access to a syscall which reports whether the
+distribution phase is active. This new syscall `sol_get_epoch_rewards_status`
+should report the value of `EpochRewards::active`. All Stake Program
```
jstarry, line 217: Rather than creating a new syscall, should we use the `sol_get_epoch_rewards_sysvar` syscall and check `active` there?

jstarry (commented):
Looks really good! Thanks for writing this up and consolidating everything.

proposals/0118-partitioned-epoch-reward-distribution.md
CriesofCarrots: Technically (at least in the current implementation), rewards calculation completes before the sysvar is populated, although both steps happen in the same block and both complete before any transactions are processed.

I think the existing comment is correct, but I am happy to make more explicit if there's additional information that would be helpful here.

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,274 @@
...
+
+### Reward Distribution
+
+The reward distribution phase happens after the reward computation phase, ie.
+during the second block in the epoch as per the current design. It lasts for `M`
+blocks (see previous section).
+
+#### `EpochRewards` Sysvar Account
+
+The `EpochRewards` sysvar account records whether the rewards distribution phase
```
HaoranYi, line 163: What's the balance on the sysvar account?

CriesofCarrots: In this SIMD, the balance on the sysvar account is the rent-exempt minimum for the amount of data in the account (which I believe is 0.00134328 SOL).

We should definitely have the conversation about whether to track distributions with the SOL balance of this sysvar here.

My personal opinion is that the existing implementation is a little scary when the balance gets low (below rent-exempt minimum or even below zero). Also, as of this SIMD, the sysvar needs to persist, so its balance can no longer be withdrawn to 0. We could fund the account `sum_distributions` on top of the rent-exempt minimum, but I wrote it up without that because I don't think it gains us much over tracking the amount distributed in the sysvar data.

proposals/0118-partitioned-epoch-reward-distribution.md
jstarry: Ok makes sense. So is it set to active before the first block of an epoch is processed?

CriesofCarrots: > So is it set to active before the first block of an epoch is processed?

Yep, correct

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,274 @@
...
+The [SipHash 1-3](https://www.aumasson.jp/siphash/siphash.pdf) pseudo-random
+function is used to hash stake account addresses efficiently and uniformly
+across the blocks in the reward distribution phase. The hashing function for an
+epoch is created by seeding a new `SipHasher` with the parent block's blockhash.
+This hashing function can then be used to hash each active stake account's
+address into a u64 hash value. The reward distribution block index `I` can then
+be computed by applying the following formula to the hash:
+
+```
+I = (M * stake_address_hash) / 2^64
```
t-nelson, line 152:
```suggestion
I = (M * stake_address_hash) / (2^64 - 1)
```

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,274 @@
...
+starting from a snapshot during the reward distribution phase. These details
+include: the parameters needed to recalculate the reward partitions, the total
+rewards to be distributed, and the amount of rewards distributed so far.
+
+The layout of `EpochRewards` sysvar is shown in the following pseudo code.
+
+```
+struct EpochRewards{
+   // whether rewards distribution is active
+   active: bool, // byte
```
t-nelson, line 174:
```suggestion
   active: bool, // byte. false: 0x00, true: 0x01
```

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,274 @@
...
+### Restrict Stake Account Mutation
+
+The stake accounts need to be preserved in credit-only state while distributions
+are in progress in order to prevent any changes interfering with reward
+distribution.
+
+Therefore, the Stake Program needs access to a syscall which reports whether the
+distribution phase is active. This new syscall `sol_get_epoch_rewards_status`
+should report the value of `EpochRewards::active`. All Stake Program
+instructions that mutate stake data or debit stake balances should be disabled
```
t-nelson, line 218:
```suggestion
instructions that mutate stake data or debit stake balances must be disabled
```

t-nelson (commented):
left a couple clarifying suggestions. the changes lgtm otherwise. thanks for committing the copy of the original simd with the proposed modifications, it made review very pleasant!

i did notice some other, non-technical changes that might make the document more clear, but probably couldn't justify their own simd. do we want to entertain those here?

proposals/0118-partitioned-epoch-reward-distribution.md
CriesofCarrots: Stet? https://github.com/solana-labs/solana/blob/f122e99c4e737b7cc1c2a697cf81e56eb39ccec7/sdk/src/epoch_rewards_hasher.rs#L42

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,274 @@
+---
```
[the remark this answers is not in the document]
CriesofCarrots: Hmm, that authors list is particularly not good. I'll see if I can force it to be a list within the table display

CriesofCarrots (commented):
> i did notice some other, non-technical changes that might make the document more clear, but probably couldn't justify their own simd. do we want to entertain those here?

I say yes. Let's get this as complete and useful as possible. I will keep things separated by commit.

proposals/0118-partitioned-epoch-reward-distribution.md
t-nelson: ah right. b- math guy here. i have to redo the calculation scratch every time :sweat_smile:

proposals/0118-partitioned-epoch-reward-distribution.md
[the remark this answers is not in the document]
t-nelson: yeah and the zero-prefixed numbers being interpreted as octal is... surprising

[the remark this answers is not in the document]
CriesofCarrots: > yeah and the zero-prefixed numbers being interpreted as octal is... surprising

Ah, that one I can fix by using single quotes. But agreed.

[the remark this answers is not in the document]
t-nelson: what does this look like?
```
accounts: Haoran Yi (Anza)<br>Justin Starry (Anza)<br>Tyera Eulberg (Anza)
```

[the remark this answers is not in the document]
t-nelson: :sob: 

how about
```
accounts: >-
Haoran Yi (Anza)
Justin Starry (Anza)
Tyera Eulberg (Anza)
```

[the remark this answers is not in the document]
jacobcreech: Looks like EIP just uses a [comma separated list](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md). I'd be fine with that since it has a precedent. I'll update the other multi-author SIMDs later if we go with that.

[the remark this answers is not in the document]
CriesofCarrots: > how about

It looks like this 😂 
Screenshot 2024-02-20 at 10 47 55 PM

I'll push comma-separated list

proposals/0118-partitioned-epoch-reward-distribution.md
jstarry: Great and maybe worth one more small clarification that if the last partition of rewards is distributed at the beginning of slot X, it will longer be active in slot X when processing transactions. Just want to avoid an off-by-one issue with other implementations.

proposals/0118-partitioned-epoch-reward-distribution.md
jstarry: I don't have a strong preference either way. I'm leaning towards keeping the full sysvar in the syscall just for consistency with other sysvars. And it's easy enough to add a function for `get_epoch_rewards_status` which wraps that syscall to pull out the `active` field.

proposals/0118-partitioned-epoch-reward-distribution.md
jstarry: I also agree that it's tricky to manage the sysvar lamport balance properly with all the edge cases. I like this new proposal of just tracking it in the data.

HaoranYi: Ok. Sounds good to me. This will also make sysvar cache management simpler too!

One thing to note is that the total capital will *not* stay stable during the epoch. Should any user programs or economics factors that depend on this assumption, they will need to be adjusted.

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -152,54 +152,59 @@ be computed by applying the following formula to the hash:
...
+include: the parameters needed to recalculate the reward partitions, the total
+rewards to be distributed, and the amount of rewards distributed so far.
 
 The layout of `EpochRewards` sysvar is shown in the following pseudo code.
 
 ```
-struct RewardRewards{
-   // total rewards for the current epoch, in lamports
-   total_rewards: u64,
+struct EpochRewards{
```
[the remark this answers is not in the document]
CriesofCarrots: Sure, wouldn't hurt.

proposals/0118-partitioned-epoch-reward-distribution.md
CriesofCarrots: Added some clarification. Let me know if it reads okay.

proposals/0118-partitioned-epoch-reward-distribution.md
CriesofCarrots: Yep, makes sense. I pushed a commit describing the `sol_get_epoch_rewards_sysvar` syscall instead.

jstarry (approved):

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,274 @@
...
+
+However, there are quite a few promising optimizations that can cut down the
+reward computation time. An experiment for reward calculation optimization
+(https://github.com/solana-labs/solana/pull/31193) showed that we can cut the
+reward calculation time plus vote reward distribution time to around 1s. This
+makes synchronous reward computation and asynchronous reward distribution a
+feasible approach. We also believe that there is room for more optimization to
+further cut down the above timing.
+
+Therefore, the following design is based on the above optimization. The reward
```
t-nelson, line 115:
```suggestion
The reward
```

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,274 @@
...
+calculation will be performed at the first block of the epoch, block height `X`.
```
t-nelson, line 116: throughout we assume that the rewards distribution algorithm is known by the implementer. i'm not sure this is a strong assumption

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,274 @@
...
+formula to round up to the nearest integer without using floating point value
+arithmetic:
+
+```
+M = ((4096 - 1)+num_stake_accounts)/4096
+```
+
+To safeguard against the number of stake accounts growing dramatically and
+overflowing the number of blocks in an epoch, the number of blocks is capped at
+10% of the number of block in an epoch (currently 432,000).
```
t-nelson, line 138:
```suggestion
10% of the total slots in an epoch.
```

t-nelson, line 138: we make no provision for how this overflow case is actually handled. iirc the original proposal was to "let the last partition take a long time," but the partitioning algorithm doesn't accommodate.

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,274 @@
...
+When booting from a snapshot, a node must check the EpochRewards sysvar account
+to determine whether the distribution phase is active. If so, the node must
+rerun the rewards calculation using the `EpochRewards::num_partitions` and
+`EpochRewards::parent_blockhash` sysvar fields as well as the `EpochStakes` in
+the snapshot. Then the runtime must resume rewards distribution from the
+partition indicated by comparing its current block height to
+`EpochRewards::distribution_starting_block_height`. This can be confirmed by
+comparing a partial sum of the rewards calculation (those partitions expected to
+have been distributed) with the `EpochRewards::distributed_rewards` field.
+
```
t-nelson, line 207:
```suggestion
Partitions for blocks prior to the current block height can be discarded.

```

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,274 @@
...
+epoch rewarding phase.
+
+The first impact is that stake accounts will see their rewards being credited a
+few blocks later in the epoch than before.
+
+The second impact is that users will only be able to credit their stake accounts
+during the epoch reward phase. Any other updates will have to wait until the end
+of the phase.
+
+Nonetheless, the overall amount of time that the user must wait before
```
t-nelson, line 245:
```suggestion
Nonetheless, the overall amount of wall time that the user must wait before
```

t-nelson (commented):
i left quite a bit of eye twitch intact for the sake of brevity.  these suggestions knock out most of the stumbling i came across that actually brought confusion

proposals/0118-partitioned-epoch-reward-distribution.md
CriesofCarrots: I assume you mean the rewards *calculation* algorithm?
I don't want to fully define that here, but if there's a good doc to link to, that would be good.

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,274 @@
...
+   per-block reward reserve sysvar accounts. Those sysvar accounts are
+   introduced to track and verify the rewards distributed per block. The
+   per-block reward reserve sysvar accounts add additional checks and safety
+   for reward distribution. However, they also add addition cost to block
+   processing, especially for the first block in the epoch. The first block is
+   already computationally heavily - it is responsible for processing all the
+   reward computation. The additional cost of those sysvars puts more burden
+   onto that block and hurt the timing for it.
+
+## Detailed Design
```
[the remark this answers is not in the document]
CriesofCarrots: I added a couple of my own edits to your suggestions, in the new commit. See what you think.

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,255 @@
...
+address into a u64 hash value. The reward distribution block index `I` can then
+be computed by applying the following formula to the hash:
+
+```
+I = (M * stake_address_hash) / 2^64
+```
+
+### Reward Distribution
+
+The reward distribution phase happens after the reward computation phase, ie.
```
HaoranYi, line 134: ie. => i.e.

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,255 @@
...
+the reward distribution phase. The `active` field is set to false after the last
+partition is distributed (and `total_rewards == distributed_rewards`), ie.
+before processing transactions in block at height `X + M`.
+
+#### Booting from Snapshot
+
+When booting from a snapshot, a node must check the EpochRewards sysvar account
+to determine whether the distribution phase is active. If so, the node must
+rerun the rewards calculation using the `EpochRewards::num_partitions` and
+`EpochRewards::parent_blockhash` sysvar fields as well as the `EpochStakes` in
```
HaoranYi, line 181: `EpochStakes` doesn't store the full stake account state. It only stores the stake delegation info, which is not enough to recompute the rewards. 
To recompute stake rewards, we have to store the full stake account state, which is as expensive as storing the rewards.
Therefore, I think we should keep the current implementation of storing the computed rewards in the snapshot.

proposals/0118-partitioned-epoch-reward-distribution.md
CriesofCarrots: Given that we have no agreed upon style guide, I'm going to stick with my stylistic preference on this.

proposals/0118-partitioned-epoch-reward-distribution.md
HaoranYi: > we are guaranteeing that no state will have changed since the first block in the epoch

No, this is not true. When the stake account received rewards, its state changes.

CriesofCarrots: Well, correct. Sorry, I left that part implied. Explain how that is an issue, please.

HaoranYi: For example, the `observed_credit` field will change after receiving rewards. 
Reward computation will depend on that field. If we want to recompute rewards later, we need to know the `observed_credit` for all stake accounts at the epoch boundary.

HaoranYi: That field is inside stake `state`. but not saved in `EpochStake`. When we load the stake account later, we can't recover the old state.

CriesofCarrots: > that will add extra complexity and make startup from snapshot slower.

The motivation of this amendment is that a little extra time booting from snapshot is preferable to adding 100s of MBs to the snapshot.
I would expect the performance impact of that subtraction to be nominal, but if it is more than that, we definitely don't need to add each entire StakeState to the snapshot to preserve just `credits_observed`.

HaoranYi: There is also code complexity and risk aspect to consider. There are a lot of if-else and corner case handling in reward calculation code. If not careful, we could easily introduce a bug. And a bug there will be very bad. 

In my opinion, the extra space in snapshot doesn't sound too bad. 
1. We only pay the cost when the snapshot is taken during reward period. For snaphot take out of reward period, we don't need to pay the cost.
2. I think the reward data should have many similar data in the fields, which will make it very compressible. After compress, it woun't take 100s of MBs.
3. It makes startup from snapshot faster.

HaoranYi: I have simulated a bank snapshot with 1M stake rewards, randomly delegated to 2K vote accounts. 

Total space for uncompressed bank snapshot took 296M, compressed space 65M. 

```
-rw-r--r-- 1 sol users 296M Feb 24 00:38 foo.snap
-rw-r--r-- 1 sol users  65M Feb 24 00:38 foo.snap.lz4
```

t-nelson: it's only partially about the space. adding snapshot fields must be adopted by all implementation. the fd team was lamenting about this at the offsite and we committed to avoid adding them whenever possible, instead preferring to store data in accounts. storing _this_ data in an account is a nonstarter though.

i don't think we even need to unwind the credits, but we might need to store a couple more fields in the sysvar. we already have the total and distributed lamports. if we also add the total and distributed credits, we can can trivally recover the distributions for the remaining partitions.

HaoranYi: I can argue that there is no free lunch. If you don't want to deal with the complexity of handling the new field, you will have to introduce the complexity of a new code path to recompute rewards. And you will have to maintain both reward code paths in future.  Besides, the field is already in v1.17. If fd, use v.1.17 and above as the reference impl, it would already have this field.

HaoranYi:
Handling the new field is much easier than recompute. And it is a one time cost. After it is implemented, you don't need to worry about it in future. However, adding another code path introduce maintaining cost in future too.

t-nelson:
the snapshot shortcut is only "easier" in the short term. it is punting the problem down the road to snapshot encoding, (de)ser, storage and distribution. success requires long term thinking, not taking the easy way out.

proposals/0118-partitioned-epoch-reward-distribution.md
CriesofCarrots: Ofc we can ask, but I think it is highly unlikely that firedancer supports partitioned epoch rewards in any fashion yet. The fact that something is already implemented does not make it the right design, especially since the existing approach was not described (approved) in the original simd, nor was that simd signed off on by the firedancer team.

HaoranYi: Heh. What's the policy for simd approval, consensus or majority? Anyway, if I need to vote, my vote is against recomputing. But if the majority thinks that recomptuing is the way to go, I am fine to accept it too. [(*^ ‿ ⁢*](https://www.emoticonstext.com/wink.html)

CriesofCarrots:
> I think there was a concern raised by Anatoly that the total capital did not stay stable during the epoch.

Total capital is not stable during an epoch now, as `run_incinerator` is triggered on every Bank freeze and decreases the capitalization by the amount in that account.

HaoranYi:
How much balance does incinerator account get per epoch? And where does it get the blance?

CriesofCarrots:
> How much balance does incinerator account get per epoch? And where does it get the blance?

I don't know how much it receives; the runtime isn't logging that data. I suppose we could write a geyser plugin that would report that (or an indexer might know already). Anyone can transfer lamports to the incinerator to burn them.

t-nelson:
i'm not sure how total capitalization nor incinerator are relevant here? we just need to be able to recover remaining partition distributions from minimal state when loading snapshots. if we have total&distributed epoch reward lamports and total&distributed epoch credits, Delegations are in accounts (or snapshotted stakes cache? :melting_face:). what else do we need do this practically?

HaoranYi:
One thing people may start noticing is that the `Sol` Supply on solana explorer will starting slowly increase at the epoch boundary. 

image

It is more noticeable than a onetime increase at the epoch boundary. Probably, people would want an explanation for that. A slow and gradual increase in the total supply of sol **per block** may make people worry about inflation...

HaoranYi:
> Total capital is not stable during an epoch now, as run_incinerator is triggered on every Bank freeze and decreases the capitalization by the amount in that account.

The capital change due to incinerator is very minimal. It doesn't make material impact on the total sol. While reward change are much more noticeable than incinerator burning.

HaoranYi:
> Total capital is not stable during an epoch now, as run_incinerator is triggered on every Bank freeze and decreases the capitalization by the amount in that account.

The capital change due to incinerator is very minimal. It doesn't make material impact on the total sol. While reward change are much more noticible than 

> i'm not sure how total capitalization nor incinerator are relevant here? we just need to be able to recover remaining partition distributions from minimal state when loading snapshots. if we have total&distributed epoch reward lamports and total&distributed epoch credits, Delegations are in accounts (or snapshotted stakes cache? :melting_face:). what else do we need do this practically?

That's one of the main changes for this SIMD. If I read the SIMD correctly, there are two major chagnes:
1. sysvar no longer carries the total reward balance. 
2. recompute rewards at restart.

CriesofCarrots:
> The capital change due to incinerator is very minimal. It doesn't make material impact on the total sol. While reward change are much more noticeable than incinerator burning.

Because that is user-dependent, we absolutely cannot rely on that assumption. However, it sounds like epoch-capitalization stability is not a technical concern in the first place; just a comms issue. Since there will need to be communication about things like stake withdrawals being unavailable during the rewards period anyway, we can also explain the new shape of supply increases.

proposals/0118-partitioned-epoch-reward-distribution.md
CriesofCarrots: What's written here does describe the current implementation. "Overflow" is actually handled by letting the number of accounts in a partition exceed 4096 by clamping the number of partitions to `slots_per_epoch/10`:
https://github.com/solana-labs/solana/blob/8ad125d0c0688aaf2b62bb95b535ff988ed7f9ac/runtime/src/bank.rs#L1162-L1165

So each slot will go a little slower, rather than distribution extending for more slots. This is probably okay for now, but we probably ought to consider making `MAX_FACTOR_OF_REWARD_BLOCKS_IN_EPOCH` tunable at some point.

proposals/0118-partitioned-epoch-reward-distribution.md
CriesofCarrots: Added this link

t-nelson (dismissed):
r+ one nit.

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,265 @@
+---
+simd: '0118'
+title: Partitioned Epoch Rewards Distribution
+authors: Haoran Yi (Anza), Justin Starry (Anza), Tyera Eulberg (Anza)
+category: Standard
+type: Core
+status: Draft
+created: 2024-02-16
+feature: (fill in with feature tracking issues once accepted)
+supersedes: '0015'
```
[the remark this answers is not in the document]
CriesofCarrots: Certainly something. I like `superceded-by`... will see if that's acceptable to Jacob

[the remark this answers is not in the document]
CriesofCarrots: I also update the simd-0015 header to status: Accepted, because that's technically what it was before this superceding simd. I'm not totally clear on what the status-change process should look like, though (cc @jacobcreech )

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,265 @@
...
+
+#### `EpochRewards` Sysvar Account
+
+The `EpochRewards` sysvar account records whether the rewards distribution phase
+is in progress, as well as the details needed to resume distribution when
+starting from a snapshot during the reward distribution phase. These details
+include: the parameters needed to recalculate the reward partitions, the total
+rewards to be distributed, and the amount of rewards distributed so far.
+
+The layout of `EpochRewards` sysvar is shown in the following pseudo code.
```
riptl, line 149: The alignment is somewhat unspecified here. In repr(C), u128 either has 8 or 16 byte alignment depending on architecture. AFAIK bool has 8 byte alignment 🤔

I suggest putting offsets like so `/* 0x08 */` as prefixes to make it unambiguous. 

The padding would obviously be zeros, but it wouldn't hurt to mention here.

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,265 @@
...
+new error code:
+
+```
+StakeError {
+   EpochRewardsDistributionActive = 15,
+}
+```
+
+Other users can access the `active` field from the
+`sol_get_epoch_rewards_sysvar` syscall to determine the distribution-phase
```
riptl, line 221: Is this syscall going to be specified by another SIMD?

riptl (approved):
Looks great. I have two nits. (Sorry am on vacation so can't use my work account @ripatel-fd)

proposals/0118-partitioned-epoch-reward-distribution.md
t-nelson: we probably need a generic "sysvar getter syscall" simd to cover stuff like this

proposals/0118-partitioned-epoch-reward-distribution.md
CriesofCarrots: We do specify 8-byte alignment in the comment on L152, but certainly doesn't hurt to add offsets, and I'll remove the "ie".
Hopefully that u128 won't be too much of a PITA. If I understand [this issue](https://github.com/rust-lang/rust/issues/54341) correctly, rust will always align u128 to 8 bytes even with repr(C), but C programs may not.

proposals/0118-partitioned-epoch-reward-distribution.md
CriesofCarrots: @ripatel-fd , what do you think of Trent's idea here? There are a lot of these sysvar-getter syscalls already, and it would be nice not to have to define new ones beyond the sysvar layout in SIMDs like this one.

proposals/0118-partitioned-epoch-reward-distribution.md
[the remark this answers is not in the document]
jacobcreech: That sounds correct. All of these SIMDs were technically "Accepted" and I'll need to update them. Let me add `superceded-by` so that the CI will play nice. This also helps a lot when we eventually get a website to display SIMDs.

godmodegalactus (approved):
Overall looks a very good direction

proposals/0118-partitioned-epoch-reward-distribution.md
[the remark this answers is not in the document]
jacobcreech: Added in https://github.com/solana-foundation/solana-improvement-documents/pull/126

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,273 @@
...
+#### `EpochRewards` Sysvar Account
+
+The `EpochRewards` sysvar account records whether the rewards distribution phase
+is in progress, as well as the details needed to resume distribution when
+starting from a snapshot during the reward distribution phase. These details
+include: the parameters needed to recalculate the reward partitions, the total
+rewards to be distributed, and the amount of rewards distributed so far.
+
+The layout of `EpochRewards` sysvar is shown in the following pseudo code.
+Prefixes indidate the offset of each field, and padding bytes are always zero.
```
lheeger-jump, line 150:
```suggestion
Prefixes indicate the offset of each field, and padding bytes are always zero.
```

lheeger-jump, line 150:
```suggestion
Prefixes indicate the offset of each field, and padding bytes are always zero.
```

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,273 @@
...
+
+```
+#[repr(C)] // C representation, 8-byte alignment
+struct EpochRewards{
+   // whether the rewards period (calculation and distribution) is active
+   // byte. false: 0x00, true: 0x01
+   /* 0x00 */ active: bool,
```
lheeger-jump, line 157: What is the size of `bool`? Can you just be explicit about the size of the type: u8, u16, u32, or u64 please

lheeger-jump: and endianness

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,273 @@
...
+   // little-endian unsigned 64-bit integer
+   /* 0x48 */ total_rewards: u64,
+
+   // distributed rewards for the current epoch, in lamports
+   // little-endian unsigned 64-bit integer
+   /* 0x50 */ distributed_rewards: u64, 
+}
+```
+
+The `EpochRewards` sysvar is repopulated at the start of the first block of the
```
lheeger-jump, line 183: We need to add detail on when the sysvar is updated during the calculation phase and the distribution phase. That is to say, the reward calculation and sysvar update therein should probably be updated before the transactions in the block (or after, i dont think it matters, actually after is better). The rewards distribution phase and sysvar update should be done _after_ the transactions like rent collection.

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,273 @@
...
+`EpochRewards::total_rewards` sysvar fields, as well as the `EpochStakes` in the
+snapshot. The recalculated rewards can be confirmed by comparing a sum of the
+rewards remaining (those partitions expected to not yet have been distributed)
+with the difference between the `EpochRewards::total_rewards` and
+`EpochRewards::distributed_rewards` fields. Partitions for blocks prior to the
+current block height can be discarded.
+
+### Restrict Stake Account Mutation
+
+The stake accounts need to be preserved in credit-only state while distributions
```
lheeger-jump, line 209: I do not understand the issue here. If the account rewards are credited at the end of the block, then all edits can proceed as normal in txns. Can you elaborate?

lheeger-jump (changes requested):

proposals/0118-partitioned-epoch-reward-distribution.md
CriesofCarrots: In this design, if a node boots from snapshot in the middle of the rewards period, it needs to recalculate the reward partitions. Therefore, Delegations need to remain unchanged until all partitions are distributed, as do the number and addresses of stake accounts.

The original implementation of SIMD-0015 prevented any use of the stake program at all during partitioned distribution. However, given the existing use-cases for depositing into stake accounts (MEV), we wanted to relax this restriction. Credit-only is a conservative approach to doing so.

Certainly merges and splits must be prevented during distribution, and we would have to be very careful about allowing delegate or deactivate. But other operations could be permitted.

That said, I lean toward establishing this SIMD with the conservative credit-only approach and amending the SIMD later as we feel confident about the safety of other operations. Wdyt?

proposals/0118-partitioned-epoch-reward-distribution.md
t-nelson: move `active` to the end?

CriesofCarrots: Ah, thanks for this, @ripatel-fd . Seems to contradict some of what I read about this.
But anyway, I'll do some reworking and update this on Monday.

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,275 @@
...
+distribution in a first block of a new epoch.
+
+## Motivation
+
+The distribution of epoch rewards at the start block of an epoch becomes a
+significant bottleneck due to the rising number of stake accounts and voting
+nodes on the network.
+
+To address this bottleneck, we propose a new approach for distributing the
+epoch rewards over multiple blocks.
```
topointon-jump, line 26: Just curious - have we measured that it is the distribution (write-back) versus the calculation that is the bottleneck?

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,275 @@
...
+`EpochRewards::parent_blockhash` sysvar fields and determining the upcoming
+partitions by comparing its current block height to
+`EpochRewards::distribution_starting_block_height`. Then the runtime must
+recalculate the remaining rewards using the `EpochRewards::total_points` and
+`EpochRewards::total_rewards` sysvar fields, as well as the `EpochStakes` in the
+snapshot. The recalculated rewards can be confirmed by comparing a sum of the
+rewards remaining (those partitions expected to not yet have been distributed)
+with the difference between the `EpochRewards::total_rewards` and
+`EpochRewards::distributed_rewards` fields. Partitions for blocks prior to the
+current block height can be discarded.
```
topointon-jump, line 207: The rewards calculation assumes that the calculation is done at the epoch boundary (for example, using `VoteState::epoch_credits`). We need to make sure that re-calculating rewards when booting off a snapshot doesn't break the calculation, as this assumption is no longer true.

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,275 @@
...
+The `EpochRewards` sysvar is repopulated at the start of the first block of the
+epoch (before any transactions are processed), as both the total epoch rewards
+and vote account rewards become available at this time. The
+`distributed_rewards` field is updated per reward distribution for each block in
+the reward distribution phase (before transaction processing). The `active`
+field is set to false after the last partition is distributed (and
+`total_rewards == distributed_rewards`), ie. before processing transactions in
+block at height `X + M`.
+
+#### Booting from Snapshot
```
topointon-jump, line 193: Could we be a bit more verbose about what logic happens in this case? In an ideal world client teams should be able to use SIMDs, without looking at code, to implement new features. Thank-you! 🙏

proposals/0118-partitioned-epoch-reward-distribution.md
lheeger-jump: Okay before is fine. I think I could have been clearer that it is preferable to keep it consistent.

proposals/0118-partitioned-epoch-reward-distribution.md
lheeger-jump: That is fine for now. This text should make it into the SIMD

proposals/0118-partitioned-epoch-reward-distribution.md
topointon-jump: In particular, I think `solana_stake_program::stake_state::calculate_stake_points_and_credits` will have to be tweaked slightly (and possibly some other places), as will anywhere which uses `VoteState::epoch_credits`.

proposals/0118-partitioned-epoch-reward-distribution.md
CriesofCarrots: Pushed

proposals/0118-partitioned-epoch-reward-distribution.md
CriesofCarrots: Updated to 16-byte alignment, including moving `active` to the end

proposals/0118-partitioned-epoch-reward-distribution.md
CriesofCarrots: Yes, definitely, although it doesn't seem like we have that data summarized concisely in any one place. It seems to be mostly in various places in the #proj-epoch-boundary-optimization channel on discord. For instance, here's a comment about the calculation time: https://discord.com/channels/428295358100013066/960593861732884520/1096146390037561414 (64ms for 550K stake accounts)
Whereas distribution is more like 5-10s.

proposals/0118-partitioned-epoch-reward-distribution.md
topointon-jump: No, not in the SIMD, as I don't think this will be fully known until it's implemented. Just wanted to flag 😄

proposals/0118-partitioned-epoch-reward-distribution.md
topointon-jump: As discussed on Slack, lets update the SIMD with more precise implementation details after a POC has been implemented 😄

proposals/0118-partitioned-epoch-reward-distribution.md
lheeger-jump: That looks good

lheeger-jump (approved):
Approved

proposals/0118-partitioned-epoch-reward-distribution.md
```diff
@@ -0,0 +1,278 @@
...
+stake-delegation amounts, must remain unchanged. There are existing use-cases
+for crediting stake accounts at epoch boundaries; so as to not disrupt these
+credits, but otherwise take the most conservative approach, account credits will
+be the only stake-account changes permitted during distribution.
+
+To limit permissible actions, the Stake Program needs access to a syscall which
+reports whether the distribution phase is active. This new syscall
+`sol_get_epoch_rewards_sysvar` should return the values of the `EpochRewards`
+sysvar. All Stake Program instructions that mutate stake data or debit stake
+balances must be disabled when `EpochRewards::active` is true.
```
t-nelson, line 222: nit
```suggestion
account `lamports` balances must be disabled when `EpochRewards::active` is true.
```

t-nelson (approved):

t-nelson (approved):
:shipit:

billythedummy:
To clarify, with this change, would the various stake pool programs (spl, marinade) need to upgrade to have their epoch update crank instruction fail if `EpochRewards.active` is true?

CriesofCarrots:
> To clarify, with this change, would the various stake pool programs (spl, marinade) need to upgrade to have their epoch update crank instruction fail if `EpochRewards.active` is true?

@billythedummy , I'm not personally familiar with the various stake-pool programs or the instruction you reference, but if they have operations that depend on rewards distribution being complete, then the answer is yes.
With this change, the Stake Program will be updated so that only credit-only stake-account operations will succeed when `EpochRewards.active`, so stake-pool operations that depend on mutating stake accounts will fail during the rewards period.

billythedummy:
@CriesofCarrots each stake pool program basically have crank instructions that on every epoch reads the new balances of the stake accounts it owns, which are supposed to have increased at the epoch boundary due to staking rewards, and updates the exchange rate between SOL and the pool’s LST. These instructions dont mutate stake accounts in all cases, so i think we would need the `EpochRewards.active` check on the code paths that only read the balances of the stake accounts.

Tagging stake pool programs maintainers here for visibility @joncinque @ochaloup

proposals/0118-partitioned-epoch-reward-distribution.md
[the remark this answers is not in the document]
LucasSte: Why is the 16-byte alignment necessary? Memcpy with 16-byte aligned values are not supported in our SBF target. This means, we have compiler error when trying to use EpochRewards.

PR #127, opened 2024-03-15
SIMD-0127: Syscall: Get-Sysvar

ripatel-fd:
This proposal outlines a new syscall interface for sysvars; one which can
retrieve dynamic ranges of sysvar data without fully copying the entire data
structure.

With a unified syscall interface to handle retrieval of sysvar data, many
existing syscalls will become obsolete, and the act of changing sysvar data
layouts or adding new sysvars will involve much less overhead.

Sysvars are special accounts that are implicitly modified by the runtime.
Certain sysvars are essential for operation of the network, and many others are
used by on-chain programs for a wide range of use cases.

The sysvar API should be completely consistent, available to all BPF programs,
and capable of more easily handling changes in the future. The syscall API for
sysvars should be designed to serve a wide range of requests on sysvar data
without bloating the overall syscall interface.

### DIFF
--- a/proposals/0127-get-sysvar-syscall.md
+++ b/proposals/0127-get-sysvar-syscall.md
@@ -145,13 +145,37 @@ uint64_t sol_get_sysvar(
 
 ```
 
-If the following conditions are not met, this syscall will return an error:
+### Control Flow
 
-- The sysvar data is available to read.
-- The sysvar data is not corrupt.
-- `offset` + `length` is in the range `[0, 2^64)`.
-- `var_addr` + `length` is in the range `[0, 2^64)`.
-- Each byte in the range starting at `var_addr` of size `length` is writable.
+The syscall aborts the virtual machine if any of these conditions are true:
+
+- Not all bytes in VM memory range `[sysvar_id, sysvar_id + 32)` are readable.
+- Not all bytes in VM memory range `[var_addr, var_addr + length)` are writable.
+- `offset + length` is not in `[0, 2^64)`.
+- `var_addr + length` is not in `[0, 2^64)`.
+- Compute budget is exceeded.
+
+The syscall returns `1` if the sysvar is not present in the Sysvar Cache.
+
+The syscall returns `2` if `offset + length` is greater than the size of the
+sysvar.
+
+Otherwise, the syscall writes the sysvar data to the virtual memory at
+`var_addr` and returns `0`.
+
+### Compute Unit Usage
+
+The syscall will always attempt to consume the same amount of CUs regardless of
+control flow.
+
+CU usage is proportional to the `length` parameter.
+
+```
+sysvar_base + (32 / cpi_per_u) + max(mem_op_base, (length / cpi_per_u))
+```
+
+- `sysvar_base`: Base cost of accessing a sysvar.
+- `cpi_per_u`: Number of account data bytes per CU charged during CPI.
 
 ### Supported Sysvars
 
--- a/proposals/0127-get-sysvar-syscall.md
+++ b/proposals/0127-get-sysvar-syscall.md
@@ -1,0 +1,180 @@
+---
+simd: '0127'
+title: sol_get_sysvar syscall
+authors:
+  - Richard Patel
+category: Standard
+type: Core
+status: Draft
+created: 2024-03-15
+---
+
+## Summary
+
+SIMD-0127 introduces the syscall `sol_get_sysvar` which allows BPF programs
+to retrieve any sysvar.
+
+New invariants are placed on the sysvar cache to enable safe retrieval.
+
+Existing syscalls to retrieve sysvars are henceforth considered obsolete.
+
+## Motivation
+
+Sysvars are special accounts that are implicitly modified by the runtime.
+Certain sysvars are essential for operation of the network.  For example, the
+system program reads the 'recent block hashes' sysvar to provide durable nonce
+functionality.
+
+Currently, programs can read sysvars can using one of these three methods:
+
+1. Via explicit account access by specifying the sysvar address in the list of
+   transaction accounts
+2. Via API calls to the sysvar cache (for native programs only)
+3. Via syscalls that copy the sysvar content to virtual machine memory
+   (for BPF programs only, does not work for all sysvars)
+
+Improvements to these APIs are motivated by the following reasons.
+
+### Allowing sysvar upgrades
+
+The function signature of a `sol_get_{...}_sysvar` syscall is as follows:
+
+```c
+uint64_t sol_get_example_sysvar( void * dest_addr );
+```
+
+When called, the syscall handler will copy the sysvar to the given memory
+location.  Notably, BPF programs do not inform the syscall handler of the
+expected size of the sysvar.  BPF programs thus effectively assume that the
+requested sysvars will forever have the same size.
+
+This prevents future upgrades that shrink or extend sysvars.
+
+### Syscall API fragmentation
+
+The following sysvars are currently accessible using syscalls:
+
+- `SysvarC1ock11111111111111111111111111111111` (via `sol_get_clock_sysvar`)
+- `SysvarEpochRewards1111111111111111111111111` (via `sol_get_epoch_rewards_sysvar`)
+- `SysvarEpochSchedu1e111111111111111111111111` (via `sol_get_epoch_schedule_sysvar`)
+- `SysvarFees111111111111111111111111111111111` (via `sol_get_fees_sysvar`)
+- `Sysvar1nstructions1111111111111111111111111` (indirectly via `sol_get_processed_sibling_instruction`)
+- `SysvarLastRestartS1ot1111111111111111111111` (via `sol_get_last_restart_slot`)
+- `SysvarRent111111111111111111111111111111111` (via `sol_get_rent_sysvar`)
+
+BPF programs cannot access the following sysvars using syscalls.
+
+- `SysvarRecentB1ockHashes11111111111111111111`
+- `SysvarRewards111111111111111111111111111111`
+- `SysvarS1otHashes111111111111111111111111111`
+- `SysvarS1otHistory11111111111111111111111111`
+- `SysvarStakeHistory1111111111111111111111111`
+
+Introducing a new syscall for every new sysvar results in bloat of the syscall
+interface.  The aforementioned function signature that most sysvar getters use
+is also inefficient for large accounts.  Copying the entire sysvar to VM memory
+might consume excessive compute units (e.g. 16392 bytes for 'stake history').
+
+### Unblocking core BPF programs
+
+As per [SIMD-0088](./0088-enable-core-bpf-programs.md), an effort is underway
+to reduce code complexity of the Solana runtime by porting native programs to
+core BPF programs.
+
+These new programs target full compatibility with their predecessors to avoid
+introducing breaking API changes.  In some cases, native programs read a sysvar
+from the sysvar cache without requiring the user to include this sysvar in the
+transaction account list.  This happens in the stake program 'redelegate'
+instruction, for example.
+
+As a result, it is currently impossible to port certain native programs to core
+BPF programs.  Another factility to expose sysvars to the virtual machine is
+required.
+
+### Optimizing existing user programs
+
+Reading sysvars with the proposed syscall is expected to be more efficient than
+an explicit account access.  Removing explicit account accesses to sysvars will
+generally result in smaller transactions.  Reducing the amount of account data
+that is mapped into virtual machine memory may also yield cost savings.
+
+## Alternatives Considered
+
+Other mechanisms considered that provide sysvar data to BPF programs are as
+follows:
+
+### Sysvar-specific syscalls
+
+TODO
+
+### New memory regions
+
+TODO
+
+## Detailed Design
+
+### Sysvar Cache
+
+TODO
+
+### Restrictions on sysvar data dependencies
+
+TODO
+
+### `sol_get_sysvar` syscalls
+
+TODO this is outdated
+
+```c
+/* sol_get_sysvar_size queries the serialized byte size of a sysvar.
+   It is equal to the length of the serialization of the sysvar value.
+   The sysvar value is taken in the sysvar cache.
+
+   sysvar_id points to the 32 byte address of the sysvar.
+
+   If a non-empty sysvar was found in the sysvar cache, returns the
+   byte size of that sysvar.
+
+   If the requested sysvar was not found, is corrupt, is
+   unsupported for the current feature set, or is zero size, returns 0.
+
+   This syscall aborts the transaction if (and only if):
+   - The compute budget was exhausted */
+
+uint64_t
+sol_get_sysvar_size( /* r1 */ void const * sysvar_id );
+
+/* sol_get_sysvar copies sysvar data to VM memory.
+
+   sysvar_id points to the 32 byte address of the sysvar.
+
+   A sysvar is valid if sol_get_sysvar_size(sysvar_id) returns non-zero.
+   If the sysvar is not valid, returns zero without writing to out.
+
+   The byte range [offset,offset+size) selects the range of data to be
+   copied. The sysvar data is the serialized value of the corresponding
+   sysvar cache entry.
+   The size of this sysvar data is returned by sol_get_sysvar_size.
+
+   Copies the sysvar data to [out,out+size) and returns 1 if the sysvar is valid.
+
+   This syscall aborts the transaction if:
+   - Range [out,out+size) is not writable
+   - The compute budget was exhausted */
+
+int
+sol_get_sysvar(
+  /* r1 */ void const * sysvar_id,
+  /* r2 */ uint8_t *    out,
+  /* r3 */ uint64_t     offset,
+  /* r4 */ uint64_t     size
+);
+```
+
+## Impact
+
+TODO
+
+## Security Considerations
+
+TODO
--- a/proposals/0127-get-sysvar-syscall.md
+++ b/proposals/0127-get-sysvar-syscall.md
@@ -185,9 +185,6 @@ in the Sysvar Cache.
 Unless specified otherwise, any new sysvars added to the Sysvar Cache in the
 future also become accessible through this syscall.
 
-Existing syscalls to fetch sysvar data will be deprecated in favor of the new
-unified syscall proposed herein.
-
 The supported list of sysvars for the proposed syscall will be as follows:
 
 - `SysvarC1ock11111111111111111111111111111111`
@@ -216,6 +213,15 @@ interface.
 Furthermore, it unblocks the efforts to give BPF programs more access to
 sysvar data, so on-chain programs gain the ability to read more sysvar data
 without increasing transaction size.
+
+One minor change is the behavior of verification of syscall availability for a
+given program. Currently, the availability of specific sysvar data can be
+verified by simply determining whether its corresponding syscall is available to
+the program during the verification step.
+
+With this change, even though the new unified syscall can be verified available
+during program verification, the actual availability of specific sysvar data is
+determined at runtime.
 
 ## Security Considerations
 
--- a/proposals/0127-get-sysvar-syscall.md
+++ b/proposals/0127-get-sysvar-syscall.md
@@ -165,7 +165,7 @@ Otherwise, the syscall writes the sysvar data to the virtual memory at
 
 ### Compute Unit Usage
 
-The syscall will always attempt to consume the same amount of CUs regardless of
+The syscall will always consume the same amount of CUs regardless of
 control flow.
 
 CU usage is proportional to the `length` parameter.
--- a/proposals/0127-get-sysvar-syscall.md
+++ b/proposals/0127-get-sysvar-syscall.md
@@ -2,7 +2,6 @@
 simd: '0127'
 title: Get-Sysvar Syscall
 authors:
-  - Richard Patel (Jump)
   - Joe Caulfield (Anza)
 category: Standard
 type: Core
@@ -112,30 +111,42 @@ region each time a VM is created (currently one per transaction processed),
 causing a degradation in runtime performance and likely reducing compute
 by far less compared to the compute savings posed by this proposal.
 
+Even with direct mapping, this change makes address translation more complex
+and therefore slows down _all_ memory accesses.
+
 ## New Terminology
 
 N/A.
 
 ## Detailed Design
 
-Defining one single `get` API to retrieve data from the sysvar cache can be used
-by all forward-facing sysvar interfaces. This `get` API should require the
-`sysvar_id` to retrieve data from, the `offset` at which to start copying data,
-and the `length` of the bytes to copy and return to the caller.
+A single `sol_get_sysvar` syscall can be defined, which can be used by all
+forward-facing sysvar interfaces to retrieve data from any sysvar.
 
-The `sysvar_id` can be the address of the sysvar account or simply a single-byte
-enum aligned with the available sysvars in the sysvar cache. This is an
-implementation detail.
+```c
+/**
+ * Retrieves a slice of data from a sysvar and copies it to VM memory.
+ *
+ * @param sysvar_id The identifier of the sysvar to retrieve data from.
+ *                  It should be a single byte representing the sysvar.
+ * @param var_addr  VM memory address to copy the retrieved data to.
+ * @param offset    The offset to start copying data from.
+ * @param length    The length of data to copy, in bytes.
+ * @return          A 64-bit unsigned integer error code:
+ *                    - 0 if the operation is successful.
+ *                    - Non-zero error code.
+ *
+ * If the operation is not successful, data will not be written to the
+ * provided VM memory address.
+ */
+uint64_t sol_get_sysvar(
+  /* r1 */ void const * sysvar_id,
+  /* r2 */ uint8_t *    var_addr,
+  /* r3 */ uint64_t     offset,
+  /* r4 */ uint64_t     length,
+);
 
-A psuedo-code representation of the `get` API is as follows, where the returned
-value is the slice of bytes from the requested sysvar.
-
-```rust
-fn get(sysvar_id: SysvarID, offset: usize, length: usize) -> &[u8];
 ```
-
-In reality, the syscall will write this slice of data to a memory address,
-exactly like the existing syscalls do for sysvar data.
 
 Generally, the following cases would be handled with errors:
 
@@ -147,27 +158,9 @@ Sysvars themselves should have internal logic for understanding their own size
 and the size of their entries if they are a list-based data structure. With this
 information, each sysvar can customize how it wishes to drive the `get` syscall.
 
-A list-based sysvar - say `SlotHashes` - could leverage this API like so:
-
-```rust
-fn get(index: usize) -> SlotHash {
-  let length = size_of::<SlotHash>();
-  let offset = index * length;
-  syscall::get(offset, length)
-}
-```
-
 Any sysvar that wishes to offer a more robust API would then be responsible for
 defining such an interface, rather than pushing that burden onto the syscall
 interface. For example, `SlotHashes` could offer something like the following:
-
-```rust
-fn get_hash(index: usize) -> Hash {
-  let offset = index * size_of::<SlotHash>() + size_of::<Slot>();
-  let length = size_of::<Hash>();
-  syscall::get(offset, length)
-}
-```
 
 Any additionally complex methods - such as binary searches - should be left to
 the on-chain program to implement locally. These should not be made available to
--- a/proposals/0127-get-sysvar-syscall.md
+++ b/proposals/0127-get-sysvar-syscall.md
@@ -113,19 +113,30 @@ by far less compared to the compute savings posed by this proposal.
 
 ## Detailed Design
 
-By simply redefining the `get` API to accept `offset` and `length` parameters,
-one single `get` syscall can be used by all sysvars to retrieve data.
+Defining one single `get` API to retrieve data from the sysvar cache can be used
+by all forward-facing sysvar interfaces. This `get` API should require the
+`sysvar_id` to retrieve data from, the `offset` at which to start copying data,
+and the `length` of the bytes to copy and return to the caller.
+
+The `sysvar_id` can be the address of the sysvar account or simply a single-byte
+enum aligned with the available sysvars in the sysvar cache. This is an
+implementation detail.
+
+A psuedo-code representation of the `get` API is as follows, where the returned
+value is the slice of bytes from the requested sysvar.
 
 ```rust
-fn get(offset: usize, length: usize) -> &[u8];
+fn get(sysvar_id: SysvarID, offset: usize, length: usize) -> &[u8];
 ```
 
-The specifics of returning optional values or errors can be left up to the
-implementation, but the following cases would be handled with errors:
+In reality, the syscall will write this slice of data to a memory address,
+exactly like the existing syscalls do for sysvar data.
+
+Generally, the following cases would be handled with errors:
 
 - The sysvar data is unavailable.
-- The sysavr data is corrupt.
-- The provided offset or length would be out of bounds on the sysvar data.
+- The sysvar data is corrupt.
+- The provided offset or length would be out of bounds on the sysvar data. 
 
 Sysvars themselves should have internal logic for understanding their own size
 and the size of their entries if they are a list-based data structure. With this
@@ -158,7 +169,6 @@ the on-chain program to implement locally. These should not be made available to
 BPF programs via the sysvar interface.
 
 The syscall interface for sysvars should only support `get` as defined above.
-
 
 ## Impact
 
--- a/proposals/0127-get-sysvar-syscall.md
+++ b/proposals/0127-get-sysvar-syscall.md
@@ -8,6 +8,7 @@ category: Standard
 type: Core
 status: Draft
 created: 2024-03-15
+feature: (fill in with feature tracking issues once accepted)
 ---
 
 ## Summary
@@ -57,7 +58,7 @@ fragments of the data can be made accessible via calls like
 Additionally, as per [SIMD-0088](./0088-enable-core-bpf-programs.md), an effort
 is underway to reduce code complexity of the Solana runtime by porting native
 programs to Core BPF programs. These programs must be 100% backwards compatible,
-therefore requiring access to this sysvar data via syscalls and *not* via
+therefore requiring access to this sysvar data via syscalls and _not_ via
 providing the account directly. Requiring the account would break the programs'
 ABIs.
 
@@ -75,7 +76,7 @@ spirit, if a new sysvar is created, a new syscall must be created.
 
 Furthermore, if certain queries are requred by BPF programs - such as
 `SlotHashes::get_slot()` and `StakeHistory::get_entry()` - those single-element
-retrievals will *also* require new syscalls.
+retrievals will _also_ require new syscalls.
 
 It's worth noting that at the time of this writing, the work required for
 [SIMD-0088](./0088-enable-core-bpf-programs.md) will require at least five new
@@ -111,6 +112,10 @@ region each time a VM is created (currently one per transaction processed),
 causing a degradation in runtime performance and likely reducing compute
 by far less compared to the compute savings posed by this proposal.
 
+## New Terminology
+
+N/A.
+
 ## Detailed Design
 
 Defining one single `get` API to retrieve data from the sysvar cache can be used
@@ -136,7 +141,7 @@ Generally, the following cases would be handled with errors:
 
 - The sysvar data is unavailable.
 - The sysvar data is corrupt.
-- The provided offset or length would be out of bounds on the sysvar data. 
+- The provided offset or length would be out of bounds on the sysvar data.
 
 Sysvars themselves should have internal logic for understanding their own size
 and the size of their entries if they are a list-based data structure. With this
@@ -146,7 +151,7 @@ A list-based sysvar - say `SlotHashes` - could leverage this API like so:
 
 ```rust
 fn get(index: usize) -> SlotHash {
-  let length = size_of::<SlotHash>(); 
+  let length = size_of::<SlotHash>();
   let offset = index * length;
   syscall::get(offset, length)
 }
@@ -172,7 +177,7 @@ The syscall interface for sysvars should only support `get` as defined above.
 
 ## Impact
 
-As mentioned as partial motiviation for this proposal, the new syscall would
+As mentioned as partial motivation for this proposal, the new syscall would
 make changing and adding new sysvars in the future much easier, which would be
 positively impactful to contributors.
 
@@ -188,7 +193,6 @@ without increasing transaction size.
 This new syscall interface does increase the chances of improper management of
 bytes and offsets, which could throw critical errors.
 
-However, this can be mitigated in testing, and the upside for reducing the numbe
-of syscalls - thereby reducing the surface area for other critical bugs or
-explots - is a decent tradeoff.
-
+However, this can be mitigated in testing, and the upside for reducing the
+number of syscalls - thereby reducing the surface area for other critical bugs
+or exploits - is a worthy tradeoff.
--- a/proposals/0127-get-sysvar-syscall.md
+++ b/proposals/0127-get-sysvar-syscall.md
@@ -150,6 +150,7 @@ Generally, the following cases would be handled with errors:
 - The sysvar data is unavailable.
 - The sysvar data is corrupt.
 - The provided offset or length would be out of bounds on the sysvar data.
+- The provided destination address `var_addr` is invalid.
 
 Sysvars themselves should have internal logic for understanding their own size
 and the size of their entries if they are a list-based data structure. With this
--- a/proposals/0127-get-sysvar-syscall.md
+++ b/proposals/0127-get-sysvar-syscall.md
@@ -148,43 +148,12 @@ uint64_t sol_get_sysvar(
 If the following conditions are not met, this syscall will return an error:
 
 - The sysvar data is available to read.
-- The sysvsr data is not corrupt.
+- The sysvar data is not corrupt.
 - `offset` + `length` is in the range `[0, 2^64)`.
 - `var_addr` + `length` is in the range `[0, 2^64)`.
 - Each byte in the range starting at `var_addr` of size `length` is writable.
 
-Sysvars themselves should have internal logic for understanding their own size
-and the size of their entries if they are a list-based data structure. With this
-information, each sysvar can customize how it wishes to drive the `get` syscall.
-
-Any sysvar that wishes to offer a more robust API would then be responsible for
-defining such an interface, rather than pushing that burden onto the syscall
-interface. For example, `SlotHashes` could offer something like the following:
-
-Any additionally complex methods - such as binary searches - should be left to
-the on-chain program to implement locally. These should not be made available to
-BPF programs via the sysvar interface.
-
-The syscall interface for sysvars should only support `get` as defined above.
-
 ### Supported Sysvars
-
-Currently, the following sysvars are available to BPF programs:
-
-- `SysvarC1ock11111111111111111111111111111111` (via `sol_get_clock_sysvar`)
-- `SysvarEpochRewards1111111111111111111111111` (via `sol_get_epoch_rewards_sysvar`)
-- `SysvarEpochSchedu1e111111111111111111111111` (via `sol_get_epoch_schedule_sysvar`)
-- `SysvarFees111111111111111111111111111111111` (via `sol_get_fees_sysvar`)
-- `Sysvar1nstructions1111111111111111111111111` (indirectly via `sol_get_processed_sibling_instruction`)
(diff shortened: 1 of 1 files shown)


t-nelson:
is the intent here to document the _existing_ sysvar getter interface or a new one?

riptl:
> is the intent here to document the _existing_ sysvar getter interface or a new one?

@t-nelson Correct me if I'm wrong, but there isn't one syscall to get a syscall currently. There are like 5 different ones and they don't scale well with sysvar size. So this proposals adds a new sysvar getter that overlaps with the existing syscalls. I suggest disabling the then deprecated syscalls for loader v4.

buffalojoec:
> > is the intent here to document the _existing_ sysvar getter interface or a new one?
> 
> @t-nelson Correct me if I'm wrong, but there isn't one syscall to get a syscall currently. There are like 5 different ones and they don't scale well with sysvar size. So this proposals adds a new sysvar getter that overlaps with the existing syscalls. I suggest disabling the then deprecated syscalls for loader v4.

For context, @ripatel-fd you're right there is not one single syscall to retrieve any sysvar response. Each `Sysvar::get()` implementation maps to one syscall per sysvar.

Here's the trait implementation that most sysvars employ (the large ones omit this).
https://github.com/anza-xyz/agave/blob/e8526f60aa80d986111f935717f003d6be7670df/sdk/program/src/sysvar/mod.rs#L221-L239

And here's `Clock` using it to map to the syscall `sol_get_clock_sysvar`.
https://github.com/anza-xyz/agave/blob/e8526f60aa80d986111f935717f003d6be7670df/sdk/program/src/sysvar/clock.rs#L134-L136

Then each getter syscall is in here.
https://github.com/anza-xyz/agave/blob/master/programs/bpf_loader/src/syscalls/sysvar.rs

It's my understanding that the intention of this SIMD is to have _one_ syscall interface that all sysvars can use? So you have `get_size` and `get_sysvar` (or `get_data`), and that can handle any of the sysvar data types?

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,180 @@
...
+system program reads the 'recent block hashes' sysvar to provide durable nonce
+functionality.
+
+Currently, programs can read sysvars can using one of these three methods:
+
+1. Via explicit account access by specifying the sysvar address in the list of
+   transaction accounts
+2. Via API calls to the sysvar cache (for native programs only)
+3. Via syscalls that copy the sysvar content to virtual machine memory
+   (for BPF programs only, does not work for all sysvars)
```
buffalojoec, line 34: More accurately: "intentionally not made available for all sysvars".

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,180 @@
...
+- `SysvarEpochSchedu1e111111111111111111111111` (via `sol_get_epoch_schedule_sysvar`)
+- `SysvarFees111111111111111111111111111111111` (via `sol_get_fees_sysvar`)
+- `Sysvar1nstructions1111111111111111111111111` (indirectly via `sol_get_processed_sibling_instruction`)
+- `SysvarLastRestartS1ot1111111111111111111111` (via `sol_get_last_restart_slot`)
+- `SysvarRent111111111111111111111111111111111` (via `sol_get_rent_sysvar`)
+
+BPF programs cannot access the following sysvars using syscalls.
+
+- `SysvarRecentB1ockHashes11111111111111111111`
+- `SysvarRewards111111111111111111111111111111`
```
buffalojoec, line 68: This may or may not be necessary for the proposal, but these two sysvars are marked deprecated.

https://github.com/anza-xyz/agave/blob/e8526f60aa80d986111f935717f003d6be7670df/sdk/program/src/sysvar/recent_blockhashes.rs#L8

https://github.com/anza-xyz/agave/blob/e8526f60aa80d986111f935717f003d6be7670df/sdk/program/src/sysvar/rewards.rs#L1

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,180 @@
...
+that is mapped into virtual machine memory may also yield cost savings.
+
+## Alternatives Considered
+
+Other mechanisms considered that provide sysvar data to BPF programs are as
+follows:
+
+### Sysvar-specific syscalls
+
+TODO
```
buffalojoec, line 108: You could just say "change nothing; this is what we currently do".

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,180 @@
...
+
+### New memory regions
+
+TODO
```
buffalojoec, line 112: Maybe something like this?

```suggestion
### New memory regions

A new memory region in the VM could be introduced to store sysvar data, which
could then be made available to BPF programs through an interface around raw
pointers, which loads the data similarly to the interface proposed herein.

This approach would depend on direct mapping of host memory to VM memory,
which has been in development for some time and requires further research
and development to introduce safely to the VM ABI. This extensive R&D period
would mean Core BPF program initiatives would remain blocked.

Without direct mapping, sysvar data would be fully copied to this memory
region each time a VM is created (currently one per transaction processed),
causing a degradation in runtime performance and likely reducing compute
by far less compared to the compute savings posed by this proposal.
```

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,180 @@
...
+
+   A sysvar is valid if sol_get_sysvar_size(sysvar_id) returns non-zero.
+   If the sysvar is not valid, returns zero without writing to out.
+
+   The byte range [offset,offset+size) selects the range of data to be
+   copied. The sysvar data is the serialized value of the corresponding
+   sysvar cache entry.
+   The size of this sysvar data is returned by sol_get_sysvar_size.
+
+   Copies the sysvar data to [out,out+size) and returns 1 if the sysvar is valid.
```
buffalojoec, line 159: It sounds like this does offer a lot of flexibility for larger sysvars, particularly when only a subset or a single element of the data is being requested. However, this still would not enable a BPF program to request the full data for a sysvar like `SlotHashes`, correct? The program would still face the current "too large to copy" resource limitation?

buffalojoec:
> @buffalojoec Thanks for the review. Would you like to take over this SIMD? I'm working on other stuff at the moment and won't be able to finish it for a while, so I'll close it for now. Feel free to reopen it with your additions.

Yeah, I'll take it over.

proposals/0127-get-sysvar-syscall.md
2501babe: if you really wanted to, you could take the whole thing to heap. i believe you said slot hashes is like 18kb? do you need more than one entry for your usecase tho? i only need one for mine

buffalojoec: ALT nope but Vote does yeah. Unfortunately, Vote needs something like `SlotHashes::get_slots() -> &[Slot]`.

buffalojoec: Aside from that, it seems like single-element lookups would suffice for all the other use cases I have. There's three I can think of across ALT, Vote, and Feature Gate.

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,184 @@
...
+by far less compared to the compute savings posed by this proposal.
+
+## Detailed Design
+
+By simply redefining the `get` API to accept `offset` and `length` parameters,
+one single `get` syscall can be used by all sysvars to retrieve data.
+
+```rust
+fn get(offset: usize, length: usize) -> &[u8];
+```
```
2501babe, line 121: note the current unified `get` api is implementation-defined, built on top of a set of sysvar-specific syscalls. so a new unified syscall would need to accept an argument indicating which sysvar to fetch and read/write from. so we would need a protocol-defined enum mapping sysvars to u8 tags, or we need to pass in the account id of the sysvar and have a protocol-defined list of allowed account ids/sysvars

on that topic im in favor of explicitly not supporting `Instructions` via this mechanism. i think the must-haves are clock, rent, slot hashes, stake history, last restart slot. im less sure about epoch schedule/rewards, assume we dont need slot history, and assume its obvious we dont need fees or recent blockhashes

we should also be clear that sysvars can be added or removed in the future... although in practice i guess we can never really remove them if programs have come to reply on them

also noting that syscalls generally (always?) return a u64 with a success/error code and the output buffer needs to be passed in as a pointer

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,184 @@
...
+
+The specifics of returning optional values or errors can be left up to the
+implementation, but the following cases would be handled with errors:
+
+- The sysvar data is unavailable.
+- The sysavr data is corrupt.
+- The provided offset or length would be out of bounds on the sysvar data.
```
2501babe, line 128: it occurs to me i dont actually know if ? inside syscalls transforms to an error code, or aborts. or if error codes have defined meanings for all syscalls or whether each syscall can define its own... in any event it would be good to define a way for the syscall to signal these error conditions to the implementation. inclined to say that if the sysvar is unavailable (if a sysvar is deprecated/removed, or specifically if we support EpochRewards) it should return an error code that the implementation can handle, but corrupt data or out of bounds read should abort

also im not sure what happens if the syscall attempts an out of bounds write. presumably it cant detect and prevent this because everything we do with this is unsafe transmute... at the least, the spec should say the caller must ensure their buffer is long enough

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,184 @@
...
+
+Sysvars themselves should have internal logic for understanding their own size
+and the size of their entries if they are a list-based data structure. With this
+information, each sysvar can customize how it wishes to drive the `get` syscall.
```
2501babe, line 132: highly in favor of the unstated implication here ie that this syscall is not made available to programs directly and must be provided through an interface (though perhaps that is already the case for all syscalls?) so we dont lock ourselves into memory layout decisions more than we already are

2501babe:
overall i like this approach better than #124 and then having to do the same as that for `SlotHashes` and then whatever else would come up in the future. it would also allow us to get rid of most of the existing sysvar syscalls, as their outputs could be produced via this one

proposals/0127-get-sysvar-syscall.md
buffalojoec: > a new unified syscall would need to accept an argument indicating which sysvar to fetch and read/write from

Good point. I missed that bit. Richie's original draft had the sysvar address being passed. Personally I don't see a reason to opt for the 32-byte address over a single-byte enum, do you? Regardless, I've added this to the proposal.

> on that topic im in favor of explicitly not supporting `Instructions` via this mechanism. i think the must-haves are clock, rent, slot hashes, stake history, last restart slot.

I think it probably makes the most sense to couple this enum, which describes which sysvars are available through the unified syscall, to the sysvar cache. I think if we start to include things from the invoke context in here it might get a bit complicated. In the future, if we decide to apply a unified interface to the invoke context data as well, we can draft a new proposal. What do you think?

I don't see a particular reason to omit anything from the sysvar cache that isn't deprecated - like the epoch schedule/rewards. If they're there, might as well make them available.

> we should also be clear that sysvars can be added or removed in the future... although in practice i guess we can never really remove them if programs have come to reply on them

I touched on adding new ones a few times in here, but yeah I specifically didn't say remove because that would be a deprecation process for the forward-facing sysvar interface, not this unified syscall.

> also noting that syscalls generally (always?) return a u64 with a success/error code and the output buffer needs to be passed in as a pointer

True, I was trying to psuedo-code demonstrate that you'd provide those params and get a slice, but you're right I should probably be a bit more explicit on how the syscall-sysvar relationship would work.

proposals/0127-get-sysvar-syscall.md
buffalojoec: > it occurs to me i dont actually know if ? inside syscalls transforms to an error code, or aborts.

It does indeed! It gets turned into a `u64` but then turned back into an `InstructionError` in the sysvar interface. So we have lots of flexibility there to define our own syscall-level errors, either with existing `InstructionError` variants or `InstructionError::Custom`.

> also im not sure what happens if the syscall attempts an out of bounds write.

The way I was thinking about this was that it would be an out of bounds _read_, and that's easy to detect because say you get `offset: 90` and your desired sysvar is only `60` bytes long, throw.

proposals/0127-get-sysvar-syscall.md
buffalojoec: Pushed changes with this feedback addressed!

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,198 @@
...
+The `sysvar_id` can be the address of the sysvar account or simply a single-byte
+enum aligned with the available sysvars in the sysvar cache. This is an
+implementation detail.
+
+A psuedo-code representation of the `get` API is as follows, where the returned
+value is the slice of bytes from the requested sysvar.
+
+```rust
+fn get(sysvar_id: SysvarID, offset: usize, length: usize) -> &[u8];
+```
```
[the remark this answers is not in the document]
buffalojoec: I can walk this away from Rust. I can't promise C. In doing so, do you have any qualms about making sysvar ID an enum value versus a 32-byte address?

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,198 @@
...
+defining such an interface, rather than pushing that burden onto the syscall
+interface. For example, `SlotHashes` could offer something like the following:
+
+```rust
+fn get_hash(index: usize) -> Hash {
+  let offset = index * size_of::<SlotHash>() + size_of::<Slot>();
+  let length = size_of::<Hash>();
+  syscall::get(offset, length)
+}
+```
```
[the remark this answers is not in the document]
buffalojoec: OK I can make this less verbose then, and remove the code snippets.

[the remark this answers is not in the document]
buffalojoec: Done.

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,198 @@
+---
+simd: '0127'
+title: Get-Sysvar Syscall
+authors:
+  - Richard Patel (Jump)
```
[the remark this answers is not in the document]
buffalojoec: Done.

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,198 @@
...
+Without direct mapping, sysvar data would be fully copied to this memory
+region each time a VM is created (currently one per transaction processed),
+causing a degradation in runtime performance and likely reducing compute
+by far less compared to the compute savings posed by this proposal.
+
+## New Terminology
+
+N/A.
+
+## Detailed Design
```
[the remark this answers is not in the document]
buffalojoec: Sorry, I guess I'm not entirely sure how much of our syscall interface is a specification or Agave-specific. For example, do all invoke contexts look the same? If so, we can expand this to say "anything from the invoke context is retrievable", but I was sort of under the impression that invoke context and sysvar cache can look different amongst clients. FD might serve syscalls from some other data source entirely when it comes to sysvars. Nobody says FD has to specifically have a sysvar _cache_. Right?

proposals/0127-get-sysvar-syscall.md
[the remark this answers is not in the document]
buffalojoec: I think I've managed to answer my own question on this.

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,198 @@
...
+
+Defining one single `get` API to retrieve data from the sysvar cache can be used
+by all forward-facing sysvar interfaces. This `get` API should require the
+`sysvar_id` to retrieve data from, the `offset` at which to start copying data,
+and the `length` of the bytes to copy and return to the caller.
```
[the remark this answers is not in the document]
buffalojoec: Duplicate of below.

proposals/0127-get-sysvar-syscall.md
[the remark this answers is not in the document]
buffalojoec: Ok fair enough. Thanks for all of that context, it's helpful for this and other efforts!

I'll include the sysvars that are accessible by the new syscall in the proposal. As for your points in the earlier message:

> 1. What about the sysvars that are not in the sysvar cache? (Of which there are a bunch) (example: "the sysvar getter does not expose sysvars that are not in the sysvar cache)

I don't particularly see a valid reason for _not_ including data outside of the sysvar cache, but being specific about it. For example, I think it's reasonable to say things from the invoke context - instructions, recent blockhash, etc. - could be made available from this getter. Perhaps everything from the sysvar cache and select items from the invoke context?

> 2. What about any newly added sysvars? (example verbiage: "Unless specified otherwise, any sysvars that are introduced into the sysvar cache in the future also become visible via the sysvar getter")

Yeah, I think this should go in as-is.

> 3. The sysvar cache contains logical objects, not a byte serialization of these objects. Underspecifying this can lead to a consensus issue. When restoring the sysvar cache from the account DB, the sysvar cache ignores trailing data when deserializing. So does the sysvar getter include or exclude this trailing data?

I would lean towards no, it should not include it. It should include whatever's available to transaction processing, and if in the future we decide the **sysvar cache** _shouldn't_ ignore those trailing bytes, they're automatically made available to this syscall via the cache.

2501babe:
while impling my own prototype i realized we need to specify more clearly what exactly the output of this syscall is supposed to be. i was proceeding on the assumption that we just want to copy out a slice from the in-memory representation of the sysvar. eg for `Clock`, `get_sysvar(Tag::Clock, 16, 8, &ptr)` would write out the third `u64`

unfortunately i was rudely reminded that `StakeHistory` the object is a vector, and its only the account data thats an array. so "write out the in-memory representation" is incoherent

there are a few options:
* specify that `length` and `offset` refer to the _account data_ rather than the memory representation. this could be ugly because it would mean we need to re-serialize the object in the `SysvarCache` or keep the buffers around on the off chance someone might request them. i guess we could also just calculate what vector entries correspond to what byte positions in the account data, serialize only those entries (or pieces of entries...) and then copy that. it does complicate things somewhat however
* specify that this syscall is vector-aware in some way
  - the weakest version is to say "for vectors you can only copy whole entries or the length parameter"
  - a stronger version would be "you can only get one entry at a time." eg stake history is essentially `StakeHistory(Vec<(epoch, u64, u64, u64)>)`. so `get(0,4)` gets length, `get(4, 36)` gets the first entry, `get(0, 2)`, `get(4, 68)` and so forth would be illegal
  - the strongest version would be for flat unboxed sysvars `length` and `offset` refer to bytes whereas for vector sysvars `length` and `offset` refer to vector slots, eg the first two `StakeHistoryEntry` would be `get(0,2)`. we could expose a mechanism to get length or just return empty

proposals/0127-get-sysvar-syscall.md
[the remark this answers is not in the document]
2501babe: imo we should specify a list of sysvars that the syscall supports and leave it at that

i was regarding the sysvar cache as an implementation detail that shouldnt be explicitly referenced in the specification of this at all, but it sounds like we need to make this explicit (and expect that firedancer will copy the sysvar cache mechanism exactly) because the cache may have consistency issues, so the source of the data could affect consensus?

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,227 @@
...
+  /* r4 */ uint64_t     length,
+);
+
+```
+
+Generally, the following cases would be handled with errors:
+
+- The sysvar data is unavailable.
+- The sysvar data is corrupt.
+- The provided offset or length would be out of bounds on the sysvar data.
```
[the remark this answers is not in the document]
buffalojoec: Sorry, is this not captured by bullet 3? Offset 900 would be out of bounds. If you request `0` length in-bounds, why error?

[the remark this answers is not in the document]
buffalojoec: Added!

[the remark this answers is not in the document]
buffalojoec: Gotcha. How's this then?

```md
If the following conditions are not met, this syscall will return an error:

- The sysvar data is available to read.
- The sysvar data is not corrupt.
- `offset` + `length` must be in the range `[0, 2^64)`.
- `var_addr` + `length` must be in the range `[0, 2^64)`.
- Each byte in the range starting at `var_addr` of size `length` must be 
  writable.
```

2501babe:
i have to think about it a bit more (and by think about it i mean probably just implement it) but. i tried prototyping this last night. i had been running off the assumption that out length/offset design referred to the in-memory representation of the objects, but i realize now this is insufficient because `StakeHistory` and `SlotHashes` are vectors

im thinking that in the spec we want to explicitly state that the length/offset refer to the data as represented in the account buffer itself. since `SysvarCache` provide already deserialized objects though, this means we may need to reserialize the data to return it

i think it might be prudent to enforce in the spec that for vector sysvars, we will only return either the length, or an undivided set of vector entries. this means, for instance, for stake history, (pseudocode) `get(0,4)` would be valid, `get(4,36)` would be valid, `get(3,68)` would be valid, but eg `get(0,36)` would be invalid, `get(3,57)` would be invalid, etc

this would both simplify implementation and also make the sematics clearer between both rust and c. the fact that vectors are boxed means they cant be just copied out as-is from the cache. but if we say that we only return the length `u32` or an unboxed array of vector entries, this problem sort of goes away. it also makes serialization in rust simpler. does this seem reasonable?

buffalojoec:
>im thinking that in the spec we want to explicitly state that the length/offset refer to the data as represented in the account buffer itself.

Yep agreed. Conceptually the way that the sysvars implement this API should mirror to partial `from_account_info` deserialization, if we desired to offer that as well with the same approach (off-topic).

> since SysvarCache provides already deserialized objects though, this means we may need to reserialize the data to return it

But if we're only returning length or an undivided entry, we can just re-serialize that slice, right?

> i think it might be prudent to enforce in the spec that for vector sysvars, we will only return either the length, or an undivided set of vector entries. this means, for instance, for stake history, (pseudocode) `get(0,4)` would be valid, `get(4,36)` would be valid, `get(4,68)` would be valid, but eg `get(0,36)` would be invalid, `get(3,57)` would be invalid, etc
> 
> this would both simplify implementation and also make the sematics clearer between both rust and c. the fact that vectors are boxed means they cant be just copied out as-is from the cache. but if we say that we only return the length `u32` or an unboxed array of vector entries, this problem sort of goes away. it also makes serialization in rust simpler. does this seem reasonable?

Yes quite reasonable! I think, to your point, maybe it's better to add specifications for each sysvar here, rather than just their IDs. It's relevant to do so since we've established that each client is going to have to adhere to the sysvar cache spec and we already know the sysvar format must exactly match.

Right now we have three classifications of sysvar layouts:
1. Static, non-list (`Clock`, `Rent`, etc.).
2. Absolute vector (`SlotHashes`, `StakeHistory`).
3. Layout with embedded vector (or whatever) (`SlotHistory`).

So we might as well add the specification for each classification, and outline exactly what you're describing.

The first one is easy, and you just laid out the second one. The third is pretty much the exact same as the second, just with some additional specs:
- The leading byte is a discriminator for a `BitVec`.
- The leading length value (length of the `BitVec`) is a `u64` (8 bytes).
- The entire `BitVec` is `131_072` bytes.
- The `BitVec` contents is followed by 8 bytes for the number of bits in the `BitVec` (static `1024 * 1024`).
- The last 8 bytes are the next slot.

buffalojoec:
@2501babe I took a stab at defining this in the SIMD with my most recent commit. I based it on your comments above and the PoC link you shared with me in DMs.

The `SlotHistory` one is awful, but the others I think are good to have and make sense. Nice suggestion!

@ripatel-fd @2501babe what are we missing still in order to get this moving?

topointon-jump:
I think the differences of opinion are coming from _where_ the checks that the user is requesting a sane byte slice happen.

Instead of having a spec defined for each sysvar in the sycall itself, it might be better to move this logic into an SDK.

- The Get-Sysvar syscall should allow the user to request a byte slice of the bincode serialisation of the current sysvar cache's entry for a given sysvar (the same representation as in the actual account buffer).
- It is up to each client if this serialisation happens on-demand, or is cached.
- We can provide a simple wrapper SDK to allow users to safely access each sysvar, which provides convenience methods with sanity checks (e.g. getting a subset of the slot history, etc).
- The syscall itself should not check if the request is aligned to the data layout of the sysvar, this can be checked in the SDK.

buffalojoec:
After offline discussions with @2501babe and @topointon-jump, as well as the recent comments by @ripatel-fd and @topointon-jump, I've reverted the most recent commit outlining sysvar implementations of this syscall.

The way both of you have articulated the spec above makes sense, and is reflected in the proposal.

Basically, as mentioned above, we just need the syscall getter to be able to return some in-bound slice of any supported sysvar data (specifically listed in the proposal). Any caller implementation details should be left up to the SDK that defines them.

I've also added a commit attempting to simplify things and more closely align with your suggestions.

Let me know if I've missed anything.

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,203 @@
...
+
+```
+
+If the following conditions are not met, this syscall will return an error:
+
+- The sysvar data is available to read.
+- The sysvar data is not corrupt.
+- `offset` + `length` is in the range `[0, 2^64)`.
+- `var_addr` + `length` is in the range `[0, 2^64)`.
+- Each byte in the range starting at `var_addr` of size `length` is writable.
```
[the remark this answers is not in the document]
buffalojoec: > Sorry, I think I commented this before, but I can't find my comment anymore --
> 
> If the user passes a non-writable range or there is a overflow, the syscall should abort the VM, not return an error. (Which in turn aborts the transaction) This is different from graceful errors, where the syscall just returns non-zero and the VM can continue.

You wrote this on 133, but yes I can update this proposal to say the same thing.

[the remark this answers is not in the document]
buffalojoec: I can try to write it a bit more succinctly, but I like the overall approach.

I'm still not sure how you calculate the CU, as I wrote in my reply to you on 133.
Once I understand that better I can update the CU bits.

> - The syscall will always attempt to consume the same amount of CUs irrespective of control flow.

I'm assuming this means that even if we throw an error early on in the execution,
the same number of CUs are consumed as if the process finished successfully?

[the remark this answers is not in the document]
buffalojoec: Yeah agreed!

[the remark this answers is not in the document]
buffalojoec: @ripatel-fd I think CU calculation should be:

```
syscall_base + (32 / cpi_per_u) + (length / cpi_per_u)
```

Since we're reading the 32-byte sysvar ID and then returning a `length` slice of data. What do you think?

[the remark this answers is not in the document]
buffalojoec: Sorry, you seem to agree on the idea in 0127 but not here. These calculations are highly similar. Why use mem_ops.rs?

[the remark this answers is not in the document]
buffalojoec: Agave source doesn't do a great job of explaining why these CU values are what they are. For example, why is `syscall_base_cost` and `sysvar_base_cost` equal to `100` while `mem_op_base_cost` is `10`?

I'm assuming there's a reason the sysvar base is much higher than mem_op, so I'm going to use that as a safety net. If anyone explicitly disagrees I'll modify it (@ripatel-fd you said you don't _really_ care).

Thus, the CU calculation I've added is:

```
max(sysvar_base, (32 / cpi_per_u) + (length / cpi_per_u))
```

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,203 @@
...
+and therefore slows down _all_ memory accesses.
+
+## New Terminology
+
+N/A.
+
+## Detailed Design
+
+A single `sol_get_sysvar` syscall can be defined, which can be used by all
+forward-facing sysvar interfaces to retrieve data from any sysvar.
```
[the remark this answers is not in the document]
buffalojoec: CUs added.

proposals/0127-get-sysvar-syscall.md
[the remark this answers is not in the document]
buffalojoec: OK here's where we landed, no givesies-backsies.
```
syscall_base + (32 / cpi_per_u) + max(mem_op_base, (length / cpi_per_u))
```

buffalojoec:
> Approving (can't send an actual approval because the PR is still under my name)

I guess I'll proxy your approval through me?

buffalojoec (approved):

topointon-jump (approved):

2501babe:
will go through the spec again with latest changes to make sure it conforms, but i have what i think is a final impl for the agave version of this, if anyone would like to inspect: https://github.com/anza-xyz/agave/pull/560

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,227 @@
...
+The syscall returns `2` if `offset + length` is greater than the size of the
+sysvar.
+
+Otherwise, the syscall writes the sysvar data to the virtual memory at
+`var_addr` and returns `0`.
+
+### Compute Unit Usage
+
+The syscall will always attempt to consume the same amount of CUs regardless of
+control flow.
```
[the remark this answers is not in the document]
buffalojoec: Sorry, forgot you requested this change!

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,227 @@
...
+### Supported Sysvars
+
+The proposed unified syscall interface will support all non-deprecated sysvars
+in the Sysvar Cache.
+
+Unless specified otherwise, any new sysvars added to the Sysvar Cache in the
+future also become accessible through this syscall.
+
+Existing syscalls to fetch sysvar data will be deprecated in favor of the new
+unified syscall proposed herein.
```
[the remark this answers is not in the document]
buffalojoec: We can move this to a future SIMD, yeah. I think maybe just dropping the line altogether is more appropriate than adding "discouraged" imo.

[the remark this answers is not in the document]
Lichtso: We have a mechanism in which we forbid new deployments to use certain syscalls but continue to support existing programs using them. See e.g.: https://github.com/anza-xyz/agave/blob/592107a907df841169f29a25b34799fd7ac9a2c9/programs/bpf_loader/src/syscalls/mod.rs#L275

And I am in favor of deprecating all other sysvar related syscalls that way.

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,227 @@
...
+can be accessed using the new syscall. For example, multiple SDKs may choose to
+allow users to access one element from a list-based sysvar at a time, while only
+a few may offer lookups or binary searches on the data.
+
+## Impact
+
+As mentioned as partial motivation for this proposal, the new syscall would
+make changing and adding new sysvars in the future much easier, which would be
+positively impactful to contributors.
+
```
Lichtso, line 212: Another minor and not so obvious impact is that this changes the verification of supported sysvar related syscalls from verification time to runtime.

proposals/0127-get-sysvar-syscall.md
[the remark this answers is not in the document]
buffalojoec: I took this line out. Deprecating seems delicate considering existing programs, and I think, based on all above comments, it falls outside the scope of this SIMD.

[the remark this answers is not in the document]
Lichtso: As explained above, deprecating in this context does not mean rug-pulling existing programs but just no new deployments of programs which contain deprecated syscalls.

proposals/0127-get-sysvar-syscall.md
buffalojoec: Added a note on this in the section you've commented.

buffalojoec:
Hey guys, @lheeger-jump and @ripatel-fd I'm going to need you guys to agree on the Sysvar Cache piece, since determining whether it's implementation- or spec- level is kind of up to Firedancer atm. Agave has to do it either way.

Also @Lichtso let me know if there's anything else you need for approval.

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,227 @@
...
+The syscall aborts the virtual machine if any of these conditions are true:
+
+- Not all bytes in VM memory range `[sysvar_id, sysvar_id + 32)` are readable.
+- Not all bytes in VM memory range `[var_addr, var_addr + length)` are writable.
+- `offset + length` is not in `[0, 2^64)`.
+- `var_addr + length` is not in `[0, 2^64)`.
+- Compute budget is exceeded.
+
+The syscall returns `1` if the sysvar is not present in the Sysvar Cache.
```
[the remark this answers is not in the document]
lheeger-jump: Its just a list of sysvars. That list of sysvars is specified here. It is confusing to use the term Sysvar Cache when really what is meant is "sysvars supported by this syscall" which is more direct and more explicit given the language in the spec. No one has defined sysvar cache before this SIMD, and this SIMD hardly defines it. No spec or doc provides any color on this either. Lets keep it simple and unambiguous: 

1. the list of supported sysvars is the list here in this SIMD.
2. any new sysvar is included in this list of supported sysvars by default, unless otherwise specified.

proposals/0127-get-sysvar-syscall.md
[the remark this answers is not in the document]
lheeger-jump: Deprecating is a separate SIMD. its an easy one to justify once this lands.

lheeger-jump:
Added another comment here. https://github.com/solana-foundation/solana-improvement-documents/pull/127#discussion_r1564868321

lheeger-jump (dismissed):
Approved.

I am not super happy with the language around the sysvar cache. It implies a nuanced difference in the encoding of sysvars (which also lacks specification for some). In the future, we should probably describe, via specification, the difference between the sysvar accounts and their sysvar cache which is highly nuanced.

lheeger-jump:
> In the future, we should probably describe, via specification, the difference between the sysvar accounts and their sysvar cache which is highly nuanced.

To be clear, this is out of scope for this SIMD

buffalojoec (dismissed):
@lheeger-jump I agree we definitely should specify that.

In the meantime I'm approving here on behalf of myself and @ripatel-fd, who originally proposed this idea and has also expressed approval (but can't). I think for a full Anza approval we need @Lichtso to stamp.

Lichtso (approved):

buffalojoec:
Let's leave this open for any remaining comments and merge on Friday 4/19.

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,233 @@
...
+
+The syscall returns `2` if `offset + length` is greater than the size of the
+sysvar.
```
2501babe, line 161: if we are introducing a distinction between hard and soft errors (successful failures?) we need to specify what order the checks happen in, because a set of inputs that could trigger both a soft and a hard error needs to trigger the same one on all impls

im inclined to want to check the _soft_ conditions first, and bail with the appropriate code if they are met, because there is no possibility of the language runtime accidentally returning them later

the more i think about this tho the more worried i get about the possibility of an attacker crafting a transaction where writing to the output buffer triggers a segfault on one validator impl but not the other

2501babe: incidentally im assuming this means that a `length` of 0 should noop and return `Ok(0)`

buffalojoec: The intention by @ripatel-fd was to have "graceful errors" that won't abort the VM, but maybe there's a case to be made to get rid of these. I believe the intention was to handle a scenario where an on-chain program may use the syscall directly (without the handler) and give them the ability to gracefully handle errors, rather than the VM panicking.

https://github.com/solana-foundation/solana-improvement-documents/pull/127#discussion_r1544351726

However, I'm pretty sure we dove into the error propagation after this discussion, and if a syscall throws an `InstructionError`, that will bubble up to a `ProgramError` in the executing program, so perhaps _all_ of these syscall errors are actually graceful?

https://discord.com/channels/428295358100013066/967028962746327060/1222861983842570294

2501babe: oh to be clear my preferred solution would be to redefine the nonzero return cases to also abort, unless theres a concrete usecase we are hoping to support with this

buffalojoec: Right I agree on condensing the control flow, but I'm saying I don't think it actually aborts. I think the error bubbles up. It would have to panic to abort.

2501babe: as far as i can tell, returning an error does abort unconditionally, after testing with `Clock::get` this afternoon and again now. this must happen somewhere inside solana-rbpf but because of that fact plus ffi, macros, the separation between syscall registration and invocation, and thiserror, it is too difficult for me to untangle and locate the place where this actually occurs

what i do know is `solana_program::syscalls::sol_get_clock_sysvar` returns a `u64` rather than a `Result`, and control never resumes inside the solana-program `Sysvar::get` impl if you return an `Err` inside `SyscallGetClockSysvar`

@Lichtso could you enlighten us?

buffalojoec: So, @2501babe you're correct that syscall errors abort the VM. I just ran a little test to investigate further. Turns out I already ran this test, lol.

Here's the block of code in question.

https://github.com/anza-xyz/agave/blob/297a7aa40a42d5dad4002d11da10ef3748876041/program-runtime/src/invoke_context.rs#L514-L534

```rust
let result = match vm.program_result {
    ProgramResult::Ok(_) => {
        stable_log::program_success(&logger, &program_id);
        Ok(())
    }
    ProgramResult::Err(ref err) => {
        if let EbpfError::SyscallError(syscall_error) = err {
            if let Some(instruction_err) = syscall_error.downcast_ref::<InstructionError>()
            {
                stable_log::program_failure(&logger, &program_id, instruction_err);
                Err(instruction_err.clone())
            } else {
                stable_log::program_failure(&logger, &program_id, syscall_error);
                Err(InstructionError::ProgramFailedToComplete)
            }
        } else {
            stable_log::program_failure(&logger, &program_id, err);
            Err(InstructionError::ProgramFailedToComplete)
        }
    }
};
```

The `vm.program_result` is a `StableResult<u64, EbpfError>`. As you can see from the snippet, the program's result can be any `Ok(u64)`, but we bury that `u64` value. We only care that it was successful.

The snippet also shows that valid `InstructionError` values can be bubbled up, although **the VM is still aborted**.

It's not possible to catch a syscall error, even if it is a valid `InstructionError`.

```rust
fn process_instruction(
    _program_id: &Pubkey,
    accounts: &[AccountInfo],
    _instruction_data: &[u8],
) -> ProgramResult {
    let mut accounts_iter = accounts.iter();

    let vote_account = next_account_info(&mut accounts_iter)?;

    msg!("Log!");

    if let Err(e) = get_epoch_stake(vote_account.key) {
        msg!("Error: {:?}", e);
    }

    msg!("Another log!");

    // Should always return Ok
    Ok(())
}
```

The reason the syscall definitions and the program stubs define the syscall as returning a `u64`, is because it _may_ return an `InstructionError` that can be bubbled up to the client, such as a `BanksClient` or `RpcClient`, but the VM has aborted nonetheless.

I guess it means we can use `InstructionError` to define specific error codes from within the syscall, if we want. Anything that throws an `EbfError`, such as [`translate`](https://github.com/anza-xyz/agave/blob/033bcfcab5696839c0ad10486ddf4c85fc73358b/programs/bpf_loader/src/syscalls/mod.rs#L480-L490) - even if it's an `EbfError::SyscallError`, is going to route to `InstructionError::ProgramFailedToComplete`.

buffalojoec: With _all of that_ being said, @ripatel-fd was making the case of gracefully returning an error, which can only be done by doing `Ok(1)`, `Ok(2)`, etc.

With that in mind, ordering shouldn't _really_ matter, since anything that throws within the syscall is automatically an abort, while returned `Ok(..)` values are returned. The handler is then responsible for translating those `Ok(T)` values into `Ok(bytes)` or `Err(ProgramError::V)`.

buffalojoec: In that spirit, I think the specification is accurate.

2501babe: returning `Ok(1)` doesnt halt execution tho. so if there is a case where one impl hits an `Err` whereas the other hits one of the `Ok` cases first, there would be a consensus failure. the order has to be de defined or else the `Ok` errors have to change to `Err`

buffalojoec: Ahhh, I see what you're saying. I think it probably makes sense to say if you provide a condition that should abort the VM, then the VM should be aborted, period.

In other words, VM violations should take precedence. One should only receive a graceful error if the VM wouldn't be violated otherwise. Open to discussion that, though.

buffalojoec: If you hate what I just added, then I can do pseudocode, but it seemed sufficient and concise.

buffalojoec: We were planning to merge today, and looks like based on above there's acceptance of what I suggested, then committed?
@2501babe @ripatel-fd

riptl (approved):
(From my alt account) @buffalojoec Looks good to me!

proposals/0127-get-sysvar-syscall.md
```diff
@@ -155,13 +155,15 @@ The syscall aborts the virtual machine if any of these conditions are true:
 - `var_addr + length` is not in `[0, 2^64)`.
 - Compute budget is exceeded.
 
-The syscall returns `1` if the sysvar is not present in the Sysvar Cache.
+All VM violations (above) are checked first, before attempting to return
+gracefully.
```
lheeger-jump, line 159:
```suggestion
All VM violations (above) are checked first, before to returning one of the following error codes. The following checks are completed in the order they appear below.
```

lheeger-jump (changes requested):
Small change to wording

proposals/0127-get-sysvar-syscall.md
buffalojoec: Note I flipped these around since we can't know the length of the sysvar data if it's not found.

topointon-jump (approved):

riptl (approved):

proposals/0127-get-sysvar-syscall.md
2501babe: think we got ourselves a simd!

2501babe (approved):

buffalojoec (approved):

jacobcreech (approved):
Looks like we've got consensus and it is now Monday EOD. Merging. Thanks to all for reviewing and getting this through!

2501babe:
i hate to be That Person, but i was triple-checking my implementation of this and had not realized that at some point `SlotHistory` was added to the list of supported sysvars. this sysvar is not presently part of the sysvar cache and is used by no native programs, only bank and consensus. after discussing with joe, we both believe it should be removed from the list as it is superfluous

does anyone object? and is there a process we should go through to amend this without another whole simd?

proposals/0127-get-sysvar-syscall.md
```diff
@@ -0,0 +1,236 @@
...
+
+### Compute Unit Usage
+
+The syscall will always consume the same amount of CUs regardless of
+control flow.
+
+CU usage is proportional to the `length` parameter.
+
+```
+sysvar_base + (32 / cpi_per_u) + max(mem_op_base, (length / cpi_per_u))
```
joncinque, line 177: Sorry for the flyby here, but I was looking at this implementation in Agave and was curious about the `(32 / cpi_bytes_per_unit)` term here. `cpi_bytes_per_unit` is currently set to `250` on mainnet, so that factor will always be 0.

It's not a big deal, and thankfully it's explicitly defined, but was there a reason to keep it separate rather than combining it with `length`? Ie instead of doing:

```
sysvar_base + max(mem_op_base, ((length + 32) / cpi_per_u))
```

buffalojoec: The intention was to apply `cpi_per_u` to the 32-byte sysvar ID, then to the length (or mem_op_base, whichever is max). So with that being said, what you've suggested would probably make more sense. Should be adjust this?

joncinque: Since this syscall is part of 2.0, which is really close to mainnet, it might be a bit late to do a quick change. That would probably require an amendment SIMD + new feature gate, which sounds like a lot of work for what's essentially a rounding difference of 1 CU on a syscall :sweat_smile: I think we can leave as is for now. Thanks for the response!

ZKcash-IrishGov: {
  "metadata": {
    "version": "0.0.1",
    "origin": "shank"
  },
  "additionalPrograms": ["proposals/0042-svm-fraud-proofs.md"],
  "standard": "codama",
  "version": "1.0.0"
}

Angelaa0609 (approved):

Angelaa0609 (approved):

PR #72, opened 2023-10-12
feature-gate-threshold-automation init

buffalojoec:
This SIMD outlines a proposal for automating the feature activation process
based on a stake-weighted support threshold, rather than manual human action.

PR #133, opened 2024-03-26
SIMD-0133: Syscall: Get Epoch Stake

buffalojoec:
A syscall to retrieve a vote account's delegated stake for the current epoch.

Currently, on-chain programs have no knowledge of the current epoch's stake and
how much active stake is delegated to a certain vote account.

If this data was available for querying by on-chain programs, it would unblock
many use cases, such as validator governance and secondary consensus mechanisms,
that were previously not possible on Solana.

Additionally, this would enable the Feature Gate program defined in
[SIMD 0089](https://github.com/solana-foundation/solana-improvement-documents/pull/72) to tally vote account stake in
support for a pending feature gate.

### DIFF
--- a/proposals/0133-syscall-get-epoch-stake.md
+++ b/proposals/0133-syscall-get-epoch-stake.md
@@ -79,6 +79,18 @@ violation.
 If the provided vote address corresponds to an account that is not a vote
 account or does not exist, the syscall will write `0` for active stake. 
 
+### Compute Unit Usage
+
+The syscall will always attempt to consume the same amount of CUs regardless of
+control flow.
+
+```
+(32 / cpi_per_u) + (8 / cpi_per_u)
+```
+
+- `syscall_base`: Base cost of a sysvall.
+- `cpi_per_u`: Number of account data bytes per CU charged during CPI.
+
 ## Impact
 
 Dapp developers will be able to query vote account stake for the current epoch
--- a/proposals/0133-syscall-get-epoch-stake.md
+++ b/proposals/0133-syscall-get-epoch-stake.md
@@ -80,15 +80,6 @@ an access violation.
 If the provided vote address corresponds to an account that is not a vote
 account or does not exist, the syscall will write `0` for active stake. 
 
-The `InvokeContext` must be modified to store a pointer to the epoch's current
-stake, which will be used to obtain the query response for the syscall. The
-epoch's current stake should be a collection of fixed-size elements of the
-following structure:
-
-- 32 bytes for the vote address.
-- 8 bytes for the delegated stake.
-
-
 ## Impact
 
 This new syscall directly unlocks highly relevant network data for a wide range
--- a/proposals/0133-syscall-get-epoch-stake.md
+++ b/proposals/0133-syscall-get-epoch-stake.md
@@ -1,0 +1,106 @@
+---
+simd: '0133'
+title: Syscall Get-Epoch-Stake
+authors:
+  - Joe Caulfield (Anza)
+category: Standard
+type: Core
+status: Draft
+created: 2024-03-25
+feature: (fill in with feature tracking issues once accepted)
+---
+
+## Summary
+
+A syscall to retrieve a vote account's delegated stake for the current epoch.
+
+## Motivation
+
+Currently, on-chain programs have no knowledge of the current epoch's stake and
+how much active stake is delegated to a certain vote account.
+
+If this data was available for querying by on-chain programs, it would unblock
+many use cases, such as validator governance and secondary consensus mechanisms,
+that were previously not possible on Solana.
+
+Additionally, this would enable the Feature Gate program defined in
+[SIMD 0089](./0089-programify-feature-gate.md) to tally vote account stake in
+support for a pending feature gate.
+
+## Alternatives Considered
+
+An alternative design has been proposed in the past for using an on-chain sysvar
+account to store all of the current epoch's stake. It was proposed in
+[SIMD 0056](https://github.com/solana-foundation/solana-improvement-documents/pull/56).
+
+Using a sysvar account to store even just the current epoch's stake introduces a
+size limitation on the number of entries that can be stored, which the number of
+validators on Solana could surpass in the future.
+
+By offering access to this information through only a syscall, we can avoid this
+account maximum size constraint. However, the syscall approach also makes the
+data more cumbersome to retrieve off-chain than the sysvar approach.
+
+While retrieving this data off-chain may not be straightforward, it remains
+entirely feasible. Conversely, on-chain programs currently do not have this
+capability at all.
+
+## New Terminology
+
+N/A.
+
+## Detailed Design
+
+The specification for the proposed syscall is as follows:
+
+```c
+/**
+ * Retrieves the total active stake delegated to a vote account for the current
+ * epoch.
+ *
+ * @param var_addr      VM memory address to copy the retrieved data to.
+ * @param vote_address  The vote account whose stake to query.
+ * @return              A 64-bit unsigned integer error code:
+ *                        - 0 if the operation is successful.
+ *                        - Non-zero error code.
+ *
+ * If the operation is not successful, data will not be written to the
+ * provided VM memory address.
+ */
+uint64_t sol_get_epoch_stake(
+  /* r1 */ uint8_t *    var_addr,
+  /* r2 */ void const * vote_address,
+);
+```
+
+`var_addr` must be the starting address of at least 8 bytes of writable VM
+memory to store the `u64` response. If not, the syscall will return an error.
+
+If the provided vote address corresponds to an account that is not a vote
+account or does not exist, the syscall will write `0` for active stake. 
+
+The `InvokeContext` must be modified to store a pointer to the epoch's current
+stake, which will be used to obtain the query response for the syscall. The
+epoch's current stake should be a collection of fixed-size elements of the
+following structure:
+
+- 32 bytes for the vote address.
+- 8 bytes for the delegated stake.
+
+
+## Impact
+
+This new syscall directly unlocks highly relevant network data for a wide range
+of protocols. Developers seeking to access vote account stake will be positively
+impacted.
+
+## Security Considerations
+
+This new syscall introduces the same security considerations as the rest of the
+syscalls in the existing interface, which manipulate raw pointers to VM memory
+and must be implemented with care.
+
+A potential pitfall in the implementation could come when comparing the act of
+writing a `0` to the VM memory for active stake versus returning a `0` for
+success. The two should not be conflated.
+
--- a/proposals/0133-syscall-get-epoch-stake.md
+++ b/proposals/0133-syscall-get-epoch-stake.md
@@ -65,9 +65,6 @@ The specification for the proposed syscall is as follows:
 uint64_t sol_get_epoch_stake(/* r1 */ void const * vote_addr);
 ```
 
-If the provided vote address corresponds to an account that is not a vote
-account or does not exist, the syscall will write `0` for active stake. 
-
 ### Control Flow
 
 The syscall aborts the virtual machine if not all bytes in VM memory range
--- a/proposals/0133-syscall-get-epoch-stake.md
+++ b/proposals/0133-syscall-get-epoch-stake.md
@@ -82,7 +82,7 @@ The syscall will always attempt to consume the same amount of CUs regardless of
 control flow.
 
 ```
-syscall_base + (32/cpi_per_u) + mem_op_base
+syscall_base + floor(32/cpi_bytes_per_unit) + mem_op_base
 ```
 
 - `syscall_base`: Base cost of a sysvall.
--- a/proposals/0133-syscall-get-epoch-stake.md
+++ b/proposals/0133-syscall-get-epoch-stake.md
@@ -74,7 +74,8 @@ uint64_t sol_get_epoch_stake(
 ```
 
 `var_addr` must be the starting address of at least 8 bytes of writable VM
-memory to store the `u64` response. If not, the syscall will return an error.
+memory to store the `u64` response. If not, the syscall will abort the VM with
+an access violation.
 
 If the provided vote address corresponds to an account that is not a vote
 account or does not exist, the syscall will write `0` for active stake. 
--- a/proposals/0133-syscall-get-epoch-stake.md
+++ b/proposals/0133-syscall-get-epoch-stake.md
@@ -57,27 +57,27 @@ The specification for the proposed syscall is as follows:
  * Retrieves the total active stake delegated to a vote account for the current
  * epoch.
  *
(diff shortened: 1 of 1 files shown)


proposals/syscall-get-epoch-stake.md
```diff
@@ -0,0 +1,106 @@
...
+ * provided VM memory address.
+ */
+uint64_t sol_get_epoch_stake(
+  /* r1 */ uint8_t *    var_addr,
+  /* r2 */ void const * vote_address,
+);
+```
+
+`var_addr` must be the starting address of at least 8 bytes of writable VM
+memory to store the `u64` response. If not, the syscall will return an error.
```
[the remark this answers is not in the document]
buffalojoec: Yeah, this happens within the [`translate`](https://github.com/anza-xyz/agave/blob/24c55f39ee75c4a64568a31088ad33129b15a70f/programs/bpf_loader/src/syscalls/mod.rs#L480-L490) helpers shared across most syscalls. I can explicitly mention it in the proposal, sure.

proposals/syscall-get-epoch-stake.md
```diff
@@ -0,0 +1,106 @@
...
+If the provided vote address corresponds to an account that is not a vote
+account or does not exist, the syscall will write `0` for active stake. 
+
+The `InvokeContext` must be modified to store a pointer to the epoch's current
+stake, which will be used to obtain the query response for the syscall. The
+epoch's current stake should be a collection of fixed-size elements of the
+following structure:
+
+- 32 bytes for the vote address.
+- 8 bytes for the delegated stake.
```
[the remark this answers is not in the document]
buffalojoec: Okay, it's not part of the transaction processor and `InvokeContext` is internal, so maybe it is implementation-level. If this bit is irrelevant for FD I can take it out and consider it Agave-only.

buffalojoec:
> Looks great. This is a highly requested feature. I suggested two changes

Changes addressed!

proposals/0133-syscall-get-epoch-stake.md
```diff
@@ -0,0 +1,98 @@
...
+size limitation on the number of entries that can be stored, which the number of
+validators on Solana could surpass in the future.
+
+By offering access to this information through only a syscall, we can avoid this
+account maximum size constraint. However, the syscall approach also makes the
+data more cumbersome to retrieve off-chain than the sysvar approach.
+
+While retrieving this data off-chain may not be straightforward, it remains
+entirely feasible. Conversely, on-chain programs currently do not have this
+capability at all.
```
CriesofCarrots, line 46:
```suggestion
Exposing epoch-stake information through a syscall avoids this
account maximum size constraint. While the syscall approach does not offer the 
easy off-chain retrieval of a sysvar, there are existing ways
to get epoch-stake data off chain. The priority of a new design should be making the data available to on-chain programs.
```
I think this is what you're trying to say.

CriesofCarrots (commented):
Mostly suggestions for conciseness and clarity.

proposals/0133-syscall-get-epoch-stake.md
buffalojoec: Yeah, exactly, thanks.

proposals/0133-syscall-get-epoch-stake.md
```diff
@@ -0,0 +1,98 @@
...
+
+The specification for the proposed syscall is as follows:
+
+```c
+/**
+ * Retrieves the total active stake delegated to a vote account for the current
+ * epoch.
+ *
+ * @param var_addr      VM memory address to copy the retrieved data to.
+ * @param vote_address  The vote account whose stake to query.
```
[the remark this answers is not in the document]
buffalojoec: Lmk if what I've updated the parameter definition with makes sense, or if you were thinking we should change the type from being a pointer.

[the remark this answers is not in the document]
CriesofCarrots: It looks like there are examples of both approaches to passing Pubkeys in existing syscalls, so I'm fine with the pointer.

proposals/0133-syscall-get-epoch-stake.md
```diff
@@ -0,0 +1,96 @@
...
+);
+```
+
+`var_addr` must be the starting address of 8 bytes of writable VM memory to
+store the `u64` response. If not, the syscall will abort the VM with an access
+violation.
+
+If the provided vote address corresponds to an account that is not a vote
+account or does not exist, the syscall will write `0` for active stake. 
+
```
lheeger-jump, line 81:
```suggestion
account or does not exist, the syscall will write `0` for active stake. 

This syscall consumes `syscall_base_cost` compute units always.

```

lheeger-jump (approved):
Please include the one change. Otherwise, approved.

proposals/0133-syscall-get-epoch-stake.md
buffalojoec: Seems like it should maybe be `syscall_base_cost + cost_of(u64)`.

Currently, we charge `syscall_base_cost + cost_of(T)` to write sysvar `T` to VM memory.. Should this `u64` be the same?

proposals/0133-syscall-get-epoch-stake.md
```diff
@@ -0,0 +1,96 @@
...
+ *                        - 0 if the operation is successful.
+ *                        - Non-zero error code.
+ *
+ * If the operation is not successful, data will not be written to the
+ * provided VM memory address.
+ */
+uint64_t sol_get_epoch_stake(
+  /* r1 */ uint8_t *    var_addr,
+  /* r2 */ void const * vote_address,
+);
```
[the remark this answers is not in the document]
buffalojoec: It does not break any existing patterns for existing syscalls, since they all seem to do
things differently.

For example, many cryptographic syscalls - such as `sol_curve_multiscalar_mul` -
use `1` to indicate an `Option::None` response from curve calculation.

https://github.com/anza-xyz/agave/blob/b1e17998b64203fe62f52efe3fe5ba87706ed446/programs/bpf_loader/src/syscalls/mod.rs#L1180-L1189

https://github.com/anza-xyz/agave/blob/b1e17998b64203fe62f52efe3fe5ba87706ed446/zk-token-sdk/src/curve25519/ristretto.rs#L227-L242

While others - such as `sol_secp256k1_recover` - use the integer value to map specific errors,
thrown intentionally based on computation. Note these are separate from errors such as memory
access or unavailable data (like in sysvars).

https://github.com/anza-xyz/agave/blob/b1e17998b64203fe62f52efe3fe5ba87706ed446/programs/bpf_loader/src/syscalls/mod.rs#L842-L853

https://github.com/anza-xyz/agave/blob/b1e17998b64203fe62f52efe3fe5ba87706ed446/sdk/program/src/secp256k1_recover.rs#L409-L422

Finally, a handful of syscalls - such as `sol_remaining_compute_units` simply return the
value in the resulting `u64`, as you've mentioned.

https://github.com/anza-xyz/agave/blob/b1e17998b64203fe62f52efe3fe5ba87706ed446/programs/bpf_loader/src/syscalls/mod.rs#L1735-L1753

proposals/0133-syscall-get-epoch-stake.md
buffalojoec: Gotcha. I was taking a look at sysvars, which is `sysvar_base` plus one CU per byte written.
https://github.com/solana-labs/solana/blob/e2d34d375bad87a95cdfbe4b07163bb13515bfeb/programs/bpf_loader/src/syscalls/sysvar.rs#L10-L16

Note `sysvar_base` and `syscall_base` are the same (`100`) right now, for what it's worth.

I was also checking out some that use length, like `set_return_data`.
https://github.com/solana-labs/solana/blob/e2d34d375bad87a95cdfbe4b07163bb13515bfeb/programs/bpf_loader/src/syscalls/mod.rs#L1249-L1255

I think the latter makes the most sense for 127. Here we can use a similar concept but make things static since it's always going to read a 32-byte address and return an 8-byte integer. Maybe:

```
syscall_base + (32 / cpi_per_u) + (8 / cpi_per_u)
```

buffalojoec: Just wanted to point out that as per SIMD 0127, this calculation _could_ be:

```
max(sysvar_base, (32 / cpi_per_u) + (length / cpi_per_u))
```

But in this case length is always `8`, so we'd get `max(100, 0.16)`, so we can just charge `(32 / cpi_per_u) + (length / cpi_per_u)` constantly.

buffalojoec:
> Needs CU documentation

CUs added.

proposals/0133-syscall-get-epoch-stake.md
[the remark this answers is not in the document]
buffalojoec: I've added a commit changing this syscall to return the data directly, in the interest of saving a memory access.

proposals/0133-syscall-get-epoch-stake.md
buffalojoec: With the above being said, as well as https://github.com/solana-foundation/solana-improvement-documents/pull/127#discussion_r1546374161:

```
syscall_base + (32/cpi_per_u) + mem_op_base
```

proposals/0133-syscall-get-epoch-stake.md
```diff
@@ -0,0 +1,104 @@
...
+### Control Flow
+
+The syscall aborts the virtual machine if not all bytes in VM memory range
+`[vote_addr, vote_addr + 32)` are readable.
+
+Otherwise, the syscall returns a `u64` integer representing the total active
+stake delegated to the vote account at the provided address.
+
+If the provided vote address corresponds to an account that is not a vote
+account or does not exist, the syscall will return `0` for active stake.
```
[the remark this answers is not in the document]
buffalojoec: Thanks. Removed.

proposals/0133-syscall-get-epoch-stake.md
```diff
@@ -57,27 +57,36 @@ The specification for the proposed syscall is as follows:
...
-account or does not exist, the syscall will write `0` for active stake. 
+account or does not exist, the syscall will return `0` for active stake.
+
+### Compute Unit Usage
+
+The syscall will always attempt to consume the same amount of CUs regardless of
+control flow.
+
+```
+syscall_base + (32/cpi_per_u) + mem_op_base
```
lheeger-jump, line 85:
```suggestion
syscall_base + floor(32/cpi_bytes_per_unit) + mem_op_base
```

proposals/0133-syscall-get-epoch-stake.md
```diff
@@ -57,27 +57,36 @@ The specification for the proposed syscall is as follows:
...
+```
+
+- `syscall_base`: Base cost of a sysvall.
+- `cpi_per_u`: Number of account data bytes per CU charged during CPI.
```
lheeger-jump, line 89:
```suggestion
- `cpi_bytes_per_units`: Number of account data bytes per CU charged during CPI.
- `mem_op_base`: Base cost of a memory operation syscall.
```

lheeger-jump (changes requested):
Small edits

CriesofCarrots (approved):

lheeger-jump (approved):

lheeger-jump:
We will merge this SIMD if there are no more comments by April 12th.

buffalojoec:
> @michaelh-laine That's a great point. I suggest we send an amendment SIMD that specifies that `sol_get_epoch_stake` returns the total active stake if a null pointer is given as the `vote_addr` argument. @CriesofCarrots wdyt?

The concept of being able to obtain the total epoch stake is salient, but I think this proposed solution sounds like overloading the syscall. Personally I'd rather see two syscalls for this, one for a vote account, one for total stake.

CriesofCarrots:
I actually like @ripatel-fd 's idea; using a null pointer to return total epoch stake, vs filtering with a provided vote address, makes semantic sense to me.

Side note: if we amend this SIMD, let's please clarify the `32` magic number in the CU calculation as the length of a vote address (and I suppose if we pursue the above idea, we need to decide how to handle compute for the null case).

buffalojoec:
> @CriesofCarrots @buffalojoec Which one of us should write the amend SIMD? I'm happy to do a first pass and then hand it off. Or before we start writing, should we get on a call and agree how the "get total" API should look like first?

https://github.com/buffalojoec/solana-improvement-documents/commit/6ce1bd995949e12335b1fcac250095a902d6d6cb

CriesofCarrots:
> [buffalojoec@6ce1bd9](https://github.com/buffalojoec/solana-improvement-documents/commit/6ce1bd995949e12335b1fcac250095a902d6d6cb)

Thanks, @buffalojoec, that is looking great to me . Let's get it PRed and reviewed. I think since SIMD-0133 is still a draft, we can just amend it in place, rather than creating a superseding SIMD. ...at least, that's my vote.

michaelh-laine:
Thanks for picking up this suggestion so quickly everyone!

hussein-aitlahcen:
Quick question regarding this: from my understanding, we have both the current and next epoch stakes available as the schedule for epoch N+1 is set once we cross epoch N, why no providing two sysvars, one for the current and next epoch. That would allow trivial implementation of light clients by aggregating the next epoch vote accounts and their stake on a program and proving that the current vote accounts signed the block containing this program (after having aggregated)? As far as I know, there is currently no way to compute the next validator set on-chain and have it signed by the current validator set (vote accounts), hence light clients are pretty difficult to build because you'd have to prove all stake accounts for epoch N to know the validator set for N+1?

You can imagine a program where you do:
```
assert(state.expected_epoch = clock.epoch);
validator_stake = get_next_epoch_stake(vote_account)
assert(!pda_exists(validator_pda(vote_account, epoch)));
init_pda(validator_pda(vote_account, epoch));
assert(validator_stake <= state.highest_state);
state.chain = hash(state.chain, (vote_account, validator_stake))
state.total_stakes += validator_stake
if(state.total_stakes = get_stake(null)))
  state.expected_epoch++
```

You would be able to prove this off-chain with a merkle inclusion proof of the block where the final validator has been aggregated in `state.chain`, where the final result for each epoch is `h(...h(h(null, (v_0, s_0)), (v_1, s_1))..., (v_n, s_n))`.
With the current validator set being around 1.5K, that would mean 1.5K tx per epoch which is pretty decent, the 1.5K pda for each epoch could be wiped on the valset has been accumulated on `state.chain`.

buffalojoec:
> Quick question regarding this: from my understanding, we have both the current and next epoch stakes available as the schedule for epoch N+1 is set once we cross epoch N, why no providing two sysvars, one for the current and one for the next epoch?

Interesting idea! I think since this SIMD has been implemented and feature-gated, the best direction for your suggested improvements would be to discuss this in channels like the Solana Tech Discord, or propose a new SIMD that expands on the syscall introduced by this SIMD, where your suggested changes are detailed.

PR #137, opened 2024-04-08
SIMD-0137: curve25519 syscalls - immediately abort on unsupported ids

samkim-crypto:
Currently, there is inconsistency in the way unsupported curve id's are handled in curve25519 and alt_bn128 syscall functions.
- curve25519 syscalls: on unsupported curve id, returns with error code 1 (`Ok(1)` in rust)
- alt_bn128 syscalls: on unsupported curve id, immediately aborts (`Err(SyscallError::InvalidAttribute)` in rust)

This proposal suggests to update the curve25519 syscalls to also immediately abort since invoking the syscalls on an unsupported curve id should be an unrecoverable error.

### DIFF
--- a/proposals/0137-curve25519-error.md
+++ b/proposals/0137-curve25519-error.md
@@ -1,6 +1,6 @@
 ---
-simd: "0134"
-title: Curve25519 - Abort on Unsupported Curve
+simd: "0137"
+title: EC Syscalls - Abort on Unsupported Curve/Ops
 authors:
   - Sam Kim
 category: Standard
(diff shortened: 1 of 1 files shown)


0x0ece:
I'd probably clarify both unknown curve and unknown operation should be unrecoverable errors.
Other than that looks good to me.

alessandrod (approved):

0xSol:
@lheeger-jump or @ripatel-fd,   please review / comment / approve. Thanks.

PR #144, opened 2024-04-16
Amend 0088 to use buffer account

buffalojoec:
#### Problem
During implementation of #88, Anza and Firedancer contributors discussed the 
benefits of using a buffer account to store the BPF implementation of a builtin 
program, rather than a deployed upgradeable program.

This protects builtin migration preparation from unexpected issues, such as 
those that may arise from attackers invoking the source program before it's 
moved in place of the target builtin.

Additionally, we wanted to take the opportunity to specify the **delayed
visibility slot** involved in migration.

#### Summary of Changes
Revise the SIMD to specify using buffer accounts as well as the one-slot delayed 
visibility during migration.

### DIFF
--- a/proposals/0088-enable-core-bpf-programs.md
+++ b/proposals/0088-enable-core-bpf-programs.md
@@ -68,6 +68,11 @@ upgrade authority and the ELF bytes of its BPF implementation. A feature gate
 is used to create the BPF program accounts, replacing the existing native
 program at its original address with a legitimate BPF Upgradeable program.
 
+In the slot immediately following the feature activation, the program will *not*
+be invocable. It will be marked as `DelayedVisibility` in the program cache.
+This status will last one slot, then the program will return to being fully
+operational.
+
 No program IDs for existing native programs are changed by this migration
 process.
 
--- a/proposals/0088-enable-core-bpf-programs.md
+++ b/proposals/0088-enable-core-bpf-programs.md
@@ -62,24 +62,27 @@ published outlining at least the following details:
   program
 - How this changes to this program will be managed after it becomes BPF
 
-**Migrating a native program to core BPF** shall consist of deploying a
-BPF version of the native program to a new arbitrary address and using a
-feature gate to move the BPF program, replacing the existing native program at
-its original program address. No program IDs for existing native programs would
-be changed by this migration process.
+**Migrating a native program to core BPF** shall consist of creating a buffer
+account, owned by the BPF Upgradeable Loader, containing the program's desired
+upgrade authority and the ELF bytes of its BPF implementation. A feature gate
+is used to create the BPF program accounts, replacing the existing native
+program at its original address with a legitimate BPF Upgradeable program.
+
+No program IDs for existing native programs are changed by this migration
+process.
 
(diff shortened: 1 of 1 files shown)


proposals/0088-enable-core-bpf-programs.md
```diff
@@ -62,24 +62,32 @@ published outlining at least the following details:
...
-its original program address. No program IDs for existing native programs would
-be changed by this migration process.
+**Migrating a native program to core BPF** shall consist of creating a buffer
+account, owned by the BPF Upgradeable Loader, containing the program's desired
+upgrade authority and the ELF bytes of its BPF implementation. A feature gate
+is used to create the BPF program accounts, replacing the existing native
+program at its original address with a legitimate BPF Upgradeable program.
+
+In the slot immediately following the feature activation, the program will *not*
+be invocable. It will be marked as `DelayedVisibility` in the program cache.
```
Lichtso, line 72: The program cache and `DelayedVisibility` are internals of the agave client, not protocol. The sentence above is sufficient.

proposals/0088-enable-core-bpf-programs.md
```diff
@@ -62,24 +62,32 @@ published outlining at least the following details:
...
-1. Verifiably build the source BPF program.
-2. Generate a new keypair for the source program.
-3. Deploy the program to the source program address.
+1. Verifiably build the ELF of the BPF implementation.
+2. Generate a new keypair for the source buffer account.
+3. Create the buffer account and write the ELF bytes to it.
 4. Generate a new keypair for the feature gate.
-5. Create a new feature gate to replace the target program with the source BPF
-   program.
+5. Create a new feature gate to replace the target program with the uploaded
```
Lichtso, line 89: I think we should keep the source & destination wording. Just replace source program with source buffer.

proposals/0088-enable-core-bpf-programs.md
buffalojoec: Removed.

proposals/0088-enable-core-bpf-programs.md
buffalojoec: Done!

lheeger-jump (approved):

jacobcreech:
Technically amendments would still deprecate the old SIMD and create a new one, but this seems to happen during development a decent amount. If we want to allow amendments, I'm ok with it. I wouldn't want to force process just for process's sake.

buffalojoec:
> Technically amendments would still deprecate the old SIMD and create a new one, but this seems to happen during development a decent amount. If we want to allow amendments, I'm ok with it. I wouldn't want to force process just for process's sake.

I guess it depends on how we envision using SIMDs as the protocol spec itself (which I think is a bad idea). If replacing SIMDs makes it easier we can do that, otherwise I think the amendment is fine, since the Git history on the proposal document will show any amendments.

Lichtso (approved):

buffalojoec:
@jacobcreech @lheeger-jump @ripatel-fd Leaving this open until Friday 4/26 before merging amendment.

lheeger-jump:
Merging. Thanks everyone

PR #148, opened 2024-04-30
SIMD-0148: `MoveStake` and `MoveLamports` instructions

2501babe:
two new instructions for moving value between stake accounts without holding `Withdrawer`

### DIFF
--- a/proposals/0148-stake-program-move-instructions.md
+++ b/proposals/0148-stake-program-move-instructions.md
@@ -16,7 +16,7 @@ We propose introducing two new instructions to the stake program for moving
 value between stake accounts with identical `Authorized` and `Lockup`:
 
 * `MoveStake`: Move a given `amount` of active stake from one active account to
-another active account, or from a active account to an inactive one, turning it
+another active account, or from an active account to an inactive one, turning it
 into an active account. If the entire source account delegation is moved, the
 source account becomes inactive. In all cases, rent-exempt balance is unaffected
 and minimum delegations are respected for accounts that end in an active state.
--- a/proposals/0148-stake-program-move-instructions.md
+++ b/proposals/0148-stake-program-move-instructions.md
@@ -99,15 +99,11 @@ effective, activating, or deactivating stake.
 
 ### `MoveStake`
 
-`MoveStake` requires 5 accounts:
+`MoveStake` requires 3 accounts:
 
 0. Source stake account: Writable, owned by the stake program
 1. Destination stake account: Writable, owned by the stake program
-2. Clock: Read-only, with the fixed address
-`SysvarC1ock11111111111111111111111111111111`
-3. Stake history: Read-only, with the fixed address
-`SysvarStakeHistory1111111111111111111111111`
-4. Stake account authority: Read-only signer
+2. Stake account authority: Read-only signer
 
 `MoveStake` instruction data is 12 bytes, containing:
 
@@ -149,15 +145,11 @@ rules as `Merge`
 
 ### `MoveLamports`
 
-`MoveLamports` requires 5 accounts:
+`MoveLamports` requires 3 accounts:
 
 0. Source stake account: Writable, owned by the stake program
 1. Destination stake account: Writable, owned by the stake program
-2. Clock: Read-only, with the fixed address
-`SysvarC1ock11111111111111111111111111111111`
-3. Stake history: Read-only, with the fixed address
-`SysvarStakeHistory1111111111111111111111111`
-4. Stake account authority: Read-only signer
+2. Stake account authority: Read-only signer
 
 `MoveLamports` instruction data is 12 bytes, containing:
 
--- a/proposals/0148-stake-program-move-instructions.md
+++ b/proposals/0148-stake-program-move-instructions.md
@@ -1,0 +1,161 @@
+---
+simd: '0148'
+title: MoveStake and MoveLamports Instructions
+authors:
+  - Hanako Mumei
+category: Standard
+type: Core
+status: Draft
+created: 2024-04-30
+feature: (fill in with feature tracking issues once accepted)
+---
+
+## Summary
+
+We propose introducing two new instructions to the stake program for moving
+value between stake accounts with identical `Authorized` and `Lockup`:
+
+* `MoveStake`: Move a given `amount` of active stake from one fully active
+account to another fully active account, or from a fully active account to an
+inactive one, turning it into an active account. In all cases, rent-exempt
+balance is unaffected and minimum delegations are respected.
+* `MoveLamports`: Move a given `amount` of excess lamports from one active or
+inactive account to another active or inactive account, where "excess lamports"
+refers to lamports that are neither delegated stake nor required for
+rent-exemption.
+
+For simplicity of implementation, we choose not to support accounts that are
+activating, deactivating, or partially active. A future SIMD may choose to
+extend this functionality should it be desirable.
+
+## Motivation
+
+Recently, a feature was activated which mandates that `Split` destinations be
+prefunded with the rent-exempt reserve, because before that, `Split` could be
+used to deactivate stake immediately, bypassing the cooldown period.
+
+However, this has introduced issues for protocols that manage stake on behalf
+of users without taking `Withdrawer` authority. Particularly, for one that
+splits user stake across many validators and periodically redelegates between
+them, every time they want to split part of a user stake to deactivate, the
+protocol must fund the rent-exemption themselves. And then when that split
+account is merged, those lamports cannot be reclaimed by the protocol, instead
+accumulating (undelegated) in the destination merge accounts.
+
+The purpose of the `MoveStake` instruction is to enable a flow whereby moving
+stake from a user's stake accounts U1 -> U2 from validator V1 to validator V2
+may proceed:
+
+* `MoveStake` the `amount` of stake from the user stake account U1 to a
+"transient" inactive account T holding sufficient lamports for rent exemption
+and minimum delegation. T instantly becomes a second active stake account
+delegated to V1 with `amount` stake.
+* `Deactivate` T and wait an epoch.
+* `DelegateStake` T to V2 and wait an epoch. T becomes an active stake account
+delegated to V2 with `amount + minimum_delegation` stake.
+* `MoveStake` the `amount` stake from T to U2. `Deactivate` T to return it to
+its initial state. Stake has moved from U1 to U2 with no outside lamports
+required and no new undelegated lamports in delegated stake accounts.
+
+The motivation for `MoveLamports` is to enable housekeeping tasks such as
+reclaiming lmaports from `Merge` destinations.
+
+## Alternatives Considered
+
+* There is a longstanding proposal we call Multistake, which would allow a
+stake account to have two delegation amounts to the same validator, the idea
+being that one serves as an onramp (and possibly offramp) for active stake on
+the account. This could support other flows to accomplish the same objective,
+such as allowing `DelegateStake` to accept an active stake account to begin
+activating an account's excess (non-rent non-stake) lamports, or allowing a
+`Split` source to `Deactivate` enough stake to cover rent-exemption for the new
+account. However, Multistake is a much larger design/engineering project, and
+we have to solve this sooner than it would be ready.
+* We discussed various proposals for allowing `Merge` to leave behind the
+source account or `Split` to split into any mergeable destination. However this
+confuses the presently clear distinction between these two operations and
+entails additional implementation risk as they are already rather complex. A
+new instruction that does one specific thing seems highly preferable.
+* The original version of this SIMD proposed a `Move` that did not take an
+`amount`, but this would require changes to `Split` to enable the first leg of
+the proposed flow.
+* Back out the changes introduced by requiring rent-exempt `Split` destinations.
+This is undesirable because that restriction was added for very good reason: an
+effectively unbounded amount of stake could be instantly deactivated through
+repeated splitting.
+
+## New Terminology
+
+`MoveStake` and `MoveLamports`, two new stake program instructions.
+
+## Detailed Design
+
+### `MoveStake`
+
+`MoveStake` requires 5 accounts:
+
+* Source stake account
+* Destination stake account
+* Clock
+* Stake history
+* Stake account authority
+
+`MoveStake` requires 1 argument:
+
+* `amount`, a `u64` indicating the quantity of lamports to move
+
+`MoveStake` aborts the transaction when:
+
+* `amount` is 0
+* Source and destination have the same address
+* Source and destination do not have identical `Authorized` and `Lockup`
+* The stake account authority is not the `Staker` on both accounts
+* The stake account authority is not a signer
+* Destination data length is not equal to the current version of `StakeState`
+* Source is not fully active (100% of delegation is effective)
+* Destination is neither fully active nor fully inactive (initialized or
+deactivated)
+* If destination is active, source and destination are not delegated to the same
+vote account
+* Moving `amount` stake would bring source below minimum delegation
+* Moving `amount` stake would fail to bring destination up to minimum delegation
+
+If all of these conditions hold, then:
+
+* Delegation and lamports on source are debited `amount`
+* Delegation and lamports on destination are credited `amount`
+* If destination is inactive, it is set to active with the same `Stake` as
+source, aside from delegation amount
+
+### `MoveLamports`
+
+Accounts and arguments are identical to the above.
+
+`MoveLamports` aborts the transaction when:
+
+* `amount` is 0
+* Source and destination have the same address
+* Source and destination do not have identical `Authorized` and `Lockup`
+* The stake account authority is not the `Staker` on both accounts
+* The stake account authority is not a signer
+* Source is neither fully active nor fully inactive
+* Destination is neither fully active nor fully inactive
+* `amount` exceeds source `lamports - stake - rent_exempt_reserve`
+
+If all of these conditions hold, then:
+
+* Lamports on source are debited `amount`
+* Lamports on destination are credited `amount`
+
+## Impact
+
+The primary utility of the proposed instructions is to support protocol
+developers in moving stake without controlling the `Withdrawer`. There is no
+loss of existing functionality.
+
+## Security Considerations
+
+Care must be taken to ensure stakes are fully active, as moving delegations
+between accounts in any kind of transient state is fraught. Otherwise this
+change should be fairly low impact, as it does not require changing any existing
+logic, in particular avoiding making `Split` or `Merge` more permissive.
--- a/proposals/0148-stake-program-move-instructions.md
+++ b/proposals/0148-stake-program-move-instructions.md
@@ -178,7 +178,8 @@ the quantity of lamports to move
 * The stake account authority is not the `Staker` on both accounts
 * The stake account authority is not a signer
 * Source is neither active nor inactive
-* Destination is neither active nor inactive
+* Destination is not a valid merge destination (active, inactive, or activating
+with zero effective stake)
 * `amount` exceeds source `lamports - effective_stake - rent_exempt_reserve`
 
 If all of these conditions hold, then:
--- a/proposals/0148-stake-program-move-instructions.md
+++ b/proposals/0148-stake-program-move-instructions.md
@@ -15,10 +15,11 @@ feature: (fill in with feature tracking issues once accepted)
 We propose introducing two new instructions to the stake program for moving
 value between stake accounts with identical `Authorized` and `Lockup`:
 
-* `MoveStake`: Move a given `amount` of active stake from one fully active
-account to another fully active account, or from a fully active account to an
-inactive one, turning it into an active account. In all cases, rent-exempt
-balance is unaffected and minimum delegations are respected.
+* `MoveStake`: Move a given `amount` of active stake from one active account to
+another active account, or from a active account to an inactive one, turning it
+into an active account. If the entire source account delegation is moved, the
+source account becomes inactive. In all cases, rent-exempt balance is unaffected
+and minimum delegations are respected for accounts that end in an active state.
 * `MoveLamports`: Move a given `amount` of excess lamports from one active or
 inactive account to another active or inactive account, where "excess lamports"
 refers to lamports that are neither delegated stake nor required for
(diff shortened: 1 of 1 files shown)


2501babe:
cc @joncinque for discussion, and if you could bring in the marinade guys

proposals/0148-stake-program-move-instructions.md
```diff
@@ -0,0 +1,164 @@
...
+
+* Source stake account
+* Destination stake account
+* Clock
+* Stake history
+* Stake account authority
+
+`MoveStake` requires 1 argument:
+
+* `amount`, a `u64` indicating the quantity of lamports to move
```
joncinque, line 108: nit: Can you specify that it is an unaligned little-endian unsigned 64-bit integer? It might even be clearest to specify the exact addresses of the expected params, similar to https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0118-partitioned-epoch-reward-distribution.md#epochrewards-sysvar-account

proposals/0148-stake-program-move-instructions.md
```diff
@@ -0,0 +1,164 @@
...
+
+`MoveStake` requires 5 accounts:
+
+* Source stake account
+* Destination stake account
+* Clock
+* Stake history
+* Stake account authority
+
+`MoveStake` requires 1 argument:
```
joncinque, line 106: Can you specify the exact instruction discriminant?

proposals/0148-stake-program-move-instructions.md
```diff
@@ -0,0 +1,164 @@
...
+## Detailed Design
+
+### `MoveStake`
+
+`MoveStake` requires 5 accounts:
+
+* Source stake account
+* Destination stake account
+* Clock
+* Stake history
```
joncinque, line 103: nit: can you specify these two addresses and that they are read-only?

proposals/0148-stake-program-move-instructions.md
```diff
@@ -0,0 +1,164 @@
...
+* Stake account authority
```
joncinque, line 104: Can you specify that this account is a read-only signer?

proposals/0148-stake-program-move-instructions.md
```diff
@@ -0,0 +1,164 @@
...
+`MoveStake` and `MoveLamports`, two new stake program instructions.
+
+## Detailed Design
+
+### `MoveStake`
+
+`MoveStake` requires 5 accounts:
+
+* Source stake account
+* Destination stake account
```
joncinque, line 101: nit: can you also specify that these are owned by the stake program and writable?

proposals/0148-stake-program-move-instructions.md
```diff
@@ -0,0 +1,164 @@
...
+deactivated)
+* If destination is active, source and destination are not delegated to the same
+vote account
+* Moving `amount` stake would bring source below minimum delegation
+* Moving `amount` stake would fail to bring destination up to minimum delegation
+
+If all of these conditions hold, then:
+
+* Delegation and lamports on source are debited `amount`
+* Delegation and lamports on destination are credited `amount`
```
joncinque, line 129: Although this is an edge case, the `credits_observed` on the destination account must be updated to the weighted average of the source and destination accounts, like during merge: https://github.com/anza-xyz/agave/blob/9403ca6f0451b670d41dee1b7592daa297727ed1/programs/stake/src/stake_state.rs#L1188, where you would do `(source_credits_observed * amount + destination_credits_observed * destination_amount + (amount + destination_amount) - 1) / (amount + destination_amount)`

proposals/0148-stake-program-move-instructions.md
```diff
@@ -0,0 +1,164 @@
...
+`MoveLamports` aborts the transaction when:
+
+* `amount` is 0
+* Source and destination have the same address
+* Source and destination do not have identical `Authorized` and `Lockup`
+* The stake account authority is not the `Staker` on both accounts
+* The stake account authority is not a signer
+* Source is neither fully active nor fully inactive
+* Destination is neither fully active nor fully inactive
+* `amount` exceeds source `lamports - stake - rent_exempt_reserve`
```
joncinque, line 146: Since this is only moving lamports, it would be possible to create a spoof account not owned by the stake program, so I think it's worth calling out that the destination must be owned by the stake program

proposals/0148-stake-program-move-instructions.md
```diff
@@ -0,0 +1,164 @@
...
+If all of these conditions hold, then:
+
+* Delegation and lamports on source are debited `amount`
+* Delegation and lamports on destination are credited `amount`
+* If destination is inactive, it is set to active with the same `Stake` as
+source, aside from delegation amount
+
+### `MoveLamports`
+
+Accounts and arguments are identical to the above.
```
joncinque, line 135: For completeness, can you also specify the instruction discriminant and the required accounts here?

proposals/0148-stake-program-move-instructions.md
```diff
@@ -0,0 +1,164 @@
...
+created: 2024-04-30
+feature: (fill in with feature tracking issues once accepted)
+---
+
+## Summary
+
+We propose introducing two new instructions to the stake program for moving
+value between stake accounts with identical `Authorized` and `Lockup`:
+
+* `MoveStake`: Move a given `amount` of active stake from one fully active
```
joncinque, line 18: nit: Someone else might bring this up, but we can probably ditch the term "fully" and just say "active" everywhere. Stakes can be active, inactive, activating, or deactivating. We could even just combine "activating or deactivating" and say "transient".

joncinque (commented):
Looks great to me! Mostly small things to address

proposals/0148-stake-program-move-instructions.md
```diff
@@ -0,0 +1,164 @@
...
+
+* `MoveStake` the `amount` of stake from the user stake account U1 to a
+"transient" inactive account T holding sufficient lamports for rent exemption
+and minimum delegation. T instantly becomes a second active stake account
+delegated to V1 with `amount` stake.
+* `Deactivate` T and wait an epoch.
+* `DelegateStake` T to V2 and wait an epoch. T becomes an active stake account
+delegated to V2 with `amount + minimum_delegation` stake.
+* `MoveStake` the `amount` stake from T to U2. `Deactivate` T to return it to
+its initial state. Stake has moved from U1 to U2 with no outside lamports
```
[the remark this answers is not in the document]
2501babe: there is; since we nixed the idea of a "full" move, the state of the source is never changed, and it has to remain a valid and active account. ill add something like "The user may then..."

proposals/0148-stake-program-move-instructions.md
2501babe: hmm i was trying to be unambigous because a stake account caught by the (de)activation warmup is partly "(de)activating" and partly "active." `Split` for instance treats any account with nonzero effective stake as "active" when determining if it can fund destination rent

proposals/0148-stake-program-move-instructions.md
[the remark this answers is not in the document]
joncinque: Oh sorry, I think there was a miscommunication then. I'm not sure if that'll cover the use-case required, where they likely want to move the full delegated amount.

proposals/0148-stake-program-move-instructions.md
joncinque: Ah I see what you mean... for the sake of documents, we might want to assume that "active" == "fully active" and not let the code's treatment of "active" create confusion.

proposals/0148-stake-program-move-instructions.md
2501babe: ill confess i dont understand the full implications of this and thought `credits_observed` was determined by, and advanced in lockstep with, `epoch_credits` on `VoteState`

i see how when two accounts are merged, the destination credits become a stake-weighted average of both accounts, but move is more complicated. checking my intuitions:
* partial move between active accounts. eg U1 and U2 have 100 sol each, U1 has C1 credits observed, U2 has C2 credits observed, and 50 sol moves U1 -> U2. this means that 33.3% of the new C2 value is provided by C1, because 1/3 of U2's stake have observed the credits seen by U1?
* full move between active accounts is identical to merge
* a move from active to inactive doesnt need to fiddle with credits observed because, much like delegation, it is a garbage value when inactive and assigned a fresh value when activated that doesnt depend on its prior value
* source credits observed never change because "credits observed" is more like an experience that a particular piece of stake has had rather than a quantity of something it carries with it

is this accurate?

joncinque: It's almost always advanced, except for a few edge cases, such as stakes that don't earn any rewards in an epoch.

And yep, you've correctly covered all of the situations and what should happen.

proposals/0148-stake-program-move-instructions.md
```diff
@@ -0,0 +1,161 @@
...
+* The stake account authority is not the `Staker` on both accounts
+* The stake account authority is not a signer
+* Destination data length is not equal to the current version of `StakeState`
+* Source is not fully active (100% of delegation is effective)
+* Destination is neither fully active nor fully inactive (initialized or
+deactivated)
+* If destination is active, source and destination are not delegated to the same
+vote account
+* Moving `amount` stake would bring source below minimum delegation
+* Moving `amount` stake would fail to bring destination up to minimum delegation
```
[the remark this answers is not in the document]
2501babe: is this really necessary to include in specifications for core program instructions? these are only ever consumed outside of consensus

defining all this would drastically increase the burden of writing specs and of implementing those specs. you would have to worry about the exact order errors are checked in, structure programs such that the language runtime and vm can never trigger their own error conditions for situations that would be exceptional before they're checked explicitly, and possibly rewrite specs for any nontrivial refactor or any improvement to error or logging information

even simple things like nested `match` expressions become a potential headache because if error A has to happen before error B, and error A is returned by the wildcard pattern but error B is in a matched block, then error B will come first by line number, and you have to start worrying "is the order here right? is the order obvious enough to someone looking at this code for the first time? should i just `unwrap_or_else` everything instead of writing structured code?"

if we say "you have to define the exact error codes returned for each abort condition and the precise log outputs that would result from any operation" then we effectively incentivize maintainers to use as few error codes as humanly possible and log nothing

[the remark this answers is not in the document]
2501babe: sorry, i dont understand how this would affect rpc if theres one bpf implementation of the stake program. wouldnt you pass through or remap any errors or log messages using the same rules as the existing rpc, without concern for the underlying program?

2501babe:
ok i updated to cover all of the resolved comments. accounts and instruction data are now defined precisely, i updated the spec to include the "full move" variant, and got rid of uses of "fully" but defined exactly what we mean by "active" and "inactive"

also while i was implementing `MoveStake` i reread the `Merge` impl and noticed that lockups arent cleared on expiration, so i updated the language to indicate that only in-force lockups must match

janlegner:
@2501babe thanks for this proposal, it looks awesome - and it will make things at Marinade much easier.

proposals/0148-stake-program-move-instructions.md
```diff
@@ -0,0 +1,201 @@
...
+reclaiming lmaports from `Merge` destinations.
+
+## Alternatives Considered
+
+* There is a longstanding proposal we call Multistake, which would allow a
+stake account to have two delegation amounts to the same validator, the idea
+being that one serves as an onramp (and possibly offramp) for active stake on
+the account. This could support other flows to accomplish the same objective,
+such as allowing `DelegateStake` to accept an active stake account to begin
+activating an account's excess (non-rent non-stake) lamports, or allowing a
```
lheeger-jump, line 70: This approach could also make slashing a bit harder to justify. That is okay though i think.

proposals/0148-stake-program-move-instructions.md
```diff
@@ -0,0 +1,201 @@
...
+* The stake account authority is not a signer
+* Source is neither active nor inactive
+* Destination is not a valid merge destination (active, inactive, or activating
+with zero effective stake)
+* `amount` exceeds source `lamports - effective_stake - rent_exempt_reserve`
+
+If all of these conditions hold, then:
+
+* Lamports on source are debited `amount`
+* Lamports on destination are credited `amount`
```
lheeger-jump, line 187: How does this affect activation?

proposals/0148-stake-program-move-instructions.md
```diff
@@ -0,0 +1,201 @@
...
+If all of these conditions hold, then:
+
+* Delegation and lamports on source are debited `amount`
+* Delegation and lamports on destination are credited `amount`
+* If `amount` constitutes the full delegation on the source, source is reset to
+an `Initialized` state
+* If destination is inactive, destination becomes active with the same `Stake`
+as source, aside from delegation amount
+* `credits_observed` must be updated on the destination according to the same
+rules as `Merge`
```
lheeger-jump, line 148: How does this affect activation?

proposals/0148-stake-program-move-instructions.md
```diff
@@ -0,0 +1,201 @@
...
+
+* `amount` is 0
+* Source or destination are not writable
+* Source or destination are not owned by the stake program
+* Source and destination have the same address
+* Source and destination do not have identical `Authorized`
+* If `Lockup` is in force, source and destination do not have identical `Lockup`
+* The stake account authority is not the `Staker` on both accounts
+* The stake account authority is not a signer
+* Source is neither active nor inactive
```
lheeger-jump, line 179: so not deactivating or activating, correct?

proposals/0148-stake-program-move-instructions.md
lheeger-jump, line 187: Does this move the two accounts in activating/deactivating?

lheeger-jump (commented):
Okay it took like 3 read throughs over the course of the last month to really understand this proposal. Some questions I have: 

- When does moved stake come active?
- How does this affect the future of slashing?
- Are there other problems this solves that the alternatives did not?
- There are a maximum amount of stake changes allowed per epoch, or am I misremembering that? Does that affect this proposal?
- Are there any drawbacks?
- If this change launches, what are the measurable effects it will have that we can look at to measure efficacy?

lheeger-jump:
This proposal is getting into very good shape. All I have are questions.

proposals/0148-stake-program-move-instructions.md
```diff
@@ -0,0 +1,201 @@
...
+becomes a second active stake account delegated to V1 with `amount` stake.
+* `Deactivate` T and wait an epoch.
+* `DelegateStake` T to V2 and wait an epoch. T becomes an active stake account
+delegated to V2 with `amount` stake.
+* `MoveStake` the `amount` stake from T to U2. T returns to its initial inactive
+state. Stake has moved from U1 to U2 with no outside lamports required and no
+new undelegated lamports becoming trapped in delegated stake accounts.
+
+The motivation for `MoveLamports` is to enable housekeeping tasks such as
+reclaiming lmaports from `Merge` destinations.
```
CriesofCarrots, line 61:
```suggestion
reclaiming lamports from `Merge` destinations.
```

proposals/0148-stake-program-move-instructions.md
```diff
@@ -0,0 +1,201 @@
...
+
+* There is a longstanding proposal we call Multistake, which would allow a
+stake account to have two delegation amounts to the same validator, the idea
+being that one serves as an onramp (and possibly offramp) for active stake on
+the account. This could support other flows to accomplish the same objective,
+such as allowing `DelegateStake` to accept an active stake account to begin
+activating an account's excess (non-rent non-stake) lamports, or allowing a
+`Split` source to `Deactivate` enough stake to cover rent-exemption for the new
+account. However, Multistake is a much larger design/engineering project, and
+we have to solve this sooner than it would be ready.
```
CriesofCarrots, line 73:
```suggestion
we have to solve the existing problem sooner than Multistake would be ready.
```

proposals/0148-stake-program-move-instructions.md
```diff
@@ -0,0 +1,201 @@
...
+### `MoveStake`
+
+`MoveStake` requires 5 accounts:
+
+0. Source stake account: Writable, owned by the stake program
+1. Destination stake account: Writable, owned by the stake program
+2. Clock: Read-only, with the fixed address
+`SysvarC1ock11111111111111111111111111111111`
+3. Stake history: Read-only, with the fixed address
+`SysvarStakeHistory1111111111111111111111111`
```
CriesofCarrots, line 109: Why do we need to pass these in as accounts?

proposals/0148-stake-program-move-instructions.md
```diff
@@ -0,0 +1,201 @@
...
+### `MoveLamports`
+
+`MoveLamports` requires 5 accounts:
+
+0. Source stake account: Writable, owned by the stake program
+1. Destination stake account: Writable, owned by the stake program
+2. Clock: Read-only, with the fixed address
+`SysvarC1ock11111111111111111111111111111111`
+3. Stake history: Read-only, with the fixed address
+`SysvarStakeHistory1111111111111111111111111`
```
CriesofCarrots, line 159: Ditto

proposals/0148-stake-program-move-instructions.md
2501babe: it doesnt. neither instruction _ever_ activates or deactivates stake at all, their only purpose is to move value around inside a authorized/lockup bubble without changing its state. _accounts_ may become active or inactive based on `MoveStake`, but the total delegation amounts never change:
* active stake stays active and moves with its underlying lamports
* lamports in an inactive account never become active just because stake is moved into it
* accounts in any activating or deactivating state, including partially active/inactive accounts, are rejected by `MoveStake` for simplicity
* moving stake from an active account to another active account requires them to be delegated to the same vote account

proposals/0148-stake-program-move-instructions.md
2501babe: correct. i defined active/inactive earlier in the simd since there was some quibbling over previous use of "fully active" (not being an official term), but the only allowed states are `Initialized`, `Stake` and fully active, or `Stake` and fully deactivated. `Uninitialized`, activating (active next epoch), deactivating (inactive next epoch), and in any warmup/cooldown state (partially active/inactive due to whole-cluster activation/deactivation) are all abort states

proposals/0148-stake-program-move-instructions.md
2501babe: it doesnt. `MoveLamports` only moves lamports that arent backing a delegation and arent required for rent exemption. just like `MoveStake` no stake ever changes activation state under any condition

2501babe: deactivating isnt allowed for either account, and activating is only allowed for the destination for `MoveLamports`. tbh i describe the allowed states from first principles for the sake of letting the simd stand alone, but a lot of these are just reusing logic from `Merge`, eg here destination is identical to a valid merge destination, and source is a valid merge source except it cant be activating because this would require calculating whether the minimum will be upheld next epoch (which is pointless because theres no usecase)

proposals/0148-stake-program-move-instructions.md
2501babe: im not sure why it would affect slashing, conceptually it isnt any different from having two stake accounts, one active and one activating, delegated to the same vote account

proposals/0148-stake-program-move-instructions.md
2501babe: hm we dont need to. i just thought it would be best practice since most instructions do this, even if they arent used. should i remove them?

incidentally when the bpf stake program comes, i was planning on making none of the instructions require sysvars anymore and just skipping them if provided

2501babe:
addressing the rest of @lheeger-jump comments:

> * When does moved stake come active?
as described in code comment replies, moved stake never changes activation status

> * How does this affect the future of slashing?
i dont believe it does, assuming slashing is a penalty applied to all the stake accounts of a vote account proportional to their active delegations. if there are other factors im missing please elaborate tho

> * Are there other problems this solves that the alternatives did not?
the main reason for doing it this way is multistake is too complicated to move through the simd process and implement in a reasonable timeframe (and may not even cover the full usecase) and the alternative to adding new instructions is changing both `Split` _and_ `Merge` to be more permissive. also on a philosophical level it would muddy the relatively clean concepts that both represent (division and unification) by making split more merge-like and merge more split-like

> * There are a maximum amount of stake changes allowed per epoch, or am I misremembering that? Does that affect this proposal?
youre thinking of warmup cooldown, which is a system where if more than x% (presently 7, used to be 25) of stake on a whole cluster is activated or deactivated in one epoch, then in the next epoch, only that % of _effective_ stake will change. the reason for the "this is required to be fully active, this is required to be fully active or inactive" etc rules for accounts with these instructions is so that none of that applies to us, we reject any accounts that are in transient states

> * Are there any drawbacks?
uhhhh. well its more instructions to support. i cant think of any negative results tho, it doesnt allow any new behavior, it just lets you shuffle value around using `Staker` when you dont have `Withdrawer`, but only between accounts that have the same authorities

> * If this change launches, what are the measurable effects it will have that we can look at to measure efficacy?
we should expect to see an appreciable reduction in the absolute number of stake accounts (not just a reduction in the rate of increase)

proposals/0148-stake-program-move-instructions.md
CriesofCarrots: I think remove them. We do have some stake-program instructions (like DeactivateDelinquent) already forgoing sysvar accounts, so no consistency reason to keep them. Might as well make these new instructions take up less transaction space from the get-go.

CriesofCarrots (approved):
Lgtm

proposals/0148-stake-program-move-instructions.md
lheeger-jump: Got it!

proposals/0148-stake-program-move-instructions.md
lheeger-jump: Got it great!

proposals/0148-stake-program-move-instructions.md
lheeger-jump: Thank you!

lheeger-jump (approved):
Lets let it sit for a bit, and then merge in a week. We should coordinate on implementation timelines separately.

agca52 (commented):
Katkılarınız için teşekkürler, bildirmek istediğiniz aktif olarak destek vermek istediğiniz konular için fikir alış verisine acığım

proposals/0148-stake-program-move-instructions.md
agca52: Thank you

proposals/0148-stake-program-move-instructions.md
agca52: Thank you

proposals/0148-stake-program-move-instructions.md
agca52: Thank you

proposals/0148-stake-program-move-instructions.md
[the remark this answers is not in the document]
agca52: Thank you

PR #149, opened 2024-05-09
SIMD-0149: Migrate Snapshot Epoch Stakes

jstarry:
Migrate bank snapshots to a new "epoch stakes" field in order to store
additional stake state needed for calculating partitioned rewards.

### DIFF
--- a/proposals/0149-migrate-snapshot-epoch-stakes.md
+++ b/proposals/0149-migrate-snapshot-epoch-stakes.md
@@ -125,8 +125,10 @@ epoch stakes entries to the new epoch stakes field. Note, however, that there
 are 3 different epoch stakes entry variants:
 
 1. Entries created during epoch boundaries which have full stake account data
-2. Entries deserialized from the old snapshot epoch stakes field which only have stake delegation state.
-3. Entries deserialized from the new snapshot epoch stakes field which have full stake state.
+2. Entries deserialized from the old snapshot epoch stakes field which only have
+    stake delegation state.
+3. Entries deserialized from the new snapshot epoch stakes field which have full
+    stake state.
 
 Only variants 1 and 3 can be serialized into the new epoch stakes field so any
 variant 2 epoch stakes entries will continue being serialized into the old epoch
--- a/proposals/0149-migrate-snapshot-epoch-stakes.md
+++ b/proposals/0149-migrate-snapshot-epoch-stakes.md
@@ -1,0 +1,157 @@
+---
+simd: '0149'
+title: Migrate Snapshot Serialized Epoch Stakes
+authors: Justin Starry (Anza)
+category: Standard
+type: Core
+status: Draft
+created: 2024-05-09
+feature: (fill in with feature tracking issues once accepted)
+---
+
+## Summary
+
+Migrate bank snapshots to a new "epoch stakes" field in order to store
+additional stake state needed for calculating partitioned rewards.
+
+## Motivation
+
+In order to properly support recalculating partitioned rewards (SIMD-0118)
+when rebooting from a snapshot in the middle of the rewards distribution
+period at the beginning of an epoch, additional stake state should be
+stored in the epoch stakes snapshot field.
+
+Since the currently used epoch stakes field doesn't support versioning, we
+propose migrating the snapshot format to use a new epoch stakes field which is
+versioned and supports storing the required stake state.
+
+## New Terminology
+
+NA
+
+## Alternatives Considered
+
+We have discussed the following alternative approach:
+
+1. The missing stake state information can be retrieved from the current epoch's
+stakes cache instead of epoch stakes. However this alternative is risky because
+the current epoch's stake cache is be updated during rewards distribution with
+newly rewarded stake accounts and possibly newly created stake accounts as well.
+
+## Detailed Design
+
+### Serialized Data Changes
+
+To be more specific, the missing stake information is the `credits_observed`
+field from the `Stake` struct:
+
+```rust
+struct Stake {
+    delegation: Delegation,
+    credits_observed: u64,
+}
+```
+
+The currently used `epoch_stakes` field in the bank snapshot serializes a map of
+epochs to `EpochStakes` structs which have special logic for serializing stake
+state:
+
+```rust
+struct (BankFieldsToDeserialize | BankFieldsToSerialize) {
+    ..
+    epoch_stakes: HashMap<Epoch, EpochStakes>,
+    ..
+}
+
+struct EpochStakes {
+    #[serde(with = "crate::stakes::serde_stakes_enum_compat")]
+    stakes: Arc<StakesEnum>,
+    total_stake: u64,
+    node_id_to_vote_accounts: Arc<NodeIdToVoteAccounts>,
+    epoch_authorized_voters: Arc<EpochAuthorizedVoters>,
+}
+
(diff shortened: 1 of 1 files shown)


proposals/0149-migrate-snapshot-epoch-stakes.md
```diff
@@ -0,0 +1,159 @@
...
+new software releases properly. The rollout will require two phases:
+
+1. Introduce support for deserializing the migrated epoch stakes field 
+2. Enable serializing epoch stakes to the new field and phase out the old field
+
+During the first phase, validator software will be updated to attempt to
+deserialize the new epoch stakes field appended at the end of the bank snapshot.
+If the field doesn't exist, validators will continue using the old deserialized
+field. If the new epoch stakes field is deserialized successfully, the epoch
+stakes entries from this field will be merged with entries from the old field.
```
CriesofCarrots, line 121: Do we actually need to support merging the two fields? ie. is there a use-case for having both the old field and the new field populated?

proposals/0149-migrate-snapshot-epoch-stakes.md
```diff
@@ -0,0 +1,159 @@
...
+
+1. Entries created during epoch boundaries which have full stake account data
+2. Entries deserialized from the old snapshot epoch stakes field which only have
+    stake delegation state.
+3. Entries deserialized from the new snapshot epoch stakes field which have full
+    stake state.
+
+Only variants 1 and 3 can be serialized into the new epoch stakes field so any
+variant 2 epoch stakes entries will continue being serialized into the old epoch
+stakes field.
```
CriesofCarrots, line 135: It might be useful to describe what cases could cause variant 2 after a validator updates to the second-phase software. Will this be the case until the next epoch boundary after update?

CriesofCarrots: Also, please make explicit that only one of the fields will be serialized into for any given snapshot.

CriesofCarrots (commented):
Generally looks good to me! A couple small comments and one question added.
I feel somewhat ambivalent about the feature gate; I do like that it potentially enables the transition to be faster, so if you can convince @brooksprumo , I'm on board.

proposals/0149-migrate-snapshot-epoch-stakes.md
jstarry: We typically keep around like 6 epochs worth of epoch stakes entries. After the migration we could have some delegation only entries still which I think we should keep storing in the old field until enough epochs pass.

proposals/0149-migrate-snapshot-epoch-stakes.md
jstarry: I will add more details to make this clearer but basically we will have a mixed bag of epoch stakes entries during the migration. The new field won't support serializing variant 2 entries so I was planning to put them in the old field still even if we have already started using the new field.

proposals/0149-migrate-snapshot-epoch-stakes.md
```diff
@@ -0,0 +1,159 @@
...
+
+We propose adding a new feature gate `migrate_epoch_stakes_snapshot_field` to
+facilitate the migration by activating phase 2 so that the implementation for
+both phases can be added to the next beta release.
```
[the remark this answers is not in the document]
jstarry: Alright, you have me convinced that using minor versions is the best way to go, thanks for explaining the context behind your stance here. I've updated the rollout description in the doc, would appreciate another look!

jstarry:
@CriesofCarrots can you take another look? I made some changes that should hopefully make the field merging part more clear

proposals/0149-migrate-snapshot-epoch-stakes.md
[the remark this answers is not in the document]
CriesofCarrots: Do we consider this a blocker for activating partitioned epoch rewards? I'm concerned about adding an additional 3-4 months before that can be activated. Every epoch boundary is scary.

proposals/0149-migrate-snapshot-epoch-stakes.md
CriesofCarrots: Thanks for adding this info; much clearer now.

CriesofCarrots (commented):
Lgtm, just the question about rollout and timing.

proposals/0149-migrate-snapshot-epoch-stakes.md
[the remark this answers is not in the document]
jstarry: I don't consider it a blocker personally, we can continue keeping the stake program fully locked down to avoid any issues popping up from our usage of stakes cache. Later when this migration is completed, we can allow new stake accounts to be created if we want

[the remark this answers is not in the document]
CriesofCarrots: Okay, cool 👍

CriesofCarrots (dismissed):

0xSol:
@lheeger-jump / @riptl  please review / approve this, likely post BP24, from FD side. Thanks.

proposals/0149-migrate-snapshot-epoch-stakes.md
```diff
@@ -0,0 +1,169 @@
...
+
+NA
+
+## Alternatives Considered
+
+We have discussed the following alternative approach:
+
+1. The missing stake state information can be retrieved from the current epoch's
+stakes cache instead of epoch stakes. However this alternative is risky because
+the current epoch's stake cache is be updated during rewards distribution with
```
lheeger-jump, line 38:
```suggestion
the current epoch's stake cache is being updated during rewards distribution with
```

proposals/0149-migrate-snapshot-epoch-stakes.md
```diff
@@ -0,0 +1,169 @@
...
+## Impact
+
+No major impact beyond backwards compatibility concerns. Snapshots will
+be a few MB larger than before.
+
+## Security Considerations
+
+Missing or corrupted epoch stakes entries caused by faulty snapshot migration
+can cause validators to fork off from the cluster or cause the cluster to lose
+consensus if sufficient stake is affected.
```
lheeger-jump, line 160:
```suggestion
Missing or corrupted epoch stakes entries caused by faulty snapshot migration
can cause validators to fork off from the cluster or cause the cluster to lose
consensus if sufficient stake is affected.

Care should be taken to ensure that implementations can detect corrupted
or incorrect values.
```

lheeger-jump (commented):
This looks good, 2 minor comments. Correct me if I am wrong but it is fundamentally just making epoch stakes a versioned field?

jstarry:
Yup, it is being made versioned and we are adding the credits observed field into snapshots

jstarry:
@lheeger-jump I applied your suggested changes

CriesofCarrots (approved):

lheeger-jump (approved):

0xSol:
Thanks for collaborating on this SIMD!  There are two approvals. Plan to merge by end of next week (June 7) unless other opinions.

jstarry:
@0xSol can you merge please

CriesofCarrots:
> @0xSol can you merge please

Oh neat, I have the power. Since the date @0xSol proposed has passed several weeks ago and we have approvals/consensus, I will merge.

PR #150, opened 2024-05-16
Amend 0089: Update program upgrade process

buffalojoec:
After further discussions with the Firedancer team around Core BPF programs,
we've decided these programs should all be upgraded using feature gates.

This amendment updates SIMD 0089 (Programify Feature Gate) to reflect this
agreement.

I've also removed the bit about first deploying a no-op program, since we wouldn't
be able to upgrade via keypair after performing the migration.

### DIFF
--- a/proposals/0089-programify-feature-gate-program.md
+++ b/proposals/0089-programify-feature-gate-program.md
@@ -59,9 +59,12 @@ changes need only be done once, eliminating this duplication of work.
 
 ## Detailed Design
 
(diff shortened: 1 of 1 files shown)


joncinque (approved):
Looks good to me, thanks!

lheeger-jump (approved):
I think later on all the core BPF programs including this one should move to a governance upgrade authority.

We can do that later though.

PR #151, opened 2024-05-20
Amend 0133: Support total cluster stake

buffalojoec:
Amend SIMD 0133 - Syscall `GetEpochStake` - to support retrieval of the
_total cluster stake_, in addition to a vote account's stake, for the current
epoch.

The main change here is that a null pointer will allow callers to retrieve the
total cluster stake for the epoch.

A potential point of contention is going to be the designated compute usage
for retrieval of total stake.

cc @michaelh-laine

### DIFF
--- a/proposals/0133-syscall-get-epoch-stake.md
+++ b/proposals/0133-syscall-get-epoch-stake.md
@@ -12,7 +12,8 @@ feature: (fill in with feature tracking issues once accepted)
 
 ## Summary
 
-A syscall to retrieve a vote account's delegated stake for the current epoch.
+A syscall to retrieve a vote account's delegated stake or the total cluster
+stake for the current epoch.
 
 ## Motivation
 
@@ -54,10 +55,18 @@ The specification for the proposed syscall is as follows:
 
 ```c
 /**
- * Retrieves the total active stake delegated to a vote account for the current
- * epoch.
+ * Retrieves the total active stake delegated to a vote account, or for the
+ * entire cluster, for the current epoch.
+ *
+ * If a valid pointer for `vote_addr` is provided, returns the total active
+ * stake delegated to the vote account at the 32-byte address found at
+ * `var_addr`.
+ *
+ * If a null pointer is provided, returns the total active stake for the
+ * cluster.
  *
  * @param vote_addr     A pointer to 32 bytes representing the vote address.
+ *                      Can be a null pointer.
  * @return              A 64-bit unsigned integer representing the total
  *                      active stake delegated to the vote account at the
  *                      provided address.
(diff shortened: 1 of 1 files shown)


michaelh-laine:
Can't comment on CU usage but overall this looks like a simple change that is very welcome and will be extremely helpful

proposals/0133-syscall-get-epoch-stake.md
```diff
@@ -67,32 +76,44 @@ uint64_t sol_get_epoch_stake(/* r1 */ void const * vote_addr);
...
+
+```
+syscall_base + floor(PUBKEY_BYTES/cpi_bytes_per_unit) + mem_op_base
+```
+
+If `var_addr` is a null pointer:
 
 ```
-syscall_base + floor(32/cpi_bytes_per_unit) + mem_op_base
+syscall_base + mem_op_base
```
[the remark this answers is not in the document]
buffalojoec: I think this makes sense. I'll add this change.

@CriesofCarrots just to touch on our discussion about compute usage, where we were ideating on charging more compute to perform the summation, I think just `syscall_base` does make sense here. Caching the total epoch stake or otherwise structuring the runtime component to serve this syscall in a performant fashion seems like it should be an implementation detail.

Furthermore, we already do cache this information, so this call is trivial for Agave.

https://github.com/anza-xyz/agave/blob/9dd01f55182ff75a9c21bf89958ba5b37c8fe937/runtime/src/epoch_stakes.rs#L47-L49

proposals/0133-syscall-get-epoch-stake.md
```diff
@@ -67,32 +76,44 @@ uint64_t sol_get_epoch_stake(/* r1 */ void const * vote_addr);
...
-account or does not exist, the syscall will return `0` for active stake.
+If `var_addr` is a null pointer, the syscall returns a `u64` integer
+representing the total active stake on the cluster for the current epoch.
 
 ### Compute Unit Usage
 
-The syscall will always attempt to consume the same amount of CUs regardless of
-control flow.
+The syscall will always attempt to consume the same amount of CUs for both
+possible `var_addr` pointers, regardless of control flow.
```
CriesofCarrots, line 94: This sentence is confusing, since the following lines define *different* amounts of CUs for the two possible `var_addr` pointers. Do we need this sentence? If so, can you reword it more simply?

buffalojoec: Updated it to be more clear!

CriesofCarrots (approved):

proposals/0133-syscall-get-epoch-stake.md
CriesofCarrots: Much better, thanks!

buffalojoec:
I'll leave this open until this Friday, 5/31, before merging.

@lheeger-jump cc'ing you since you were the only other active reviewer on the original 0133 PR, in case you wanted to have a look.

lheeger-jump (approved):

lheeger-jump:
Looks good.

PR #152, opened 2024-06-04
SIMD-0152: Precompiles

0x0ece:
## Summary

Define a unified behavior for precompiles, and highlight a few minor
changes to the existing precompiles to minimize differences.

## Motivation

Precompiles are special native programs designed to verify additional
transaction signatures.
They run without the VM and without loading any account, and they can access
data from other instructions, within the same transaction.

At the time of writing, two precompiles exist to verify Ed25519 and
Ethereum-like Secp256k1 signatures, and another one is being proposed to
support Secp256r1 signatures for FIDO Passkeys.

Historically, the two precompiles were built at different times and
by different people, so naturally there are some subtle differences
in how they behave, especially in edge cases.

The main goal of this document is to provide a specification for how
a precompile should behave, remove differences and provide guidelines
for future proposals.

In addition, we highlight 3 minor changes to the existing precompiles
that will simplify their behavior and make it easier to develop
alternative validator clients.

### DIFF
--- a/proposals/0152-precompiles.md
+++ b/proposals/0152-precompiles.md
@@ -1,0 +1,270 @@
+---
+simd: '0152'
+title: Precompiles
+authors:
+  - Emanuele Cesena
+category: Standard
+type: Core
+status: Draft
+created: 2024-06-03
+feature:
+supersedes:
+superseded-by:
+extends:
+---
+
+## Summary
+
+Define a unified behavior for precompiles, and highlight a few minor
+changes to the existing precompiles to minimize differences.
+
+## Motivation
+
+Precompiles are special native programs designed to verify additional
+transaction signatures.
+They run without the VM and without loading any account, and they can access
+data from other instructions, within the same transaction.
+
+At the time of writing, two precompiles exist to verify Ed25519 and
+Ethereum-like Secp256k1 signatures, and another one is being proposed to
+support Secp256r1 signatures for FIDO Passkeys.
+
+Historically, the two precompiles were built at different times and
+by different people, so naturally there are some subtle differences
+in how they behave, especially in edge cases.
+
+The main goal of this document is to provide a specification for how
+a precompile should behave, remove differences and provide guidelines
+for future proposals.
+
+In addition, we highlight 3 minor changes to the existing precompiles
+that will simplify their behavior and make it easier to develop
+alternative validator clients.
+
+## Alternatives Considered
+
+Leave as is.
+
+## New Terminology
+
+n/a
+
+## Detailed Design
+
+### Generic Precompile
+
+Precompiles are special native programs designed to verify additional
+signatures. Each precompile consists of a single `verify` instruction.
+
+Precompiles are executed right after transaction signature verification,
+they run without the VM and without loading any account.
+
+If a transaction contains more than 8 precompile signatures, it must fail.
+
+```
+MAX_ALLOWED_PRECOMPILE_SIGNATURES = 8
+```
+
+The precompile instruction `verify` accepts the following data:
+
+```
+struct PrecompileVerifyInstruction {
+  num_signatures:  u8,                       // Number of signatures to verify
+  padding:         u8,                       // Single byte padding
+  offsets:         Array<PrecompileOffsets>, // Array of offsets
+  additionalData?: Bytes,                    // Optional additional data, e.g.
+                                             // signatures included in the same
+                                             // instruction
+}
+
+struct PrecompileOffsets {
+  signature_offset: u16 LE,                  // Offset to signature
+  signature_instruction_index: u16 LE,       // Instruction index to signature
+  public_key_offset: u16 LE,                 // Offset to public key
+  public_key_instruction_index: u16 LE,      // Instruction index to  public key
+  message_offset: u16 LE,                    // Offset to start of message data
+  message_length: u16 LE,                    // Size of message data
+  message_instruction_index: u16 LE,         // Instruction index to message
+}
+```
+
+The behavior of the precompile instruction `verify` is as follow:
+
+1. If instruction `data` is empty, return error.
+2. The first byte of `data` is the number of signatures `num_signatures`.
+3. If `num_signatures` is 0, return error.
+4. Expect (enough bytes of `data` for) `num_signatures` instances of
+   `PrecompileOffsets`.
+5. For each signature:
+   a. Read `offsets`: an instance of `PrecompileOffsets`
+   b. Based on the `offsets`, retrieve `signature`, `public_key`, and
+      `message` bytes. If any of the three fails, return error.
+   c. Invoke the actual `sigverify` function. If it fails, return error.
+
+To retrieve `signature`, `public_key`, and `message`:
+
+1. Get the `instruction_index`-th `instruction_data`
+   - The special value `0xFFFF` means "current instruction"
+   - If the index is invalid, return Error
+2. Return `length` bytes starting from `offset`
+   - If this exceeds the `instruction_data` length, return Error
+
+If the precompile `verify` function returns any error, the whole transaction
+should fail. Therefore, the type of error is irrelevant and is left as an
+implementation detail.
+
+In pseudo-code:
+
+```
+fn verify() {
+  if data_length == 0 {
+    return Error
+  }
+  num_signatures = data[0]
+  if num_signatures == 0 {
+    return Error
+  }
+  if data_length < (2 + num_signatures * size_of_offsets) {
+    return Error
+  }
+
+  all_tx_data = { data, instruction_datas }
+
+  data_position = 2
+  for i in 0..num_signatures {
+    offsets = (PrecompileOffsets)
+      data[data_position..data_position+size_of_offsets]
+    data_position += size_of_offsets
+
+    signature = get_data_slice(all_tx_data,
+                               offsets.signature_instruction_index,
+                               offsets.signature_offset
+                               signature_length)
+    if !signature {
+      return Error
+    }
+
+    public_key = get_data_slice(all_tx_data,
+                                offsets.public_key_instruction_index,
+                                offsets.public_key_offset,
+                                public_key_length)
+    if !public_key {
+      return Error
+    }
+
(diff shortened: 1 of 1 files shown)


proposals/0152-precompiles.md
```diff
@@ -0,0 +1,270 @@
...
+
+The KeccakSecp256k1 precompile currently accepts an input of `[0]`, as
+in "verify 0 signatures", which is a useless instruction.
+With Change #3 we'll avoid this anomaly.
+
+We note that the KeccakSecp256k1 precompile has a slightly different
+struct for offset, with instruction indexes of a single byte (and, as
+a result, no special value of `0xFFFF` to indicate the "current instruction").
+This is for historical reasons, and since modifying it would break
+existing users, we recommend to NOT change the existing behavior.
```
seanyoung, line 257: This is not true. The struct is defined as
```rust
#[derive(Default, Serialize, Deserialize, Debug, Eq, PartialEq)]
pub struct SecpSignatureOffsets {
    /// Offset to 64-byte signature plus 1-byte recovery ID.
    pub signature_offset: u16,
    /// Within the transaction, the index of the instruction whose instruction data contains the signature.
    pub signature_instruction_index: u8,
    /// Offset to 20-byte Ethereum address.
    pub eth_address_offset: u16,
    /// Within the transaction, the index of the instruction whose instruction data contains the address.
    pub eth_address_instruction_index: u8,
    /// Offset to start of message data.
    pub message_data_offset: u16,
    /// Size of message data in bytes.
    pub message_data_size: u16,
    /// Within the transaction, the index of the instruction whose instruction data contains the message.
    pub message_instruction_index: u8,
}
```
The padding rules mean that a padding byte is introduced after each `u8`. So, by changing every `u8` to `u16` the layout does not change. Assuming that the clients insert zeros for the padding byte, nothing will break. 

So I think the offsets structs can be made identical.

I have not tested this.

proposals/0152-precompiles.md
```diff
@@ -0,0 +1,270 @@
...
+## Alternatives Considered
+
+Leave as is.
+
+## New Terminology
+
+n/a
+
+## Detailed Design
+
```
samkim-crypto, line 53:
```suggestion

We first propose a new specification for a generic precompile. Then we propose changes to the existing precompiles with respect to the new specification.

```
Very minor point, but it might be confusing on first read whether the "Generic Precompile" section is describing an existing precompile or the new spec. A sentence or two to clarify that this should be the new spec going forward will help with the readability I think. We can perhaps rename "Generic Precompile" to either "Generic Precompile Specification" or just "Precompile Specification", wdyt?

samkim-crypto (approved):
Looks good to me and sorry for the delay on this.

proposals/0152-precompiles.md
0x0ece: Sounds good, will rephrase

seanyoung (approved):

proposals/0152-precompiles.md
0x0ece: Added, and changed the title to Precompile Specification (actually removed "generic precompile" entirely).

proposals/0152-precompiles.md
```diff
@@ -0,0 +1,273 @@
...
+MAX_ALLOWED_PRECOMPILE_SIGNATURES = 8
+```
+
+The precompile instruction `verify` accepts the following data:
+
+```
+struct PrecompileVerifyInstruction {
+  num_signatures:  u8,                       // Number of signatures to verify
+  padding:         u8,                       // Single byte padding
+  offsets:         Array<PrecompileOffsets>, // Array of offsets
```
ptaffet-jump, line 77: Can you clarify that `Array<PrecompileOffsets>` is not length prefixed? Maybe calling it `PrecompileOffsets[]` would be more clear

proposals/0152-precompiles.md
```diff
@@ -0,0 +1,273 @@
...
+  additionalData?: Bytes,                    // Optional additional data, e.g.
+                                             // signatures included in the same
+                                             // instruction
+}
+
+struct PrecompileOffsets {
+  signature_offset: u16 LE,                  // Offset to signature
```
ptaffet-jump, line 84: Offset could be a bit more clear. It's the offset within specified instruction data, right?

proposals/0152-precompiles.md
```diff
@@ -0,0 +1,273 @@
...
+1. Get the `instruction_index`-th `instruction_data`
+   - The special value `0xFFFF` means "current instruction"
+   - If the index is invalid, return Error
+2. Return `length` bytes starting from `offset`
+   - If this exceeds the `instruction_data` length, return Error
+
+If the precompile `verify` function returns any error, the whole transaction
+should fail. Therefore, the type of error is irrelevant and is left as an
+implementation detail.
+
```
ptaffet-jump, line 118: I think you should explicitly mention the case in which fields overlap. Is that allowed?

proposals/0152-precompiles.md
```diff
@@ -0,0 +1,273 @@
...
+Define a unified behavior for precompiles, and highlight a few minor
+changes to the existing precompiles to minimize differences.
+
+## Motivation
+
+Precompiles are special native programs designed to verify additional
+transaction signatures.
+They run without the VM and without loading any account, and they can access
+data from other instructions, within the same transaction.
```
ptaffet-jump, line 26: If I remember correctly, signatures verified from a precompile are also charged in the base fee and don't require compute units. Maybe a statement like that doesn't belong in this document.

0x0ece: We have the same sentence in the specs below. I added a comment about cost there, instead of in the motivation.

ptaffet-jump (approved):

proposals/0152-precompiles.md
```diff
@@ -0,0 +1,279 @@
...
+MAX_ALLOWED_PRECOMPILE_SIGNATURES = 8
+```
+
+The precompile instruction `verify` accepts the following data:
+
+```
+struct PrecompileVerifyInstruction {
+  num_signatures:  u8,                       // Number of signatures to verify
+  padding:         u8,                       // Single byte padding
+  offsets:         PrecompileOffsets[],      // Array of `num_signatures` offsets
```
topointon-jump, line 79: `signatures` might be a clearer name?

proposals/0152-precompiles.md
```diff
@@ -0,0 +1,279 @@
...
+If a transaction contains more than 8 precompile signatures, it must fail.
+
+```
+MAX_ALLOWED_PRECOMPILE_SIGNATURES = 8
+```
+
+The precompile instruction `verify` accepts the following data:
+
+```
+struct PrecompileVerifyInstruction {
```
topointon-jump, line 76: Could we also specify these structs size and alignment? 🙏

proposals/0152-precompiles.md
```diff
@@ -0,0 +1,279 @@
...
+  public_key_instruction_index: u16 LE,      // Instruction index to  public key
+  message_offset: u16 LE,                    // Offset to start of message data
+  message_length: u16 LE,                    // Size of message data
+  message_instruction_index: u16 LE,         // Instruction index to message
+}
+```
+
+The behavior of the precompile instruction `verify` is as follow:
+
+1. If instruction `data` is empty, return error.
```
topointon-jump, line 99: It might be worth reiterating that this data should be of the shape `PrecompileVerifyInstruction`.

proposals/0152-precompiles.md
```diff
@@ -0,0 +1,279 @@
...
+2. The first byte of `data` is the number of signatures `num_signatures`.
+3. If `num_signatures` is 0, return error.
+4. Expect (enough bytes of `data` for) `num_signatures` instances of
+   `PrecompileOffsets`.
```
topointon-jump, line 103: It might be worth specifying that if this is not the case the instruction fails.

proposals/0152-precompiles.md
```diff
@@ -0,0 +1,279 @@
...
+Reduce the complexity of existing precompiles, to simplify building
+different validator clients.
+
+## Security Considerations
+
+All 3 changes are straightforward and have no impact on security.
+
+## Backwards Compatibility
+
+All 3 changes require a feature gate.
```
topointon-jump, line 279: Should we do these in one feature gate, or 3 separate ones? I think a single feature gate is probably easier for users to track.

proposals/0152-precompiles.md
```diff
@@ -0,0 +1,279 @@
...
+
+Precompiles are special native programs designed to verify additional
+signatures. Each precompile consists of a single `verify` instruction.
+
+Precompiles are executed right after transaction signature verification,
+they run without the VM and without loading any account. From a cost
+perspective, they're included in the transaction fee (each signature to
+verify counts as a transaction signature), but don't require any compute units.
+
+If a transaction contains more than 8 precompile signatures, it must fail.
```
topointon-jump, line 67: Could we add the the reason behind the number 8 to this SIMD?

topointon-jump (approved):
Nice! Left some minor comments 😄

proposals/0152-precompiles.md
0x0ece: No reason. 8 is what's implemented.

proposals/0152-precompiles.md
0x0ece: Anza said 3 gates.

proposals/0152-precompiles.md
0x0ece: 🤷 It's called this way in the code, I don't think it's worth changing everywhere in the doc.

proposals/0152-precompiles.md
0x0ece: There's no alignment, this is in code, not in vm. Or am I missing something?
Added note on size.

samkim-crypto (approved):

samkim-crypto:
@jacobcreech, it seems like we have reached consensus, but I cannot seem to merge it. Can you help with merging 🙏 ?

joncinque (approved):

lheeger-jump:
Yay great work @0x0ece!

PR #159, opened 2024-07-25
SIMD-0159: Relax Precompile Failure Constraint

jstarry:


jstarry:
This is technically part of https://github.com/solana-foundation/solana-improvement-documents/pull/82 but I think it's good to break up that SIMD a bit

proposals/0159-relax-precompile-failure-constraint.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Motivation
+
+Ensure that block producers are sufficiently incentivized to include transactions
+which make use of precompile instructions by allowing such transactions to be
+included in a block even if verification fails.
+
+## Alternatives Considered
+
+NA
```
apfitzge, line 27: Have we considered to also relax the constraint similarly for all non-fee-payer signatures. i.e. transaction.signatures()[1..] in addition to any pre-compile signatures

jstarry: Yeah that's a good idea as well, I wouldn't mind expanding this SIMD to include those if that's what you're suggesting. It's a very similar constraint.

apfitzge: Yeah think we make the relaxations part of the same SIMD.

jstarry:
@apfitzge any other comments?

proposals/0159-relax-transaction-signature-constraints.md
```diff
@@ -0,0 +1,76 @@
...
+status: Draft
+created: 2024-07-25
+feature: (fill in with feature tracking issues once accepted)
+---
+
+## Summary
+
+The Solana protocol currently prohibits block producers from including
+transactions with any invalid signature whether that signature be the fee-payer
+signature, other transaction-level signatueres or precompile instruction-level
```
apfitzge, line 16: typo:
```suggestion
signature, other transaction-level signatures or precompile instruction-level
```

proposals/0159-relax-transaction-signature-constraints.md
```diff
@@ -0,0 +1,76 @@
...
+    successfully for the transaction to be executed successfully
+- Non-fee payer transaction-level signature:
+    Any signature included in the top-level signatures field in a transaction
+    besides the first listed signature which denotes the fee-payer signature.
+
+## Detailed Design
+
+The Solana protocol will no longer reject blocks with transactions that have
+failed precompile verification or have invalid non-fee payer transacation-level
+signatures. Such transactions will be allowed to be recorded in a block and
```
apfitzge, line 48: Would be good to be more explicit about what "invalid" means here.
For example, I don't think we're relaxing the constraint(s) on number of tx-level signatures matching the num_required_signatures in the header; i.e. "missing" is not included in this now allowed "invalid" category.

jstarry: Sounds good, let me know if my latest update is explicit enough

proposals/0159-relax-transaction-signature-constraints.md
```diff
@@ -44,11 +44,16 @@ valid, the block producer is able to collect fees for including the transaction.
 ## Detailed Design
 
 The Solana protocol will no longer reject blocks with transactions that have
-failed precompile verification or have invalid non-fee payer transacation-level
-signatures. Such transactions will be allowed to be recorded in a block and
-signature verification failures should be handled similarly to how SVM program
-errors are handled. The transaction will be committed, fees deducted, but no
-other account state changes should be persisted.
+failed precompile verification or have non-fee payer transacation-level
```
apfitzge, line 47:
```suggestion
failed precompile verification or have non-fee payer transaction-level
```

apfitzge (approved):
commented on typo, but otherwise lgtm

jstarry:
Added a section about durable nonce transactions still requiring the nonce authority to sign the transaction in order to be committed. I feel this is related enough to add into the SIMD but in reality this nonce authority constraint is one of many nonce related constraints that would prevent a transaction from being committed right now.

proposals/0159-relax-transaction-signature-constraints.md
```diff
@@ -0,0 +1,86 @@
...
+persisted.
+
+The signature count included in the bank hash calculation should still include
+the number of transaction-level signatures for all transactions included in a
+block regardless of verification success.
+
+Since the nonce authority for durable nonce transactions may be different from
+the fee payer, such transactions may need an additional signature verification
+in order to be committed to a block and avoid unauthorized usage of on-chain
+nonce accounts.
```
apfitzge, line 61: If the nonce authority signature is invalid or not present, could we not just skip execution and additionally not advance the nonce?

jstarry: If you don't advance the nonce then you can keep adding the same transaction to future blocks (once it's purged from the status cache) and take more fees

apfitzge: yeah that's true. Realistically you should never sign to pay fees if not also signing as nonce-authority. But if you do it shouldn't be a wallet-draining mistake.

proposals/0159-relax-transaction-signature-constraints.md
```diff
@@ -0,0 +1,111 @@
...
+The signature count included in the bank hash calculation should still include
+the number of transaction-level signatures for all transactions included in a
+block regardless of verification success.
+
+Since the nonce authority for durable nonce transactions may be different from
+the fee payer, such transactions may need an additional signature verification
+in order to be committed to a block and avoid unauthorized usage of on-chain
+nonce accounts. 
+
+### Transaction Errors
```
apfitzge, line 63: as far as I am aware, transaction errors are not under consensus.

Think the "should" in the description below indicates this is reccomended (not required). If that's our reccomendation, we should indicate the reason why we are reccomending it - I assume for consistency in reporting between clients?

jstarry: Correct, updated

apfitzge (approved):

proposals/0159-relax-precompile-failure-constraint.md
jstarry: As discussed offline, we cannot relax the non-fee payer signature constraint because those signatures are not signed by the fee payer and could be tampered with in-flight.

jstarry:
@apfitzge can you take another look? I removed the non-fee payer signature constraint changes.

proposals/0159-relax-precompile-failure-constraint.md
```diff
@@ -0,0 +1,88 @@
...
+## Motivation
+
+Ensure that block producers are sufficiently incentivized to include transactions
+which make use of precompile instructions by allowing such transactions to be
+included in a block even if verification fails.
+
+## Alternatives Considered
+
+Also considered relaxing the constraint of having valid non-fee payer
+transaction-level signatures but those signatures are not signed the fee payer
```
apfitzge, line 29: signed by?

apfitzge (approved):

topointon-jump (approved):

joncinque:
This has approval from both Anza and Firedancer. If no one objects in the meantime, I will merge this a week from today

joncinque (approved):
Thanks for your contributions everyone!

PR #160, opened 2024-07-25
SIMD-0160: Static Instruction Limit

apfitzge:


ptaffet-jump (approved):
Fantastic, thanks! You always say "more than 64," so I think it should be pretty clear, but it might be worth calling out explicitly that exactly 64 is fine. The inequality is strict.

proposals/0160-static-instruction-limit.md
```diff
@@ -0,0 +1,66 @@
...
+in the context of CPIs.
+It is bad user experience to allow creation and submission of transactions that
+have no chance of being executed successfully.
+Several other checks worst-case performance scales with the number of
+top-level instructions in a transaction, so limiting this number can help the
+worst-case performance.
+
+## Alternatives Considered
+
+Do nothing.
```
jstarry, line 35: Why not allow them to be included but limit processing to fee payer validation? So before we load the program cache, we check the instruction count and fail the transaction immediately if it exceeds the limit but still allow fee collection?

apfitzge: I view it similarly to having the number of signatures not match the required signatures field in the header; its' a basic sanitization check that can be checked extremely cheaply.

checking it does not scale with any aspects of the transaction such as number of instructions, number of account lookup tables, etc.

jstarry: So long as we do it really early then I'm fine with prohibiting such transactions from being included. Can you write up a brief summary of this comment thread in the alternatives section?

apfitzge: 4398665f2ae72df927fd532effcb6b145903700a

jstarry (approved):

apfitzge:
@Benhawkins18 can we merge this? we have approval from both FD and Anza teams

jacobcreech (approved):

PR #161, opened 2024-07-26
SIMD-XXXX: Explicit versioning and feature gating of SBPF programs

Lichtso:


proposals/0161-sbpf-feature-gating.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Detailed Design
+
+Every program must signal which SBPF version it was compiled for via
+the `e_flags` field in the
+[ELF header](https://refspecs.linuxfoundation.org/elf/gabi4+/ch4.eheader.html).
+Block explorers are recommended to display this SBPF version field for program
+accounts.
+
+For each SBPF version two feature gates will control its availability:
```
alessandrod, line 57: I'm not sure that we need such a strict system. I think we could probably have several versions active and deployable at the same time, without forcing people to update toolchain and redeploy or pay extra costs. 

Take the following examples:

- dynamic frames: from the vm perspective, it _is_ cheaper to use dynamic frames, because programs potentially page in less memory if an X-deep call stack fits in 1 page vs taking X pages because of fixed framing. We might want to push people to update to increase performance.

- the `sub imm` operand swap change: this doesn't materially change anything. It's a nice thing for us to fix for aesthetic reasons, but it doesn't really change anything for developers. 

I think I'd prefer an ad-hoc system where depending on the impact of a new SBPF version, we decide whether and how to rug earlier versions (in practice I expect this to happen very rarely).

Lichtso: Mandating that all deployments are targeting the newest version does not force anybody to redeploy. It only means that dapp devs have to keep their toolchain updated when they re/deploy. You are right in that it is not necessary to be this strict, but it has the nice effect of allowing us to get more testing exposure for the newest toolchain version.

proposals/0161-sbpf-feature-gating.md
```diff
@@ -0,0 +1,138 @@
...
+complexity again. This feature gate can be shared by multiple SBPF versions,
+effectively deprecating larger blocks of versions in order to reduce the amount
+of redeployment required from dapp developers.
+
+### Example
+
+| SBPF version | becomes deployable and executable | ceases to be deployable |
+| ------------ | --------------------------------- | ----------------------- |
+| v1           |                                   | Feature gate A          |
+| v2           | Feature gate A                    | Feature gate B          |
```
buffalojoec, line 72: Is this purely an example, or are we considering only one version to be able to deploy at a time? It might be helpful to include the number of simultaneous SBPF versions that the network will allow.

Lichtso: Yes the proposal right now says that there can only be one deployable version at a time (but many executable versions). See the discussion above.

proposals/0161-sbpf-feature-gating.md
topointon-jump: I agree that we shouldn't force re-deployments. Mandating that when you do deploy, you can only deploy the latest SBPF version is a reasonable tradeoff between usability and actually getting new versions adopted.

topointon-jump (commented):
Overall looks good. It's not clear how exactly deprecating older versions will work - could we elaborate on this a bit more in the SIMD?

proposals/0161-sbpf-feature-gating.md
```diff
@@ -0,0 +1,146 @@
...
+effectively deprecating larger blocks of versions in order to reduce the amount
+of redeployment required from dapp developers.
+
+### Version discriminator
+
+Currently the protocol deems every value of `e_flags` which is not `0x0020` as
+being SBPF v0 and thus valid. This clearly does not scale for multiple versions
+and must therefore be changed. With the activation of the first feature gate
+which enables the deployment and execution of the first new SBPF version the
+discriminator must switch to directly interpret `e_flags` as SBPF version.
```
topointon-jump, line 73: Can we describe exactly how to parse and interpret this field?

0x0ece (approved):

proposals/0161-sbpf-feature-gating.md
Lichtso: Added a line to describe that via example.

Benhawkins18:
Not a protocol change: closing

PR #162, opened 2024-07-31
SIMD-0162: Remove Accounts `is_executable` Flag Checks

Lichtso:


proposals/0162-remove-accounts-executable-flag-checks.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Alternatives Considered
+
+None.
+
+## Impact
+
+Error codes of these conditions, which are rarely triggered, will change.
+
+The only consensus relevant change is that it will become possible (for
```
HaoranYi, line 131: Will this change open doors for user to withdraw funds from program accounts?

HaoranYi: Also, it seems that this change will open doors for modifying `data` on executable account, and reset `executable` flag itself on executable account. Are there any concerns about this for consensus? Should we document the impact of these changes in the SIMD?

Lichtso: Users will remain unable to: Withdraw funds from program accounts, reset the `is_executable` flag, or change the data directly without going through the loaders program managements instructions. That is because for that you need to be the owner, but programs are owned by a loader. Thus, the loader decides what happens to program accounts.

HaoranYi: Ah. Make sense.

proposals/0162-remove-accounts-executable-flag-checks.md
```diff
@@ -0,0 +1,137 @@
...
+None.
+
+## Detailed Design
+
+This proposal aims to unblock loader-v4 with minimal impact on the ecosystem by
+only removing checks of the `is_executable` flag. A complete removal of the
+flag can be addressed in a subsequent proposal. Thus, the following must remain
+unaffected for now:
+
+- Setting of the `is_executable` flag during program deployment
```
buffalojoec, line 65: Can we specify that this is for both loaders - v3 and v4?

proposals/0162-remove-accounts-executable-flag-checks.md
```diff
@@ -0,0 +1,137 @@
...
+the `is_executable` flag set
+
+These checks of the `is_executable` flag must be removed:
+
+- `ExecutableLamportChange` during execution (transaction succeeds instead)
+
+These checks of the `is_executable` flag are irrelevant to consensus as
+transactions which try to invoke accounts, which do not contain programs,
+continue to fail without them. They should still be removed because all
+implementations should aim to produce the same error codes:
```
buffalojoec, line 79: Might be worth mentioning that changing these error codes is non-consensus-breaking. It's implied here, but could be good to be specific.

proposals/0162-remove-accounts-executable-flag-checks.md
```diff
@@ -0,0 +1,137 @@
...
+- `IncorrectProgramId` for unrecognized built-in programs
+
+All in all, the following error messages related to invocation of program
+accounts will be coalesced into `UnsupportedProgramId`:
+
+- `InvalidProgramForExecution`
+- `AccountNotExecutable`
+- `InvalidAccountData`
+- `IncorrectProgramId`
+- `UnsupportedProgramId`
```
buffalojoec, line 121: Will these be deprecated, then?

buffalojoec (commented):
Looks good. I'm just curious how buffer accounts will be handled going forward. I'm assuming that it will simply be the header of the account data `UpgradeableLoaderState::Buffer`, but maybe mentioning how loaders v3 and v4 distinguish between buffer accounts and programs might be useful, even if unchanged, with this flag removed.

proposals/0162-remove-accounts-executable-flag-checks.md
Lichtso: Some of them (such as `InvalidAccountData`) can still be produced by other built-in / core programs, e.g. program management instructions.

buffalojoec: Okay sounds good. No change needed here imo, then.

proposals/0162-remove-accounts-executable-flag-checks.md
Lichtso: Loader-v4 will get its own SIMD explaining it in full detail.

buffalojoec (approved):
Nice, looks good!

HaoranYi (approved):
I like it. lgtm!

With this SIMD, it will greatly simply [SIMD-0094](https://github.com/solana-foundation/solana-improvement-documents/pull/94/files#top), which no longer need to worry about "simulating" executable flag, after deprecating 'executable' flag update in bpf loader.

topointon-jump (approved):
It's great to simplify things here!

A bit concerned that it's easy to make mistakes here and unintentionally introduce consensus violations, so there will need to be very thorough testing around the failure cases. We have a lot of tooling which can help here.

joncinque:
It looks like there's consensus here. If no one objects within the next week, I'll merge this on 25 November.

proposals/0162-remove-accounts-executable-flag-checks.md
```diff
@@ -0,0 +1,139 @@
...
+only removing checks of the `is_executable` flag. A complete removal of the
+flag can be addressed in a subsequent proposal. Thus, the following must remain
+unaffected for now:
+
+- Setting of the `is_executable` flag during program deployment in loader-v3
+- Calculation of the account hashes
+- Minimization of snapshots
+- Serialization of instruction accounts `is_executable` flag for dapps
+- CPI ignores changes made by the caller to instruction accounts which have
+the `is_executable` flag set
```
jstarry, line 70: I'm confused why this was left in because by allowing executable accounts to be written, it becomes important that any changes to executable accounts are handled properly.

Without this change, not only does CPI ignore changes made my the caller, it also ignores changes made by the callee. In the latter case, when the caller returns, the runtime will fail the transaction because it will think that the caller tried to revert the callee changes.

Also imagine if during a CPI, both the caller and callee program sent a lamport to an executable account from a non executable writable account. The caller sends the lamport first, then when the callee is cpi'd it hasn't seen the lamport change so when it sends a lamport, the balance is updated to the value that was already updated to by the caller. When the caller returns, the executable account has only received 1 lamport but the sender non executable account has sent two lamports. This leads to a surprising lamport imbalance error.

Lichtso: Sorry for the ambiguous formulation. What is meant is not that `is_executable` flag can be set (you can not change the `is_executable` flag inside the VM anyway), instead it means that CPI behavior was dependent on the state of the `is_executable` for instruction accounts, and we do not change any of that in this SIMD ("the following must remain unaffected for now").

Lichtso: We had to leave it in because direct mapping would break otherwise. We can only remove it after activating and cleaning up the feature gate of direct mapping.

jstarry: > Sorry for the ambiguous formulation. What is meant is not that is_executable flag can be set (you can not change the is_executable flag inside the VM anyway)

I think the SIMD was clear. I wasn't thinking you meant that the `is_executable` flag can be set. I meant that other fields in executable accounts can be modified now but this CPI behavior would cause surprising errors when executable accounts are modified in otherwise valid ways (e.g. adding lamports as you noted in the impact section)

> We had to leave it in because direct mapping would break otherwise. We can only remove it after activating and cleaning up the feature gate of direct mapping.

Surely there was some way to make it work before direct mapping is enabled without breaking direct mapping? Especially if we can remove after direct mapping is activated?

PR #165, opened 2024-08-12
Initial draft of Async Vote Execution SIMD.

wen-coding:
Add initial draft of Async Vote Execution SIMD.

### DIFF
--- a/proposals/0154-async-vote-execution.md
+++ b/proposals/0154-async-vote-execution.md
@@ -1,0 +1,109 @@
+---
+simd: '0154'
+title: Async Vote Execution
+authors:
+  - Wen Xu
+category: Standard
+type: Core
+status: Draft
+created: 2024-08-11T00:00:00.000Z
+feature: null
+supersedes: null
+superseded-by: null
+extends: null
+---
+## Summary
+Separate the execution of vote and non-vote transactions in each block. The
+vote transactions will be verified and executed first, then the non-vote
+transactions will be executed later asynchronously to finalize the block.
+
+## Motivation
+Currently the vote transactions and non-vote transactions are mixed together in
+a block, the vote transactions are only processed in consensus when the whole
+block has been frozen and all transactions in the block have been verified and
+executed. This is a problem because slow running non-vote transactions may affect
+how fast the votes are processed and then affect the ability of consensus to
+pick the correct fork.
+
+With different hardware and running environment, there will always be some
+difference on speed of transaction execution between validators. Generally
+speaking, because vote transactions are so simple, the variation between vote
+execution should be smaller than that between non-vote executions. Therefore,
+if we only execute vote transactions in a block before voting on the block,
+it is more likely validators can reach consensus faster.
+
+Even with async vote execution, forks can still happen because of
+various other situations, like network partitions or mis-configured validators.
+This work just reduces the chances of forks caused by variance in non-vote
+transaction executions.
+
+The non-vote transactions do need to be executed eventually. Even though it's
+hard to make sure everyone executes every block within 400ms, on average majority
+of the cluster should be able to keep up.
+
+## New Terminology
+- `VED`: Vote Execution Domain, Vote transactions and all its dependencies (e.g.
+fee payers for votes).
+- `VED Bankhash`: The hash calculated after executing only vote transactions in
+a block. If there are no votes, use default hash.
+- `UED`: User Execution Domain, currently everything other than votes. We may
+have more domains in the future.
+- `UED Bankhash`: The hash calculated after executing only non-vote transactions
+in a block. If there are no non-vote transactions, use default hash.
+
+## Detailed Design
+
+### Allow leader to skip execution of transactions (Bankless Leader)
+There is already on-going effort to totally skip execution of all transactions
+when leader pack new blocks. See SIMD 82, SIMD 83, and related trackers:
+https://github.com/anza-xyz/agave/issues/2502
+
+Theoretically we could reap some benefit without Bankless Leader, the leader
+pack as normal, while other validators only replay votes first, then later
+execute other transactions and compare with the bankhash of the leader. But in
+such a setup we gain smaller speedup without much benefits, it is a possible
+route during rollouts though.
+
+### Separating vote transactions and dependencies into a different domain
+To make sure vote transactions can be executed independently, we need to
+isolate its dependencies.
+
+#### Remove clock program's dependency on votes
+Introduce new transaction `ClockBump` to remove current clock program's
+dependency on votes.
+
+The transaction `ClockBump` is sent by a leader with at least 0.5% stake
+every 12 slots to correct the clock drift. A small script can be used to
+refund well-behaving leaders the cost of the transactions.
+
+#### Split vote accounts into two accounts in VED and UED respectively
+We need to allow users move money in and out of the vote accounts, but
+we also need the vote accounts to vote in VED. So there will be two accounts:
+- `VoteTowerAccount`: tracks tower state and vote authority, it will be
+in `VED`, it is updated by vote transactions and tracks vote credits.
+- `VoteAccount`: everything else currently in vote accounts, it will be
+in `UED`, users can move funds in and out of `VoteAccount` freely.
+
+The two accounts are synced every Epoch when the rewards are calculated.
+
+### Separate the VED and UED Domains
+- Only Vote or System program can read and write accounts in `VED`
+- Other programs can only read accounts in `VED`
+- Users can't directly access accounts in `VED` but they can move accounts
+from `VED` to `UED` and vice versa. Moving accounts from one domain to
+another takes 1 Epoch, and the migration happens at Epoch boundary
+
+### Enable Async Vote Executions
+1. The leader will no longer execute any transactions before broadcasting
+the block it packed. We do have block-id (Merkle tree root) to ensure
+everyone receives the same block.
+2. Upon receiving a new block, the validator computes the `VED bankhash`,
+then vote on this block and also gives its latest `UED bankhash` on the
+same fork. The `UED bankhash` will most likely be hash of an ancestor of
+the received block.
+3. A block is not considered Optimistically Confirmed or Finalized until
+some percentage of the validators agree on the `UED bankhash`.
+4. Add assertion that confirmed `UED bankhash` is not too far away from the
+confirmed `VED bankhash` (currently proposed at 1/2 of the Epoch)
+5. Add alerts if `UED bankhash` differs when the `VED bankhash` is the same.
+This is potentially an event worthy of cluster restart.


aeyakovenko:
High level seems right

proposals/0165-async-vote-execution.md
```diff
@@ -0,0 +1,120 @@
...
+every 12 slots to correct the clock drift. A small script can be used to
+refund well-behaving leaders the cost of the transactions.
+
+#### Split vote accounts into two accounts in VED and UED respectively
+
+We need to allow users move money in and out of the vote accounts, but
+we also need the vote accounts to vote in VED. So there will be two accounts:
+
+- `VoteTowerAccount`: tracks tower state and vote authority, it will be
+in `VED`, it is updated by vote transactions and tracks vote credits.
```
carllin, line 92: A flow diagram/description of how users will pay for vote fees would be useful here. A step by  step from vote account creation, to funding the account, to which domains it's created/moved into, etc.

proposals/0165-async-vote-execution.md
```diff
@@ -0,0 +1,192 @@
...
+In this step the validators will check the fee payers of vote transactions. So
+each vote transaction is executed twice: once in the optimistical voting stage
+*without* checking fee payer, and once in this stage *with* checking fee payer.
+If a staked validator does not have vote fee covered for specific votes, we
+will not accept the vote today, while in the future we accept the vote in fork
+selection, but does not actually give vote credits because the transaction
+failed.
+
+A side benefit of this design is that the sysvars have all been calculated and
+determined in the optimistical voting stage, also the fork of blocks are known,
```
aeyakovenko, line 154: Typo: s/optimistical/optimistic

proposals/0165-async-vote-execution.md
```diff
@@ -0,0 +1,192 @@
...
+* `ephemeral_slot`: This is the slot where the ephemeral_hash is calculated.
+
+This step is optimistic in the sense that validators do not check the fee
+payers when executing the vote transactions in a block. They assume vote
+transactions will not fail due to insufficient fees, apply the execution
+results to select the correct fork, then immediately vote on the bank with
+only the hash result of the vote transactions.
+
+This is safe because the list of validators and their corresponding stake
+has already been determined at the beginning of the Epoch. The stake we used
```
aeyakovenko, line 108: The leader schedule for the current epoch was computed at the beginning of the previous epoch.  Not this epoch.

The fork choice stake weights should match the LS computation, otherwise there will be a stall waiting for async to finish at the start of the epoch.

aeyakovenko:
Async Program Execution: APE 🦧🦧🦧

proposals/0165-async-vote-execution.md
wen-coding: Done.

proposals/0165-async-vote-execution.md
wen-coding: Done.

wen-coding:
> Async Program Execution: APE 🦧🦧🦧

The intention is limit the scope of this SIMD to just the vote part, because we have several SIMDs for the non-vote part.

proposals/0165-async-vote-execution.md
```diff
@@ -0,0 +1,218 @@
...
+This allows us to more quickly converge on one chain of blocks, so that
+validators don't have to execute any blocks not on selected fork. This saves
+CPU and memory resource needed in replay, it also ensures that the cluster
+will have fewer forks that are caused by slow transaction execution.
+
+## Motivation
+
+Currently the vote transactions and non-vote transactions are mixed together in
+a block, the vote transactions are only processed in consensus when the whole
+block has been frozen and all transactions in the block have been verified and
```
AshwinSekar, line 32: nit: votes are processed per entry batch, doesn't have to wait for the bank to be frozen. as a result we still partially process votes even if the block later ends up dead.
e.g. agave's impl https://github.com/anza-xyz/agave/blob/1f06fbdbe3b72f330f50ad93a15c1116f5021392/ledger/src/blockstore_processor.rs#L177

proposals/0165-async-vote-execution.md
```diff
@@ -0,0 +1,218 @@
...
+
+Two new fields will be added to `TowerSync` vote transaction:
+
+* `replay_tip_hash`: This is the hash as we know it today.
+* `replay_tip_slot`: This is the slot where the replay tip hash is calculated.
+
+The `hash` and `slot` in the `TowerSync` transaction will be updated to
+the vote only hash. The vote only hash is calculated as follows:
+
+1. Sort all vote accounts with non-zero stake in the current epoch by
```
AshwinSekar, line 103: non-zero in the previous epoch right? since stakes are offset by 1 epoch?

proposals/0165-async-vote-execution.md
```diff
@@ -0,0 +1,218 @@
...
+vote account pubkey.
+
+2. Calculate vote account hash by hashing (vote account pubkey,
```
AshwinSekar, line 106: nit: specify the hash fn

proposals/0165-async-vote-execution.md
```diff
@@ -0,0 +1,218 @@
...
+serialized vote state) in the order given.
```
AshwinSekar, line 107: do we have to hash the whole vote state? only the tower / root will change

proposals/0165-async-vote-execution.md
```diff
@@ -0,0 +1,218 @@
...
+that the ephemeral hash vote doesn't vote on a block which will be marked
+dead later. Currently a block can be dead for the following reasons:
+
+1. Unable to load data from blockstore
+2. Invalid block (wrong number of ticks, duplicate block, bad last fec, etc)
+3. Error while set root
+4. Invalid transaction
+
+For the first two, the same check can be performed computing ephemeral hash.
+We will set root on a bank only when it has full hash computed later, so the
```
AshwinSekar, line 137: what does "set root" mean in this context?

proposals/0165-async-vote-execution.md
```diff
@@ -0,0 +1,218 @@
...
+replaying blocks on the selected fork. The replay process is the same as today,
+all transactions (vote and non-vote) will be executed to determine the final
+bankhash. The computed bankhash will be attached to vote instructions. So we
+can still detect non-determinism (same of set of instructions leading to
+different results) like today, only that maybe at a later time.
+
+To guarantee the blockchain will halt when full replay is having problems, we
+propose:
+
+1. If full replay is behind vote only replay by more than 1/2 epoch or vice
```
AshwinSekar, line 169: re: "vice versa" can full replay run before vote only replay? I think we should ensure that full replay cannot run on a block that has not been vote only replayed

proposals/0165-async-vote-execution.md
```diff
@@ -0,0 +1,218 @@
...
+everyone receives the same block.
+2. Upon receiving a new block, the validator executes only the vote
+transactions without checking the fee payers. The result is immediately
+applied in consensus to select a fork. Then votes are sent out for the
+selected fork with the `Vote only bankhash` for the tip of the fork and the
+most recent `Replay tip bankhash`.
+3. The blocks on the selected forks are scheduled to be replayed. When
+a block is replayed, all transactions are executed with fee payers checked.
+This is the same as the replay we use today.
+4. Optimisticly confirmed or finalized on `Vote only bankhash` and `Replay tip
```
AshwinSekar, line 197: Might want to specify the listener changes here, we expect clients to track commitment statuses on both hashes? What does it mean to be OC/finalized on the replay tip bankhash, will invalid fee payer votes count towards this?

proposals/0165-async-vote-execution.md
```diff
@@ -0,0 +1,218 @@
...
+bankhash` will be both exposed through RPC so users can select accordingly.
+5. Add assertion that confirmed `Replay tip bankhash` is not too far away from
+the confirmed `Vote only bankhash` (currently proposed at 1/2 of the Epoch)
+6. Add alerts if `Replay tip bankhash` differs when the `Vote only bankhash` is
+the same. This is potentially an event worthy of cluster restart. If more than
+1/3 of the validators claim a different `Replay tip bankhash`, halt and exit.
+
```
AshwinSekar, line 204: Specify how the feature program will look with APE:
- Normal pathway features will work same as before
- APE features will require 2 epochs of advance activation

AshwinSekar, line 204: Specify the modifications to fork choice. Fork choice will now *only* read vote only replayed votes. Needs to be keyed by block-id or vote only hash instead.

AshwinSekar, line 204: Specify the modifications to repair, we will now ingest vote only replayed votes in repair weighting. Also changes to repair peer selection, is EpochSlots updated after vote only replay now?

proposals/0165-async-vote-execution.md
AshwinSekar: Also specify changes to the commitment service, do we still aggregate when we vote on the block / root a block? or is there a separate pathway taken only when we full replay the block.

AshwinSekar (commented):
high level looks good, just needs some more specification

proposals/0165-async-vote-execution.md
wen-coding: Yeah, now I think about it, it's less about processing votes as fast as possible, more about a lot of things don't happen until the block is frozen, changes this part.

proposals/0165-async-vote-execution.md
wen-coding: A bit worried the vote authority isn't copied correctly in some implementation, not a lot else other than the tower/root I guess.

proposals/0165-async-vote-execution.md
wen-coding: There is no reason a full replay can't start before the bank is vote only frozen, they write into completely different set of states.

AshwinSekar: makes sense, we should outline the interaction here:
> Once a validator has determined the fork it will vote on, it can prioritize
replaying blocks on the selected fork

maybe point out that if we haven't completed vote replay we replay all available forks equally. also outline what happens in the case that we're full replaying a fork but end up vote replaying and voting on a different fork.

proposals/0165-async-vote-execution.md
AshwinSekar: what I meant is that consensus uses the stakes from the previous epoch, so if we're only interested in processing consensus votes we should match the logic here. Basically only consider vote accounts with non-zero stake in the [previous epoch](https://github.com/anza-xyz/agave/blob/9d9be7169209c981fb29ecabada71d02ca885c6b/core/src/consensus/heaviest_subtree_fork_choice.rs#L1257)

If we ever implement the handoff then we should consider the previous previous epoch as well

wen-coding: We should invent a name for this thing, this is so confusing. I think you and me both mean "that epoch stakes we use for voting in this Epoch", right? Then let's say in the future we decide to use epoch stakes of Epoch X for voting in Epoch X+2 we don't need to chase down all documents again and change it from "previous epoch" to "previous previous epoch". We just stick to one name.

I'm confused: why don't we need stakes for the previous epoch (previous previous epoch in your description) if we don't implement handoff?

proposals/0165-async-vote-execution.md
wen-coding: Specified sha256.

proposals/0165-async-vote-execution.md
wen-coding: I think what you described is reasonable, but I prefer not to specify in what order the full replay should happen in the protocol. Anza and Jump may decide to use different fork weight algorithm to pick which block to send to full replay next, I think the protocol would still work. And from what I heard, playing multiple forks at the same time or playing into the distant future is also a possibility.

proposals/0165-async-vote-execution.md
wen-coding: This is SetRootError in BlockstoreProcessorError. I think we will keep this check in the new world.

AshwinSekar: it doesn't mark the block dead though, just kills the replay thread.

proposals/0165-async-vote-execution.md
```diff
@@ -131,12 +131,13 @@ dead later. Currently a block can be dead for the following reasons:
...
-3. Error while set root
+3. Error while performing set root on the new root
 4. Invalid transaction
 
-For the first two, the same check can be performed computing ephemeral hash.
-We will set root on a bank only when it has full hash computed later, so the
-behavior will be the same as now.
+For the first three, the same check can be performed computing vote only
+hash. We will add a new check that the new root must be vote only replayed
+and fully replayed, this may mean the tower has more than 32 slots
```
AshwinSekar, line 139: Can you explain this part? We avoid setting the root if the block has not been fully replayed, so the tower can contain thousands of slots?
Why can't we set the root like we do today and modify pruning to keep up to the latest fully replayed root? Essentially moving the smr to prune by to be the replayed-smr

proposals/0165-async-vote-execution.md
wen-coding: Hmm that's true, removed this check

proposals/0165-async-vote-execution.md
wen-coding: We need to save the slots which should be kicked out of tower but haven't been fully replayed somewhere. We can only save snapshot on a fully replayed slot, and we need to keep ancestor relationships between blocks. It could be somewhere not on the tower, are you mostly worried about the space cost if we keep it on the tower?

proposals/0165-async-vote-execution.md
wen-coding: Added to User visible changes.

proposals/0165-async-vote-execution.md
wen-coding: Added to user visible changes.

wen-coding: It's covered in "The result is immediately applied in consensus to select a fork." I think whether we key our internal data structure by block-id or hash is implementation details we don't need in SIMD. I've added that to point 2 in "Enable Async Vote Executions".

proposals/0165-async-vote-execution.md
AshwinSekar: we have all that in bank forks right? the tower is just a list of slot #s

wen-coding: We do. I think the main question is whether we want it in vote transactions. We do want replay tip in there, but maybe not the full list of slots.

proposals/0165-async-vote-execution.md
wen-coding: EpochSlots just specified which slots I have to serve repair right? I don't see why we can't update it after vote only replay.

wen-coding: I think repair weighting is also internal choice Anza and Jump may have different choices on, we surely should do what you suggested but we don't need it in the SIMD if it's not visible on the line.
Added that we will add banks to EpochSlots after vote only replay.

HaoranYi:
After reading the SIMD, this is my take-away on adopting chili-pepper into
async execution if we want to.

Overall, we can add "chili-pepper" check into vote only execution in general,
similar to the CU check. However, there is one important difference between CU
check and chili-pepper check. CU check is "stateless", while "chili-pepper"
check is "stateful". Current block's chili-pepper value depends on previous
block's chili-pepper values. Therefore, if we want to incorporate
"chili-pepper" check into the vote only execution, we will need to move the
update the "chili-pepper" from the end of full execution to the end of vote
only execution. And a side effect is that updating all account's chili pepper
state in the block will slow down vote only execution. We probably need to
profile it to see how much the slow down is.

PR #168, opened 2024-08-21
template: move alternatives section below detailed design

buffalojoec:
#### Problem
The current template for SIMDs places the "Alternatives Considered" section before any details about the proposed design, including "Detailed Design" in particular.

This can be confusing for readers, since alternative designs usually reference the detailed design proposed in the "Detailed Design" section for contrast.

#### Summary of Changes
Move the "Alternatives Considered" section below "Detailed Design". With this setup, readers can consume the proposed design, then read the alternatives, after digesting the proposed design details. This should make the context much more clear when a reader reaches the "Alternatives Considered" section.

### DIFF
--- a/XXXX-template.md
+++ b/XXXX-template.md
@@ -24,11 +24,6 @@ A brief summary of what the feature is.
 Why are we doing this? What use cases does it support? What is the expected
 outcome?
 
-## Alternatives Considered
-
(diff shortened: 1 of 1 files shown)


Benhawkins18:
This makes sense to me. @jacobcreech any concerns?

Benhawkins18 (approved):

PR #170, opened 2024-08-26
SIMD-0170: Reserve minimal CUs for builtins

tao-stones:


buffalojoec (commented):
The motivation makes sense to me, but does this proposal take into account builtin programs who CPI to other builtin programs? The address lookup table program, for example, consumes default CUs of 750, but a few instructions (create & extend) will CPI to the system program.

It might be difficult and/or brittle to hard-code all default CUs for builtin programs including via CPI. If we were going to go this route, I'd advocate for moving away from blanket CU usage across all instructions, and instead configuring a CU value per-instruction, which might make this benchmarking process a little safer.

tao-stones:
> The motivation makes sense to me, but does this proposal take into account builtin programs who CPI to other builtin programs? The address lookup table program, for example, consumes default CUs of 750, but a few instructions (create & extend) will CPI to the system program.
> 
> It might be difficult and/or brittle to hard-code all default CUs for builtin programs including via CPI. If we were going to go this route, I'd advocate for moving away from blanket CU usage across all instructions, and instead configuring a CU value per-instruction, which might make this benchmarking process a little safer.

Excellent point. This proposal calls for "A builtin instruction that might call other instructions (CPI) would fail without explicitly requesting more CUs." (In `Detailed Design` section, Example 2).

Budget was moved from "per instruction" to "per transaction", it might be good idea to revisit it. Another possible option to handle "builtin program that CPIs" is the second one in `Alternatives Considered`. But asking user to explicitly request cu limit seems to be most straightforward atm.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
```diff
@@ -0,0 +1,127 @@
...
+
+   [Example 1](https://github.com/anza-xyz/agave/pull/2746/files#diff-c6c8658338536afbf59d65e9f66b71460e7403119ca76e51dc9125e1719f4f52R13403-R13429):
+   A transaction consists of a Transfer and an expensive non-builtin instruction.
+If the non-builtin instruction requires more than DEFAULT_INSTRUCTION_COMPUTE_UNIT_LIMIT
+CUs, the over-budgeting for Transfer allows the expensive instruction to
+execute to completion, forcing an upward adjustment.
+
+   [Example 2](https://github.com/anza-xyz/agave/pull/2746/files#diff-c6c8658338536afbf59d65e9f66b71460e7403119ca76e51dc9125e1719f4f52R13431-R13455):
+   A builtin instruction that might call other instructions (CPI) would fail
+without explicitly requesting more CUs. However, this isn't currently happening.
```
apfitzge, line 90: Under this proposal, this transaction would fail, right?

That seems to indicate to me that cost-model is broken for builtins (even with this proposal).
If we are claiming these builtins have known static costs, we should use those costs instead of creating a weird UX where some native/builtin instructions **require** the use of some other one.

In addition to this, why can we not fix the cost-tracking done in the VM?
If we are going to continue claiming in cost-model that some builtin program always uses X CUs, then that's what it should use - regardless of internal behavior which may CPI.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Alternatives Considered
+
+- One possible alternative approach would be to maintain the current allocation
+of compute budget for builtin instructions but add logic to the cost tracker
+to account for the discrepancy during tracking. However, this would add
+complexity and could introduce additional corner cases, potentially leading to
+more issues.
+
+- Another alternative would be to treat builtin instructions the same as other
```
apfitzge, line 53: there's already plans underway to move the builtins to bpf.

at that point it'd effectively be the same as this alternative since builtins would be treated the same as other user programs - (?) possible exception for compute-budget which is a configuration not an instruction (imo).

Thus we will need some way to address the over-reserving described here.
There have been discussions on ramping down the default ix CUs over time, what is the issue with such an approach now if it is eventually the path we have to take?

proposals/simd-0170-builtin-instruction-cost-and-budget.md
```diff
@@ -0,0 +1,127 @@
...
+being executed when they should fail due to exceeding the budget. Consequently,
+the cost tracker may need to account for additional compute units toward block
+limits after transaction execution, potentially producing blocks that exceed
+those limits.
+
+Furthermore, maintaining consistency in transaction costs between the banking
+stage and SVM would simplify the code logic and make reasoning more
+straightforward.
+
+## Alternatives Considered
```
apfitzge, line 45: If we were to take this approach of using static costs per builtin program, and alternative is to just fix the implementation of CU usage in execution.

If our cost-model says that builtin program Z always uses X CUs, then that should be what is actually used by the execution, regadless of what it does internally, including CPI.

tao-stones: > If we were to take this approach of using static costs per builtin program, and alternative is to just fix the implementation of CU usage in execution.

Acknowledging that several alternatives exist, this proposal focuses on the specific issue: "If the VM must consume builtin.default_compute_units, then it should allocate exactly that amount for the builtin, rather than a fixed 200,000 units." This approach addresses block overpacking in the short term, while longer-term solutions are still being explored.

> If our cost-model says that builtin program Z always uses X CUs, 

To clarify, it's the cost model *being told* that a builtin program Z always uses X CUs.

> then that should be what is actually used by the execution, regadless of what it does internally, including CPI.

Yea,  that'd be ideal. I am not 100% how and when this will happen tho. Happy to discuss more.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
tao-stones: Not sure if moving builtins to bpf solves the particular issue (mentioned above) this proposal aims in immediate term, cause still need to allocate "proper" budget for builtin, or any instruction in that scenario.

But lower default cu/ix will help - lowered default per ix will ease over reserving. What would the new "default" be is arbitrary. In extreme case, if it is lowered to `0`, we'd be in the ideal world that every transaction will have to set (in other words, configure) its cu limit.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
tao-stones: > Under this proposal, this transaction would fail, right?

Yes, it'd fail under this proposal, kinda of intentionally.

> If we are claiming these builtins have known static costs, we should use those costs instead of creating a weird UX where some native/builtin instructions require the use of some other one.

Yea, builtins that CPI to other instructions is a tricky case, perhaps it can't be said to have a static cost, unless the number of CPIs are static. I don't have a good immediate solution for it, but asking/pushing for every transaction to explicitly call set_compute_unit_limit.

> In addition to this, why can we not fix the cost-tracking done in the VM?

To reduce confusion, let's use "cost tracking" for block packing side of business, and "budget consuming" when relates to VM takes CUs.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
apfitzge: right, but I think there's simplicity in saying "all programs are treated the same" vs "builtins are treated differently AND these specific builtin instructions are treated even different from that"

tao-stones: Absolutely. to be at "all programs are treated the same" in terms of allocating budget for VM, it needs to shift to 100% request-based mode. More tools need to be built to help devs to request cu limits easily and accurately, also lower the default cu/ix when < 100% transaction requests cu limit. This proposal serves stopgap before all that land.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
buffalojoec: > In addition to this, why can we not fix the cost-tracking done in the VM?
> If we are going to continue claiming in cost-model that some builtin program always uses X CUs, then that's what it should use - regardless of internal behavior which may CPI.

Technically the VM is doing its designated job correctly, since the `InvokeContext` is created per-instruction, and it handles the accrual of CU usage across CPI stack frames.

I feel like the solution would have to be at the SVM (runtime) level, where we keep some list of builtin CUs that are reserved for each instruction. Something like this might do it (sorry, bunch of pseudo-code here).

```rust
// ALT program:
mod address_lookup_table_program {
    const CUS_BASE_CREATE_LOOKUP_TABLE: u64 = 750;
    const CUS_TOTAL_CREATE_LOOKUP_TABLE: u64 = 750 + 150 + 150; // Includes CPI to System.

    enum Instruction {
        CreateLookupTable,
        /* ... */
    }

    impl Instruction {
        fn get_cus_to_allocate(ix_data: &[u8]) -> BuiltinComputeUnits {
            match ix_data.first() {
                Some(&0) => CUS_TOTAL_CREATE_LOOKUP_TABLE,
                /* ... */
            }
        }
    }

    // Processor for `CreateLookupTable`:
    fn process_create_lookup_table(/* ... */) {
        // Base CUs. CPIs will consume their corresponding base CUs.
        consume_cus(CUS_BASE_CREATE_LOOKUP_TABLE);
    }
}


// SVM:
fn process_tx(tx: SVMTransaction, /* ... */) {
    let cus_to_allocate = tx
        .instructions()
        .iter()
        .map(|ix| {
            if ix.program_id == address_lookup_table_program::id() {
                address_lookup_table_program::Instruction::get_cus_to_allocate(ix.data)
            }
            // Default allocation 200k.
            200_000
        })
        .sum();
    allocate_cus_for_transaction(cus_to_allocate);
}
```

proposals/simd-0170-builtin-instruction-cost-and-budget.md
apfitzge: (agave implementation specific)
I believe we'd need to modify the `declare_process_instruction` macro so that we can somehow have a call that doesn't consume CUs and one that does - either through some flag or separate call.
Calls from inside our builtins would then need to call the correct one that does not consume.
Would also need to make sure that actual user-code cannot call into the non-consuming version.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
tao-stones: There's possibility "builtin instruction" can be deprecated all together. But if it stays, then allocating budget by builtin "instructions" not by "program" would be next step. ZK program [consumes CUs by instruction](https://github.com/anza-xyz/agave/blob/master/programs/zk-token-proof/src/lib.rs#L186), not by program_id.

buffalojoec: > ZK program [consumes CUs by instruction](https://github.com/anza-xyz/agave/blob/master/programs/zk-token-proof/src/lib.rs#L186), not by program_id.

Awesome! I'm in favor.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
buffalojoec: @apfitzge you can just provide zero and it becomes a no-op.
https://github.com/anza-xyz/agave/blob/a72f981370c3f566fc1becf024f3178da041547a/program-runtime/src/invoke_context.rs#L71-L76

A good standard would be to follow the [ZK proof program's instructions](https://github.com/anza-xyz/agave/blob/master/programs/zk-token-proof/src/lib.rs#L186) (shared previously by @tao-stones). We would just need to sort out how to represent the constants to get some kind of builtin-program dictionary of CUs per instruction.

apfitzge: but then I think then it'd always consume 0, since it is a macro defining a function, not a function itself.

ptaffet-jump:
Overall looks pretty good. I agree with Andrew's comment 

> If our cost-model says that builtin program Z always uses X CUs, then that should be what is actually used by the execution, regadless of what it does internally, including CPI.

As for the UX strangeness that this causes, I'd propose one of the following two:

1. Expose a CPI-inclusive number to the cost tracker and a CPI-exclusive number to the VM. This may have to be per-instruction then.
2. Make CPIs from native programs not consume CUs.

I'd be okay with either, though there's some complexity involved with per-instruction costs (suppose you distinguish instruction by first byte, then what if the instruction data is empty or not one of the known bytes? You throw the transaction out?).

apfitzge:
Of @ptaffet-jump's 2 suggestions, I don't feasibly see how the per-instruction will work well. Especially given the last case he mentioned - what if the ix variant is invalid?

For option 2, if @tao-stones agrees it is reasonable approach, think we'll need to bring in someone from agave VM team to comment on how difficult it would be. And also how unsafe it would be - we definitely do not want to create a bug where user txs can CPI into native programs for free.
Just CPI **from** native programs being free.

tao-stones:
Thanks for all the helpful inputs. It looks like the primary issue is handling builtins that make CPIs without introducing confusing or inconsistent user experiences. The potential solutions are converging too. @ptaffet-jump's option 1 is similar to @buffalojoec pseudo-code, his option 2 is inline with @apfitzge suggestion. 

I am inclined toward the first option, which avoids introducing special cases into the VM and instead focuses on making builtin programs more transparent about their compute requirements, and most the logics are implemented within `builtin-default-costs` crate:

1. Changes to builtin programs: 
  - Expose `DEFAULT_COMPUTE_UNIT` per instruction (instead of currently per program), similar to ZK as mentioned above.
  - Expose CPI instruction Array. Additionally, builtin programs should expose an array of instructions they invoke via CPIs. For example, `create_address_lookup_table` instruction that makes three CPIs to the `system` program would expose `[system_ix, system_ix, system_ix]`. Others might expose an empty array. 
  - This makes builtin programs more transparent about what they do.
  
2. Changes to `builtin-default-costs` crate:
  - Dictionary of Instruction Costs and CPIs: Maintain a static dictionary with the structure `<instruction, {ix_default_compute_units, cpi_list}>` to store the default compute units and associated CPI instructions for each builtin instruction.
  - Helper Function for CU Calculation that calculates the appropriate number of compute units to allocate per instruction based on the dictionary data, pseudo: 
  ```
  fn get_cu_for_allocation( &ix ) -> Result<u64> {
      let entry = get_dictionary_entry( ix ) ? ;
      let mut allocation_size = entry.value.ix_default_compute_units;
      for cpi_ix in entry.value.cpi_list {
          let cpi_ix_cost = get_cu_for_allocation( cpi_ix ) ? ;
          alloacation_size += cpi_ix_cost;
      }
      allocation_size
  }
  ```
  
3. Call-Site Implementation:
  - Instruction Type Lookup: at the call-site, such as within the compute budget or cost model, determine the type of builtin instruction. If the instruction type cannot be determined, returns `Err(invalid_instruction_data_error)`.
  - CU Allocation Calculation: If the instruction type is valid, use function provided by the builtins-default-costs crate to calculate the correct amount of compute units to allocate for that instruction.
  - transaction's `program_id_index` is checked with above process only once, result is cached for reuse.
    
4. No Changes to VM:

wdyt?

apfitzge:
All sounds reasonable except the error handling here:

> Instruction Type Lookup: at the call-site, such as within the compute budget or cost model, determine the type of builtin instruction. If the instruction type cannot be determined, returns Err(invalid_instruction_data_error).

Dropping these on invalid ix data would be an attack vector.
Seems they should just have some "fallback" program cost, which represents the cost to deserialize/match on ix data enum variant.
and then let the tx error out at runtime.

tao-stones:
> Dropping these on invalid ix data would be an attack vector.

It just returns an error at early stage of process pipeline (before execution, like compute-budget [is doing currently](https://github.com/anza-xyz/agave/blob/master/runtime-transaction/src/compute_budget_instruction_details.rs#L133)), leaders can decide to pack them and charge the fee, or drop them. If leaders can't do this yet, probably can keep current "per program default cost" as fallback.

apfitzge:
> > Dropping these on invalid ix data would be an attack vector.
> 
> It just returns an error at early stage of process pipeline (before execution, like compute-budget [is doing currently](https://github.com/anza-xyz/agave/blob/master/runtime-transaction/src/compute_budget_instruction_details.rs?rgh-link-date=2024-09-04T16%3A16%3A41Z#L133)), leaders can decide to pack them and charge the fee, or drop them. If leaders can't do this yet, probably can keep current "per program default cost" as fallback.

Cannot do that right now. Code would be the same that we've already implemented for #82 - code is effectively done on our side, but that SIMD has not been agreed upon yet.

tao-stones:
> Unfortunately, this isn't as straightforward to represent in an array like this. Programs may not always CPI each time they're invoked. Consider an instruction that may CPI once, may CPI twice, or may not CPI at all, considering some account state or input data.
> 
> For this reason, I think we should gear the pattern(s) toward using the maximum CUs possible by an instruction. In the above example, the instruction would define MAX_CUS_WITH_CPI (or whatever) as the worst-case, ie. 2 CPIs.

Thanks for bringing this up. I was assuming builtin instructions have rather fixed CPIs schema, not aware there are instances that dynamically based on account states. I only know that "create lookup table" always CPIs "system" 3 times, and "extend lookup account" CPI "system" once. Most likely I am not up to date with builtins, if there are more dynamic scenarios, then `MAX_CUS_WITH_CPI` is a good idea to me.

> A few more suggestions from my side for contributors' QoL:

A great list of TODOs! To add to it: 
- forbid builtin from nested CPIs (builtin CPIs to anothe rBuiltin that CPIs to another builtin); to extend that, possible to limit builtin to only statically CPIs at top level?
- is it possible to add static assertion, or tests, to ensure newly created builtin program, or instruction, that complies with all this buitins rules? And are included in the "dictionary"
( Maybe this all belong to separate SIMD)

buffalojoec:
> Thanks for bringing this up. I was assuming builtin instructions have rather fixed CPIs schema, not aware there are instances that dynamically based on account states. I only know that "create lookup table" always CPIs "system" 3 times, and "extend lookup account" CPI "system" once.

We could go through and profile all of the processors to make sure they're fixed, but we'd also have to impose this constraint on any new instructions/processors. Considering your last bullet (below), it might also be harder to programmatically enforce.

> is it possible to add static assertion, or tests, to ensure newly created builtin program, or instruction, that complies with all this buitins rules? And are included in the "dictionary"
( Maybe this all belong to separate SIMD)

Yeah, I think some kind of interface (trait for Agave) for builtins and a testing standard (check instruction stack height for example) can accomplish this.

IMO we probably don't need a separate SIMD, we can introduce the constraints in this one, and mention that all builtins are already compliant as-is. Since the introduction of these constraints doesn't inherently change anything about the current protocol, I lean toward not requiring they be proposed in a new SIMD.

tao-stones:
> We could go through and profile all of the processors to make sure they're fixed, but we'd also have to impose this constraint on any new instructions/processors.

Yea, I take it back, such constraint is unnecessarily restrictive. Make builtin programs to expose worse-case CUs, as you suggested, is better. 

If no other objects, I'll include updated [option one](https://github.com/solana-foundation/solana-improvement-documents/pull/170#issuecomment-2329402430) to proposal.

tao-stones:
> IMO we probably don't need a separate SIMD, we can introduce the constraints in this one, and mention that all builtins are already compliant as-is. Since the introduction of these constraints doesn't inherently change anything about the current protocol, I lean toward not requiring they be proposed in a new SIMD.

For the sake of documentation, the constrains all current and future builtins should comply, and testing standard they must follow, deserve its own SIMD. Would work better for multiple clients too. ( Plus I am not the right person to draft these rules for builtins 😄 )

proposals/simd-0170-builtin-instruction-cost-and-budget.md
```diff
@@ -0,0 +1,177 @@
...
+    Instead of the current dictionary of [builtin program, DEFAULT_COMPUTE_UNIT],
+    a new dictionary will be created with [instruction, MAX_COMPUTE_UNIT].
+    This will allow the system to accurately calculate compute allocation for
+    each instruction, factoring in its potential CPIs.
+
+  1.3 Call-Site Implementation:
+
+    - Instruction Type Lookup: At the call-site (e.g., in the compute budget or
+    cost model), the type of builtin instruction will be determined.
+    - If the instruction type cannot be identified, a small number of CUs will
```
tao-stones, line 130: "charging small amount of CUs" for basic works done on transaction, is a change apply to all transaction that'd fail during sanitization, or before VM execution. It is beyond this proposal.

buffalojoec (commented):
Lgtm @tao-stones, just a few nits on my side! Also want to get FD's opinion on the additional static value.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
```diff
@@ -0,0 +1,105 @@
...
+counted against block limits during the banking stage. However, historically,
+builtin instructions have been allocated DEFAULT_INSTRUCTION_COMPUTE_UNIT_LIMIT
+units of compute budget. This discrepancy between the allowed consumption and
+the actual usage tracked for block limits has led to several issues that need
+to be addressed.
+
+Allocating DEFAULT_INSTRUCTION_COMPUTE_UNIT_LIMIT instead of the actual
+DEFAULT_COMPUTE_UNITS for builtin instructions distorts the tracking of
+transaction compute budgets. This can result in more expensive instructions
+being executed when they should fail due to exceeding the budget. Consequently,
```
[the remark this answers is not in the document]
tao-stones: Good question, I should word it more clearly. The issue is if block packer reserves `150` cu for `transfer`, but CU meter is allocateed `200_000` for it, the following BPF ix can take additional `199_850` CU without burst CU meter, that'd make block packer to adjust *up* `199_850` CUs. Adjust-up is considered bug.

In case of two BPF ixs, both block packer and CU meter see `400_000` CUs, so even 2nd ix can take advantage of left-over CUs from 1st, there wouldn't be case of adjust-up.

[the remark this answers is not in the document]
buffalojoec: Okay gotcha, thanks, that makes sense! So by allocating `ALLOCATE_COMPUTE_UNITS` instead of `200_000`, in the case where `ALLOCATE_COMPUTE_UNITS` > `EXECUTE_COMPUTE_UNITS`, is there still an adjust-up between `ALLOCATE_COMPUTE_UNITS` and `EXECUTE_COMPUTE_UNITS`?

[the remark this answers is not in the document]
tao-stones: In that case, BPF instruction can still consume the left-over CUs (eg, up to `ALLOCATE_COMPUTE_UNITS` - total executed ); but since block producer should reserved `ALLOCATE_COMPUTE_UNITS` for builtin, and default 200K for BPF instruction, which is the same CUs allocated for compute-budget for VM, the most CUs combined instructions can consume is the sum of `ALLOCATE_COMPUTE_UNITS` and `200K`, so there can't be adjust-up.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
```diff
@@ -0,0 +1,106 @@
...
+path. However, this introduces new constraints for existing and future
+built-ins, requiring updates to comply with these rules, which could
+overcomplicate the design.
+
+## Detailed Design
+
+1. **Statically define CUs per instruction**: Assign a fixed CU consumption
+   (DEFAULT_BUILTIN_INSTRUCTION_COMPUTE_UNITS) to each built-in instruction
+rather than per built-in program.
+2. **Set a static maximum CU allocation**: Propose a global limit of 5,000 CUs
```
tao-stones, line 82: `5000` is based on existing builtins. AddressLookupTable's `CreateLookupTable` makes most CPI calls (up to 3 calls to `System`), that results up to `750 + 3*150 = 1200 CU`; Upgradeable Loader's `ExtendProgram` may CPI `System` once, in that case it'd need `2370 + 150 = 2520 CU`, which is the highest of all existing builtins. `5000` nearly doubles that, to leave room for future builtins.

tao-stones:
Some mainnet-beta data analysis: 

- Among `653,281,183` non-vote transactions without setting `compute_unit_limit`, 
- `1,199,965` transactions *would* fail if builtin instructions are allocated `5,000` CU (per proposal) instead of `200_000` CU (as currently is). That is `0.18%` non-vote transaction that w/o `compute_unit_limit` would fail. 
- Among these *would fail* transactions, the average "available data bytes" (which is 1232 - tx-data-bytes) is 623 bytes, with Max 815 bytes, Min 0.
- look further, only `370` such transactions have less than 32 bytes available, which means `0.000057%` transactions would fail hard under this proposal.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
buffalojoec: Is the `5000` figure enough to increase block density by an adequately significant margin? Or should it be tighter? Seems like this would be a pretty dramatic improvement, right?

proposals/simd-0170-builtin-instruction-cost-and-budget.md
```diff
@@ -0,0 +1,106 @@
...
+allow fine-tuned CU allocation based on each instruction’s potential execution
+path. However, this introduces new constraints for existing and future
+built-ins, requiring updates to comply with these rules, which could
+overcomplicate the design.
+
+## Detailed Design
+
+1. **Statically define CUs per instruction**: Assign a fixed CU consumption
+   (DEFAULT_BUILTIN_INSTRUCTION_COMPUTE_UNITS) to each built-in instruction
+rather than per built-in program.
```
buffalojoec, line 81: So to be totally clear, the transaction CU meter (VM) will consume `CU[ix]`, where `ix` is the instruction invoked. Meanwhile, the producer will allocate static `M` for any top-level builtin instruction. Is that correct?

tao-stones: Exactly! 

Taking `CreateLookupTable` as example, itself consumes `750` CU, it also CPIs up to 3 `System` instructions, each consumes `150` CUs. 

Both producer and CU Meter will allocate `5,000` CU for it to start with. Invoking `CreateLookupTable` itself takes `750` out of `5,000`, then each time `System` is invoked, `150` is taken out from remaining of `5,000 - 750`.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
tao-stones: Yes, the `5,000` yields a good result so far. 

In testing, I looked at 3 metrics: 
1. Should have zero case of "adjust-up" at producer side, and
2. Produced blocks should be (nearly) full, and
3. Count of transaction retries (due to "would-exceed-limits") should not be too high (indicating not over allocating);

Compare to original `200,000`, setting builtin allocation to `5000` on both producer and CU meter have a good result: no adjust-up, blocks are always (near) full, and number of retries were up very slightly. 

Setting it tighter to `3,000` didn't make too much improvements from `5,000`.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
topointon-jump: Could we explain this in the SIMD? This wasn't clear to me upon first glance either.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
```diff
@@ -0,0 +1,110 @@
...
+None
+
+## Detailed Design
+
+1. Statically define CUs per instruction: Assign a fixed CU consumption
+   (DEFAULT_BUILTIN_INSTRUCTION_COMPUTE_UNITS) to each builtin instruction
+rather than per builtin program.
+2. Set a static maximum CU allocation: Propose a global limit of 5,000 CUs
+   (MAX_BUILTIN_ALLOCATION_COMPUTE_UNIT_LIMIT) to cover worst-case scenarios,
+including CPI operations.
```
topointon-jump, line 63: Could we explain that:

- `MAX_BUILTIN_ALLOCATION_COMPUTE_UNIT_LIMIT` is the same for each builtin program instruction
- `MAX_BUILTIN_ALLOCATION_COMPUTE_UNIT_LIMIT` is used to allocate the CU meter and set the block producer limits

topointon-jump: Could we document what value `MAX_BUILTIN_ALLOCATION_COMPUTE_UNIT_LIMIT` will take, and the data used to derive this?

proposals/simd-0170-builtin-instruction-cost-and-budget.md
```diff
@@ -0,0 +1,110 @@
...
+
+## New Terminology
+
+None
+
+## Detailed Design
+
+1. Statically define CUs per instruction: Assign a fixed CU consumption
+   (DEFAULT_BUILTIN_INSTRUCTION_COMPUTE_UNITS) to each builtin instruction
+rather than per builtin program.
```
topointon-jump, line 60: Could we document that when a builtin instruction is invoked by VM, the `DEFAULT_BUILTIN_INSTRUCTION_COMPUTE_UNITS` is deducted from CU Meter?

topointon-jump: Could we also specify the values that each `DEFAULT_BUILTIN_INSTRUCTION_COMPUTE_UNITS` will take? And ideally the justification for these.

ptaffet-jump:
In general, I think Firedancer is on board with the proposal. I would have preferred a tighter limit, but I'm okay with 5000 because in our block packing, I think we'll continue packing transactions that only invoke built-ins in the same way, which continues to be compatible with this approach. Thank you for eliminating the adjust-up cases!

I agree with Tom that I think it's important for the actual values of `DEFAULT_BUILTIN_INSTRUCTION_COMPUTE_UNITS` to be in this SIMD, since this SIMD is unimplementable without those values.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
tao-stones: aa70ccf9d061eabb282c3846a5d60a12a506325c

buffalojoec:
> If there're strong feeling that no additional complexity would be added to existing builtin instructions, or new builtins added, I'm happy to remove the buffer, tighten limit to 3,000.

I feel pretty strongly about this for all programs except stake and loader v4, however I anticipate neither would include so much complexity that it would affect CUs.

On the Agave side, we can avoid this footgun with traits or constants that call out to contributors the CU constraint imposed by cost model.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
```diff
@@ -0,0 +1,126 @@
...
+superseded-by:
+extends:
+---
+
+## Summary
+
+1. Builtin programs should consume a predefined number of CUs for each
+   instruction.
+2. Since builtins can invoke other programs (CPI), they should allocate enough
+   but granular CUs to ensure successful execution without over-allocation.
```
jstarry, line 21: Let's simplify this to something like "If a transaction doesn't define a compute budget limit, then for each builtin program instruction allocate 3000 compute units rather than 200,000."

proposals/simd-0170-builtin-instruction-cost-and-budget.md
```diff
@@ -0,0 +1,126 @@
...
+from both the CU meter and block limits after execution. However, the CU meter
+allocates a much larger value (DEFAULT_INSTRUCTION_COMPUTE_UNIT_LIMIT) to each
+builtin instruction. This discrepancy creates imprecise CU tracking and may
+require block producers to account for additional CUs after a transaction's
+execution, potentially causing block limits to be exceeded.
+
+   Furthermore, builtins can conditionally CPI into other programs, introducing
+variability in CU consumption. This unpredictability makes it difficult to
+allocate CUs upfront solely based on CUs builtin consumes during execution, also
+adds complexity to the tracking process.
```
jstarry, line 39: This is basically just an agave scheduler bug which isn't subject to consensus rules. Let's remove this section from the SIMD and just keep the 2nd point.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
```diff
@@ -0,0 +1,126 @@
...
+## New Terminology
+
+None
+
+## Detailed Design
+
+1. Retain the existing DEFAULT_COMPUTE_UNITS for builtin programs.
+   When the virtual machine (VM) invokes a builtin instruction, the
+statically defined DEFAULT_COMPUTE_UNITS is consistently deducted from the CU
+Meter.
```
jstarry, line 61: Let's leave this out of this SIMD because it's existing behavior which isn't changing. I think if firedancer is interested in how this existing behavior works, we can detail it in the follow up SIMD that @tao-stones will be writing for per-ix cu costs.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
```diff
@@ -0,0 +1,126 @@
...
+CU demand of 1,200 CUs (750 + 3 × 150). Similarly, `UpgradeableLoader`’s
+`ExtendProgram`, which may invoke the System program once, requires up to
+2,520 CUs (2,370 + 150), representing the most resource-intensive operation
+among current built-ins. The proposed 3,000 CU limit slightly exceeds this
+requirement, allowing for controlled flexibility without an excessive margin.
+
+3. Handling invalid CU requests: Transactions will fail if they request:
+   - More than MAX_COMPUTE_UNIT_LIMIT per transaction, or
+   - Less than the sum of all included builtin instructions'
+     MAX_BUILTIN_ALLOCATION_COMPUTE_UNIT_LIMIT
```
jstarry, line 83: I think this can also be pulled out and solved in a separate SIMD

proposals/simd-0170-builtin-instruction-cost-and-budget.md
```diff
@@ -0,0 +1,126 @@
...
+builtins, requiring updates to comply with these rules, which could
+overcomplicate the design.
+
+## Impact
+
+Users who previously relied on including builtin instructions, instead of
+explicitly setting compute-unit limits, to allocate budget for their
+transactions may experience an increase in transaction failures. To avoid this,
+users are encouraged to use set_compute_unit_limit to explicitly request the
+necessary budget for their transactions.
```
jstarry, line 120: If those impacted users have issues fitting in the set compute unit limit instruction into their transactions due to tx data size limits, they can also migrate to using address lookup tables to fit in the compute budget instruction call.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Detailed Design
+
+Establish a static maximum CU allocation: Define a global compute unit limit
+of 3,000 CUs, denoted as MAX_BUILTIN_ALLOCATION_COMPUTE_UNIT_LIMIT, to
+accommodate worst-case execution scenarios, including potential Cross-Program
+Invocations (CPIs). This uniform limit is applied to each builtin program
+instruction and is used to configure both CU meter allocations and block
+producer CU limits reservation for all builtin instructions.
+
```
jstarry, line 49: The static list of builtin program id's that will have 3000 compute units allocated are:
- `Stake11111111111111111111111111111111111111`
- `Config1111111111111111111111111111111111111`
- `Vote111111111111111111111111111111111111111`
- `11111111111111111111111111111111`
- `ComputeBudget111111111111111111111111111111`
- `AddressLookupTab1e1111111111111111111111111`
- `BPFLoaderUpgradeab1e11111111111111111111111`
- `BPFLoader1111111111111111111111111111111111`
- `BPFLoader2111111111111111111111111111111111`
- `LoaderV411111111111111111111111111111111111`
- `KeccakSecp256k11111111111111111111111111111`
- `Ed25519SigVerify111111111111111111111111111`

Note that there are a few builtin programs not in the list including the upcoming zk programs as well as the feature program.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
```diff
@@ -0,0 +1,99 @@
...
+allocation in #2.
+
+## Alternatives Considered
+
+1. Treat builtins like regular instructions: Another approach would be to
+   allocate DEFAULT_INSTRUCTION_COMPUTE_UNIT_LIMIT for builtins as for other
+instructions. However, this fails to address over-allocation. Unlike regular
+instructions, the execution of builtins is more predictable and usually
+requires significantly fewer CUs than BPF instructions. This approach would
+allocate more CUs than necessary, undermining the goal of efficient CU usage.
```
jstarry, line 71: 1. Do nothing and continue allocating 200k compute unit's for all non-compute budget instructions. However, this fails to address over-allocation. Unlike regular instructions, the execution of builtins is more predictable and usually requires significantly fewer CUs than BPF instructions. This approach would allocate more CUs than necessary, undermining the goal of efficient CU usage.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
```diff
@@ -0,0 +1,120 @@
...
+
+Establish a static maximum allocation of 3,000 CU's for builtin programs denoted
+as `MAX_BUILTIN_ALLOCATION_COMPUTE_UNIT_LIMIT`, which is larger than all actual
+compute unit costs of builtin programs as well as the potential Cross-Program
+Invocations (CPIs) they do. This uniform limit is applied to each builtin
+program instruction and is used to configure both CU meter allocations and block
+producer CU limits reservation for all builtin instructions.
+
+The static list of builtin program id's that will have 3000 compute units
+allocated are:
```
buffalojoec, line 54: Can you add a line here that also states the builtin must be owned by the native loader for this CU allocation to apply? Avoids any footguns when they move to BPF.

buffalojoec: The list can still be static, we just also want that explicit requirement.

tao-stones: added in 0e6e0caa2d35a915ae2455dadf66a4ad8c1e8ae6

jstarry: I think we should update this list of builtin program id's when each of the builtin-to-bpf feature gates are activated. We're not going to check the owner of the builtin program each time during cost calculation.

proposals/simd-0170-builtin-instruction-cost-and-budget.md
```diff
@@ -0,0 +1,121 @@
...
+Establish a static maximum allocation of 3,000 CU's for builtin programs denoted
+as `MAX_BUILTIN_ALLOCATION_COMPUTE_UNIT_LIMIT`, which is larger than all actual
+compute unit costs of builtin programs as well as the potential Cross-Program
+Invocations (CPIs) they do. This uniform limit is applied to each builtin
+program instruction and is used to configure both CU meter allocations and block
+producer CU limits reservation for all builtin instructions.
+
+The static list of builtin program id's that will have 3,000 compute units
+allocated are listed below, note that the builtins must be owned by the native
+loader.
```
jstarry, line 55:
```suggestion
The static list of builtin program id's that will have 3,000 compute units
allocated are listed below, note that when builtins are migrated to sBPF
programs, they MUST be removed from this list and have the default 200k
compute units allocated instead.
```

topointon-jump (approved):

proposals/simd-0170-builtin-instruction-cost-and-budget.md
buffalojoec: Works for me, as long as an ID can be popped out on feature activation.

jstarry: Yup, @tao-stones please make sure @buffalojoec gets added as a reviewer on your impl pr

tao-stones: yep, will do

apfitzge (approved):

apfitzge:
@Benhawkins18 can we get this merged? @tao-stones is already implementing it. We have 2 folks from Anza, and @topointon-jump from Jump approving.

Benhawkins18 (approved):
I see approvals from both jump and Anza. Merging

PR #173, opened 2024-09-05
SIMD-0173: SBPF instruction encoding improvements

Lichtso:


proposals/0173-sbpf-instruction-encoding-improvements.md
```diff
@@ -0,0 +1,98 @@
...
+type: Core
+status: Draft
+created: 2024-09-05
+feature: TBD
+extends: SIMD-0161
+---
+
+## Summary
+
+There are some instruction with questionable encodings, that when slightly
```
LucasSte, line 16:
```suggestion
There are some instructions with questionable encodings, that when slightly
```

proposals/0173-sbpf-instruction-encoding-improvements.md
```diff
@@ -0,0 +1,98 @@
...
+adjusted, could significantly simplify verification and execution of programs.
+
+## Motivation
+
+The instruction `lddw dst, imm` is currently the only instruction which takes
+two instruction slots. This proposal splits it into a two one-slot instruction
+sequence: `mov32 dst, imm` and an introduced `hor64 dst, imm`. This way all
+instructions will be exactly one slot long which will simplyfiy:
```
LucasSte, line 24:
```suggestion
instructions will be exactly one slot long which will simplify:
```

proposals/0173-sbpf-instruction-encoding-improvements.md
```diff
@@ -0,0 +1,98 @@
...
+- Calculating the number of instructons in a program will no longer require a
+full linear scan. A division of the length of the text section by the
+instruction slot size will suffice.
+- The instruction meter will no longer have to skip one instruction slot when
+counting a `LDDW` instruction.
+- Jump and call instructions will no longer have to verify that the desination
+is not the second half of a `LDDW` instruction.
+- The verifier will no longer have to check that `LDDW` instructions are
+complete and its first or second half does not occur without the other on its
+own.
```
LucasSte, line 35: Removing LDDW does not eliminate relocations. When the program contains an address load with the new instruction pair `mov32/hor64`, we may still need a relocation. In this case, I believe we should check (at least during the relocation process) we are writing the relocated address to a `mov32/hor64` pair instead of overwriting another instruction.

Lichtso: That is a problem of the toolchain. In general we should aim to apply all relocations at link time.

proposals/0173-sbpf-instruction-encoding-improvements.md
```diff
@@ -0,0 +1,122 @@
...
+
+- `le dst, 16` behaves the same as `and32 dst, 0xFFFF`
+- `le dst, 32` behaves the same as `and32 dst, 0xFFFFFFFF`
+- `le dst, 64` behaves the same as `mov64 dst, src`
+
+The `CALLX` instruction encodes its source register in the immediate field.
+This is makes the instruction decoder more complex because it is the only case
+in which a register is encoded in the immediate field, for no reason.
+
+With all of the above changes and the ones define in SIMD-0174, the memory
```
LucasSte, line 49:
```suggestion
With all of the above changes and the ones defined in SIMD-0174, the memory
```

proposals/0173-sbpf-instruction-encoding-improvements.md
```diff
@@ -0,0 +1,122 @@
...
+### Changes to Execution
+
+The introduced `HOR64` instruction (opcode `0xF7`) must take its immediate
+value, shift it 32 bit towards the MSBs (multiplication-like left shift) and
+then bitwise OR it into the given `dst` register.
+
+For the `CALLX` instruction (opcode `0x8D`) the jump destination must be read
+from the `src` register field instead of the `imm` immediate field.
+
+The execution behavior of the moved opcodes is mapped to their new encodings:
```
LucasSte, line 106: I think you should use say that the behavior will be transferred to the new opcodes. `mapped` may sound as if you would map the old instruction to new new ones, if the former appears in a programs.

proposals/0173-sbpf-instruction-encoding-improvements.md
```diff
@@ -0,0 +1,123 @@
...
+When a `CALLX` instruction (opcode `0x8D`) is encountered during verification,
+the `src` register field must be verified instead of the `imm` immediate field.
+Otherwise, the verification rule stays the same: The src register must be in
+the inclusive range from R0 to R9.
+
+### Changes to Execution
+
+The introduced `HOR64` instruction (opcode `0xF7`) must take its immediate
+value, shift it 32 bit towards the MSBs (multiplication-like left shift) and
+then bitwise OR it into the given `dst` register.
```
topointon-jump, line 101: We should aim to bundle all of the ISA/opcode changes in the current set of ABIv1 SIMDs into the same SBPF version.

LucasSte: The idea to split them in different versions was to avoid huge modifications and facilitate debugging in case an unforeseen problem arises after the ISA changes have been deployed.

0x0ece (approved):

proposals/0173-sbpf-instruction-encoding-improvements.md
```diff
@@ -0,0 +1,123 @@
...
+
+- the `HOR64` instruction (opcode `0xF7`)
+- the moved opcodes:
+  - `0x27`, `0x2C`, `0x2F` (`STB`, `LDXB`, `STXB`)
+  - `0x37`, `0x3C`, `0x3F` (`STH`, `LDXH`, `STXH`)
+  - `0x87`, `0x8C`, `0x8F` (`STW`, `LDXW`, `STXW`)
+  - `0x97`, `0x9C`, `0x9F` (`STDW`, `LDXDW`, `STXDW`)
+
+When a `CALLX` instruction (opcode `0x8D`) is encountered during verification,
+the `src` register field must be verified instead of the `imm` immediate field.
```
niooss-ledger, line 93: For information, (upstream) LLVM recently changed the encoding of `CALLX` to use `dst` register instead of the `imm` immediate field: https://github.com/llvm/llvm-project/pull/81546 (this landed in LLVM 19.1.0).

Would modifying the proposal to use `dst` instead of `src` be acceptable?

niooss-ledger: > We could but it would require backports to v2.2.

Thanks for the information. I was wondering whether SIMD-0173 was already frozen/not subject to change, as this Pull Request is not merged. It seems SBPF v2 (and v3 ?) are already frozen standards and a version bump would be needed to make other changes.

It feels like much effort for only few improvements to, to change the register used by `CALLX`. I don't have the bandwidth (or motivation) to work on this. Feel free to mark this comment as resolved.

> Also, the "dst" field? The destination is for registers to be written to, which callx does not. They probably thought: Oh, the jump destination, that must be the destination field.

I don't know the rationale behind the decision of choosing `dst` instead of `src`. Nevertheless, in the Linux kernel, the behavior of `CALL` (with an immediate ; opcode `0x85`) changes depending on `src`:

- `src=0` is used to encode calls to BPF helper kernel functions (`imm` contains the ID).
- `src=BPF_PSEUDO_CALL=1` is used to encode relative calls (`imm` contains a relative offset).
- `src=BPF_PSEUDO_KFUNC_CALL=2` is used to encode calls to kernel functions using their BTF identifiers.

(cf. https://web.git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/bpf.h?h=v6.14#n1339 )

While the Linux kernel does not support indirect calls (`CALLX reg`, opcode `0x8d`) yet, I can imagine people working on eBPF not willing to re-use `src` to store the target address.

proposals/0173-sbpf-instruction-encoding-improvements.md
```diff
@@ -0,0 +1,123 @@
...
+
+A program containing one of the following instructions must **not** throw
+`VerifierError::UnknownOpCode` during verification anymore:
+
+- the `HOR64` instruction (opcode `0xF7`)
+- the moved opcodes:
+  - `0x27`, `0x2C`, `0x2F` (`STB`, `LDXB`, `STXB`)
+  - `0x37`, `0x3C`, `0x3F` (`STH`, `LDXH`, `STXH`)
+  - `0x87`, `0x8C`, `0x8F` (`STW`, `LDXW`, `STXW`)
+  - `0x97`, `0x9C`, `0x9F` (`STDW`, `LDXDW`, `STXDW`)
```
LucasSte, line 90: Probably a little late for my comment, but the critique is still valid. Repurposing instruction classes brings no immediate benefit for either the validator or developers. We can only reap potential positive effects of this change if coupled with a future SIMD. Having said that, modifications like this could be postponed to when the gains are clear and well documented.

Benhawkins18 (approved):

proposals/0173-sbpf-instruction-encoding-improvements.md
niooss-ledger: For information (for future readers stumbling on this thread), modifying the encoding of callx to use the destination register (and to revert SIMD-0173) is now being discussed in [SIMD-0377: eBPF ISA compatibility](https://github.com/solana-foundation/solana-improvement-documents/pull/377). The sBPF implementation was already modified: https://github.com/anza-xyz/sbpf/pull/89

PR #174, opened 2024-09-06
SIMD-0174: SBPF arithmetics improvements

Lichtso:


proposals/0174-sbpf-arithmetics-improvements.md
```diff
@@ -0,0 +1,154 @@
...
+- the `UDIV64` instruction (opcode `0x56` and `0x5E`)
+- the `UREM32` instruction (opcode `0x66` and `0x6E`)
+- the `UREM64` instruction (opcode `0x76` and `0x7E`)
+- the `LMUL32` instruction (opcode `0x86` and `0x8E`)
+- the `LMUL64` instruction (opcode `0x96` and `0x9E`)
+- the `SHMUL64` instruction (opcode `0xB6` and `0xBE`)
+- the `SDIV32` instruction (opcode `0xC6` and `0xCE`)
+- the `SDIV64` instruction (opcode `0xD6` and `0xDE`)
+- the `SREM32` instruction (opcode `0xE6` and `0xEE`)
+- the `SREM64` instruction (opcode `0xF6` and `0xFE`)
```
LucasSte, line 79: I think these instruction opcodes is a bit of a departure from the pattern we had until now. In BPF, opcode for 32-bit and 64-bit instructions is in the least significant bits: `mul64 dest, reg` is `0x2F` and `mul32 dest, reg` is `0x2C`.

For these new instructions, we change the most significant bits. Maintaining the same pattern simplifies code generation.

proposals/0174-sbpf-arithmetics-improvements.md
```diff
@@ -0,0 +1,154 @@
...
+- the `SREM64` instruction (opcode `0xF6` and `0xFE`) produces the remainder of
+a signed 64 bit division (`dst / imm` and `dst / src`).
+
+Runtime exceptions are:
+
+- If the divisor (`imm` or `src`) of any division is zero,
+`EbpfError::DivideByZero` must be thrown.
+- If the dividend (`dst`) of a signed division has the minimal value for its
+bit-width and the divisor (`imm` or `src`) is `-1`, then
+`EbpfError::DivideOverflow` must be thrown.
```
LucasSte, line 125: Why is it important to error out for this case? We don't error for any other overflows (add, mul, and sub).

proposals/0174-sbpf-arithmetics-improvements.md
Lichtso: Yes, but that bit is part of the instruction class. Meaning the 32 bit ALU instruction class and the 64 bit ALU instruction class are two different instruction classes. And this SIMD introduces a new instruction class for PQR because they don't fit into the ALU classes.

proposals/0174-sbpf-arithmetics-improvements.md
```diff
@@ -0,0 +1,154 @@
...
+status: Draft
+created: 2024-09-06
+feature: TBD
+extends: SIMD-0161
+---
+
+## Summary
+
+This proposal introduces wide multiplication, signed division and explicit sign
+extension to SBPF.
```
topointon-jump, line 17: Let's specify which SBPF version this will be released in.

topointon-jump (approved):

proposals/0174-sbpf-arithmetics-improvements.md
0x0ece: SDIV64_IMM and SREM64_IMM are exceptions. IMM is 32-bit, so it can't be 64-bit -1. I think it'd be better to remove the check for these.

Lichtso: There is sign extension of the inputs (not the outputs). Thus the name "**signed** division".

0x0ece (approved):

proposals/0174-sbpf-arithmetics-improvements.md
```diff
@@ -0,0 +1,154 @@
...
+instructions: `UDIV32_IMM`, `UDIV64_IMM`, `UREM32_IMM`, `UREM64_IMM`,
+`SDIV32_IMM`, `SDIV64_IMM`, `SREM32_IMM`, `SREM64_IMM`.
+
+### Changes to Execution
+
+#### PQR Instruction Class
+
+A new instruction class product, quotient and remainder (PQR) is introduced:
+
+- the `UHMUL64` instruction (opcode `0x36` and `0x3E`) produces the 64 MSBs of
```
0x0ece, line 92: Just an impl note for `UHMUL64_IMM`: `imm` is sign extended to 64-bits.

Practically speaking, this means that to do 64x64=128 bit umul one should use `UHMUL64_REG`.

0x0ece (approved):

LucasSte (approved):

Benhawkins18 (approved):

PR #166, opened 2024-08-19
SIMD-0166: SBPF Dynamic stack frames

LucasSte:

PR #178, opened 2024-10-03
SIMD-XXXX: SBPF Static Syscalls

LucasSte:


proposals/0176-static-syscalls.md
```diff
@@ -0,0 +1,87 @@
...
+
+None.
+
+## Impact
+
+The changes proposed in this SIMD are transparent to dApp developers. The 
+compiler toolchain will emit correct code for the specified SBF version. 
+Static syscalls obviate relocations for call instructions and move the virtual 
+machine closer to eliminating relocations altogether, which can bring 
+considerable performance improvements.
```
buffalojoec, line 83: The motivation states that the intention is to remove syscall-related relocations. Maybe it's best to explicitly state in the "Impact" section here that, since the syscalls are statically registered in the VM, these relocations are removed.

This line as-is makes it seem like it's merely an incremental step.

proposals/0176-static-syscalls.md
```diff
@@ -0,0 +1,87 @@
...
+
+For simplicity, syscalls must be represented as a natural number greater than 
+zero, so that they can be organized in a lookup table. This choice allows for 
+quick retrieval of syscall information from integer indexes. An instruction 
+`syscall 2` must represent a call to the function registered at position two 
+in the lookup table.
+
+Consequently, system calls in the Solana SDK and in any related compiler tools 
+must be registered as function pointers, whose address is a natural number 
+greater than zero, representing their position in a syscall lookup table. The
```
buffalojoec, line 59: Sorry if I'm missing the extra context here, but is this syscall lookup table part of this SIMD proposal, or another?

proposals/0176-static-syscalls.md
LucasSte: The most pronounced relocation we do is related to syscalls. There are a few other we also do that can be resolved directly in the linker, but they require small changes in the ELF loader. We plan to introduce relocation free programs in an SIMD @Lichtso is going to write about ELF headers and a refactored loader.

proposals/0176-static-syscalls.md
LucasSte: It is part of this one. The idea is to remove the hash map we use for lookups today. There was a request to use integer indexing for syscalls to avoid hashes. The underlying data structure does not impact the protocol and is more implementation specific. Did my wording sound too vague?

proposals/0176-static-syscalls.md
buffalojoec: Okay, in that case this is probably fine as-is, but I still wonder if "obviate" makes sense here (maybe it does). Does this proposal make syscall relocation obvious, or does it eliminate them?

proposals/0176-static-syscalls.md
buffalojoec: I think the wording is totally fine, I just wonder if we should have a table of all of those integer IDs in this proposal itself, since this would definitely need to be matched on all clients. What do you think?

proposals/0176-static-syscalls.md
LucasSte: By "obviate" I meant make them unnecessary, so that we can eliminate them.

proposals/0176-static-syscalls.md
LucasSte: Yes, I'll add the table.

LucasSte: Added.

buffalojoec: Thanks! Looks great. Any particular ordering, or just arbitrarily selected?

proposals/0178-static-syscalls.md
```diff
@@ -0,0 +1,145 @@
...
+|   sol_remaining_compute_units            |    31    |
+|   sol_create_program_address             |    32    |
+|   sol_try_find_program_address           |    33    |
+|   sol_get_sysvar                         |    34    |
+|   sol_get_epoch_stake                    |    35    |
+|   sol_get_clock_sysvar                   |    36    |
+|   sol_get_epoch_schedule_sysvar          |    37    |
+|   sol_get_last_restart_slot              |    38    |
+|   sol_get_epoch_rewards_slot             |    39    |
+|   sol_get_fees_sysvar                    |    40    |
```
buffalojoec, line 128: I think`sol_get_rent_sysvar` is missing here, right?

proposals/0176-static-syscalls.md
LucasSte: > Any particular ordering, or just arbitrarily selected?

I tried grouping related syscalls to similar numbers and I wanted those declared in the compiler to have the lower numbers.

proposals/0178-static-syscalls.md
LucasSte: Yes, it was missing. Thanks for catching that!

buffalojoec (approved):
Looks great to me! I think we just need to land on which SBPF version this goes into?

proposals/0178-static-syscalls.md
```diff
@@ -0,0 +1,146 @@
...
+us to resolve syscalls without relocations.
+
+## New Terminology
+
+None.
+
+## Detailed Design
+
+The following must go into effect if and only if a program indicates the SBPF 
+version XX or higher in its ELF header e_flags field, according to the
```
topointon-jump, line 40: Should we specify which version XX is?

proposals/0178-static-syscalls.md
```diff
@@ -0,0 +1,146 @@
...
+## Motivation
+
+The resolution of syscalls during ELF loading requires relocating addresses, 
+which is a performance burden for the validator. Relocations require an entire 
+copy of the ELF file in memory to either relocate addresses we fetch from the 
+symbol table or offset addresses to after the start of the virtual machine’s 
+memory. Moreover, relocations pose security concerns, as they allow the 
+arbitrary modification of program headers and programs sections. A new 
+separate opcode for syscalls modifies the behavior of the ELF loader, allowing 
+us to resolve syscalls without relocations.
```
topointon-jump, line 31: 🎉

topointon-jump (commented):
This is awesome! The only thing which makes me a bit sad is the opcode change, which complicates/slows down our implementation (we have a static jump table). We should try and keep these to a minimum/only make these changes once.

proposals/0178-static-syscalls.md
```diff
@@ -0,0 +1,146 @@
...
+This new instruction comes together with modifications in the verification 
+phase. `call imm` (opcode `0x85`) instructions must only refer to internal 
+calls and its immediate field must only be interpreted as a relative address 
+to jump from the program counter. 
+
+### New return instruction
+
+The opcode `0x9D` must represent the return instruction, which supersedes the 
+`exit` instruction. The opcode (opcode `0x95`), previously assigned to the 
+`exit` instruction, must now be interpreted as the new syscall instruction.
```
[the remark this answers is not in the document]
topointon-jump: Also, changing the name from `exit` to `return` when it is the same instruction could be confusing (I have already seen this confused in other SIMDs).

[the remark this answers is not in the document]
topointon-jump: Side note - we should bundle large sets of proposed ISA changes together into the same SBPF version upgrade, so that clients don't have to support a mis-mash of ISAs based on feature flags.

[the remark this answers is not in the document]
Lichtso: Motivation is that `exit` was occupying the slot in the instruction class for controlflow with immediate values and it does not take an immediate value. The new `syscall` opcode however does, so it took its place.

proposals/0178-static-syscalls.md
```diff
@@ -0,0 +1,146 @@
...
+Consequently, system calls in the Solana SDK and in any related compiler tools 
+must be registered as function pointers, whose address is a natural number 
+greater than zero, representing their position in a syscall lookup table. The 
+verifier must enforce that the immediate of a syscall instruction points to a 
+valid syscall, and throw `VerifierError::InvalidSyscall` otherwise.
+
+This new instruction comes together with modifications in the verification 
+phase. `call imm` (opcode `0x85`) instructions must only refer to internal 
+calls and its immediate field must only be interpreted as a relative address 
+to jump from the program counter.
```
ravyu-jump, line 67: Does this mean there is no longer a need to hash the immediates?

Lichtso: It does and is the intention.

proposals/0178-static-syscalls.md
```diff
@@ -0,0 +1,146 @@
...
+
+|           Syscall name                   |  Number  |
+|------------------------------------------|----------|
+|   abort                                  |    1     |
+|   sol_panic_                             |    2     |
+|   sol_memcpy_                            |    3     |
+|   sol_memmove_                           |    4     |
+|   sol_memset_                            |    5     |
+|   sol_memcmp_                            |    6     |
+|   sol_log                                |    7     |
```
0x0ece, line 95:
```
sol_log_
sol_log_64_
```

LucasSte: Fixed

proposals/0178-static-syscalls.md
```diff
@@ -0,0 +1,146 @@
...
+|   sol_big_mod_exp                        |    30    |
+|   sol_remaining_compute_units            |    31    |
+|   sol_create_program_address             |    32    |
+|   sol_try_find_program_address           |    33    |
+|   sol_get_sysvar                         |    34    |
+|   sol_get_epoch_stake                    |    35    |
+|   sol_get_clock_sysvar                   |    36    |
+|   sol_get_epoch_schedule_sysvar          |    37    |
+|   sol_get_last_restart_slot              |    38    |
+|   sol_get_epoch_rewards_slot             |    39    |
```
Lichtso, line 127: Think this one should be `sol_get_epoch_rewards_sysvar`, see:
https://github.com/anza-xyz/agave/blob/3890ce5bc594d1b81e4e3fa57174e919fb9257ae/programs/bpf_loader/src/syscalls/mod.rs#L394

LucasSte: Thanks for catching that. Fixed!

0x0ece (approved):

proposals/0178-static-syscalls.md
```diff
@@ -0,0 +1,146 @@
...
+The opcode `0x9D` must represent the return instruction, which supersedes the 
+`exit` instruction. The opcode (opcode `0x95`), previously assigned to the 
+`exit` instruction, must now be interpreted as the new syscall instruction.
+
+The verifier must detect an SBPF V1 program containing the `0x9D` opcode and 
+throw a `VerifierError::UnknowOpCode`. Likewise, if, by any means, a V1 
+program reaches the execution stage containing the `0x9D` opcode, an 
+`EbpfError::UnsupportedInstruction` must be raised.
+
+### Syscall numbering convention
```
deanmlittle, line 80: Integer is not a great idea. SVM will continue to diverge more and more as non-mainnet SVM chains and L2s develop. If they wish to push their own syscalls, not only will they need to write their own tooling to handle it, if Solana mainnet adds another syscall that clashes and that same binary is shipped to another chain, or vice-versa, people could lose money. It would make way more sense to use the Murmur3 hash. If they decide to launch a hash collision, at least we tried to stop them.

alessandrod: Murmur3 is what the original implementation did, and I was against switching to indexes, but then I got busy on other stuff (I just noticed that master was switched to indexes).

Is the best argument for indexes that lookup is faster? And if that's the argument, isn't it moot considering that we JIT?

LucasSte: Hey @deanmlittle,
Thanks for your input. I didn't follow why teams would have to develop their own tooling. On the Solana SDK, changing from consecutive integers to a murmur hash is simply assigning another constant to the function pointers. The compiler toolchain can deal with both cases without changes.

On the other hand, I agree that using consecutive numbers may hinder the development of SVM chains. In Agave, we considered that using a contiguous array would add to much complexity to the code just to handle inactive syscalls or deprecated ones, so we are still using a BTree, as there are only 40 syscalls to lookup for.

Agave's implementation does not prevent SVM external users from calculating the murmur32 hash for their own syscalls, as any 32-bit integers can be used for indexing. The numbering convention they use does not need to follow ours, provided that the numbers don't coincide.

Using consecutive numbers was a request from Firedancer. I believe either @topointon-jump, @ripatel-fd or @0x0ece can elaborate more on the reasons.

alessandrod: > @alessandrod There is a case for allowing zero-copy execution out of a bytecode buffer (i.e. interpreter).

you don't _have_ to fixup in place right?

I understand your perspective. From the anza POV tho, there would be no overhead in having hashes since we always JIT. From the devs POV, I think we can agree that having SVM compatibility is desirable if it doesn't come at a huge cost?

So is it worth making a protocol compromise to accommodate for fd's _current_ implementation?

deanmlittle: @ripatel-fd Sounds reasonable. I am also unsure the necessity of the symbol table if we went with this design?

Lichtso: > Otherwise implies you can omit the symbol table and have "raw" murmur32 hashes that aren't checked for string collisions

This is how it is planned for now. Collisions have to be handled in the SDK and toolchain anyway, so there is no use in encoding them in the ELFs too.

LucasSte: > Could you clarify in the SIMD whether static syscalls are verified during ELF loading or bytecode verification? If we won't have these measures in place to support enums, I agree with Dean that we should keep the hash instead.

As @Lichtso, syscalls are no longer supposed to appear in the symbol table. During the loader initialization, we register every valid syscall and the verifier checks the program bytecode against them. This procedure is briefly described in lines 58 to 63.

> I think we should do the central syscall registry repo regardless.

The registry repo is exactly what I've proposed we should do internally at Anza. An initial sketch of it is here: https://github.com/anza-xyz/agave/blob/master/sdk/define-syscall/src/codes.rs

alessandrod: I vote for going back to murmur32 since it makes it easier to interoperate with "svm extensions"

LucasSte: I don't have a hard preference, so I fine reverting the spec to mumur32. @ripatel-fd and @Lichtso are you OK with it?

0x0ece: out of curiosity, who are these alternative SVMs, L2s and syscalls?
(and btw, there's no requirement for the indexes to be continuous, an L2 could for example start from 50 to avoid collisions. Or they could also add the syscall to the L1 SVM since this is Solana and not Ethereum...)

deanmlittle: To name a few: Soon, SonicSVM, Atlas, Eclipse, and <redacted>. There are many use cases not well-served by mainnet-beta due to protocol development grinding to a halt for the past 2 years due to obvious reasons. People are starting to innovate elsewhere, and that's actually a very good thing, as if any of these experiments gain adoption or unlock new use cases, we have the ability to adopt them later after being tested in the wild. Making an absolute mess of the enum by padding the indices may sound like a great idea today, but the 20 different versions of the enum depending upon the chain and the inevitable 50th mainnet syscall breaking everything will eventually prove otherwise. If cross-SVM compatibility is of any importance, the Murmur hash is a much better solution for avoiding collisions.

LucasSte: Thanks for all the feedback. I'll revert the syscalls code back to murmur32 hash. I'll update the SIMD text shortly.

proposals/0178-static-syscalls.md
```diff
@@ -0,0 +1,98 @@
...
+scheme does not require symbol relocations and obviates the need for symbols 
+to be referenced in the table.
+
+### New return instruction
+
+The opcode `0x9D` must represent the return instruction, which supersedes the 
+`exit` instruction. The opcode (opcode `0x95`), previously assigned to the 
+`exit` instruction, must now be interpreted as the new syscall instruction.
+
+The verifier must detect an SBPF V1 program containing the `0x9D` opcode and
```
0x0ece: nit: SBPF V1 -> V3 (2x in this paragraph)

LucasSte: It is V1, isn't it? `0x9D` is exclusive to V3, so the verifier must throw `VerifierError::UnknownOpCode` for any non-V3 programs that contain `0x9D`.

I'll update the paragraph to mention non-V3, instead of V1.

0x0ece: Ah ok, I misinterpreted the sentence. All good, thanks.

Lichtso (approved):

joncinque:
It looks like there's approval for this from Firedancer and Anza. I will merge the SIMD next Friday to give time for any final objections.

deanmlittle:
Is the idea still to add the 'syscall' instruction? I think that's a very dumb idea. We can already just use call with different parameters to avoid forking ebpf bytecode for no reason.

Lichtso:
> Is the idea still to add the 'syscall' instruction?

Yes, the SIMD currently proposes the separation of `call` and `syscall` into two separate opcodes. Previously, they were differentiated by external calls having relocations. Now, with all relocations removed, internal calls continue to encode their target as a relative address (like other branches and jumps), while external calls encode the hash code of their target.

We did consider abusing other fields such as the `src` or `dst` register fields to have two flavors of the instruction, but ultimately decided it was cleaner to have two opcodes instead.

alessandrod:
> We did consider abusing other fields such as the src or dst register fields to have two flavors of the instruction, but ultimately decided it was cleaner to have two opcodes instead.

Abuse? It's how upstream does it, and how my implementation of static syscalls did it. If you didn't think of the design, it's not automatically abuse.

Lichtso:
> If you didn't think of the design, it's not automatically abuse.

Has nothing to do with who thought of what. Ideas have value independent of who spoke them first.

> Abuse?

You can also call it "misappropriation" if you like that word better. Still, it is not what these register fields are intended for. The opcode is what discriminates instructions.

alessandrod:
> > If you didn't think of the design, it's not automatically abuse.
> 
> Has nothing to do with who thought of what. Ideas have value independent of who spoke them first.
> 
> > Abuse?
> 
> You can also call it "misappropriation" if you like that word better. Still, it is not what these register fields are intended for. The opcode is what discriminates instructions.

So to recap: upstream thinks it's fine; I think it's fine and not worth diverging with upstream, and do the work to update compiler, vm and validator accordingly; the only people outside Anza who work on a toolchain think it's fine.

You disagree with everyone, go and actively undo my work for literally no tangible benefit because according to you "it's not elegant".

Doesn't seem great does it?

Lichtso:
> You disagree with everyone

IIRC, this specific request to split them into two opcodes came from the Firedancer team during Breakpoint last year. And I agreed. There are different parties and they have different opinions, please refrain from driving individual people into a corner just because you disagree.

> go and actively undo my work ...

Calm down, it is only a slight modification to the instruction encoding. Most of the logic of the static syscalls feature remains.

deanmlittle:
The biggest problem with all of these SIMDs is that they start out as a good ideas and devolve into a bad implementation full of unnecessary experiments that really don't need to be tried.

ripatel-fd:
> to avoid forking ebpf bytecode for no reason [...] I think that's a very dumb idea

@deanmlittle Just for context, Solana Labs' and Anza's prior design philosophy so far has been to acknowledge that the SVM is a completely different environment than kernel eBPF, and go with what's simplest/makes the most sense for Solana. The SVM has a runtime privilege model similar to x86 rings, where the slot boundary has highly privileged access to everything, the transaction executor has reduced access, and the VM has minimal access. Accordingly, it makes perfect sense to have another opcode for a context switch/privilege change, considering we've had a compile toolchain independent of upstream eBPF for the last two years or so.

So for one camp, it's uncontroversial and obvious byte code evolution, afor another camp it's "very dumb idea". And the reason is that you and the author have different design philosophy. You and Alessandro seem to care a lot about maintaining eBPF compatibility, and the authors of this SIMD follow previously agreed upon design guidance of treating sBPF as it's own thing. 

I really don't care how we solve it in the end. The acute pain of having an overcomplicated error-prone VM is larger than breaking compatibility with unused LLVM code. But if we can avoid breaking LLVM code, even better I guess.

@alessandrod @deanmlittle If you strongly care about this, then please submit a competing SIMD proposal doing this change in a way you think is cleaner. An alternative proposal is more productive than fighting in the comments how one proposal should look like, IMO.

deanmlittle:
@riptl good point. Will consider making a competing SIMD. That said, I don't wholly agree that is what Anza thinks, even if that is what Anza has ended up doing. Anza originally hired @alessandrod with the express goal of being able to use upstream ebpf bytecode and solve for Solana-specific idiosyncrasies in our own VM to end toolchain hell. Most (all?) core devs agree this is a very good idea. For whatever reason, these VM-related SIMDs are constantly conflating good ideas, like static syscalls, with bad ones, like inventing new mnemonics for the sake of it. Further divergence from upstream because "i feel like it" without any tangible benefit definitely deserves heavy scrutiny, and in absence of a competing SIMD, this is the most reasonable forum to voice those concerns.

Lichtso:
Two months ago I asked you on Discord to elaborate what you want and formulate that in SIMDs.

Estimated from your comments so far, how does the following sound to you?

- Revert SBPFv2 ([SIMD-0173](https://github.com/solana-foundation/solana-improvement-documents/pull/173), [SIMD-0174](https://github.com/solana-foundation/solana-improvement-documents/pull/174)) entirely
- Start at SBPFv1 ([SIMD-0166](https://github.com/solana-foundation/solana-improvement-documents/pull/166)) bytecode, to retain the dynamic stack feature
- But with SBPFv3 ELF parsing ([SIMD-0189](https://github.com/solana-foundation/solana-improvement-documents/pull/189)), to eliminate relocations
- And SBPFv3 verification ([SIMD-0179](https://github.com/solana-foundation/solana-improvement-documents/pull/179))
- Turn `e_machine` back to `EM_BPF`
- Split `call` and `syscall` by `src` reg instead of by opcode

deanmlittle:
> Two months ago I asked you on Discord to elaborate what you want and formulate that in SIMDs.
> 
> Estimated from your comments so far, how does the following sound to you?

Really appreciate you putting these together in a constructive way. I think the SIMDs are awkward in the sense that they are not small enough to support/reject individual pieces, but are also fragmented enough to make it hard to organize a conversation about them, as each SIMD spans multiple changes.

> * Revert SBPFv2 ([SIMD-0173](https://github.com/solana-foundation/solana-improvement-documents/pull/173), [SIMD-0174](https://github.com/solana-foundation/solana-improvement-documents/pull/174)) entirely

Yes, imo, >64 bit math should be done with a bignum math syscall that operates on a stack allocation the size of two operands and writes the result in little endian to the second operand, zeroing or overflowing into the space of the first operand for a double wide result making it super easy to check for overflow and/or reuse the first operand in a following operation. The syscall should then be JITed into the fastest possible instruction set, like AVX or similar. We do not need to fork the bytecode to achieve great performance and solve a lot of bignumber problems for fast cryptography and defi.

> * Start at SBPFv1 ([SIMD-0166](https://github.com/solana-foundation/solana-improvement-documents/pull/166)) bytecode, to retain the dynamic stack feature

I am pretty ambivalent about this one. I'm not sure if it creates further incompatibility issues with upstream ebpf or not. I am fair surely that it does not solve a real problem though. Stack issues are primarily a result of our least performant library, Anchor, incentivizing poor program design through the use of dynamic types and copying everything to the stack instead of mutating it in place, along with users not knowing what any of those words mean, or how to use zero copy, or box accounts to copy them to the heap instead. If the total amount of stack space is not increased, I don't think this solves a real problem for our users. This issue is likely best addressed by fixing Anchor, something I also work on from time to time.

> * But with SBPFv3 ELF parsing ([SIMD-0189](https://github.com/solana-foundation/solana-improvement-documents/pull/189)), to eliminate relocations

Yes, text based relocations are definitely bad.

> * And SBPFv3 verification ([SIMD-0179](https://github.com/solana-foundation/solana-improvement-documents/pull/179))

I don't see any good reason for strict header verification. Especially not the whole defining the stack/heap part of it. If the need arises to do this later to diverge from VM defaults, we can always do that.

> * Turn `e_machine` back to `EM_BPF`

Yes, we don't need to invent a fake machine ID. We should try to maintain compatibility with ebpf bytecode so as to not break debugger/decompiler/whatever external tooling.

> * Split `call` and `syscall` by `src` reg instead of by opcode

Yes, this is how upstream does it and it makes perfect sense to maintain this standard.

topointon-jump:
> I think the SIMDs are awkward in the sense that they are not small enough to support/reject individual pieces, but are also fragmented enough to make it hard to organize a conversation about them, as each SIMD spans multiple changes.

Completely agree with this. Imo in future we should use https://github.com/solana-foundation/solana-improvement-documents/discussions for high-level/contentious/strategic discussions about _what_ we want to do (e.g. compatibility with upstream), then each SIMD can be about _how_ we do it.

Lichtso:
>  Imo in future we should use https://github.com/solana-foundation/solana-improvement-documents/discussions

Took the liberty to condense all of this in https://github.com/solana-foundation/solana-improvement-documents/discussions/346

ripatel-fd:
> Yes, imo, >64 bit math should be done with a bignum math syscall that operates on a stack allocation the size of two operands and writes the result in little endian to the second operand, zeroing or overflowing into the space of the first operand for a double wide result making it super easy to check for overflow and/or reuse the first operand in a following operation

@deanmlittle I don't think this is a good idea. Why do we need to spill to memory? We should be able to translate multiplies to x86 directly. `IMUL` (64x64->128) has awesome 1 cycle throughput. Spilling to memory is probably going to break that nice ILP. Most instruction sets have the concept of 128-bit return values, why not eBPF?

> The syscall should then be JITed into the fastest possible instruction set, like AVX or similar. We do not need to fork the bytecode to achieve great performance and solve a lot of bignumber problems for fast cryptography and defi.

x86 does not have 64-bit wide multiplies in SIMD hardware, the widest you can get is 52-bit with the IFMA engine. And from my experience using parallel 32-bit multiplies is practically slower for cryptographic code than just using IMUL. I don't think it's possible to do these sorts of transformations general-purpose anyways. You would have to rewrite a larger algorithm to stay in 52-bit AVX10 or 32-bit AVX2 representation and schedule multiple independent multiplies per op. This is such a difficult data flow problem that Clang cannot do this automatically, you have to write these with SIMD intrinsics by hand. And then you have to deal with all these vector register dependencies during JIT compilation. So AVX stuff should definitely not be in a JIT with Solana's security requirements.

So for on-chain cryptography, which is basically always large finite field math, you basically need 3 ingredients: 
- the widest, highest throughput possible multiply (either AVX512-IFMA or scalar IMUL on x86)
- fast add with carry chains (e.g. ADCX/ADOX on x86)
- Do the two above with good ILP (meaning the generated x86 byte code is arranged in a way that the CPU frontend can schedule to multiple ALUs and a bunch of other things)

I might be missing a few specializations, but I'm pretty sure you can build everything with these primitives. And spilling to memory kinda goes against point 3

LucasSte:
I've updated this SIMD with the agreement we had on our offline discussion: to keep compatibility with the eBPF ISA.

Having said that, my last commit has these changes:
1. Do not deprecate the `exit` instruction anymore nor create a `return` instruction.
2. Do not create a new opcode for syscalls.
3. Use the `call` (opcode `0x85`) with the source field set to zero for representing external calls (syscalls).

ripatel-fd:
@jacobcreech @LucasSte @alessandrod Let's merge this this week. No comments since 30 days, seems like everyone is happy?

ripatel-fd:
cc @deanmlittle

deanmlittle:
I sincerely appreciate previous feedback being addressed and incorporated. This SIMD now cleanly deprecates dynamic relocations for syscalls in exactly the way a BPF developer would expect, while maintaining ISA compatibility and reducing long-term tooling burden. Excellent work.

I don’t wish to delay this any further, but I do have one follow-up that feels worth floating:

It appears it may actually be straightforward to safely back-port this to SBPF v0 so we could ship this sooner without needing to complete the entire v3 roadmap. From what I can tell, the toolchain always emits call (0x85) with src = 1, and users cannot even change this in assembly. The only way to vary src would be to hand-edit generated bytecode, something no one would do in practice. Bit flippers trying to break the VM are not really high on my list of customers we should be caring about supporting.

Unless there’s some historical exception I’m unaware of, this should mean any case where `src = 0` and `imm` is a known static call target can already safely be treated as a static syscall. Even if this is not the case, it would also be trivial to gate static syscalls behind the absence of a dynamic relocation table/dynamic program header. A simple regression test combining both static targets and dynamic relocations would confirm that relocations take precedence, which I believe matches the intended semantics.

Overall, this is in excellent shape. I’d love to see it merged. It will certainly make our life much easier on the linker side of things!

Lichtso:
I suppose we could introduce static syscalls globally, instead of having them coupled to SBPFv3. Though I would have to confirm that it really wouldn't break any MNB programs first.

deanmlittle:
> I suppose we could introduce static syscalls globally, instead of having them coupled to SBPFv3. Though I would have to confirm that it really wouldn't break any MNB programs first.

Let's do it! We'll also help to confirm from our side

deanmlittle:
In over 14,000 programs tested, we were unable to find a single instance of a currently deployed program under Loader V3 containing a `call` instruction with a `src` register other than `1`. Thus, I see no reason why we shouldn't activate static syscalls globally. I think if we are all in agreement here, we should ship this ASAP. Super easy win for smaller binaries and much simpler linker logic.

Lichtso:
> unable to find a single instance of a currently deployed program under Loader V3 containing a call instruction with a src register other than 1

That is the problem, the syscalls should have `src=0` and only internal calls should be `src=1`. Thus, I came to the conclusion, that programs do not correctly differentiate between `call` and `syscall` by `src` register. See specs for reference: https://www.ietf.org/archive/id/draft-thaler-bpf-isa-00.html#name-jump-instructions or https://www.kernel.org/doc/html/next/bpf/instruction-set.html#jump-instructions

In fact it gets worse, the old ELF parser does implicitly [expect syscalls do be marked by `imm=-1` instead](https://github.com/anza-xyz/sbpf/blob/163efda92dcb4a7281fe462127170931efc0dc84/src/elf.rs#L902) and expects [internal calls to use relative addressing](https://github.com/anza-xyz/sbpf/blob/163efda92dcb4a7281fe462127170931efc0dc84/src/elf.rs#L903-L905). Furthermore, the situation is complicated by [internal call targets being internally Murmur3 hashed](https://github.com/anza-xyz/sbpf/blob/163efda92dcb4a7281fe462127170931efc0dc84/src/program.rs#L152). And there is the `BpfRelocationType::R_Bpf_64_32` relocation [messing with what counts as `call` or `syscall`](https://github.com/anza-xyz/sbpf/blob/163efda92dcb4a7281fe462127170931efc0dc84/src/elf.rs#L1091-L1117) as well.

So, while I agree that it would be very nice to cleanup this mess for all programs (not just SBPFv3), I don't see it being possible with what is currently deployed on mainnet-beta.

deanmlittle:
> That is the problem, the syscalls should have `src=0` and only internal calls should be `src=1`. Thus, I came to the conclusion, that programs do not correctly differentiate between `call` and `syscall` by `src` register. See specs for reference: https://www.ietf.org/archive/id/draft-thaler-bpf-isa-00.html#name-jump-instructions or https://www.kernel.org/doc/html/next/bpf/instruction-set.html#jump-instructions

I'm glad we agree that the runtime does not correctly differentiate between `call` and `syscall` by `src` register. That is the entire crux of my argument.

>In fact it gets worse, the old ELF parser does implicitly [expect syscalls do be marked by imm=-1 instead](https://github.com/anza-xyz/sbpf/blob/163efda92dcb4a7281fe462127170931efc0dc84/src/elf.rs#L902) and expects [internal calls to use relative addressing](https://github.com/anza-xyz/sbpf/blob/163efda92dcb4a7281fe462127170931efc0dc84/src/elf.rs#L903-L905).

This confirms that not only does the tooling never emit`call` instructions with a `src` other than `1` for dynamically relocated syscalls, in relocations, it also already has an explicit `imm` requirement in parsing. This further supports my argument.

>Furthermore, the situation is complicated by [internal call targets being internally Murmur3 hashed](https://github.com/anza-xyz/sbpf/blob/163efda92dcb4a7281fe462127170931efc0dc84/src/program.rs#L152). And there is the BpfRelocationType::R_Bpf_64_32 relocation [messing with what counts as call or syscall](https://github.com/anza-xyz/sbpf/blob/163efda92dcb4a7281fe462127170931efc0dc84/src/elf.rs#L1091-L1117) as well.

Problems in the old ELF Parser that only exist because, perplexingly, we seem to parse instructions to guess if they *might* need relocation (???) instead of just parsing the relocation table and applying them, are really not relevant here. This is an entirely self-inflicted problem that is completely optional and under our control. If that were not the case, perhaps we'd have something sane that works already.

>So, while I agree that it would be very nice to cleanup this mess for all programs (not just SBPFv3), I don't see it being possible with what is currently deployed on mainnet-beta.

This assertion may stem from you misunderstanding what is being proposed here. The outcomes of enabling static syscalls within the limitations of the current system are, in fact, precisely the opposite to what you state here.

*Because* of the following conditions:
- all `call` instructions currently use `src=1`
- consequently, there are also zero deployed call instructions using `src=0`
- all internal calls provide relative offsets
- the only distinction between internal calls and syscalls is the existence of a corresponding dynamic relocation table entry
- the precondition for dynamic relocations is the presence of a DYNAMIC header in the binary

This means:
- Any call where src=0 can always safely be assumed to be a static syscall on any currently active sbpf target arch
- Where a DYNAMIC header is not present, we can simply ignore dealing with dynamic relocations entirely (read: soft deprecation)
- Where a DYNAMIC header is present, along with a relocation table entry, and it points to a call with src=1, relocations win.

Unless anyone could propose a case where the above conditions are somehow confounded in a way that makes this unviable, I see nothing that would make it non-trivial for us to deliver this feature for all targets rather than having to execute on the entire v3 roadmap to get it live.

Lichtso:
> may stem from you misunderstanding what is being proposed here

I think what I was missing is that you want to precondition all of this on the presence / absence of a dynamic symbol table. That is almost the same as differentiating by the version flag in the header.

> I see nothing that would make it non-trivial for us to deliver this feature for all targets rather than having to execute on the entire v3 roadmap to get it live.

SBPFv3 has been cut down to only be eBPFv3 upstream compatible + dynamic stack frames + static syscalls & reduced headers. All of that is implemented, we only have to iron out a few details. I don't think that making one more intermediate version in which we only have static syscalls & reduced headers but not eBPFv3 upstream compatibility is going to speed things up.

What I was interested in, is a solution which allows us to stop hashing and relocating the local / internal calls for all (already) deployed programs. Which might be possible with some workarounds, but is tricky to get right.

deanmlittle:
>I think what I was missing is that you want to precondition all of this on the presence / absence of a dynamic symbol table. That is almost the same as differentiating by the version flag in the header.

That's not at all what I was saying. I don't think this is a requirement at all. I was simply pointing out that there is already a good way to know a binary doesn't require dynamic relocations without having to bump a version, and we should probably be doing this to avoid unnecessary relocations today anyway.

>SBPFv3 has been cut down to only be eBPFv3 upstream compatible + dynamic stack frames + static syscalls & reduced headers. All of that is implemented, we only have to iron out a few details. I don't think that making one more intermediate version in which we only have static syscalls & reduced headers but not eBPFv3 upstream compatibility is going to speed things up.

Historically, it has been difficult to get agreement on many smaller details that slip between the cracks when it comes to this broader roadmap. If it is possible to agree upon something that can easily be shipped today, we should do it to deliver on small, incremental improvements ASAP.

>What I was interested in, is a solution which allows us to stop hashing and relocating the local / internal calls for all (already) deployed programs. Which might be possible with some workarounds, but is tricky to get right.

So stop hashing them? What matters is unique symbols, not how you achieve those unique symbols. I don't think this is particularly relevant to the discussion?

proposals/0178-static-syscalls.md
```diff
@@ -0,0 +1,87 @@
...
+
+The changes proposed in this SIMD are transparent to dApp developers. The 
+compiler toolchain will emit correct code for the specified SBF version. 
+Static syscalls obviate relocations for call instructions and move the virtual 
+machine closer to eliminating relocations altogether, which can bring 
+considerable performance improvements.
+
+## Security Considerations
+
+None.
```
ripatel-fd, line 105: All of these runtime SIMDs have security considerations. 
The risk generally with bytecode format changes are confusion issues both in validator implementations and external tooling. 
For example, static analysis tooling that auditors use might break, which is an indirect security concern. 

But don't insist on adding a pamphlet here. The rest of the document looks great.

ripatel-fd:
@Lichtso @LucasSte Programs built with the Anza SDK depend on dynamic stack frames, right? What is the long term plan here if dynamic stack frames are being removed with sBPFv3?

Lichtso:
> Programs built with the Anza SDK depend on dynamic stack frames, right? What is the long term plan here if dynamic stack frames are being removed with sBPFv3?

See the opt-in "dynamic stack frames" in #377

topointon-jump (approved):
Let's ship it!

proposals/0178-static-syscalls.md
```diff
@@ -0,0 +1,87 @@
...
+`VerifierError::InvalidSyscall` otherwise.
+
+This new instruction comes together with modifications in the semantics of 
+the instruction opcode `0x85` with source register set to one, which  must 
+only refer to internal calls and its immediate field must only be interpreted 
+as a relative address to jump from the program counter.
+
+Syscall names must NOT be present in the symbol table anymore, since the new 
+scheme does not require symbol relocations and obviates the need for symbols 
+to be referenced in the table.
```
deanmlittle:
```suggestion
As static syscalls don't use dynamic text-based relocations, the `PT_DYNAMIC` 
program header, along with the `.dynamic`, `.dynsym`, `.dynstr` and `.rel.dyn` 
section header entries will no longer be required to invoke static syscalls 
for any SBPFv3 program and will be omitted by default.
```

As it stands, the SIMD lacks any mention of the PT_DYNAMIC program header being removed. I think this is an apt place to put that. As we will no longer do any checks on section header entries in SBPFv3, technically, "must NOT be present" is actually incorrect, as you will actually be able to put whatever you want in the program section header entries, we just no longer have this requirement for syscalls moving forward.

proposals/0178-static-syscalls.md
```diff
@@ -0,0 +1,87 @@
...
+## Motivation
+
+The resolution of syscalls during ELF loading requires relocating addresses, 
+which is a performance burden for the validator. Relocations require an entire 
+copy of the ELF file in memory to either relocate addresses we fetch from the 
+symbol table or offset addresses to after the start of the virtual machine's 
+memory. Moreover, relocations pose security concerns, as they allow the 
+arbitrary modification of program headers and programs sections. Introducing 
+static syscalls allows us to resolve all program relocations during link time.
+
```
deanmlittle, line 40:
```suggestion
## Dependencies

This proposal depends on the following previously accepted proposal:

- **[SIMD-0189]: SBPF stricter ELF headers**

    Deprecation of PT_DYNAMIC header for SBPFv3 programs

[SIMD-0189]: https://github.com/solana-foundation/solana-improvement-documents/pull/189

```

Consider making this SIMD depend upon SIMD-0189 to make the activation order more explicit.

proposals/0178-static-syscalls.md
```diff
@@ -0,0 +1,98 @@
...
+The following must go into effect if and only if a program indicates the SBPF 
+version `0x03` or higher in its ELF header e_flags field, according to the 
+specification of SIMD-0161.
+
+### Static syscall instruction
+
+
+We follow the encoding referenced in the 
+[eBPF specification](https://www.ietf.org/archive/id/draft-thaler-bpf-isa-00.html#name-jump-instructions)
```
t-nelson: is there a pdf version of this reference that we can archive here in the repo? internet archive snapshot would be a close second. mention at the bottom of the document

proposals/0178-static-syscalls.md
```diff
@@ -0,0 +1,98 @@
...
+for encoding static syscalls. That means we must use the call instruction 
+(opcode `0x85`) with the source register field set to zero. The immediate field 
+must be filled with a registered syscall hash code. For more reference on the 
+SBF ISA format, see the 
+[spec document](https://github.com/solana-labs/rbpf/blob/main/doc/bytecode.md).
+
+We define the hash code for a syscall as the murmur32 hash of its respective 
+name. The 32-bit immediate value of the call instruction must be the 
+integer representation of such a hash. For instance, the code for `abort` is 
+given by `murmur32("abort")`, so the instruction hexadecimal representation
```
t-nelson, line 65: how are collisions handled?

Lichtso: All syscalls are registered in a registry of the loader, before any ELF program is loaded. We will simply never introduce any syscalls with conflicting names and we would notice immediately in the tests because the registry would complain.

proposals/0178-static-syscalls.md
Lichtso: This is the latest capture by the internet archive: https://web.archive.org/web/20250117081057/https://www.ietf.org/archive/id/draft-thaler-bpf-isa-00.html

proposals/0178-static-syscalls.md
deanmlittle: Also worth noting there is no mechanism of handling hash collisions between internal call and external calls today.  You cannot call to an offset of -1 for example because it is reserved as a marker for syscalls. This new solution is explicitly better than what we have in this regard as it follows the upstream practice of differentiating internal and external calls by src register.

proposals/0178-static-syscalls.md
deanmlittle: Yes. This outdated proposal was upgraded to an RFC in 2024: https://www.rfc-editor.org/rfc/rfc9669.pdf

LucasSte: > is there a pdf version of this reference that we can archive here in the repo?

I'll include The PDF @deanmlittle mentioned, and link it to the SIMD. From an organization standpoint, where should this PDF be stored? Should we create a `documents` folder in the repository?

t-nelson: > Should we create a documents folder in the repository?

that sounds appropriate to me. @jacobcreech this is your house, wdyt?

proposals/0178-static-syscalls.md
t-nelson: ok yeah i don't really care how collisions are handled so much as i didn't seem them mentioned here despite their existing pretty clearly being possible. i'm guessing something like "care must be taken not to introduce syscall names that produce hash collisions" would be appropriate?

proposals/0178-static-syscalls.md
LucasSte: > that sounds appropriate to me.

I did just that in https://github.com/solana-foundation/solana-improvement-documents/pull/178/commits/4f80c3750c931074c370a6356d7f04fa8b7a4a36.

proposals/0178-static-syscalls.md
LucasSte: In https://github.com/solana-foundation/solana-improvement-documents/pull/178/commits/9c37a5a3254a9bef1773002359061491d839fb9d, I condensed all the three previous remarks in a paragraph about collisions.

proposals/0178-static-syscalls.md
jacobcreech: Seems reasonable to get out the door.

bw-solana:
Latest looks good from my side 👍

buffalojoec (approved):

PR #179, opened 2024-10-03
SIMD-0179: SBPF stricter verification constraints

LucasSte:


proposals/0179-stricter-verification.md
```diff
@@ -0,0 +1,105 @@
...
+verification rules to prevent any unwarranted behavior.
+
+## New Terminology
+
+None.
+
+## Detailed Design
+
+The following must go into effect if and only if a program indicates the SBF 
+version XX or higher in its ELF header e_flags field, according to the
```
topointon-jump, line 38: Let's specify this version

proposals/0179-stricter-verification.md
```diff
@@ -0,0 +1,105 @@
...
+All jump instructions, except for `call` (opcode `0x85`) and `callx` (opcode 
+`0x8D`), must now jump to a code location inside their own function. Jumping 
+to arbitrary locations hinders a precise program verification. 
+`VerifierError::JumpOutOfCode` must be thrown for offending this rule.
+
+`call imm` (opcode `0x85`) must only be allowed to jump to a program counter 
+previously registered as the start of a function. Otherwise 
+`VerifierError::InvalidFunction` must be thrown. Functions must be registered 
+if they are present in the symbol table. The entrypoint to the program must 
+also define a valid function.
```
topointon-jump, line 62: Will this require a double pass through the bytecode to pick up all call destinations?

proposals/0179-stricter-verification.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Runtime check for callx
+
+The jump destination of `callx` (opcode `0x8D`) must be checked during 
+execution time to match the initial address of a registered function. If this 
+is not the case, a `EbpfError::UnsupportedInstruction` must be thrown. This 
+measure is supposed to improve security of programs, disallowing the malicious 
+use of callx.
```
topointon-jump, line 70: This has performance implications - is this absolutely necessary? Long-term we can maybe have the compiler mark the beginning of valid jump destinations by emitting a marker bytecode opcode.

topointon-jump (commented):
Some of these stricter rules, especially those around restricting jump destinations, have performance implications. Commented on these.

proposals/0179-stricter-verification.md
```diff
@@ -0,0 +1,105 @@
...
+program counter, resulting in a fallthrough into another function’s code. 
+Offending this new validation condition must throw an 
+`VerifierError::InvalidFunction` error.
+
+### Restrict jump instruction destination
+
+All jump instructions, except for `call` (opcode `0x85`) and `callx` (opcode 
+`0x8D`), must now jump to a code location inside their own function. Jumping 
+to arbitrary locations hinders a precise program verification. 
+`VerifierError::JumpOutOfCode` must be thrown for offending this rule.
```
topointon-jump, line 56: How is this enforced? Is this enforced in the verifier or at runtime?

topointon-jump (commented):
Can we specify exactly how these rules are enforced:

- Whether the check is performed at runtime or verification time
- The exact logic (in pesudo-code)

A lot of these seem like they would have performance implications, and precisely specifying them would help to determine if this is the case or not.

proposals/0179-stricter-verification.md
Lichtso: The symbol table is not the bytecode. Call destinations in the bytecode are not registered as functions.

proposals/0179-stricter-verification.md
Lichtso: This is a check at verification time (see error code) and is easy to enforce:
The verifier scans all instructions and keeps track of which function it is currently in, then checks jump targets against that range. The current function tracking can be done in constant time during the scan by checking the next symbol in the dynamic synbol table and then advancing that pointer once a function boundary is crossed.

Lichtso:
- Almost all of the checks are verification time, only exception is the `callx` target location check, that is at runtime
- All our checks are O(1), O(log n) or O(n) and can be performed in [a single forward scan of the bytecode](https://github.com/solana-labs/rbpf/blob/69a52ec6a341bb7374d387173b5e6dc56218fe0c/src/verifier.rs#L227-L419)
- About the `callx` target location runtime check:
  - Our suggestion: We register functions in the dynamic symbol table, then `call` target verification can be done in the single scan together with all other instructions and `callx` target runtime check is O(log n)
  - Your suggestion: We register functions in the bytecode, then `call` target verification requires a second scan of the bytecode, and the `callx` target runtime check becomes O(1)

proposals/0179-stricter-verification.md
Lichtso: Better wording might be:
> Functions are registered by presence in the symbol table.

Lichtso:
> * Almost all of the checks are verification time, only exception is the `callx` target location check, that is at runtime
> 
>     * All our checks are O(1), O(log n) or O(n) and can be performed in [a single forward scan of the bytecode](https://github.com/solana-labs/rbpf/blob/69a52ec6a341bb7374d387173b5e6dc56218fe0c/src/verifier.rs#L227-L419)
> 
>     * About the `callx` target location runtime check:
>       
>       * Our suggestion: We register functions in the dynamic symbol table, then `call` target verification can be done in the single scan together with all other instructions and `callx` target runtime check is between O(1) and O(log n)
>       * Your suggestion: We register functions in the bytecode, then `call` target verification requires a second scan of the bytecode, and the `callx` target runtime check becomes O(1)

proposals/0179-stricter-verification.md
LucasSte: > Functions are registered by presence in the symbol table.

I was going to adopt your wording, but basically I already say this:

> Functions must be registered if they are present in the symbol table.

Is this dubious or ambiguious?

proposals/0179-stricter-verification.md
topointon-jump: Could we add this description^ to the SIMD? 🙏

proposals/0179-stricter-verification.md
topointon-jump: Yea - could we specify that we are treating the symbol table as the source of truth for what is considered a valid call dest? 🙏

topointon-jump:
> Almost all of the checks are verification time, only exception is the callx target location check, that is at runtime
> All our checks are O(1), O(log n) or O(n) and can be performed in [a single forward scan of the bytecode](https://github.com/solana-labs/rbpf/blob/69a52ec6a341bb7374d387173b5e6dc56218fe0c/src/verifier.rs#L227-L419)

Could we add more details about these checks and their performance properties to the SIMD itself, in pseudo-code? I was reading the SIMD without looking at the code, and a lot of these performance properties weren't clear until I read the code.

After reading the code most of my concerns have been addressed - it would be great to move towards a world where SIMDs can be read independently of the code.

topointon-jump:
> Your suggestion: We register functions in the bytecode, then call target verification requires a second scan of the bytecode, and the callx target runtime check becomes O(1)

To be clear, our suggestion was
- Change the bytecode emitted to have a special marker bytecode, signifying the start of the function
- We can then just check for that bytecode marker in the next instruction right before we jump

proposals/0179-stricter-verification.md
topointon-jump: Having implemented a POC version of this I think this is unnecessary now, as the check is quite cheap given that we have a bitset of function calls.

proposals/0179-stricter-verification.md
LucasSte: I included more details on this.

proposals/0179-stricter-verification.md
LucasSte: Added.

proposals/0179-stricter-verification.md
Lichtso: I have grown to like this idea over the dynamic symbol table, thus I removed that from SIMD-0189.

proposals/0179-stricter-verification.md
```diff
@@ -28,19 +28,39 @@ The new syscall instruction introduced in SIMD-0178 permits the
...
+version `0x3` or higher in its ELF header e_flags field, according to the 
 specification of SIMD-0161.
 
+### Restrict functions' first instruction
+
+Since SIMD-0166, SBPF functions can modify the stack pointer using the 
+`add64 r10, imm` (opcode `0x07`) with the restriction that `r10` is the 
+destination register and the source is an immediate value.
+
+The instruction `add64 r10, imm` must be used as a marker for a function
```
Lichtso, line 53: This should also mention that a function can not have another `add64 r10, imm` as that would count as the start of the next function / split the function in two.

LucasSte: Functions have two `add64 r10, imm` instructions. One for decrementing the stack pointer and another for incrementing the stack pointer.

Why don't we restrict the start of a function to when `imm` is either zero or negative?

Lichtso: `r10` is restored upon `return` automatically (it is on the VMs shadow stack: https://github.com/anza-xyz/sbpf/blob/main/src/interpreter.rs#L579). Thus, there is no need to counter act the stack allocation with a second stack bump in the opposite direction I think.

Lichtso (approved):

proposals/0179-stricter-verification.md
```diff
@@ -0,0 +1,133 @@
...
+destination register and the source is an immediate value.
+
+The instruction `add64 r10, imm` must be used as a marker for a function 
+start. Therefore, the verifier must enfroce that functions start with 
+an `add64 r10, imm` instruction that modifies the stack pointer `r10`, without 
+any restriction for the immediate value.
+
+The only exception to this requirement is the first function in the ELF 
+text area. Since the first instruction in the ELF can only represent a 
+function beginning, the verifier must not require it to be an `add64 r10, imm`.
```
0x0ece: Remove this sentence? We're checking that the 1st ix is `add64 r10, imm`.
https://github.com/anza-xyz/sbpf/blob/v0.12.2/src/verifier.rs#L233-L235

0x0ece (approved):

proposals/0179-stricter-verification.md
LucasSte: Fixed in https://github.com/solana-foundation/solana-improvement-documents/pull/179/commits/cedc1107ebcb54e094867a248462bd651da59075.

joncinque:
It looks like there's approval for this from Firedancer and Anza. I will merge the SIMD next Friday to give time for any final objections.

Lichtso:
Given the ongoing discussions in #178 we should not merge this one for now.

LucasSte:
I incorporated in the SIMD the new design for #178 and #377. I left the section for jumping in the middle of LDDW in place, since this is still an open point for discussion.

LucasSte:
This SIMD won't be implemented anymore.

PR #184, opened 2024-10-15
block writeable account data limit

bw-solana:


jacobcreech:
@bw-solana do we know the current impact if this change was live the past 30 days?

bw-solana:
> @bw-solana do we know the current impact if this change was live the past 30 days?

Good question, I don't know. I'll get some data

proposals/0184-block-writeable-account-data-limit.md
```diff
@@ -0,0 +1,107 @@
...
+
+None
+
+## Detailed Design
+
+This feature would check the amount of account data marked as writeable for each
+transaction being handled by the block producer (for inclusion in a block) or
+validator (replaying transactions in a block), add it to a running total of
+total writeable account data for the block, and compare it against some
+threshold (recommend 2GB to start) to decide if the transaction is allowed to be
```
ilya-bobyr, line 43: Do we have statistics on what is the current distribution of writable data on the mainnet?
I wonder if we could start with an even lower value.

proposals/0184-block-writeable-account-data-limit.md
```diff
@@ -0,0 +1,107 @@
...
+  writeable for each transaction. This can be done similarly to how
+  `data_bytes_len_total` is computed inside `get_transaction_cost` but limited
+  to accounts marked writeable.
+2. Update the cost tracker to check the current writeable account data in the
+  block being produced + the amount of account data that would be written with
+  the current transaction against the total block limit in the would_fit
+  function
+3. Store the write limit in a constant such as
+  `MAX_BLOCK_ACCOUNTS_DATA_SIZE_WRITEABLE` and set this to `2_000_000_000`
+  (2GB).
```
ilya-bobyr, line 66: 2GB is `2_147_483_648` :P

proposals/0184-block-writeable-account-data-limit.md
```diff
@@ -0,0 +1,107 @@
...
+4. Add a new `CostTrackerError` type `WouldExceedAccountDataWriteableBlockLimit`
+  and return this error if block limit is exceeded.
+5. Treat this new error type (`WouldExceedAccountDataWriteableBlockLimit`) as
+  retryable in execute_and_commit_transactions_locked
```
ilya-bobyr, line 70:
```suggestion
5. Treat this new error type (`WouldExceedAccountDataWriteableBlockLimit`) as
  retryable in `execute_and_commit_transactions_locked()`.
```

proposals/0184-block-writeable-account-data-limit.md
```diff
@@ -0,0 +1,107 @@
...
+
+There is an edge case where if a single transaction tries to consume more than
+the entire per block writeable account data budget, it should not be marked as
+retryable. This might not be possible today given tx size limits and maximum
+account size limits, but this could be exploitable in the future if either of
+this constraints change.
```
ilya-bobyr, line 76: Would it make sense to introduce a separate constraint that limits the amount of writable data a single transaction can access?
We have `MAX_LOADED_ACCOUNTS_DATA_SIZE_BYTES` equal to 64MB which limits the total amount of data a transaction can load.
We could add `MAX_LOADED_ACCOUNTS_WRITEABLE_DATA_SIZE_BYTES` (with an obvious semantics).

proposals/0184-block-writeable-account-data-limit.md
```diff
@@ -0,0 +1,107 @@
...
+various resources transactions can consume. This will allow block producers to
+determine when a block is "full" and validators to mark slots dead that do not
+adhere to these limits.
+
+This feature could be implemented like so (using Agave client as reference):
+
+1. Update the cost model to compute the aggregate size of all accounts marked
+  writeable for each transaction. This can be done similarly to how
+  `data_bytes_len_total` is computed inside `get_transaction_cost` but limited
+  to accounts marked writeable.
```
ilya-bobyr, line 59: I think this is inaccurate.

Looking at the `get_transaction_cost()` in `cost-model`, I think, `data_bytes_len_total` only calculates the size of the instruction data passed into instructions.
It does not look at the account sizes:
```rust
            data_bytes_len_total =
                data_bytes_len_total.saturating_add(instruction.data.len() as u64);
```
https://github.com/anza-xyz/agave/blob/f9f8b60ca15fa721c6cdd816c99dfd4e9123fd77/cost-model/src/cost_model.rs#L187-L188

The loaded accounts data size is only included in the transaction cost after the transaction execution.
Actual value is passed into the `calculate_cost_for_executed_transaction()` as the `actual_loaded_accounts_data_size_bytes` argument.
And it is populated and checked in the `blockstore_processor.rs`:
```rust
                Some(CostModel::calculate_cost_for_executed_transaction(
                    tx,
                    committed_tx.executed_units,
                    committed_tx.loaded_account_stats.loaded_accounts_data_size,
                    &bank.feature_set,
                ))
```
https://github.com/anza-xyz/agave/blob/f9f8b60ca15fa721c6cdd816c99dfd4e9123fd77/ledger/src/blockstore_processor.rs#L246-L251

We could record have a similar counter for the amount of writable data that the transaction touched.

So at the high level it can be computed similarly to what we are already doing.
It is just that the details would be different.

Not sure if there are any implications for the DX, due to the fact that we can only compute this value after a transaction execution.

proposals/0184-block-writeable-account-data-limit.md
```diff
@@ -0,0 +1,107 @@
...
+## Detailed Design
+
+This feature would check the amount of account data marked as writeable for each
+transaction being handled by the block producer (for inclusion in a block) or
+validator (replaying transactions in a block), add it to a running total of
+total writeable account data for the block, and compare it against some
+threshold (recommend 2GB to start) to decide if the transaction is allowed to be
+included in the block (if sum is below the limit for the case of block producer)
+or if the slot needs to be marked dead (if sum is above the limit for the case
+of validator).
```
2501babe, line 46: andrew will have to look at this because if it can invalidate blocks then it may interfere with asynchronous execution

proposals/0184-block-writeable-account-data-limit.md
2501babe:
```suggestion
  `MAX_BLOCK_ACCOUNTS_DATA_SIZE_WRITEABLE` and set this to 2GiB (`2 * 1024 * 1024 * 1024`).
```

proposals/0184-block-writeable-account-data-limit.md
bw-solana: We don't have perfect stats, but I posted comment with some analysis below (https://github.com/solana-foundation/solana-improvement-documents/pull/184#issuecomment-2432814006)

2GiB feels right as a starting point. It should constrain worst case reasonably well without impacting current typical behavior on mainnet. We could ratchet things down (or up) after observing behavior

proposals/0184-block-writeable-account-data-limit.md
bw-solana: We could. We'll effectively be constrained to the 64MiB per tx for updated account data and 100MiB per block for newly allocated account data, so we already have some guard rails here, but we could be more explicit

proposals/0184-block-writeable-account-data-limit.md
bw-solana: Yep, you're right. We don't have any precedent for understanding actual account data before loading/executing transactions. We rely on either scraping top level instructions (e.g. for allocate) or estimating and then adjusting. 

This would seemingly have to fall into the estimate + adjust case (unless we want to do something radical like grab account data size before execution), which would likely impact DX if we push this onto the user similar to how we ask for CU/Data budget estimates.

One option is to leverage the existing Data budget value for writes as well (defaults to 2MiB per tx or can be set by user). I think this is likely small enough of a default to not bottleneck scheduling before the data budget gets refunded post execution. We would need to study this.

proposals/0184-block-writeable-account-data-limit.md
2501babe: its unfortunately quite different, transaction data size (sum of all accounts required for a transaction) can be calculated on the transaction level, and if the transaction exceed its limit, it is treated as invalid. so in a world with async execution, the transaction could be included in a block before verifying it, but all nodes will see that it violates its limits and treat it the same (charge fees and not execute it, once the feature gate for fee-only transactions is active)

but if we added a data size limit at the block level, you would need to know the running total of all previous transactions to determine whether you can put the transaction in the block. the purpose of removing constraints to enable async execution is that block-packers which put "bad" transactions in a block might make suboptimal blocks, but they can always be sure the block is valid if they follow constraints that dont depend on execution (are transactions syntactically valid, are blockhashes valid, etc)

i believe we could have a block-level constraint as proposed here if instead of "if you exceed the block data size limit, the block is invalid" it was like "if you exceed the block data size limit, no further transactions in the block will be executed." this could be user-hostile tho, because it lets block producers invalidate transactions that should otherwise succeed. especially if we charge fees, the block producer now has an incentive to exceed the limit and farm fees without having to execute transactions

alternatively we could very easily charge more for writable data at the transaction level, but im not sure that accomplishes the goals of this simd. andrew may have better suggestions than me tho

2501babe: yep, think we are on the same page! the way transaction data size works now is theres a default budget (64MiB), adjustable by compute budget instruction, and at execution time the accounts are totalled and the transaction is invalid if it exceeds its limit. but this doesnt affect the block, so it doesnt interfere with async execution

we could add a transaction-level writable accounts budget with a smaller limit (assume 2MiB like you suggest), and it would likewise be async-safe. we could easily have block-level constraints based on those requested budgets, which would also be async-safe. the main risk is people requesting enormous budgets and crowding other transactions out of the block, which we could defray with either a per-transaction maximum limit** or by exponentially ramping the cost per byte

if you make changes to the transaction data size calculation code, note that the formula is quite tricky. i have an open simd to fix this #186 

**(16-20MiB would allow writing to a 10Mib account plus plenty of breathing room, but i dont actually know how big the largest accounts are onchain. 10MiB is the max allocation size but i believe you can pump accounts larger than that with realloc)

Benhawkins18:
stale

PR #185, opened 2024-10-17
SIMD-0185: Vote Account v4

jstarry:


proposals/0185-vote-account-v4.md
```diff
@@ -0,0 +1,243 @@
...
+pub struct VoteStateV4 {
+    pub node_pubkey: Pubkey,
+    pub authorized_withdrawer: Pubkey,
+
+    /// NEW: the collector accounts for validator income
+    pub inflation_rewards_collector: Pubkey,
+    pub block_fees_collector: Pubkey,
+    pub block_tips_collector: Pubkey,
+
+    /// NEW: percentages (0-100) that represent how much of each income source
```
buffalu, line 122: nice to have, but not required: bps

proposals/0185-vote-account-v4.md
```diff
@@ -0,0 +1,243 @@
...
+first 4 bytes.
+
+#### `Authorize`, `AuthorizeChecked`, `AuthorizeWithSeed`, `AuthorizeCheckedWithSeed`
+
+Existing authorize instructions will be processed differently when setting new
+authorized voters. Rather than purging authorized voter entries from the
+`authorized_voters` field that correspond to epochs less than the current epoch,
+only purge entries less than the previous epoch (current epoch - 1). This will
+mean that the `authorized_voters` field can now hold up to 4 entries for the
+epochs in the range `[current_epoch - 1, current_epoch + 2]`. Keeping the
```
AshwinSekar, line 222: is there a reason we need to store `current_epoch + 2`?

jstarry: Yeah that's the current design and I think it kinda makes sense because otherwise you could set a new authorized voter at the end of an epoch and immediately start using it in the next block in the new epoch and leaders would need an up to date view on any new authorized voters for each fork crossing the epoch boundary

proposals/0185-vote-account-v4.md
jstarry: Yeah I considered that but it seems like even jito mev tip commissions are usually just integer percentages?

diman-io:
I'm sorry, why is `block_tips_collector` here? Are tips somehow planned in the protocol level / voting program?

jstarry:
> I'm sorry, why is block_tips_collector here? Are tips somehow planned in the protocol level / voting program?

Up for debate still, but I put it there for now since it's easier to update the vote account all in one go. If that field remains unused by the protocol it can be used for something else in the future.

diman-io:
Is this about a new instruction to pay the leader who landed a successful transaction?

proposals/0185-vote-account-v4.md
segfaultdoc: When dealing with money I think best practice to use bps:

https://github.com/jito-foundation/jito-programs/blob/master/mev-programs/programs/tip-distribution/src/state.rs#L47

segfaultdoc (commented):
Overall good proposal, trying to think longer term if the Tips commission piece makes sense or becomes obsolete

jstarry:
I updated the proposal to use basis points for commission values and to remove any special handling for block tips for now.

jstarry:
@AshwinSekar I've updated the SIMD to only remove the `prior_voters` field in a new version of vote state. This way the SIMD is much simpler and we can make progress towards adding other new fields to the vote state without needing to reallocations for existing vote accounts like we had to for TVC.

AshwinSekar:
> @AshwinSekar I've updated the SIMD to only remove the `prior_voters` field in a new version of vote state. This way the SIMD is much simpler and we can make progress towards adding other new fields to the vote state without needing to reallocations for existing vote accounts like we had to for TVC.

Got it, happy to approve. I do think it would be a lot faster if we could get away with one migration to remove `prior_voters` and add the new fields. Are you facing any hurdles with that approach?

lidatong (approved):
approving for FD

jstarry:
> Got it, happy to approve. I do think it would be a lot faster if we could get away with one migration to remove prior_voters and add the new fields. Are you facing any hurdles with that approach?

Yeah maybe that's still the best approach. There's just a lot of changes I want to make:
1. Remove `prior_voters` to make room for other state without increasing vote state size
2. Additionally store the authorized voter for the previous epoch now that prior voters is going away
3. Store new fee collector addresses, commissions, and delegator rewards pool bump seed
4. New instructions for setting collector addresses and commission
5. When setting block revenue commission, create the delegator rewards pool if it's not already created

Would we use a single feature gate for all of these changes? I think that could be fine. We will have a separate feature gate for SIMD-0123 anyways to actually make use of these new fields.

dimandotsol:
Hi @jstarry 
Could you please explain more about pending delegator rewards: how is it related to Sysvar, and when is this value (and Sysvar) planned to be updated?

AshwinSekar (approved):
> Would we use a single feature gate for all of these changes? I think that could be fine. We will have a separate feature gate for SIMD-0123 anyways to actually make use of these new fields.

This sounds like a good approach. lgtm

michaelh-laine:
Can this be incorporated into this SIMD?

https://github.com/solana-foundation/solana-improvement-documents/discussions/254

jstarry:
> Can this be incorporated into this SIMD?
> 
> #254

I already have a separate SIMD written up for this actually: https://github.com/solana-foundation/solana-improvement-documents/pull/249

jstarry:
Just pushed another update to this SIMD to add more detail to the required changes section and some other minor cleanup and renaming. It'd be great to get another review pass from anza devs @AshwinSekar @buffalojoec @t-nelson and fd devs @lidatong @topointon-jump

proposals/0185-vote-account-v4.md
```diff
@@ -0,0 +1,245 @@
...
+Additionally, since the `prior_voters` field is removed from vote state v4,
+there's no need to read or modify this field when processing authorize
+instructions.
+
+#### `Withdraw`
+
+The existing withdraw instruction MUST be modified to completely zero vote
+account data for fully withdrawn vote accounts. The old behavior partially
+zeroed the account data following the vote state version discriminant and is
+less intuitive.
```
jstarry, line 209: I felt this was a nice to have change. This way we know that any vote account with the version discriminant bytes `03 00 00 00` is still initialized. Other vote state versions require checking fields for initialized values to be sure of initialization which is a footgun IMO.

proposals/0185-vote-account-v4.md
```diff
@@ -0,0 +1,245 @@
...
+    // ..
+}
+```
+
+If a modified vote account's size is smaller than `3762` bytes, first resize the
+account to `3762` bytes before updating the account data. The vote program does
+not need to check if the resulting account is rent exempt, the runtime will
+enforce that check. This differs from the prior vote program implementation which
+falls back to store vote state as v2 if the account size cannot be resized while
+keeping the account rent exempt.
```
jstarry, line 165: I felt this was nice to have because it simplifies the vote program implementation and ensures that all v4 vote accounts have the same size. Here's the current code in question that this will clean up:

https://github.com/anza-xyz/agave/blame/0df4a2d963e06ced795247719ef1aa16c6d71d15/programs/vote/src/vote_state/mod.rs#L40-L54

proposals/0185-vote-account-v4.md
```diff
@@ -0,0 +1,244 @@
...
+    /// commission: u8,
+
+    /// NEW: the collector accounts for validator income
+    pub inflation_rewards_collector: Pubkey,
+    pub block_revenue_collector: Pubkey,
+
+    /// NEW: basis points (0-10,000) that represent how much of each income
+    /// source should be given to this VoteAccount
+    pub inflation_rewards_commission_bps: u16,
+    pub block_revenue_commission_bps: u16,
```
buffalojoec, line 120: I'm not sure if what I'm about to propose is a "good" pattern, because we might want to explicitly use a new `VoteState` variant (discriminator) each time we update vote state layouts. However, we could make this section also an enum, and then if we ever need to adjust commission configs again we could do so on VoteState v4 directly.

```rust
enum CommissionConfig {
    V1 {
        block_revenue_collector: Pubkey,
        block_revenue_commission_bps: u16,
        inflation_rewards_collector: Pubkey,
        inflation_rewards_commission_bps: u16,
    }
    V2 {
        // More configs in the future
    }
}
```

proposals/0185-vote-account-v4.md
```diff
@@ -0,0 +1,244 @@
...
+    pending_delegator_rewards: 0u64,
+
+    // ..
+}
+```
+
+If a modified vote account's size is smaller than `3762` bytes, first resize the
+account to `3762` bytes before updating the account data. The vote program does
+not need to check if the resulting account is rent exempt, the runtime will
+enforce that check. This differs from the prior vote program implementation which
```
buffalojoec, line 163: Only v2 state should need a rent top-up, right? The other two are large enough already?

proposals/0185-vote-account-v4.md
```diff
@@ -0,0 +1,244 @@
...
+falls back to store vote state as v2 if the account size cannot be resized while
+keeping the account rent exempt.
```
buffalojoec, line 165: I'm ok with this, but do we care about getting stuck with v2 state in active vote accounts? It might be an annoyance, but seems like it's harmless.

proposals/0185-vote-account-v4.md
buffalojoec: Seems reasonable to me, unless there's some reason it's done like this currently? Seems strange to keep that discriminator set afterward.

proposals/0185-vote-account-v4.md
jstarry: Yeah I'm not sure either. For now I prefer making serialization changes explicit with a new state version to keep one level of versioning. This intuitively feels less error prone and avoids situations where a downstream client starts failing to deserialize a vote state version they were previously able to deserialize.

proposals/0185-vote-account-v4.md
jstarry: No explicit reason that I'm aware of. We will need an extra look from auditors at this though.

proposals/0185-vote-account-v4.md
jstarry: No, don't care. There aren't any active vote accounts with v2 state on any cluster that I'm aware of. Will double check this.

proposals/0185-vote-account-v4.md
buffalojoec: OK, that works for me as well. If we end up eating our words we can try again on Vote State v5!

proposals/0185-vote-account-v4.md
buffalojoec: OK, so you are proposing here that we throw an error if a vote account has not been resized and is attempted to be modified? And this is NBD because that error would only throw on a v2 or older vote account, which we're fairly certain there are none of in active operation.

I think just explicitly mention that the program will error here, and this proposal is g2g from me. The bit about "runtime will enforce this" is a little unclear at first.

buffalojoec (commented):
Thanks for updating. I left one more question, then I can stamp.

proposals/0185-vote-account-v4.md
jstarry: > OK, so you are proposing here that we throw an error if a vote account has not been resized and is attempted to be modified?

Nah I was proposing that we *do* resize the account but then *not* do anything (including throw an error) if the resulting account is not rent exempt because the runtime checks for rent exemption anyways. But taking your comment as a suggestion, I think it's probably best to just keep the rent exemption check in the vote program and return an error if it's not satisfied, will update!

jstarry: https://github.com/solana-foundation/solana-improvement-documents/pull/185/commits/09d8017394b0d65d1d5e6981bc073d757fb83bac

buffalojoec (approved):

jstarry:
> Approved as long as we are certain there are little to no Vote State v2 or older vote accounts in active use. Throwing an error on those accounts not being rent-exempt could affect those nodes' timely vote credits.

New vote accounts must be created as v3 and note that v2 vote state accounts wouldn't have timely vote credits because v3 was introduced to add that feature. Because of that, we don't need to be certain about whether v2 accounts are in active use for this proposal, they would already be severely disadvantaged in consensus when tvc was enabled and have plenty of incentive to upgrade if they hadn't already. But in any case, I did check mainnet and there are no active vote accounts using v2 still.

lidatong (approved):

buffalojoec (approved):

proposals/0185-vote-account-v4.md
```diff
@@ -122,6 +122,9 @@ pub struct VoteStateV4 {
     /// NEW: reward amount pending distribution to stake delegators
     pub pending_delegator_rewards: u64,
 
+    /// NEW: compressed bls pubkey for alpenglow
+    pub bls_pubkey_compressed: Option<[u8; 48]>
```
wen-coding, line 126: Looks good, compressed should be fine. But I think @AshwinSekar proposed we make this field mandatory...

AshwinSekar: As I said in the other thread option is fine. Since V4 migration happens automatically on first vote instruction so we can't enforce that this field is populated by operators, we'll have to detect it in Agave when alpenglow is active and instruct operators to fill it in.

AshwinSekar (commented):
Looks good. If possible can we also squeeze in the additional instruction here?
```
    /// Update the BLS Pubkey for the vote account
    ///
    /// # Account references
    ///   0. `[WRITE]` Vote account to be updated
    ///   1. `[SIGNER]` Withdraw authority
    UpdateBLSPubkey([u8; 48]),
```
Can also defer this to the Alpenglow SIMD if you prefer.

jstarry:
> Looks good

Thanks for looking!

> If possible can we also squeeze in the additional instruction here?
> [..]
> Can also defer this to the Alpenglow SIMD if you prefer.

I prefer to defer this instruction to another SIMD. I've deferred the new instructions for updating commission and collector accounts to other SIMD's as well.

proposals/0185-vote-account-v4.md
```diff
@@ -11,7 +11,7 @@ feature: (fill in with feature tracking issues once accepted)
 
 ## Summary
 
-Add a new version of vote account state which clears unused state to make
+Add a new version of vote account state which removes unused state to make
```
buffalojoec: Tbh, I thought "clears" made more sense. "removes" makes me think the account is being shrunk. Why not just make this summary statement more about the intended goals of the new fields? Something like:

"Add a new version of vote account state to support block revenue distribution."

proposals/0185-vote-account-v4.md
```diff
@@ -217,10 +217,20 @@ The builtin stake program reads vote account state when creating, delegating,
...
 accounts.
 
-### Other
+### Runtime
+
+Commission rates will now be stored in basis points, but higher precision values
+cannot be set until [SIMD-0249] is adopted. From the runtime's perspective,
+commission rates will remain limited to multiples of 100 basis points,
+equivalent to integer percentages. Therefore, commission calculations
+should continue to use integer percentage values for now.
```
buffalojoec: It's not immediately clear to me why this has anything to do with SIMD-0249.

Earlier in the proposal, under `UpdateCommission`, it's stated that the program will still only support integer percentage points. You could just reference that here instead.

proposals/0185-vote-account-v4.md
```diff
@@ -153,8 +153,8 @@ above:
 VoteStateV4 {
     // ..
 
-    inflation_rewards_collector: Pubkey::default(),
-    block_revenue_collector: Pubkey::default(),
+    inflation_rewards_collector: vote_pubkey,
```
buffalojoec, line 156: This makes sense to me, since when a v3 account is converted to v4, their inflation rewards collector should remain the same. This also guarantees they're protected through subsequent block revenue distribution feature activations if conversion happens during some Vote program instruction.

buffalojoec (commented):
This is ready to go from my side, but I left a few small comments on the recent additions. None affect the spec.

proposals/0185-vote-account-v4.md
jstarry: Went with this, thoughts?

> Add a new version of vote account state to support alpenglow, block revenue
distribution, and commission improvements

proposals/0185-vote-account-v4.md
jstarry: Oh yeah that was supposed to link to SIMD-0291, my mistake. I referenced `UpdateCommission` as you suggested but still kept a note about 291 adding support for higher precision.

proposals/0185-vote-account-v4.md
buffalojoec: Yeah that works for me.

buffalojoec (approved):
Lgtm, ship it.

PR #177, opened 2024-10-01
SIMD-0177: Program Runtime ABI v2

Lichtso:

PR #186, opened 2024-10-21
SIMD-0186: Loaded Transaction Data Size Specification

2501babe:
precisely define how transaction data sizes are calculated


proposals/0186-transaction-data-size-specification.md
```diff
@@ -0,0 +1,143 @@
...
+feature: (fill in with feature tracking issues once accepted)
+---
+
+## Summary
+
+Before a transaction can be executed, every account it may read from or write to
+must be loaded, including any programs it may call. The amount of data a
+transaction is allowed to load is capped, and if it exceeds that limit, loading
+is aborted. This functionality is already implemented in the validator. The
+purpose of this SIMD is to explicitly define how transaction size is calculated.
```
apfitzge, line 19: Think should be more precise here:
"transaction size" => "transaction account data size" or "transaction loaded data size"

proposals/0186-transaction-data-size-specification.md
```diff
@@ -0,0 +1,143 @@
...
+be amended to define the required semantics before LoaderV4 is enabled on any
+network.
+
+## Alternatives Considered
+
+* Transaction data size accounting is already enabled, so the null option is to
+enshrine the current Agave behavior in the protocol. This is undesirable because
+the current behavior is highly idiosyncratic, and LoaderV3 program sizes are
+routinely undercounted.
+* Builtin programs are backed by accounts that only contain the program name as
```
apfitzge, line 119: it's also in the works to make (some/most) builtins just normal programs, so adding any special cases here would probably be nullified in the future anyway

proposals/0186-transaction-data-size-specification.md
```diff
@@ -0,0 +1,143 @@
...
+## Security Considerations
+
+Security impact is minimal because this SIMD merely simplifies an existing
+feature.
+
+This SIMD requires a feature gate.
+
+## Backwards Compatibility
+
+Transactions that call LoaderV3 programs via CPI and are extremely close to the
```
apfitzge, line 142: do you happen to know if there are any such programs on mnb?

2501babe: ill edit this to be clearer, but its a property of the transaction, not the program

for instance if i have a transaction that normally uses 63 1mb accounts and a 64th account which is a 1mb program used for cpi, right now (ignoring loader sizes as not relevant) the total transaction size would be 63mb + 36 bytes because we incorrectly dont charge for the program data. with this change the calculated transaction size would become 63mb + 1mb + 36 bytes and the transaction would fail

it seems extremely unreasonable to me that any legitimate transaction would actually use >60mb of data, but i dont know how we would be able to find that out

2501babe: shoutout to tao for help with metrics, looks like in the past 30 days we saw one tx 110mb (?!) one 65mb one 57mb, three in the 40s, everything else 20s and below. change seems safe to me, this shouldnt break any real workflows. for point of comparison jupiter v6 is 3mb so with a 30mb transaction you would need to cpi into a dozen different jupiter-sized programs to even be near the limit

2501babe: shoutout to tao for help with metrics, looks like in the past 30 days grouped by hour, we saw one max size 110mb (?!) one 65mb one 57mb, three in the 40s, everything else 20s and below. change seems safe to me, this shouldnt break any real workflows. for point of comparison jupiter v6 is 3mb so with a 30mb transaction you would need to cpi into a dozen different jupiter-sized programs to even be near the limit

2501babe:
@brooksprumo adding you as a second reviewer since youve done research on transaction sizes

2501babe:
@topointon-jump also i would add you but it doesnt seem like github lets me add people not in the suggested list (unless this ping fixes that)

2501babe:
i added another existing issue that exists in the current implementation to the problem description. the proposal itself is unchanged

proposals/0186-transaction-data-size-specification.md
```diff
@@ -0,0 +1,152 @@
...
+---
+
+## Summary
+
+Before a transaction can be executed, every account it may read from or write to
+must be loaded, including any programs it may call. The amount of data a
+transaction is allowed to load is capped, and if it exceeds that limit, loading
+is aborted. This functionality is already implemented in the validator. The
+purpose of this SIMD is to explicitly define how loaded transaction data size is
+calculated.
```
apfitzge, line 20: Last 2 sentences here make it sound like the SIMD just documents the existing behavior.
Let's change the last sentence so its' clear there is a proposal for a new algorithm for calculating loaded account size.

proposals/0186-transaction-data-size-specification.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Detailed Design
+
+The proposed algorithm is as follows:
+
+1. Every account explicitly included on the transaction accounts list is counted
+once and only once.
+2. A valid program owned by LoaderV3 also includes the size of its programdata.
+3. Other than point 2, no accounts are implicitly added to the total data size.
+
```
apfitzge, line 78: edge-case for consideration.

Does it matter if the account is used as a top-level or instruction account? Realistically we should not see these if people construct txs reasonably, but if an account is:
1. not the fee-payer
2. not passed to any instructions
3. not used a program

we could reasonably choose to not load it at all?

With the proposed algorithm we would still need to load it in order to find the size.

2501babe: as-written it wouldnt matter, all accounts are treated the same

but thats an interesting idea, i wasnt thinking about account loading with this simd. i was only focused on how to arrive at a correct total given a set of accounts. but youre right, i cant think of a reason why we should load these kinds of accounts at all, other than the fact that we already do it now

if we do that (im in favor), there are several formulas that seem plausible to me

Option 1a: sum the sizes of the unique set of accounts `instruction accounts + program ids + fee payer`. any account in that set which is a LoaderV3 program account also includes its programdata

Option 1b: the same as above but programdata is only added to program size if the program is valid for execution in the current slot

Option 2: sum the sizes of the unique set of accounts `instruction accounts + fee payer`. add it to the sum of the sizes of the unique set of program ids, which include programdata sizes for LoaderV3 program ids

i think Option 1 is preferable. Option 2 fails to count cpi programs, but it sidesteps any issues with tombstoned programs because (after fee-only transactions... i say that a lot lately) the sizes dont matter, transactions that use them as program ids would fail anyway

Option 1a and 1b behave the same for closed programs since there _is_ no programdata. they differ on handling programs which fail validation and programs that are changed in-slot. this is especially germaine to simd83 since transactions will be able to mess with programs in the same batch

all the options can double-count LoaderV3-owned accounts but i think this is appropriate, the compiled program and the account data if accessed in an instruction are basically their own separate things 

im undecided between 1a and 1b. do you have an opinion?

apfitzge: Difference of a and b is just with valid programs v3 program's programdata?

A bit of a detail, but can we determine if it's valid or not without loading the programdata?

apfitzge: Yeah 1a seems better option to me in that case

proposals/0186-transaction-data-size-specification.md
```diff
@@ -0,0 +1,152 @@
+---
+simd: '0186'
+title: Transaction Data Size Specification
```
jstarry, line 3: This title reads as if the proposal specifies the size of a serialized transaction rather than size of the account data loaded for processing a transaction. How about `Transaction Account Data Size Specification`?

proposals/0186-transaction-data-size-specification.md
jstarry: Agreed on 1a. And I think I agree that it's best to double count programdata accounts if they are included in the ix account list since the SVM would effectively keep 2 copies in memory anyways.

proposals/0186-transaction-data-size-specification.md
2501babe: went with `Loaded Transaction Data Size Specification`

2501babe: went with `Loaded Transaction Data Size Specification`

proposals/0186-loaded-transaction-data-size-specification.md
```diff
@@ -0,0 +1,163 @@
...
+## Detailed Design
+
+The proposed algorithm is as follows:
+
+1. Given a transaction, take the unique set of account keys which are used as:
+    * An instruction account.
+    * A program ID for an instruction.
+    * The fee-payer.
+2. Each account's size is determined solely by the byte length of its data prior
+to transaction execution.
```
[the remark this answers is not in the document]
tao-stones: iirc, original motivation is primarily focus on memory footprint and how that chunk of memory is used, loading from accountdb and serializing wasn't discussed much then, it might make sense now

proposals/0186-loaded-transaction-data-size-specification.md
```diff
@@ -0,0 +1,163 @@
...
+3. For any `LoaderV3` program account, add the size of the programdata account
+it references, if it exists.
+4. The total transaction size is the sum of these sizes.
+
+Transactions may include a
+`ComputeBudgetInstruction::SetLoadedAccountsDataSizeLimit` instruction to define
+a data size limit for the transaction. Otherwise, the default limit is 64MiB
```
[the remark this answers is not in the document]
tao-stones: Just fyi, https://github.com/anza-xyz/agave/pull/1355 attempted to introduce DEFAULT_LOADED_ACCOUNTS_DATA_SIZE_BYTES. Leader currently used "actual" loaded accounts size after execution in block packing, so a lower DEFAULT is critical atm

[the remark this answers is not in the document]
2501babe: im definitely on board with reducing the default, but it should be done separately from this simd. after we implement the new algorithm, we can run it against ledger history to collect new data about what transaction sizes would look like with it enabled. calculated sizes will increase nontrivially and we dont want to pick a number that will cause significant breakage

for example, in #1355 2MiB is suggested as a good number that only causes 5% breakage. however, the jupiter v4 program is 2.5MiB and jupiter v6 is close to 3MiB. this means they must be passing their program as an instruction account, which due to the current algorithm, means the program is only counted as 36 bytes. it also means under the new algorithm which properly counts programs, a 2MiB default limit would break 100% of jupiter transactions

once we fix the size calculations, we will be in a better position to judge what the real sizes of transactions actually are

proposals/0186-loaded-transaction-data-size-specification.md
```diff
@@ -0,0 +1,165 @@
...
+## Detailed Design
+
+The proposed algorithm is as follows:
+
+1. Given a transaction, take the unique set of account keys which are used as:
+    * An instruction account.
+    * A program ID for an instruction.
+    * The fee-payer.
+2. Each account's size is determined solely by the byte length of its data prior
+to transaction execution, irrespective of it is used on the transaction.
```
jstarry, line 83:
```suggestion
2. Add each account's size, which is determined solely by the byte length of its data prior
to transaction execution, irrespective of how it is used on the transaction.
```

proposals/0186-loaded-transaction-data-size-specification.md
```diff
@@ -0,0 +1,165 @@
...
+3. For any loaded account identified as a `LoaderV3` program account, add the
+size of the programdata account it references to its own size, irrespective of
+how the program account is used on the transaction.
```
jstarry, line 86:
```suggestion
3. For any loaded account identified as a `LoaderV3` program account, add the
size of the programdata account it references, irrespective of
how the program account is used in the transaction.
```

proposals/0186-loaded-transaction-data-size-specification.md
```diff
@@ -0,0 +1,165 @@
...
+4. The total transaction size is the sum of these sizes.
```
jstarry, line 87:
```suggestion
4. The total transaction loaded account data size is the sum of these sizes.
```

proposals/0186-loaded-transaction-data-size-specification.md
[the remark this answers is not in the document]
jstarry: How about 96 bytes? From https://github.com/anza-xyz/agave/blob/4e7f7f76f453e126b171c800bbaca2cb28637535/programs/bpf_loader/src/serialization.rs#L417, I see we have to serialize 3 bytes for `is_writable`, `is_signer`, and `is_executable`, 4 bytes for original data length, 8 bytes each for `data length`, `lamports`, and `rent_epoch`, 32 bytes for the `account_key`, and 32 bytes for the `owner` and we adjust for alignment to the nearest multiple of 8. cc @Lichtso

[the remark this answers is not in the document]
Lichtso: Yes that is the size of the serialized account metadata. But have you thought about the resize / realloc padding (`MAX_PERMITTED_DATA_INCREASE`)?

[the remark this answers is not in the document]
jstarry: Will that be used in loaderv4 too?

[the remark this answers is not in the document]
Lichtso: Yes, serialization will stay ABIv1 in loader-v4.
ABIv2 is its own [SIMD](https://github.com/solana-foundation/solana-improvement-documents/pull/177).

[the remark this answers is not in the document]
2501babe: realloc headroom is quite different conceptually from "loaded data" tho isnt it? space is reserved, but no data is loaded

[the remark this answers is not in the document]
jstarry: Sounds like we need a (loose?) definition for what loaded data is intended to mean. It could be the amount of bytes read from accounts db (in that case maybe programdata is only counted once?) or it could be roughly the amount of bytes loaded into memory before tx execution?

In either case I think it makes sense to include the account metadata overhead of approx. 64 bytes and exclude the overhead of the account key itself and the realloc buffer since they're not really loaded in any sense.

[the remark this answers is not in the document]
2501babe: im equally fine counting programdata once or twice, and after this pr lets us refactor we should only have to actually load it once. main concerns are that the algorithm is very simple implement correctly, and always counts programdata if you use the program

maybe a new definition (counting programdata once) like:
1. Given a transaction, take the unique set of account keys which are used as:
    * An instruction account.
    * A program ID for an instruction.
    * The fee-payer.
2. For all LoaderV3 program accounts, add the key of the programdata account it references to the set of account keys, if it does not already exist.
3. Each account's size is defined as the byte length of its data prior to transaction execution, plus 64 bytes to account for metadata. This is irrespective of how the account is used on the transaction.
4. The total transaction loaded account data size is the sum of these sizes.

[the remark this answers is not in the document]
jstarry: I'm also fine either way, slight bias to counting it once now if we only need to load it once from accounts-db. Do we agree on 64 bytes overhead?

proposals/0186-loaded-transaction-data-size-specification.md
```diff
@@ -0,0 +1,165 @@
...
+programdata account.
+
+## Detailed Design
+
+The proposed algorithm is as follows:
+
+1. Given a transaction, take the unique set of account keys which are used as:
+    * An instruction account.
+    * A program ID for an instruction.
+    * The fee-payer.
```
jstarry, line 81: We may have to include every account in the transaction actually if this SIMD is accepted: https://github.com/solana-foundation/solana-improvement-documents/pull/163 (cc @Lichtso)

Lichtso: Yes, it is generally bad to try to have extra conditions for TX account depending on how they are used. That is how we got to the write lock demotion and only top-level-instructions counting loader-v3 programdata account mess we have.

2501babe: cool, ill revert that change

also i love this simd, i was thinking the other week it would be so nice if you could specify accounts you might want to execute but dont need to see instruction data for

2501babe:
ive updated the language for the algorithm again and i think i finally got the declarative rather than procedural description ive been aiming for

proposals/0186-loaded-transaction-data-size-specification.md
apfitzge: Nice catch!

proposals/0186-loaded-transaction-data-size-specification.md
```diff
@@ -0,0 +1,164 @@
...
+## Detailed Design
+
+The proposed algorithm is as follows:
+
+1. The set of accounts that determine loaded transaction data size is defined as
+the unique intersection of:
+    * The set of account keys explicitly specified on the transaction,
+irrespective of how they are used.
+    * The set of programdata accounts referenced by the LoaderV3 program
+accounts specified on the transaction.
```
jstarry, line 83: Oh just realized we should probably also include ALT account sizes as well

proposals/0186-loaded-transaction-data-size-specification.md
```diff
@@ -0,0 +1,164 @@
...
+them.
+
+Read-only and writable accounts are treated the same. In the future, when direct
+mapping is enabled, this SIMD may be amended to count them differently.
+
+As a consequence of 1 and 3, for LoaderV3 programs, programdata is counted twice
+if a transaction explicitly references the program account and its programdata
+account. This is done partly for simplicity, and partly to account for the cost
+of maintaining the compiled program in addition to the actual bytes of
+the programdata account.
```
jstarry, line 111: I think we settled on just counting them once right?

proposals/0186-loaded-transaction-data-size-specification.md
```diff
@@ -0,0 +1,164 @@
...
+for these. However, they must be made available for all programs anyway, and
+most of them are likely to be ported to BPF eventually, so this adds complexity
+for no real benefit.
+* Several slightly different algorithms were considered for handling LoaderV3
+programs in particular, for instance only counting programs that are valid for
+execution in the current slot. However, this would implicitly couple transaction
+data size with the results of ELF validation, which is highly undesirable.
+* We considered loading and counting sizes for accounts on the transaction
+account list which are not used for any purpose. This is the current behavior,
+but there is no reason to load such accounts at all.
```
jstarry, line 142: This section is no longer relevant

proposals/0186-loaded-transaction-data-size-specification.md
jstarry: Hmm but ALTs are not read or write locked by the transaction so they could be extended in parallel so we can't use the actual size. Maybe just use the max size of `LOOKUP_TABLE_META_SIZE + LOOKUP_TABLE_MAX_ADDRESSES * 32`?

2501babe: is it possible to know whether an ALT was used at all in transaction processing? my understanding is resolution happens too early and all the information is erased. im not sure if there is an easy solution to this or if we should defer adding it to transaction data size after the ALT async execution simd changes how that code works

proposals/0186-loaded-transaction-data-size-specification.md
2501babe: yes i forgot to delete this

proposals/0186-loaded-transaction-data-size-specification.md
jstarry: It's possible, we could simply include that information in static meta and make it available to the account loader. The ALT SIMD changes will most likely only really cover failure cases so I don't think we need to defer this decision.

apfitzge: If we're just using the max size of a table, we don't need to store any additional information in static meta. We have the number of ALTs in the message itself.

and I (accidently) have already exposed getting that information in `SVMMessage` trait, so we should be able to get it in our specific implementation easily.

2501babe: (based on discussion we are considering a flat 8192 byte cost for ALT usage, pending feedback from @tao-stones)

jstarry: > (based on discussion we are considering a flat 8192 byte cost for ALT usage ...

Should be 8248 byte cost because there is space in the ALT accounts for 56 bytes of metadata as well

tao-stones: `8248 * 8 / 32000 = 2 CU`, that doesn't sound bad at all.

jstarry: Cool, let's go forward with that in the proposal for now then

proposals/0186-loaded-transaction-data-size-specification.md
```diff
@@ -0,0 +1,155 @@
...
+a table plus 56 bytes for metadata.
+4. The total transaction loaded account data size is the sum of these sizes.
+
+Transactions may include a
+`ComputeBudgetInstruction::SetLoadedAccountsDataSizeLimit` instruction to define
+a lower data size limit for the transaction. Otherwise, the default limit is
+64MiB (`64 * 1024 * 1024` bytes).
+
+If a transaction exceeds its data size limit, the transaction is failed. Fees
+will be charged once `enable_transaction_loading_failure_fees` is enabled.
```
topointon-jump, line 93: Where will this be checked? Do transactions that fail this check make it into blocks? Do you mind clarifying this in the SIMD? :pray:

topointon-jump (approved):
Looks good to me! Great to simplify things here.

proposals/0186-loaded-transaction-data-size-specification.md
2501babe: ok i updated the wording to:
>If a transaction exceeds its data size limit, a loading failure occurs. This SIMD does not change any aspect of how such a failure is handled. At time of writing, such a transaction would be excluded from the ledger. When enable_transaction_loading_failure_fees is enabled, it will be written to the ledger and charged fees as a processed, failed transaction.

the idea is we arent changing _any_ of the logic about how loading works here, the existing flow stays the same (loading happens at the same time in the same place of transaction processing, what kind of error you get if you exceed the limit, how that error is handled and how its reflected in ledger history, etc)

the only thing that changes is what the number of bytes you arrive at will be and how you determine what that number is

2501babe:
@jacobcreech i believe we have consensus

2501babe:
the last commit cleared the approvals (was just a minor clarification) so if youd be so kind as to approve again i will merge this next week. planning on implementing it in 2.2

apfitzge (approved):

PR #189, opened 2024-10-23
SIMD-0187: SBPF stricter ELF headers

Lichtso:


proposals/0187-sbpf-stricter-elf-headers.md
```diff
@@ -0,0 +1,131 @@
+---
+simd: '0187'
```
nickfrosty, line 2: Can you update the SIMD number (and PR title) to match the PR number, per https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0001-simd-process.md#draft

proposals/0189-sbpf-stricter-elf-headers.md
```diff
@@ -0,0 +1,131 @@
...
+- `p_vaddr` must match the `p_vaddr` of the entry in the table above
+- `p_paddr` must match the `p_vaddr` of the entry in the table above
+- `p_filesz` must be:
+  - `0` if the section is writable (the `PF_W` flag is set)
+  - `p_memsz` otherwise (the `PF_W` flag is clear)
+- `p_filesz` must not be greater than `file.len() as u64 - p_offset`
+- `p_memsz` must fit in 32 bits / be less than `1 << 32`
+- `p_align` is ignored
+
+If any check fails `ElfParserError::InvalidProgramHeader` must be thrown.
```
topointon-jump, line 92: Do you mind specifying where in the program lifecycle this will be thrown? (i.e. will these new checks be enforced at deployment time? when loading into the program cache for execution? both?)

proposals/0189-sbpf-stricter-elf-headers.md
```diff
@@ -0,0 +1,131 @@
...
+
+The `e_entry` filed of the file header must be a `STT_FUNC` entry in the
+dynamic symbol table, otherwise `ElfParserError::InvalidFileHeader` must be
+thrown.
+
+## Impact
+
+The toolchain linker will use a new linker script to adhere to these
+restrictions defined here and thus the change will be transparent to the dApp
+developers.
```
topointon-jump, line 124: What is the plan for currently deployed programs?

proposals/0189-sbpf-stricter-elf-headers.md
topointon-jump: Or is this change pinned to an SBPF version?

topointon-jump: Could we specify in the SIMD where the checks will be performed, and what happens if they fail?

proposals/0189-sbpf-stricter-elf-headers.md
```diff
@@ -0,0 +1,131 @@
...
+If any check fails `ElfParserError::InvalidFileHeader` must be thrown.
+
+### Program headers
+
+|  purpose  |    p_type    |   p_flags  | p_vaddr |
+| --------- | ------------ | ---------- | ------- |
+| bytecode  | PT_LOAD      | PF_X       | 0 << 32 |
+| ro data   | PT_LOAD      | PF_R       | 1 << 32 |
+| stack     | PT_GNU_STACK | PF_R, PF_W | 2 << 32 |
+| heap      | PT_LOAD      | PF_R, PF_W | 3 << 32 |
```
topointon-jump, line 73: Why do we allow these headers? These regions are configured independently of the ELF. Do we allow these headers just so that we can reduce LLVM modifications? Just curious.

Lichtso: It is so that the memory regions are correctly reflected in the ELF header. This makes it easier for tooling to detect issues (such as overlap in section allocation) and could simplify the validator implementation in the future.

Also, with the upcoming CPI syscall restrictions (necessary for direct mapping) we will limit pointers provided by the user to come from the stack and heap. Thus their layout will be further solidified.

proposals/0189-sbpf-stricter-elf-headers.md
```diff
@@ -0,0 +1,131 @@
...
+- `p_filesz` must be:
+  - `0` if the section is writable (the `PF_W` flag is set)
+  - `p_memsz` otherwise (the `PF_W` flag is clear)
+- `p_filesz` must not be greater than `file.len() as u64 - p_offset`
+- `p_memsz` must fit in 32 bits / be less than `1 << 32`
+- `p_align` is ignored
+
+If any check fails `ElfParserError::InvalidProgramHeader` must be thrown.
+
+### Dynamic symbol table
```
LucasSte, line 94: Why did you choose the dynamic symbol table entry `SHT_DYNAMIC` and not the symbol table elf entry `SHT_SYMTAB`?

Lichtso: https://github.com/solana-labs/rbpf/pull/624#discussion_r1833048810

proposals/0189-sbpf-stricter-elf-headers.md
deanmlittle: Not all programs make use of stack/heap/symbols. A noop program, for example, should be just 128 bytes (ELF header + 1 PT_LOAD PF_X + 1 exit insn). Enforcing 5 headers, 4 of which aren't even used, would bloat it to 352 bytes - even larger than the current mainnet-beta minimum of 336 bytes. 

On the other hand, it's great to be able to handle for memory layouts that diverge from the default of a specific version without having to bump versions should the need arise. As such,. I think it would be better not to strictly enforce these headers, but to instead opt for a solution like:

1. First header is always PT_LOAD PF_X. This is the only mandatory header.
2. For anything with .rodata, add a PT_LOAD PF_R header.
3. For anything with symbols, add PT_NONE 0x00.
4. For anything with non-standard stack layout for its version, add PT_GNU_STACK PF_R | PF_W
5. For anything with non-standard heap layout for its version, add PT_LOAD PF_R | PF_W

This solution would make the minimum program use 1 program head, the average program use 3, and if we ever had memory layout issues, they could be solvable for the small price of 1 or 2 extra program headers.

LucasSte (dismissed):

proposals/0189-sbpf-stricter-elf-headers.md
```diff
@@ -0,0 +1,108 @@
...
+- `e_type` must be `ET_DYN` (`0x0003`)
+- `e_machine` must be `EM_SBPF` (`0x0263`)
+- `e_version` must be `EV_CURRENT` (`0x00000001`)
+- `e_entry` must point to a function start marker in the text section,
+see SIMD-0179 for details on how a function start is marked.
+- `e_phoff` must be `size_of::<Elf64Ehdr>()` (64 bytes)
+- `e_shoff` is not checked
+- `e_flags` see SIMD-0161
+- `e_ehsize` must be `size_of::<Elf64Ehdr>()` (64 bytes)
+- `e_phnum` must not be less than `0x0005`
```
ripatel-fd: Are the indexes of the predefined program headers specified below arbitrary or do they have to be the first 4 entries?

Why not put an upper bound on e_phnum?

Lichtso: They must be the first four (index `0`, `1`, `2`, `3`). We ignore all higher indices, thus `e_phnum` is not limited.

ripatel-fd: @Lichtso Nice. But not specified by the SIMD that the predefined program headers have specific indexes. Would you mind amending?

proposals/0189-sbpf-stricter-elf-headers.md
```diff
@@ -0,0 +1,109 @@
...
```
ravyu-jump: Shouldn't it be "not less than `0x0004`"?

Lichtso: good catch, yes, I will correct it

joncinque:
It looks like there's approval for this. I will merge the SIMD next Friday to give time for any final objections.

Lichtso:
Given the ongoing discussions in #178 we should not merge this one for now.

proposals/0189-sbpf-stricter-elf-headers.md
```diff
@@ -0,0 +1,109 @@
...
+If any check fails `ElfParserError::InvalidFileHeader` must be thrown.
+
+### Program headers
+
+| index |  purpose  |   p_flags  | p_vaddr |
+| ----- | --------- | ---------- | ------- |
+| 0     | bytecode  | PF_X       | 0 << 32 |
+| 1     | ro data   | PF_R       | 1 << 32 |
+| 2     | stack     | PF_R, PF_W | 2 << 32 |
+| 3     | heap      | PF_R, PF_W | 3 << 32 |
```
ripatel-fd: Can we remove the stack and heap segments? 
They're currently specified by the compute budget program instructions anyways. I don't want conflicting information for heap and stack size in the VM.

Lichtso: Yes, also we should swap the bytecode and read-only data in the virtual address space so that rodata can be accessed by 32 bit immediate loads.

ripatel-fd: Ah nice, we're bringing back the deprecated LD instruction class?

LucasSte: I just asked @cyanpencil about the tooling he was using (refer to his comment here https://github.com/solana-foundation/solana-improvement-documents/pull/358#issuecomment-3323426372). He said he was having trouble with the stack and heap addresses when using Ghidra. I think maintaining the stack and heap addresses in the ELF file would make binaries better for debugging purposes, wouldn't it?

Lichtso: yes it would, however only the start addresses would be reflected accurately as the length are runtime defined.

ripatel-fd:
@jacobcreech Looks like Anza and Firedancer are aligned on this proposal. I'm going to start implementing this loader in Firedancer.

proposals/0189-sbpf-stricter-elf-headers.md
```diff
@@ -49,16 +49,16 @@ otherwise `ElfParserError::OutOfBounds` must be thrown.
 - `e_ident.ei_osabi` must be `ELFOSABI_NONE` (`0x00`)
 - `e_ident.ei_abiversion` must be `0x00`
 - `e_ident.ei_pad` must be `[0x00; 7]`
-- `e_type` must be `ET_DYN` (`0x0003`)
-- `e_machine` must be `EM_SBPF` (`0x0263`)
+- `e_type` must be `ET_REL` (`0x0001`)
```
LucasSte: Why did we change to `ET_REL`?

LucasSte: We won't have any relocations to solve in the final objects, right? So, it isn't a relocatable file.

Lichtso: Technically the type "executable" would be the correct one to chose (as we do have a dedicated entrypoint). However, as you mentioned this would mean all dApp devs would have to restructure their programs from lib.rs to main.rs, add `#![no_main]`, change the crate type in Cargo.toml and copy .elf instead of .so files for deployment.

So yes, we could continue to stick with `ET_DYN` to minimize friction for dApp devs, even if that is also somewhat incorrect as our programs are not pure libraries. Or maybe we should just ignore this field, but that would complicate a `memcmp` based check of the file header.

LucasSte: I agree that executable is closer in definition to our usage.

Considering whether to resolve relocations or not, `ET_REL` and `ET_DYN` have different consequences in the linking stage. Since we want a binary without runtime relocations, `ET_DYN` is easier to work with off-the-shelf linkers.

LucasSte: I can't say that our final object is exactly PIC, since it is loaded to a fixed memory layout and does not have a GOT. You can view it as PIC in one sense, because all sections have fixed offsets from a value well known to runtime (our GOT-like feature here), such that relocations are no longer necessary.

On Linux, it is [a requirement](https://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html) for all shared libraries to be PIC.

ET_REL files are not required to have relocations, but typically they contain code that has not yet been linked to any other file.

Retrieving the explanation from the [ELF spec](https://refspecs.linuxfoundation.org/elf/elf.pdf):

> • A relocatable file holds code and data suitable for linking with other object files to create an executable or a shared object file.
> • An executable file holds a program suitable for execution.
> • A shared object file holds code and data suitable for linking in two contexts. First, the link editor may process it with other relocatable and shared object files to create another object file. Second, the dynamic linker combines it with an executable file and other shared objects to create a process image.

Our final objects resembles more either an executable or a shared library. Given the possible changes necessary on Rust programs to deal with executables, I would stay with ET_DYN.

proposals/0189-sbpf-stricter-elf-headers.md
```diff
@@ -0,0 +1,105 @@
...
+- `e_machine` must be `EM_BPF` (`0x00F7`)
+- `e_version` must be `EV_CURRENT` (`0x00000001`)
+- `e_entry` must point to a function start marker in the text section,
+see SIMD-0179 for details on how a function start is marked.
+- `e_phoff` must be `size_of::<Elf64Ehdr>()` (64 bytes)
+- `e_shoff` is not checked
+- `e_flags` see SIMD-0161
+- `e_ehsize` must be `size_of::<Elf64Ehdr>()` (64 bytes)
+- `e_phnum` must be `0x0002`
+- `e_phoff + e_phnum * size_of::<Elf64Phdr>()` must be less than the file size
```
intrigus-lgtm:
```suggestion
- `e_phoff + e_phnum * size_of::<Elf64Phdr>()` must be less than equal the file size
```

Doesn't our range look like this?
[`e_phoff`, `e_phoff + e_phnum * size_of::<Elf64Phdr>()`)
Meaning that `e_phoff + e_phnum * size_of::<Elf64Phdr>()` is exclusive and therefore should be okay to also be equal to the file size?

Verification of such an ELF will still fail later with `ElfParserError::InvalidProgramHeader`, but IMHO a header where `e_phoff + e_phnum * size_of::<Elf64Phdr>()` == `file.size()` holds should not fail with `ElfParserError::InvalidFileHeader` but `ElfParserError::InvalidProgramHeader`.

proposals/0189-sbpf-stricter-elf-headers.md
ripatel-fd: > However, as you mentioned this would mean all dApp devs would have to restructure their programs from lib.rs to main.rs, add #![no_main], change the crate type in Cargo.toml and copy .elf instead of .so files for deployment.

If we use @deanmlittle's sbpf linker, can't we just do this in some post-process step?

LucasSte: > > However, as you mentioned this would mean all dApp devs would have to restructure their programs from lib.rs to main.rs, add #![no_main], change the crate type in Cargo.toml and copy .elf instead of .so files for deployment.
> 
> If we use @deanmlittle's sbpf linker, can't we just do this in some post-process step?

Yes, that would be a possibility. We are considering submitting a patch to LLD so that it accepts eBPF code. If the patch is accepted, I believe we should choose something that works there as well.

Lichtso: We could just ignore the field, as it does not really carry any meaningful information anyway.

LucasSte: > We could just ignore the field, as it does not really carry any meaningful information anyway.

That is a good idea. It allows us to change to a binary in the future without any modification in the VM.

proposals/0189-sbpf-stricter-elf-headers.md
intrigus-lgtm: @Lichtso am I missing something?

Lichtso: I can adjust the comparison to be inclusive. But it is not really a useful case because if the program headers reach the end of the file, then their contents must be empty.

proposals/0189-sbpf-stricter-elf-headers.md
buffalojoec: My two cents is that - as mentioned by both @Lichtso and @LucasSte - Solana programs are much more closely aligned with executables (BPF kernel programs, `executable` account types). Therefore I agree `ET_EXEC` is the most accurate `e_type`.

`ET_DYN` is one of the more complex file layouts because of the extra indices (`.dynamic`, GOT, etc.) `ET_EXEC` makes those things optional. As we move into things like static syscalls and simplified ELFs, we can stay compliant with lots of tooling by making `ET_EXEC` the required `e_type` for Solana programs. From what I can tell, this SIMD is already pretty well-aligned with the `ET_EXEC` standard.

The shift from `.so` to `.elf` is max pain for devs, so we can include a grace period where our toolchain will build both an `.elf` and `.so`, both of which follow the `ET_EXEC` standard. Then we can slowly wane off supporting `.so` files by adding things like CLI errors that say "Point to the `.elf` instead".

All that to say, I suggest making `ET_EXEC` required in this SIMD while deprecating `ET_DYN` and `.so`. The proposal can explicitly state the intent to eventually remove support for the latter. What do you guys think?

buffalojoec: Also not to be "that guy" but the SIMD title is "Stricter ELF Headers" while relaxing the check on `e_type` is actually _less_ strict. 🤓

LucasSte: > ET_DYN is one of the more complex file layouts because of the extra indices (.dynamic, GOT, etc.) ET_EXEC makes those things optional

They are also optional for ET_DYN.

> All that to say, I suggest making ET_EXEC required in this SIMD while deprecating ET_DYN and .so. The proposal can explicitly state the intent to eventually remove support for the latter. What do you guys think?

Although I believe the migration can occur, it must not be linked to this change. As I mentioned before, the migration incurs changes in the SDK/Pinocchio, and potentially minor changes in the toolchain/debugger. I think it is best to keep these things separated for now. If you want, I believe it would be best to start a discussion thread either on Agave repo or in the SIMD repo for us to identify what needs to be done for programs to migrate to a binary, and then come up with a SIMD if that is the case.

I might be shortsighted, but, if we keep this same file ELF layout, I don't see any immediate gains for program size or CU consumption by switching to another type.

buffalojoec: I'm just thinking about how we might be able to make sure our compiled binaries can be analyzed by upstream-compatible tooling. If our ELF layout is pretty closely aligned with `ET_EXEC`, we might as well lean into that and use the proper `e_type` so that off-the-shelf tools can analyze Solana binaries directly from the chain (or compiled locally).

I completely agree we shouldn't be too strict and force breakage on all of our tools/SDKs, which is why I was suggesting an intermediate step of deprecation and compiling both `.so` and `.elf`. We could even just support both for a bit and deprecate `.so` later.

LucasSte: > I'm just thinking about how we might be able to make sure our compiled binaries can be analyzed by upstream-compatible tooling. If our ELF layout is pretty closely aligned with `ET_EXEC`, we might as well lean into that and use the proper `e_type` so that off-the-shelf tools can analyze Solana binaries directly from the chain (or compiled locally).

On one hand, the kernel expects a ET_REL, not a ET_EXEC. I asked the type not be ET_REL, because I believe linkers won't solve relocations for ET_REL files. On the other, llvm tools and gnu-binutils tools work with eBPF code regardless of the ELF type. As far as I can tell, Asymmetric Research was using Ghidra with our SBPFv0 ET_DYN  files without any problems related to the ELF type (they had problems with relocations though).

Was there any tool that you tried and did not work with ET_DYN?

> I completely agree we shouldn't be too strict and force breakage on all of our tools/SDKs, which is why I was suggesting an intermediate step of deprecation and compiling both `.so` and `.elf`. We could even just support both for a bit and deprecate `.so` later.

Either way, relaxing the restriction is a step towards that direction. I don't think discussing when deprecation or migration might happen belongs in this SIMD. I suggest we create a discussion thread instead.

proposals/0189-sbpf-stricter-elf-headers.md
deanmlittle: resolved as headers are now removed and rodata header is now optional.

deanmlittle (approved):
lgtm! thanks for all the work on this one 🙏

PR #193, opened 2024-11-12
Update 0133-syscall-get-epoch-stake.md

pucedoteth:
updated SIMD 0089 link (link404)


buffalojoec (commented):
Thanks! I think the linter is just barking at you.

Benhawkins18:
@pucedoteth can you add a line break right after the link so the linter can be happy? If you can do that I'd like to merge this

buffalojoec (approved):

Benhawkins18 (approved):
Looks good to me. Thanks for cleaning this up.

PR #194, opened 2024-11-14
SIMD-0194: Deprecate rent exemption threshold

deanmlittle:


proposals/XXXX-deprecate-rent-exemption-threshold.md
```diff
@@ -0,0 +1,123 @@
...
+implementation will be unaffected.
+
+## Alternatives Considered
+
+* Leave things as they are.
+  - Calculating rent exemption on a program remains an 
+expensive operation.
+
+* Allow users to make the assumption that `2.0` will remain stable and do `u64` 
+math themselves.
```
buffalojoec, line 94: We could stablize this, but imo the field will soon become a misnomer, considering rent collection is being disabled. I think your proposed design makes more sense.

proposals/XXXX-deprecate-rent-exemption-threshold.md
```diff
@@ -0,0 +1,123 @@
...
+## Drawbacks
+
+None.
+
+## Backwards Compatibility
+
+This feature is backwards compatible. It does not change the absolute value 
+required for an account to be rent exempt and the current way of calculating 
+the threshold will continue to evaluate to the same value. Any deployed program 
+will be unaffected.
```
buffalojoec, line 123: Technically is a breaking change if anyone's program is reading fields individually from the rent sysvar.

buffalojoec (commented):
Love the change, just a few open questions. I'm not sure what the best deprecation process is for the field.

proposals/XXXX-deprecate-rent-exemption-threshold.md
febo: If a program is reading the fields individually, it should still expect the values to change, right? So if it relies on the fields having a specific "fixed" value, then it is more an issue with the program implementation than a breaking change. The rent exempt lamports amount will stay the same and the current way to determine the amount will continue to work.

buffalojoec: Fair point, I guess it's technically _not_ breaking. If anyone is expecting the values to be stable though, maybe we should at least call out that, as per the design, the values are obviously going to change.

brooksprumo (commented):
The whole concept of Rent is going away. Since no accounts are rent-paying currently, the rent lamports can be considered a state bond. A big goal with state economics is to enable this state bond to be dynamic and reduce 100x or so. Under normal load this drops "rent" to be much much cheaper, while also allowing higher loads to exponentially increase fees to allocate state.

I *think* we'd want new sysvars/etc for programs to be able to read, so this change to the exemption threshold is probably fine.

proposals/0194-deprecate-rent-exemption-threshold.md
```diff
@@ -0,0 +1,123 @@
...
+pub fn minimum_balance(&self, data_len: usize) -> u64 {
+    (ACCOUNT_STORAGE_OVERHEAD + data_len as u64) * self.lamports_per_byte
+}
+```
+
+Validator implementations should stop using `exemption_threshold` and only use
+the `lamports_per_byte` value. Any existing program using the current 
+implementation will be unaffected.
+
+## Alternatives Considered
```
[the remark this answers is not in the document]
deanmlittle: We can actually skip the equivalence check entirely if we just decide that the best outcome for this SIMD is to ossify it at `2.0`. The thing if, if we're already going to do that, it feels much cleaner to change it to `1.0` so it can be mathematically simplified out of the fee calculation formula as opposed to just being ignored.

proposals/0194-deprecate-rent-exemption-threshold.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+ Remove all `f64` math from Rent calculations in all SDKs and onchain programs 
+moving forwards, replacing it with simple `u64` math, ie:
+
+```rs
+/// Minimum balance due for rent-exemption of a given account data size.
+pub fn minimum_balance(&self, data_len: usize) -> u64 {
+    (ACCOUNT_STORAGE_OVERHEAD + data_len as u64) * self.lamports_per_byte
+}
```
topointon-jump, line 80: Making this calculation faster is great. I'm just wondering if renaming `lamports_per_byte_year` to `lamports_per_byte` might be confusing for people grepping for `lamports_per_byte_year`. As rent is going away anyway I wonder if this would add more short-term confusion?

If the goal of this SIMD is to remove the floating point operations from this calculation, I wonder if it would be simpler to mark `exemption_threshold` as deprecated and make the following change in the SDK to remove the floating-point calculation:

```suggestion
    /// Minimum balance due for rent-exemption of a given account data size.
    pub fn minimum_balance(&self, data_len: usize) -> u64 {
        let bytes = data_len as u64;
        // exemption_threshold has been deprecated and is always 2
        let exemption_threshold = 2;
        ((ACCOUNT_STORAGE_OVERHEAD + bytes) * self.lamports_per_byte_year) * exemption_threshold
    }
```

Thoughts? That said, I am ok with the proposed change, just wanted to throw a simpler alternative out there.

deanmlittle: Yes, this was one of the discussed alternatives [above](https://github.com/solana-foundation/solana-improvement-documents/pull/194#discussion_r1853351246). If we simply decide "This number is now 2 forever, we promise not to change it", it has the least moving parts. It doesn't require a feature gate either. We can simply change all our libraries to ignore the f64 value stored in the Rent account/sysvar and substitute with our own u64. The main downsides are that we waste 1 extra CU multiplying by 2 every time unnecessarily, and instead of removing existing protocol crust, we introduce a new piece where we have this random number 2 thrown in there for no reason. I prefer changing the value to 1 to remove it from the equation, but both are fine changes imo.

topointon-jump: > If we simply decide "This number is now 2 forever, we promise not to change it", it has the least moving parts. It doesn't require a feature gate either. We can simply change all our libraries to ignore the f64 value stored in the Rent account/sysvar and substitute with our own u64.

I would prefer this option, but I am happy with either.

> What if we were to just hardcode the value of rent in our libraries too?

Do you mean making the minimum balance independent of the size of the account? I think that we should still take the size of the account into consideration.

deanmlittle: No, I mean making the lamports per bytes for rent exemption static. Right now, the reason we have a syscall and/or Rent account at all is to tell us:
1. how many `lamports_per_byte_year` we are paying, which is multiplied by
2. the `exemption_threshold` of `2.0f64` years

If these values will not change between now and when Rent is removed altogether, we could simply hardcode both values into our libraries enabling you to statically derive the correct amount of rent to pay without invoking a syscall at all.

nazreen:
What's the latest on this SIMD? It does seem excessive to have to make a SysVar call to get constants

joncinque:
I haven't heard about how rent will evolve in the future since this proposal was created, so I'm now inclined to go toward the original formulation of this SIMD, of changing the threshold to 1 and doubling the lamports_per_byte_year.

@brooksprumo you initially mentioned that rent was going away soon, is that still the case? How does the proposed change sound to you now?

brooksprumo:
> @brooksprumo you initially mentioned that rent was going away soon, is that still the case? How does the proposed change sound to you now?

Yep, rent is gone now. Well, gone in the way a normal person thinks of rent 😅 Now we can treat it as a state bond.

I think this SIMD is fine to proceed. Or at least *I* don't know of a reason not to.

topointon-jump (approved):

nazreen (approved):

Algorhythmic1 (approved):

buffalojoec (approved):

amalnathsathyan (approved):

jacobcreech (approved):

PR #195, opened 2024-11-14
Added SIMD for tpu vote using QUIC

lijunwangs:
Added SIMD for tpu vote using QUIC

### DIFF
--- a/proposals/0034-tpu-vote-using-quic.md
+++ b/proposals/0034-tpu-vote-using-quic.md
@@ -1,0 +1,107 @@
+---
+simd: '0034'
+title: TPU Vote using QUIC
+authors:
+  - Lijun Wang <lijun@anza.xyz>
+category: Standard
+type: Core
+status: Draft
+created: 2024-11-13
+development: 
+  - Anza - WIP
+  - Firedancer - Not started
+---
+
+## Summary
+
+Use QUIC for transporting TPU votes among Solana validators. This requires
+supporing receiving QUIC based vote TPU packets on the server side and sending
+QUIC-based TPU vote packets on the client side.
+
+
+## Motivation
+
+As timely vote credits are awarded to validators, they might be incentived to
+increase the TPU vote traffic to ensure their votes are received in a timely
+manner. This could cause congestions and impact overall TPU vote processing
+effectiveness. The concurrent UDP based TPU vote does not have any flow control
+mechanism.
+
+We propose to apply the pattern taken for TPU transaction processing to TPU vote
+processing -- by utlizing the flow control mechanism which were developed including
+built-in QUIC protocol level flow control, and application-level rate limiting on
+connections and packets.
+
+## Alternatives Considered
+
+There is no readily-available alternative to QUIC which addresses some of the
+requirements such as security (reliability when applying QOS), low latency and
+flow control. We could solve the security and flow control with TLS over TCP
+the concern is with the latency and head-of-line problems. We could also
+customize and build our own rate limiting mechanism based on the UDP directly,
+this is non-trivial and cannot solve the security problem without also rely on
+some sort of crypto handshaking. 
+
+
+## Detailed Design
+
+On the server side, the validator will bind to a new QUIC endpoint. Its
+corresponding port will be published to the network in the ContactInfo via
+Gossip. The client side will use the TPU vote QUIC port published by the server
+to connect to the server.
+
+The TPU vote will be using the same QUIC implementation used by regular
+transaction transportation. The client and server both uses their validator's
+identity key to sign the certificate which is used to validate the validator's
+identity especially on the server side for the purpose of provding QOS based on
+the client's stakes by checking the client's Pubkey -- stake weighted QOS.
+
+Once a QUIC connection is established, the client can send vote transaction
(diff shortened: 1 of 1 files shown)


proposals/0034-tpu-vote-using-quic.md
```diff
@@ -0,0 +1,112 @@
...
+On the server side, the validator will bind to a new QUIC endpoint. Its
+corresponding port will be published to the network in the ContactInfo via
+Gossip. The client side will use the TPU vote QUIC port published by the server
+to connect to the server.
+
+The TPU vote will be using the same QUIC implementation used by regular
+transaction transportation. The client and server both uses their validator's
+identity key to sign the certificate which is used to validate the validator's
+identity especially on the server side for the purpose of provding QOS based on
+the client's stakes by checking the client's Pubkey -- stake weighted QOS.
```
wen-coding, line 60: Just saying: validator id to vote account association and the stake distribution could change at the Epoch boundary. At Epoch boundary you probably need to accept if the validator has take in either the old Epoch or the new Epoch.

proposals/0034-tpu-vote-using-quic.md
```diff
@@ -0,0 +1,112 @@
...
+
+Once a QUIC connection is established, the client can send vote transaction
+using QUIC UNI streams. In this design, a stream is used to send one single Vote
+transaction. After that the stream is closed.
```
wen-coding, line 64: How fast is it to open and close a stream? The QUIC connection will be persistent I assume?

proposals/0034-tpu-vote-using-quic.md
```diff
@@ -0,0 +1,112 @@
...
+
+The server only supports connections from the nodes which has stakes who can
+vote. Connections from unstaked nodes are rejected with `disallowed` code.
+
+The following QOS mechanisms are employed:
+
+* Connection Rate Limiting from all clients
+* Connection Rate Limiting from a particular IpAddress
```
wen-coding, line 72: What are the limits here? I thought in QUIC we rate limit by pubkey instead of IP? Because others can spoof your IP.

proposals/0034-tpu-vote-using-quic.md
```diff
@@ -0,0 +1,112 @@
...
+* Total concurrent connections from all clients -- this is set to 2500
```
wen-coding, line 73: Is it dangerous to increase this limit? What if we suddenly have >2500 clients?

proposals/0034-tpu-vote-using-quic.md
```diff
@@ -0,0 +1,112 @@
...
+* Max concurrent connections from a client Pubkey -- this is set to 1 for votes.
+* Max concurrent streams per connection -- this is allocated based on the ratio
+of the validator's stake over the total stakes of the network.
```
wen-coding, line 76: What is the reasoning here? Why does higher-staked validator need more concurrent streams per connection?

proposals/0195-tpu-vote-using-quic.md
```diff
@@ -0,0 +1,112 @@
...
+
+The following QOS mechanisms are employed:
+
+* Connection Rate Limiting from all clients
+* Connection Rate Limiting from a particular IpAddress
+* Total concurrent connections from all clients -- this is set to 2500
+* Max concurrent connections from a client Pubkey -- this is set to 1 for votes.
+* Max concurrent streams per connection -- this is allocated based on the ratio
+of the validator's stake over the total stakes of the network.
+* Maximum of vote transactions per unit time which is also stake weighted
```
ripatel-fd, line 77: Firedancer won't enforce any of these limits server-side. 
These should more like implementation-specific behavior, so I suggest removing them from the SIMD.

proposals/0034-tpu-vote-using-quic.md
lijunwangs: Stake info is periodically updated and is eventual consistent. It applies the QOS based on the latest info as it see it. I do not think we need to complicate it to look at potentially two sets of data.

proposals/0034-tpu-vote-using-quic.md
lijunwangs: Open and close stream is light-weight operation. Can be done in one shot: open stream, send and close

proposals/0034-tpu-vote-using-quic.md
lijunwangs: For stake validation we are validating the pub key. This is only for connection  rate limiting before handshake completion

proposals/0034-tpu-vote-using-quic.md
lijunwangs: This is up to the implementation. And we have plan to make this configurable. 2500 is more than enough for votes in mb

proposals/0034-tpu-vote-using-quic.md
lijunwangs: Give them more bandwidth as reward and ensure they have better chance landing their votes?

proposals/0195-tpu-vote-using-quic.md
lijunwangs: Removed specific numbers

Benhawkins18:
stale

PR #197, opened 2024-11-19
SIMD-0197: Chili Peppers

topointon-jump:
Initial SIMD for Chili Peppers proposal for state bandwidth used by transactions.

### DIFF
--- a/proposals/0197-chilli-peppers.md
+++ b/proposals/0197-chilli-peppers.md
@@ -2,7 +2,9 @@
 simd: '0197'
 title: Chili Peppers
 authors:
-  - Firedancer Team
+  - Josh Siegel
+  - Jeff Bezaire
+  - Tom Pointon
 category: Standard
 type: Core
 status: Review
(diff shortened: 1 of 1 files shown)


HaoranYi:
Can we fix the markdown lint, please?

proposals/0197-chilli-peppers.md
```diff
@@ -0,0 +1,147 @@
...
+a precise mechanism to quantify and limit the resources consumed by transactions in terms of state memory usage.
+Note that Chili Peppers required for a transaction are not the size of hot state touched by the transaction, but rather the amount of cold
+state made hot or new state allocated. Transactions that are only accessing already hot state would require 0 Chili Peppers.
+ 
+Block Chili Pepper Limit - The maximum number of Chili Peppers that can be requested in a single block.
+ 
+Block Chili Pepper Clock - a cumulative measure of the total Chili Peppers requested since the genesis of the blockchain. At the beginning of each block, the Block State Unit Clock is updated to reflect the cumulative total of Chili Peppers consumed up to that point, ensuring a monotonically increasing record of state consumption.
+ 
+Account Chili Pepper Clock Timestamp - each account in the "hot" state needs to keep track of the value of the Block Chili Pepper Clock the last time it was accessed (read or written to). This allows determination of which accounts are hot and which are cold.
```
HaoranYi, line 47: How to determine whether the account is "hot"?

proposals/0197-chilli-peppers.md
```diff
@@ -0,0 +1,147 @@
...
+ 
+Hot Cache Size - A new predefined constant, termed "Hot Cache Size" corresponds approximately to the size of the hot state supported
```
HaoranYi, line 49: How is the size determined?

proposals/0197-chilli-peppers.md
```diff
@@ -0,0 +1,147 @@
...
+ 
+The Block Chili Pepper Clock serves as a cumulative measure of the total Chili Peppers requested since the genesis of the blockchain. At the beginning of each block, the Block State Unit Clock is updated to reflect the cumulative total of Chili Peppers requested up to that point, ensuring a monotonically increasing record of state consumption.
+ 
+Computing the block Chili Pepper clock:
+ 
+```python
+block_chili_pepper_clock = prev_block_chili_pepper_clock + sum(txn.requested_chili_peppers for txn in block.txns)
+```
+ 
+Implemented as a 64-bit unsigned integer (uint64), this clock is updated at the beginning of every block to reflect the total Chili Peppers requested since the chain's genesis. This monotonically increasing value is stored in a dedicated system variable (sysvar), ensuring that it remains accessible and immutable throughout the blockchain's operation.
```
HaoranYi, line 80: begining or end?

Without actually looking up the account, how would a block know how many chili peper are actually used?

proposals/0197-chilli-peppers.md
```diff
@@ -0,0 +1,147 @@
...
+ 
+ 
+## Alternatives Considered
+ 
+What alternative designs were considered and what pros/cons does this feature
+have relative to them?
+ 
+## Impact
+ 
+How will the implemented proposal impacts dapp developers, validators, and core contributors?
```
HaoranYi, line 137: We will need to expand this section. 
How will this impact dapp developers?

proposals/0197-chilli-peppers.md
```diff
@@ -0,0 +1,147 @@
...
+ 
+## Security Considerations
+ 
+What security implications/considerations come with implementing this feature?
```
HaoranYi, line 141: Answers for these questions?

proposals/0197-chilli-peppers.md
```diff
@@ -0,0 +1,147 @@
...
+- **Error Description**: A transaction specifies an invalid number of Chili Peppers, either by requesting more than a predefined maximum limit per transaction or by formatting the request improperly.
+- **Handling**: The transaction is invalidated, and an "Invalid Chili Pepper Request" error message is issued. Developers must ensure that Chili Pepper requests conform to protocol specifications, including proper formatting and adherence to maximum limits.
+ 
+#### Account State Unit Clock Synchronization Failure
+ 
+#### Accessing Cold Account
+ 
+- **Error Description**: A transaction attempts to interact with an account that has been designated as "cold" due to its Chili Pepper Clock Timestamp falling below the "Hot Cache Size" threshold relative to the current Block Chili Pepper Clock, without requesting sufficient Chili Peppers.
+- **Handling**: The transaction is rejected with a "Cold Account Access Attempted" error.
+
```
jumpsiegel, line 128: Accessing a cold account and a non-existent account without sufficient chili peppers to account for the meta-data associated with an account should return the same error.  This would allow us to keep the meta-data in a cold state as well and prevents the dos attack of referencing large numbers of non-existent  accounts just to crush the slower back end database where the cold is stored.

proposals/0197-chilli-peppers.md
jumpsiegel, line 141: chili peppers are a consumable resource without cost?   Does this mean I can 

1) put a single txn with all the chili peppers possible on a block 
2) a high enough priority fee to guarantee it gets put into the block 

and lock out everybody elses ability to retrieve cold accounts?

I would guess we would limit the max chili peppers a single txn could request to be the max number of accounts it could reference times 10m...  Given ALUT, that would be a lot of chili peppers.. all of them

I would suggest we make chili peppers be representative of actual SOL so that it would cost to do this attack where we only actually charge for actual cold access?!  a real head scratcher...

proposals/0197-chilli-peppers.md
jumpsiegel: accessing the meta-data for an account needs to be accounted for via chili-peppers as well?

proposals/0197-chilli-peppers.md
jumpsiegel: A block builder sees a txn, notices it references unknown accounts, notices it does not have the chili peppers required to account for the meta data for those unknown accounts, does not include that txn

I am now thinking hitting cold meta data needs to be super expensive..

proposals/0197-chilli-peppers.md
jumpsiegel: Most likely by deciding what the "standard" size of a solana network box should be and backing into how much should be used for account state.   For example, given the small amount of new-state actually referenced per block, I would guess we could set this at 50gb and it would be 99% good...

proposals/0197-chilli-peppers.md
jumpsiegel: calculating chilli peppers could be mostly hidden by the client sdk so I would guess once people update to them, there should be almost zero effect

HaoranYi: > every transaction on the Solana network will be required to use the ComputeBudget program to request the maximum number of Chili Peppers they require.

Wouldn't the dapp specify the Chili pepper in the transactions? In order to do so, will they need to know whether the account is hot or cold?

proposals/0197-chilli-peppers.md
HaoranYi: Thinking a bit more. "hot cache size" (50gb) is the physical memory capacity, right?

chili pepper meter measures just the bytes of the account's data loaded from disk. But there are other extra cost in addition to the account's data bytes to load, such as account's index, account's chili-pepper store etc. The actual physical memory will be larger than the chili pepper bytes?

If we are targeting 50gb physical memory, should we take into accounts of those extra cost and reduce them from the physical memory budget? If so, how much should we reduce?

PR #198, opened 2024-11-20
SIMD-0198: Define CUs for Builtin instructions

tao-stones:


topointon-jump (approved):

topointon-jump (changes requested):
Accidentally approved, this SIMD should actually be closed as the builtins are being migrated to BPF.

tao-stones:
Close it since all builtins are migrating to bpf; after migration, they will consume actual compute units as any other bpf instructions, instead of use predefined static CUs. Making this proposal obsolete. 

For devs who want to have a better understanding/estimation of current builtin instructions' cost, https://github.com/anza-xyz/agave/issues/3364 tracks internal benchmark projects that provide stats.

PR #199, opened 2024-11-21
SIMD-0199: Inflation

anselsol:
How to read this doc?

You can either :

- [for community members] Go to the [formatted file](https://github.com/anselsol/solana-improvement-documents/blob/main/proposals/0199-dynamic-inflation.md)
- [for devs] click on "Files changed" on the tab just above this comment, this will show you a file diff

### DIFF
--- a/proposals/0199-dynamic-inflation.md
+++ b/proposals/0199-dynamic-inflation.md
@@ -1,0 +1,61 @@
+---
+simd: '0199'
+title: Dynamic Inflation
+authors:
+  - A2KDefi
+  - Ansel (Tokamai)
+category: Standard
+type: Core
+status: Review
+created: 2024-11-21
+---
+
+## Summary
+
+The purpose of this feature is to enable Solana reward emission to be dynamic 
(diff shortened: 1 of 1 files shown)


anselsol:
@levicook that's a very good question. Cosmos has this kind of super complex formula to stabilize oscillations depending on the multiple threshold (not just 50%) - we need to learn more about the implications to be able to reply to this (source : https://forum.cosmos.network/t/proposal-48-accepted-inflation-rate-change-correction/4834)

Right now the curve ([you can find it here](https://solana.com/docs/economics/inflation/inflation-schedule)) goes to 1.5% in 6 years. The idea was to go there much quicker to stabilize the total supply.

From what we could gather a big part of the staking APY do come from fees right now (and the 50% fees burn will soon be removed).

The current inflation is around 4.8%. I think the staking APY (maybe you're talking about this) takes into account auto compounding + base tx fees + priority fees (+ jito tips if running Jito).

Let's take this to the the discussion tab 😅 I'll copy this in the first comment

PR #204, opened 2024-11-26
SIMD-0XXX: Slashable event verification

AshwinSekar:
SIMD to deploy a program to record and verify slashable events on chain

**This proposal does not modify any stakes or rewards, the program will only verify and log infractions.**


proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,297 @@
...
+- `0x00`, a fixed-value byte acting as the instruction discriminator
+- `offset`, an unaligned eight-byte little-endian unsigned integer indicating
+  the offset from which to read the proof
+- `slot`, an unaligned eight-byte little-endian unsigned integer indicating the
+  slot in which the violation occured
+- `node_pubkey`, an unaligned 32 byte array representing the public key of the
+  node which committed the violation
+
+We expect the contents of the `proof_account` when read from `offset` to
+deserialize to a struct of two byte arrays representing the duplicate shreds.
```
joncinque, line 77: Although this is clear to me, let's also specify that another on-chain program must be used to upload the proof data into an account. We can also link the record program as a usable example.

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,297 @@
...
+### Incident reporting
+
+After verifying a successful proof we store the results in a program derived
+address for future use. The PDA is derived using the `node_pubkey`, `slot`, and
+the violation type:
+
+```rust
+let (pda, _) = find_program_address(&[
+  node_pubkey.to_bytes(),
+  slot.to_le_bytes(),
```
joncinque, line 182: Let's specify that it's a little-endian 64-bit unsigned int

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,297 @@
...
+  ViolationType::DuplicateBlock.to_u8(),
```
joncinque, line 183: Let's say the specific byte here

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,297 @@
...
+])
+```
+
+At the moment `DuplicateBlock` is the only violation type but future work will
+add additional slashing types.
+
+If the `pda` account has non-zero lamports, then we abort as the violation has
+already been reported. Otherwise we create the account, with the slashing program
```
joncinque, line 191: The non-zero lamport check is exploitable -- we should check to see if it's already a valid proof. Otherwise, I could plan a duplicate block in slot X and just pre-fund the account at that PDA with some lamports.

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,297 @@
...
+
+### Incident reporting
+
+After verifying a successful proof we store the results in a program derived
+address for future use. The PDA is derived using the `node_pubkey`, `slot`, and
+the violation type:
+
+```rust
+let (pda, _) = find_program_address(&[
+  node_pubkey.to_bytes(),
```
joncinque, line 181: Let's specify that it's a 32-byte public key

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,297 @@
...
+add additional slashing types.
+
+If the `pda` account has non-zero lamports, then we abort as the violation has
+already been reported. Otherwise we create the account, with the slashing program
+as the owner. In this account we store the following:
+
+```rust
+struct ProofReport {
+  reporter: Pubkey,              // Fee payer, to allow the account to be closed
+  epoch: Epoch,                  // Epoch in which this report was created
```
joncinque, line 197: Let's specify it's an unaligned unsigned 64-bit little-endian int

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,297 @@
...
+  pubkey: Pubkey,                // The pubkey of the node that committed the violation
+  slot: Slot,                    // Slot in which the violation occured
```
joncinque, line 199: Let's specify it's an unaligned unsigned 64-bit little-endian int

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,297 @@
...
+  violation_type: u8,            // The violation type
+  proof: Vec<u8>                 // The serialized proof
```
joncinque, line 201: Let's specify this clearly as an unaligned unsigned 32-bit little-endian int giving the size followed by that many bytes for the proof data

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,297 @@
...
+  proof_account: Option<Pubkey>, // Optional account where proof is stored instead
```
joncinque, line 202: Rather than using an `Option`, let's just always include this, with the caveat that the data at `proof_account` might get modified. And let's add it before `proof` to make it easier to find

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,297 @@
...
+
+The `pubkey` is populated with the `node_pubkey`. For future violation types that
+involve votes, this will instead be populated with the vote account's pubkey.
+The work in SIMD-0180 will allow the `node_pubkey` to be translated to a vote account
+if needed.
+
+Note that PDA's can only be created with a 10kb initial size.
+Although not a problem for `DuplicateBlockProofData`, if future proof types require
+more space, we allow the proof to be stored in a separate account, and linked back
+to the PDA using the `proof_account` field.
```
joncinque, line 218: This does kind of worry me for traceability... but let's save that discussion for when it's needed

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,297 @@
...
+
+---
+
+### Closing the incident report
+
+After the slashing violation has been processed by the runtime, the initial fee
```
joncinque, line 224: nit: let's be super clear and say that the runtime processing is out of scope

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,297 @@
...
+- Deriving the pda using `pubkey`, `slot`, and `ViolationType::DuplicateBlock`
+  as outlined above does not result in the adddress of `report_account`
+- `report_account` is not writeable
+- `report_account` does not deserialize cleanly to `ProofReport`
+- `report_account.reporter` is not a signer
+- `report_account.epoch + 3` is greater than the current epoch reported from
+  the `Clock` sysvar. We want to ensure that these accounts do not get closed before
+  they are observed by indexers and dashboards.
+
+Otherwise we close the `report_account` and credit the `lamports` to `destination`
```
joncinque, line 255: nit: can be super clear and specify that closing entails setting the owner to the system program and reallocating the account to 0 bytes

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,297 @@
...
+  slot which was reported
+- `pubkey`, an unaligned 32 byte array representing the public key of the node
+  which was reported
+
+We abort if:
+
+- `violation_type` is not `0x00` (corresponds to `DuplicateBlock` violation)
+- Deriving the pda using `pubkey`, `slot`, and `ViolationType::DuplicateBlock`
+  as outlined above does not result in the adddress of `report_account`
+- `report_account` is not writeable
```
joncinque, line 248: We probably don't need this check -- the runtime will do this for us when we try to write to it. On the flip side, we probably want to check that it's owned by the slashing program

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,297 @@
...
+
+---
+
+### Closing the incident report
+
+After the slashing violation has been processed by the runtime, the initial fee
+payer may wish to close their `ProofReport` account to reclaim the lamports.
+
+They can accomplish this via the `CloseProofReport` instruction which requires
+2 accounts:
```
joncinque, line 228: Sounds like it needs one more account -- the reporter to sign. But I wonder if we can allow this to be totally permissionless by having the reporter provide the lamport destination when they create the report. That way we're sure slash reports are always cleaned up.

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,297 @@
...
+1. `destination`: Writable account to reclaim the lamports
+
+`CloseProofReport` has an instruction data of 42 bytes, containing:
+
+- `0x01`, a fixed-value byte acting as the instruction discriminator
+- `violation_type`, a one byte value acting as the violation type discriminator
+- `slot`, an unaligned eight-byte little-endian unsigned integer indicating the
+  slot which was reported
+- `pubkey`, an unaligned 32 byte array representing the public key of the node
+  which was reported
```
joncinque, line 241: Are these parameters actually needed? I think we'll be OK without them

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,297 @@
...
+
+- `violation_type` is not `0x00` (corresponds to `DuplicateBlock` violation)
+- Deriving the pda using `pubkey`, `slot`, and `ViolationType::DuplicateBlock`
+  as outlined above does not result in the adddress of `report_account`
+- `report_account` is not writeable
+- `report_account` does not deserialize cleanly to `ProofReport`
+- `report_account.reporter` is not a signer
+- `report_account.epoch + 3` is greater than the current epoch reported from
+  the `Clock` sysvar. We want to ensure that these accounts do not get closed before
+  they are observed by indexers and dashboards.
```
joncinque, line 253: Is there any other rationale for this number, like association to the leader schedule or time for stake / vote accounts to get penalized? It's ok if there isn't, but it would be good to spell it out.

joncinque (commented):
Really great work! Most of my comments are minor nits for clarification

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,297 @@
...
+1. Create a new program account at `S1ashing11111111111111111111111111111111111`
+  with an upgrade authority set to the system program
+  `11111111111111111111111111111111`
+
+2. Verify that the program account
+  `8sT74BE7sanh4iT84EyVUL8b77cVruLHXGjvTyJ4GwCe` has a verified build hash of
+  `<FILL IN AFTER IMPLEMENTATION>` [\[1\]](#notes)
+
+3. Copy the contents of `8sT74BE7sanh4iT84EyVUL8b77cVruLHXGjvTyJ4GwCe` into
+  `S1ashing11111111111111111111111111111111111`
```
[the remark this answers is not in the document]
AshwinSekar: Not familiar with the loaders so i'll defer to your wisdom :), just let me know which one makes the most sense and I can include it here.

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,297 @@
...
+  `DuplicateBlockProofData`.
+- The resulting shreds do not adhere to the Solana shred format [\[2\]](#notes)
+  or are legacy shred variants.
+- The resulting shreds specify a slot that is different from `slot`.
+- The resulting shreds specify different shred versions.
+
+After deserialization the slashing program will attempt to verify the proof, by
+checking that `shred1` and `shred2` constitute a valid duplicate proof for
+`slot` and are correctly signed by `node_pubkey`. This is similar to logic used
+in Solana's gossip protocol to verify duplicate proofs for use in fork choice.
```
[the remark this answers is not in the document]
AshwinSekar: It is yeah, clients are expected to be able to deserialize and verify the proof messages sent in gossip and update their fork choice accordingly.
If they do not detect or ingest these messages they're at risk of having their fork choice diverge from the rest of the cluster.

proposals/0204-slashable-event-verification.md
AshwinSekar: I like that approach, can definitely see people forgetting to close their reports

AshwinSekar: Could we just send it back to the initial reporter? Or is it worth carrying around a separate address as the destination

proposals/0204-slashable-event-verification.md
AshwinSekar: Got it, i've modified it to check if the account has any data and is owned by the slashing program.

proposals/0204-slashable-event-verification.md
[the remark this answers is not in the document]
joncinque: Let's go with the default loader (bpf-loader-upgradeable) for now. We can amend it if loader-v4 becomes the default sooner than that.

proposals/0204-slashable-event-verification.md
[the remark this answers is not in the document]
joncinque: Makes sense, thanks!

proposals/0204-slashable-event-verification.md
joncinque: I'd keep them separate -- it gives extra flexibility and clarity at the cost of an extra 32 bytes, which shouldn't be too bad, especially considering the accounts will get cleaned up

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,313 @@
...
+in Solana's gossip protocol to verify duplicate block proofs for use in fork choice.
+
+#### Proof verification
+
+`shred1` and `shred2` constitute a valid duplicate block proof if any of the
+following conditions are met:
+
+- Both shreds specify the same index and shred type, however their payloads
+  differ
+- Both shreds specify the same FEC set, however their merkle roots differ
```
lidatong, line 124: will there be a feature gate to deprecate / drop all shreds except chained merkles? that would simplify this check to just the comparing signature (`memcmp` 64 bytes vs. 1228) which could add up

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,313 @@
...
+- Both shreds specify the same FEC set and are coding shreds, however their
+  erasure configs conflict
+- At least one shred is a coding shred, and its erasure meta indicates an FEC set
```
lidatong, line 127: in particular, the shred with the lower FEC set index has to be the coding shred right? and the other can be data or coding

AshwinSekar: Yeah that's correct I'll update the language

proposals/0204-slashable-event-verification.md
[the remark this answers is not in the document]
AshwinSekar: Specified the default loader and your comment about moving the program data account, let me know if it looks correct f4f8ab7a66a95f60d5a267246629a9498bc316dc

proposals/0204-slashable-event-verification.md
AshwinSekar: Updated caf905835be80d97e219129bd3a3839aae968175

proposals/0204-slashable-event-verification.md
lidatong: > That could work since we already validate the merkle root -> signature as part of the sigverify section.

yep exactly, and actually meant that would also obviate the need to compare the payloads with two shreds with the same index as well. so could potentially merge these two conditions into a single "Both shreds specify the same FEC set index, however their signatures differ"

(assuming only chained merkle)

AshwinSekar: the merkle root is only 32 bytes, so I don't think it saves us anything here vs comparing the signature. It requires some compute to put together, but we have to do it anyway for the signature verification

AshwinSekar: discussed offline about the payload check:
- we cannot remove it because a leader could specify 2 shreds with the same index but different fec set indices bypassing the merkle root check
- Once we have fixed size fec sets, this becomes simple to address as we know the range of indices per fec sets
- the payload check is not technically wrong, just inefficient. We can remove it once we have fixed size fec sets

proposals/0204-slashable-event-verification.md
AshwinSekar: separated it here f89a01ba5262bb8b2e79d79f373a84e2dd534d3a

AshwinSekar:
@joncinque @lidatong this should be ready for final review.
I've updated the verifiable build hash and linked the release.

buffalojoec (commented):
Sorry for barging in, but I was curious about slashing and decided to read through the proposals.

I left a few comments related to program development concepts, but overall I think the design looks good. Take them or leave them!

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,351 @@
...
+
+If both proof and signer verification succeed, we continue on to store the incident.
+
+---
+
+#### Incident reporting
+
+After verifying a successful proof we store the results in a program derived
+address for future use. The PDA is derived using the `node_pubkey`, `slot`, and
+the violation type:
```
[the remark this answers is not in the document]
AshwinSekar: Good question, a duplicate block report is submitted when we detect 2 different blocks for the same slot from a leader. In theory the leader could submit as many different blocks for the slot as they want, however the extra consensus load to resolve this situation is roughly the same, so we don't choose to track or add an extra punishment for this. It is also difficult to accurately determine how many versions of the block they submitted as one individual validator might not see all versions. 

in the future once we add the voting related slashing violations it is possible for the voter to commit multiple of the same `violation_type` infractions in the same slot, but we choose to still only count 1 violation per `violation_type` per slot. Although there could be 10 bad votes for slot `X`, the consensus impact is still the same, as the voting program will naturally only accept one vote per slot. The remaining votes will fail and be charged a transaction fee for their network load. 

 If you're curious about how the economics will play out, the current proposal (SIMD-0212 https://github.com/solana-foundation/solana-improvement-documents/blob/297bcd403d3502c01e1c55239971c6cdc81490b4/proposals/0212-slashing.md) assigns weights to the `violation_type` of each infraction and uses that to derive the stake fraction that will be slashed

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,351 @@
...
+struct DuplicateBlockProofData {
+  shred1_length: u32      // Unaligned four-byte little-endian unsigned integer,
+  shred1: &[u8]           // `shred1_length` bytes representing a shred,
+  shred2_length: u32      // Unaligned four-byte little-endian unsigned integer,
+  shred2: &[u8]           // `shred2_length` bytes representing a shred,
+}
+```
+
+Users are expected to populate the `proof_account` themselves, using an onchain
+program such as the Record program.
```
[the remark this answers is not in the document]
AshwinSekar: I did originally implement account creation and write in the slashing program but after discussion with Jon we decided it would be better to leave it up to the user.

These proof accounts are ephemeral, after the proof is verified by the slashing program a report account which includes the proof is created,
We don't require any restrictions or special control over the proof accounts, so in the future if there's some use case for composability or reuse of these proof accounts it seems better to give the user freedom here.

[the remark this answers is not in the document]
buffalojoec: Ok, if it's ephemeral then I guess it's okay. It just feels weird to me that a user would be creating a proof account under a different program, then paying to create a report account under this program, then closing that report under this program to reclaim rent, then closing the proof account under another program to reclaim rent.

The "close" instructions could be combined on this program.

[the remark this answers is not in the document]
buffalojoec: You could also completely avoid this edge case:
https://github.com/solana-foundation/solana-improvement-documents/pull/204/files#diff-e9a92f44cc9e8d1789705b2a4a51bba175df745484c6638e475514d14eefe716R267-R269

proposals/0204-slashable-event-verification.md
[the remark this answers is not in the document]
buffalojoec: Ok nice, thanks for the explanation! I starting exploring SIMD-0212 and will continue doing so. 🫡

proposals/0204-slashable-event-verification.md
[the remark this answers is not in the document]
joncinque: I'll give some more background for this. Since the proofs are larger than the MTU, we'd need to upload it in chunks, which means implementing some logic similar to the loaders for writing a slice into an account.

Rather than re-implementing chunking / writing logic in this program, we can just lean on any other program that can write to a large buffer and copy them much more simply and cheaply. Does that makes sense?

proposals/0204-slashable-event-verification.md
```diff
@@ -0,0 +1,351 @@
...
+                                      is closed.
+  epoch: Epoch,                    // Unaligned unsigned eight-byte little endian
+                                      integer representing the epoch in which this
+                                      report was created
+  violator: Pubkey,                // 32 byte array representing the pubkey of the
+                                      node that committed the violation
+  slot: Slot,                      // Unaligned unsigned eight-byte little endian
+                                      integer representing the slot in which the
+                                      violation occured
+  violation_type: u8,              // Byte representing the violation type
```
joncinque, line 261: nit: there's a missing `//` after the first line of the comments

joncinque (approved):
Looks good to me!

proposals/0204-slashable-event-verification.md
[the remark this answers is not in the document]
buffalojoec: The bit about re-using existing code is salient, but afaict it takes ~20 lines to write a chunk of data in the record program if you don't need to realloc. That doesn't sound to me like too bad of a tradeoff for the elimination of fractured ownership edge cases.

You could add a single instruction here with the expectation of the account being allocated to the full size of the proof already. Then, reimplement the little bit of chunking code, and just build out the proof under this program, *not* copying it. You could even combine the proof and report accounts as well, making this program even simpler and cheaper.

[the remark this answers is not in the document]
joncinque: Writing chunks into the account is certainly not that tough, but it's unnecessary extra code. The whole proof gets copied into the report account in the current implementation, but it *only* gets copied from account data, and never instruction data. And the proof gets copied all at once, so we can never be in a state of partial writes.

We can add the ability to read from instruction data too in future work, but I wouldn't hold up this proposal on that point.

joncinque (approved):

PR #206, opened 2024-11-26
SIMD 0033: Move to activated

AshwinSekar:
Feature has been activated on all clusters


Benhawkins18 (approved):
LGTM Thanks for updating

PR #207, opened 2024-12-05
SIMD-0207: Raise Block Limits to 50M

apfitzge:


proposals/0207-raise-block-limits-to-50M.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Alternatives Considered
+
+- Leave the block limits as they are
+  - This leaves capacity on the table, we want to get ready for future
+    increases.
+- Double limit to 96M CUs
+  - Viewed as too aggressive at this time, and may cause unforeseen issues
+    particularly in infrastructure supporting the network users.
+  - We instead plan to increase the limits incrementally as the network
```
mertimus, line 64: Can we get a more concrete idea of what this looks like? I think doing a small increase first is prudent, but I think the increases proceeding that should be non-linear and increasing in %

proposals/0207-raise-block-limits-to-50M.md
```diff
@@ -0,0 +1,85 @@
...
+| Max Block Accounts Data Size Delta | 100MB | 100MB |
+
+## Alternatives Considered
+
+- Leave the block limits as they are
+  - This leaves capacity on the table, we want to get ready for future
+    increases.
+- Double limit to 96M CUs
+  - Viewed as too aggressive at this time, and may cause unforeseen issues
+    particularly in infrastructure supporting the network users.
```
mertimus, line 63: which infrastructure specifically? are we referring to RPCs?

7layermagik:
How about making it so the increases are contingent on cluster vote latency being good enough?

You can propose a doubling in block limit to 96mil cu and then increase the cu limit linearly each epoch (1 or 2mil cu's for example). Each epoch you can increment it and if the 80 percentile (could be something else) by stake vote latency is too slow that indicates it shouldn't be increased the next epoch until the cluster performance is good enough. For example, say right now 1.6 slot vote latency is our 80th percentile, we keep increasing cu's each epoch but once it goes over say 1.8 slot vote latency it stops increasing until latency drops again. There should also be a criteria that blocks are at least 80 or 90% full or so on avg during the epoch because we don't want to be comparing vote latencies between two epochs where the avg traffic is a lot different. Anyway, just some quick ideas.

proposals/0207-raise-block-limits-to-50M.md
apfitzge: > unforeseen

RPCs, indexers, historical transaction data providers come to mind, but idk what's out there. Hoping people will either alert us or fix their system if increased limits are coming close to breaking their stuff.

proposals/0207-raise-block-limits-to-50M.md
apfitzge: It's gonna depend how things go, and how quickly we can improve performance.
To the best of my knowledge, we have no plans for some automatic increasing limit at this time; any further changes will go through same SIMD process.

7layermagik:
> > How about making it so the block limit cu increases are contingent on cluster vote latency being within a good enough range?
> > You can propose a doubling in block limit to 96mil cu and then increase the cu limit linearly each epoch (1 or 2mil cu's for example). Each epoch you can increment it and if the 80 percentile (could be something else) by stake vote latency is too slow that indicates it shouldn't be increased the next epoch until the cluster performance is good enough. For example, say right now 1.6 slot vote latency is our 80th percentile, we keep increasing cu's each epoch but once it goes over say 1.8 slot vote latency it stops increasing until latency drops again. There should also be a criteria that blocks are at least 80 or 90% full or so on avg during the epoch because we don't want to be comparing vote latencies between two epochs where the avg traffic is a lot different and blocks aren't even close to max (because vote latency will go up once they are fuller so that's risky to base assumptions on not very full blocks). Anyway, just some quick ideas.
> 
> that seems like something for a next SIMD after seeing how this one goes



> > How about making it so the block limit cu increases are contingent on cluster vote latency being within a good enough range?
> > You can propose a doubling in block limit to 96mil cu and then increase the cu limit linearly each epoch (1 or 2mil cu's for example). Each epoch you can increment it and if the 80 percentile (could be something else) by stake vote latency is too slow that indicates it shouldn't be increased the next epoch until the cluster performance is good enough. For example, say right now 1.6 slot vote latency is our 80th percentile, we keep increasing cu's each epoch but once it goes over say 1.8 slot vote latency it stops increasing until latency drops again. There should also be a criteria that blocks are at least 80 or 90% full or so on avg during the epoch because we don't want to be comparing vote latencies between two epochs where the avg traffic is a lot different and blocks aren't even close to max (because vote latency will go up once they are fuller so that's risky to base assumptions on not very full blocks). Anyway, just some quick ideas.
> 
> that seems like something for a next SIMD after seeing how this one goes

That'll add many months though. The SIMD process seems to take forever -- might as well make the proposal a more thorough one

proposals/0207-raise-block-limits-to-50M.md
```diff
@@ -0,0 +1,92 @@
...
+
+The following table shows the current block limits and the proposed block
+limits:
+
+| Type | Current Block Limit | Proposed Block Limit |
+|------|----------------------|----------------------|
+| Max Block Units | 48M | 50M |
+| Max Writable Account Units | 12M | 12M |
+| Max Vote Units | 36M | 36M |
+| Max Block Accounts Data Size Delta | 100MB | 100MB |
```
godmodegalactus, line 54: What does increasing Block Accounts Data Size Delta affect ?

proposals/0207-raise-block-limits-to-50M.md
```diff
@@ -0,0 +1,92 @@
...
+
+This proposal only changes the `Max Block Units` limit.
```
godmodegalactus, line 56: What is the Block limit for non-vote transaction after this increase ?

apfitzge:
> That'll add many months though. The SIMD process seems to take forever -- might as well make the proposal a more thorough one. 2mil more cu's isn't going to stress things much at all either I imagine -- idk what value there is increasing it only that much

Goal with this is kind of a "foot in the door" from my perspective.

With this simd, we set up system so we can easily add features for block-limits. I'm hoping to  backport this.
Future SIMDs adjusting limits can just be backported and activated more quickly. Regular procedure of taking eons to activate be damned.

proposals/0207-raise-block-limits-to-50M.md
apfitzge: We're not increasing it. It limits overall state growth.

mmmahne (approved):

proposals/0207-raise-block-limits-to-50M.md
blockiosaurus: Is the feature gate actual required? It would be nice to have automatic progression and so long as you use a static calculation based off of numbers already under consensus (e.g. Epoch or Slot) the validators should be able to easily calculate the current block limit.

steviez: The feature gate gives us the ability to raise the limit at a time when we feel the cluster is healthy and can handle the additional load that the higher limit could introduce. Adding in "raise limit at epoch N" adds unknowns; what if the cluster is struggling when epoch N comes up or what if the entire validator set hasn't adopted a new enough version by the time epoch N rolls around.

> It would be nice to have automatic progression

This is somewhat similar to what 7layermagik mentioned [here](https://github.com/solana-foundation/solana-improvement-documents/pull/207#issuecomment-2521143758). But to quote Andrew from [here](https://github.com/solana-foundation/solana-improvement-documents/pull/207#issuecomment-2521252932), we are viewing this as a "foot in the door" feature. Implementing this SIMD will establish the logic to raise the limit which will enable future SIMDs to raise the limit to be easier

steviez (approved):

ripatel-fd (approved):
Approving even if the increase is merely symbolic. 48M CUs is far too low and holding back performance of potentially faster clients.

bw-solana (approved):

bw-solana (approved):
Approved by Anza

noelportillo (commented):
Good luck

noelportillo (commented):
Yeah I'm sure

PR #208, opened 2024-12-06
SIMD-0208: Reopen Closed Programs

blockiosaurus:
This proposal is the migration of the discussion found at https://github.com/solana-labs/solana/pull/27460, regarding closed program accounts not being recoverable. This SIMD proposes adding in the ability to reopen a closed program through the same mechanism as standard program deploys.

### DIFF
--- a/proposals/0208-reopen-closed-programs.md
+++ b/proposals/0208-reopen-closed-programs.md
@@ -6,7 +6,7 @@ authors:
   - justin.starry@icloud.com
 category: Standard
 type: Core
-status: Draft
+status: Review
 created: 2024-12-06
 feature:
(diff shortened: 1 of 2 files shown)


Lichtso:
We won't do any changes to loader-v3 (the one where closing is final) anymore since it is about to be deprecated.
But, in the loader-v3 to v4 migration we could potentially resurrect old v3 programs if they still have an authority set.

Benhawkins18:
Closing: This is not a full SIMD with sufficient detail. Please use this [repo's discussion page](https://github.com/solana-foundation/solana-improvement-documents/discussions) if you are still on the idea phase

PR #209, opened 2024-12-10
SIMD-0207: Add feature-gate and update status

apfitzge:
Updating status of SIMD-0207 with the tracking feature-key and link to issue/PR in agave


Benhawkins18 (approved):
Only a status update. Approving and merging.

PR #212, opened 2024-12-17
SIMD-0205: Slashing

AshwinSekar:
Follow up to SIMD-0204, implements penalties for validators that commit slashable offenses.


Benhawkins18:
This change in its current form is no longer applicable: Closing

PR #217, opened 2025-01-04
SIMD-02XX: Intermediate vote credits

AshwinSekar:


bji:
As the author of the public mods that most are using to perform the vote-stuffing/backfilling for extra credit, I thoroughly support this proposal.

proposals/0218-intermediate-vote-credits.md
```diff
@@ -0,0 +1,116 @@
...
+extends: SIMD-0033 Timely Vote Credits
+---
+
+## Summary
+
+This feature aims to build upon the effort of timely vote credits in order
+to improve the vote credit algorithm.
+
+Credits are awarded for a block voted on by a validator that ultimately gets
+finalized. For consensus purposes a vote for a block is implicitely a vote for
```
lostintime101, line 20: #
```suggestion
finalized. For consensus purposes a vote for a block is implicitly a vote for
```

proposals/0218-intermediate-vote-credits.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Motivation
+
+Although greatly improved via TVC, the credit accounting algorithm is still lacking
+in measuring the contribution of validators to consensus. There still exist
+gaps whereby validators can gained increased credits through modifications that
+do not have any consensus benefits. Running unaudited code poses a risk to the network.
+
+Since there is no consensus benefits to these modifications, by aligning the credit
+algorithm to more accurately reflect consensus contriubtion, we hope to reduce the
```
lostintime101, line 32:
```suggestion
algorithm to more accurately reflect consensus contribution, we hope to reduce the
```

proposals/0218-intermediate-vote-credits.md
```diff
@@ -0,0 +1,116 @@
...
+```
+[100] - [101] - [103] - [104] - [105]
+ v 2     v 2                     v 2
+            \ - [102]
+                 v
+```
+
+Assuming this top fork gets finalized, the validator will receive 3 * 16 = 48 credits.
+Although the validator did not vote on 103 or 104, its timely vote on 105 conveyed
+its vote on these intermediate blocks implicitely. Thus we propose that the
```
lostintime101, line 62:
```suggestion
its vote on these intermediate blocks implicitly. Thus we propose that the
```

michaelh-laine:
100% support this, thank you Ashwin

Spud957:
If I understand this correctly, the penalty for voting on a bad fork will only be a higher latency for the missed slots on the correct fork?

bji:
> As a validator in the example you illustrate do I even need to vote on slot 102 (there's no credit for it) but rather just wait a bit longer to see which fork gets heavier quicker?

Because of TVC, waiting costs vote credits.  So yeah if you think on average you will earn more credits from waiting than not waiting, you would wait; but this is by no means a given, and my personal belief is that it would be a losing strategy most of the time.

bji:
> If I understand this correctly, the penalty for voting on a bad fork will only be a higher latency for the missed slots on the correct fork?

The primary "penalty" for being on the wrong fork is not being allowed to vote for a span of slots on the right fork.

Spud957:
> > If I understand this correctly, the penalty for voting on a bad fork will only be a higher latency for the missed slots on the correct fork?
> 
> The primary "penalty" for being on the wrong fork is not being allowed to vote for a span of slots on the right fork.

That's what I'm wondering, doesn't this almost get rid of that penalty? You get to implicitly vote on all the slots you would have missed from the correct fork, it's just at a higher latency.

kingsichi:
Will this proposal not lead to validators purposefully skipping votes on initial blocks of a fork knowing that if they perform a single timely vote later down the line they will receive full credits for those blocks they missed..? And so encouraging validators not to vote where there is fork contention.

kingsichi:
Or will those credits only be awarded for locked out slots they couldn't vote for due to their voting on the alternative fork.

kingsichi:
Just realised that this concern has already been raised above.

AshwinSekar:
> You propose storing credits alongside latency in LandedVote and are not suggesting increasing the number of lockouts in the state, correct?

There's actually no extra storage that needs to occur. Essentially when we root a slot we update credits based on latency. At this point we should also issue credits for any intermediate slots between the old root and the new root.

lidatong (approved):
approving for FD

jstarry:
> Why does it make sense to have a lower penalty for lockouts? Doesn't this incentivize a new type of mod that recklessly votes on all slots since the penalty is so low?

No, it won't incentivize that behavior because there is still a penalty. Also the longer your lockout on an unconfirmed fork, the greater the penalty. So as before, validators are incentivized to quickly switch forks as soon as they detect that another fork has a heavier weight and higher chance of getting confirmed.

Because the penalty is lower for temporarily voting on the wrong fork, everyone should be a little more incentivized to vote eagerly when there is a fork. This should lead to forks getting resolved sooner because fewer validators will be tempted to withhold their votes and wait to see which fork the cluster is more likely to confirm. @AshwinSekar does this sound correct?

AshwinSekar:
> Because the penalty is lower for temporarily voting on the wrong fork, everyone should be a little more incentivized to vote eagerly when there is a fork. This should lead to forks getting resolved sooner because fewer validators will be tempted to withhold their votes and wait to see which fork the cluster is more likely to confirm. @AshwinSekar does this sound correct?

In general yes this is correct.  IVC slightly lowers the penalty for voting on a wrong fork initially, and helps prioritizes timely votes over being always correct.

Spud957:
> > Why does it make sense to have a lower penalty for lockouts? Doesn't this incentivize a new type of mod that recklessly votes on all slots since the penalty is so low?
> 
> No, it won't incentivize that behavior because there is still a penalty. Also the longer your lockout on an unconfirmed fork, the greater the penalty. So as before, validators are incentivized to quickly switch forks as soon as they detect that another fork has a heavier weight and higher chance of getting confirmed.
> 
> Because the penalty is lower for temporarily voting on the wrong fork, everyone should be a little more incentivized to vote eagerly when there is a fork. This should lead to forks getting resolved sooner because fewer validators will be tempted to withhold their votes and wait to see which fork the cluster is more likely to confirm. @AshwinSekar does this sound correct?

It absolutely does incentivize mods that prioritize voting faster over voting on the correct fork. If you vote on the wrong fork one time, there is no penalty as long as you vote on the next correct slot within two slots (3 slot grace period). As TVC (as well as some stake pools) puts a premium on low latency, there is definitely an incentive to vote first and make sure it was the correct vote later. As long as you don't follow a bad for more than a couple slots, the penalty for voting on a bad fork is much less than taking a longer time to vote.

jstarry (approved):

jstarry:
> It absolutely does incentivize mods that prioritize voting faster over voting on the correct fork. If you vote on the wrong fork one time, there is no penalty as long as you vote on the next correct slot within two slots (2 slot grace period). As TVC (as well as some stake pools) puts a premium on low latency, there is definitely an incentive to vote first and make sure it was the correct vote later. As long as you don't follow a bad fork for more than a couple slots, the penalty for voting on a bad fork is much less than taking a longer time to vote.

I think you mean that it incentivizes mods that prioritize voting on blocks with lower slots when there is a fork. This is because a vote on a block with a lower slot might not lock you out of voting on the fork with a higher (more recent) slot. But this is actually the happy path and the default voting strategy already covers this well, you wouldn't need a mod to vote quickly on forks with a low slot.

The hard part is that a quick voting strategy could lead you to voting on a wrong fork which has a higher slot than the correct fork, and the you lock yourself out of receiving some vote credits on the correct fork. The "vote first, make sure later" strategy doesn't work in this case.

The default voting strategy already basically does the eager voting strategy that you're describing but with an added stake weighted fork selection algorithm and lockout threshold checks. So relative to that default strategy, I was explaining that there isn't really a new incentive to make a mod based on the changes from this SIMD.

That said, this SIMD isn't really able (or trying) to eliminate all of the different potential mods. Consensus is a game which rewards the participants who correctly predict what the confirmed chain will be and there will always be people looking for an edge and I think that's a good thing as long as those mods help consensus. I support this SIMD because it removes a particular modding incentive which doesn't help consensus in any way and I don't think it really adds any additional incentives to mod beyond what already existed before.

Spud957:
> > It absolutely does incentivize mods that prioritize voting faster over voting on the correct fork. If you vote on the wrong fork one time, there is no penalty as long as you vote on the next correct slot within two slots (2 slot grace period). As TVC (as well as some stake pools) puts a premium on low latency, there is definitely an incentive to vote first and make sure it was the correct vote later. As long as you don't follow a bad fork for more than a couple slots, the penalty for voting on a bad fork is much less than taking a longer time to vote.
> 
> I think you mean that it incentivizes mods that prioritize voting on blocks with lower slots when there is a fork. This is because a vote on a block with a lower slot might not lock you out of voting on the fork with a higher (more recent) slot. But this is actually the happy path and the default voting strategy already covers this well, you wouldn't need a mod to vote quickly on forks with a low slot.
> 
> The hard part is that a quick voting strategy could lead you to voting on a wrong fork which has a higher slot than the correct fork, and the you lock yourself out of receiving some vote credits on the correct fork. The "vote first, make sure later" strategy doesn't work in this case.
> 
> The default voting strategy already basically does the eager voting strategy that you're describing but with an added stake weighted fork selection algorithm and lockout threshold checks. So relative to that default strategy, I was explaining that there isn't really a new incentive to make a mod based on the changes from this SIMD.
> 
> That said, this SIMD isn't really able (or trying) to eliminate all of the different potential mods. Consensus is a game which rewards the participants who correctly predict what the confirmed chain will be and there will always be people looking for an edge and I think that's a good thing as long as those mods help consensus. I support this SIMD because it removes a particular modding incentive which doesn't help consensus in any way and I don't think it really adds any additional incentives to mod beyond what already existed before.

That makes sense, I’m supportive of the simd in general, I just think slots where the validator would be locked out shouldn’t be backfilled.

jstarry:
> That makes sense, I’m supportive of the simd in general, I just think slots where the validator would be locked out shouldn’t be backfilled.

I think this is fair but I think it's sufficient that they just can't get full credits for slots that they were locked out of. At this point it's difficult to implement a change that prevents backfilling slots that a validator is locked out of.

andrii-kl:
This proposal looks very interesting, but it may have broader implications than it initially appears. The current draft doesn’t clearly explain how certain edge cases will be handled in the new model.

1) What is the maximum allowable length for a backfill gap?

2)  How are vote credits calculated for blocks that are retroactively filled in such a gap?
       For example, if there is a gap of 10 consecutive slots that is later backfilled automatically, how is the credit (and corresponding reward) assigned to each of those slots?

Benhawkins18:
stale and not needed with alpenglow

PR #215, opened 2024-12-20
SIMD-0215: Accounts Lattice Hash

brooksprumo:

PR #223, opened 2025-01-13
SIMD-0223: Removes Accounts Delta Hash

brooksprumo:


jumpsiegel (approved):

jeffwashington (approved):
lgtm!!!

ibhatt-jumptrading (approved):

Benhawkins18 (approved):

ncitron:
One issue with this (as well as #215) is that we won't be able to build light clients for bridging purposes.

Right now, one can use the accounts epoch hash to update the validator balances and build a light client for verifying a given block, and then use opening proofs of the accounts delta hash to prove certain bridging actions have occurred.

By removing the accounts delta hash and migrating the accounts epoch hash to LtHash which doesn't support inclusion proofs, this will no longer be possible.

A solution here is to keep supporting the account delta hash (or allow some way for an account to optionally have its deltas merklized to a different hash at an additional cost) and to add a validator epoch hash.

This means we can support opening proofs to validator balances so we can keep track of the consensus state, and opening proofs to certain accounts that are involved in a bridging transaction. The size of the validator epoch hash should be small (and updated infrequently) so should not add much additional cost. We can also keep the size of a minimal "optional account delta hash" small by making creation/updating of these accounts more expensive so they will only be used by accounts involved in bridging operations.

PR #228, opened 2025-01-16
SIMD-0228: Market-Based Emission Mechanism

tjain-mcc:


proposals/simd0221
```diff
@@ -0,0 +1,132 @@
...
+There are two major implications of Smart Emissions: 
+
+Smart Emissions dynamically incentivizes participation when stake drops to secure the network.
+Smart Emissions minimize SOL issuance  to the Minimum Necessary Amount (MNA) to secure the network.
+
+This is good for the Solana network and network stakers for four reasons:
+
+High inflation can lead to more centralized ownership. To illustrate the point, imagine a network with an exceedingly high inflation rate of 10,000%. People who do not stake are diluted and lose ~99% of their network ownership every year to stakers. The higher the inflation rate, the more network ownership is concentrated in stakers’ hands after compounding for years.
+
+Reducing inflation spurs SOL usage in DeFi, which is ultimately good for the applications and stimulates new protocol development. Additionally, a high staking rate can be viewed as unhealthy for new DeFi protocols, since it means the implied hurdle rate is the inflation cost. Lowering the “risk free” inflation rate creates stimulative conditions and allows new protocols to grow.
```
kevinheavey, line 28: 1. This has no impact on real rates, just nominal rates.
2. Worth bearing in mind that since we don't have negative nominal rates, if you have 3% _deflation_ then real rates are unable to go below 3%, if you have 5% _inflation_ then real rates can't go below negative 5% etc

So moderate stable inflation is good. High stable inflation is unnecessary but not really a problem for defi as it gets priced in

tjain-mcc: 1. The nominal rates have a real market impact as described in the proposal 
2. I could see a future where we do have negative nominal rates. That is not explicitly in this proposal but I don't see why 0% nominal rate needs to be a lower bound. If people are willing to stake at a negative nominal rate because MEV is sufficient to incentivize them, why can't nominal rates be negative?

t-nelson:
ci output is broken/retarded. plz wrap all lines at 80char

Benhawkins18:
@tjain-mcc thanks for making a proposal. A few nits and clean up items: 

- SIMD is the wrong number and should match the PR number: SIMD-0224 -> SIMD-0228
- Title is too long. Something like "SIMD-0228: Market-Based Emission Mechanism" would be perfect
- Filename is in the wrong format and has a different number then the PR and the SIMD title. "simd0221" -> "0228-market-based-emission-mechanism"
- 1st line needs updating: "SIMD ###: Introducing a Programmatic, Market-Based Emission Mechanism Based on Staking Participation Rate" -> "SIMD-0228: Market-Based Emission Mechanism"
- there are numerous linting errors that need to be fixed.

proposals/simd0221
```diff
@@ -0,0 +1,132 @@
...
+
+The max function ensures that inflation is at least zero, and the min function ensures that the inflation does not rise above the current issuance curve.
+This design offers several key benefits:
+Consensus Safety: Adjusting inflation based on staking participation ensures sufficient validator incentives to maintain network security, prioritizing consensus safety.
+Market-Based Flexibility: The model adapts to the network's economic activity, making it more responsive to changing market conditions. It’s possible to imagine a future where stakers are earning enough from MEV that no SOL emissions are necessary. 
+Validator Retention: It accommodates Solana-aligned validators who are willing to stake even with lower emissions, recognizing that they can earn more through MEV in higher economic activity ecosystems.
+This dynamic approach balances the need for a secure, decentralized network with the flexibility to thrive in a competitive market.
+##Alternatives Considered
```
t-nelson, line 102: was an alternative considered based on burn rate instead of stake rate? throw out [simd0096](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0096-reward-collected-priority-fee-in-entirety.md), add something [simd0109](https://github.com/solana-foundation/solana-improvement-documents/pull/109)-ish with burn. use utility as the measure, with security as the goal.

proposals/simd0221
tjain-mcc: 1. I'm not sure I agree that borrow rates and inflation rates are so tightly linked. Imagine an inflation rate of 0%, would you expect a borrow rate of 0%? Borrow rate also needs to incorporate a view on MEV yield which is volatile 
2. Yeah I'm not proposing that. Just saying it is theoretically possible though I have not explored the technical feasibility

kevinheavey: 1. If inflation was 0 then the nominal rate would equal the real rate

tjain-mcc: I disagree, I think there are other factors which affect the nominal borrow rate other than the inflation rate. MEV yield and derivatives funding rates are both big factors.

proposals/simd0221
michaelh-laine: > throw out [simd0096]

ser... we're like 4 days out

michaelh-laine:
Thank you Tushar and Vishal for this proposal.

I appreciate the depth of thought that has gone into the reasoning for this proposal, though I still wish it were possible (maybe it is, maybe not) to develop empirical data showing the impacts of the current and proposed inflation schedules.

This being said, I support a long-term sustainable dynamic inflation schedule, but if we make a change it needs to be well thought out and sustainable. The current inflation schedule, while simple, has worked, where decreasing stake leads to higher real yields, resulting in staking yield (net of MEV) remaining relatively stable over the years in the 6.5-7.5% range. It lends itself to a gradual release of staked capital into DeFi with the equilibrium rate being around 7%, as nominal inflation drops more stake is released into the open market.

My other point here is that I see this being interlinked with a need for block reward distribution (#123) - in the long-term simple inflationary yield and MEV alone will likely become insufficient to attract stake, or put anothe way, with dynamic inflation the inflation rate may need to become sufficiently high to remain economically attractive, while we could see zero or negative nominal rates, as you mentioned in an earlier comment, if staking rewards are supplemented by block rewards, i.e. stakers pay to delegate to a certain validator, where they then receive a share of block rewards. That seems a bit of an extreme suggestion but theoretically the market forces could lead to this.

proposals/simd0221
```diff
@@ -0,0 +1,132 @@
...
+Solana’s current emission mechanism is a fixed, time-based formula that was activated on epoch 150, a year after genesis on February 10, 2021. The mechanism is not aware of network activity, nor does it incorporate that to determine the emission rate. Simply put, it’s “dumb emissions.” Given Solana’s thriving economic activity, it makes sense to evolve the network’s monetary policy with “smart emissions.” 
+
+There are two major implications of Smart Emissions: 
+
+Smart Emissions dynamically incentivizes participation when stake drops to secure the network.
+Smart Emissions minimize SOL issuance  to the Minimum Necessary Amount (MNA) to secure the network.
+
+This is good for the Solana network and network stakers for four reasons:
+
+High inflation can lead to more centralized ownership. To illustrate the point, imagine a network with an exceedingly high inflation rate of 10,000%. People who do not stake are diluted and lose ~99% of their network ownership every year to stakers. The higher the inflation rate, the more network ownership is concentrated in stakers’ hands after compounding for years.
```
bji: But inflation will never be that high, and never so high that this effect is significant.

Inflationary rewards are a backstop on defi gains.  If a stakeholder has something profitable to do with their SOL in an epoch, then they do that.  If they do not, then they stake.  Therefore staking represents the minimum return that anyone will get from their SOL, not the maximum.  Anyone who is not staked is presumably earning even more than staking returns from their staked SOL, otherwise *they would not stake and instead do whatever that other thing is that earns even more rewards*.

Therefore anyone staking is actually financially disadvantaged compared to anyone earning even more rewards from more profitable use of SOL.

There is a third category though, of SOL-holders which are using the SOL for some purpose which doesn't generate on-chain profit but somehow supplies some other utility to them.  And yes, these SOL-holders are "spending" inflationary rewards, and will not earn inflationary income in the same way that stakeholders do.  But this is a conscious choice of choosing some other utility over inflationary rewards, and I don't think that we can "have our cake and eat it too" -- you can't both earn inflationary rewards and *use* your SOL at the same time.

So in other words, I think the reasoning here with the 10,000% inflation rate is flawed and does not support the SIMD.

proposals/simd0221
```diff
@@ -0,0 +1,132 @@
...
+
+Targeting a staking participation rate of 50% ensures sufficient stake for consensus safety while maintaining the network’s security and decentralization.
+
+Below 33%, we potentially risk network safety because a supermajority of all SOL has explicitly not voted on any given block and this opens the edge case possibility of long range attacks. It is important to note that these long range attacks are entirely theoretical and we have not seen one in practice. There are other mechanisms in Solana to protect against long range attacks.
+
+This proposal is the first in a series of steps to make Solana’s consensus more secure and economics more market driven. The successor to this proposal is another SIMD that introduces the concept of long-term staking, which seeks to improve network security. The option to unstake SOL on a relatively short notice (i.e., a short cool down period) poses a potential risk to networks’ stability and safety, particularly in extreme circumstances where a significant amount of SOL is unstaked within a brief timeframe. The combination of these two SIMDs address these concerns while improving network security and economic activity.
```
bji: I don't support this proposal.  Inflation is primarily intended to incentivize secure and reliable validator operation.  It is on a reducing schedule because the plan was that transaction fees would take on a greater and greater role in incentivizing validator operation over time, so inflation could be reduced to compensate.  And this has turned out to be exactly the case - most validators earn much more from transaction/priority/MEV fees than from inflationary rewards.

When the inflationary reward rate falls, the incentive to keep SOL staked is reduced.  But it is reduced equally for everyone.  So while the overall staking rate may drop, the relative rate between stakeholders should stay the same; there is no reason for different stakeholders to be more or less likely to stake because they are all subject to the same rewards rate.

Therefore, everyone is likely to stake less; but stake less in equal proportion.  And since the only thing that really matters is the relative voting power that stake supplies, and since that relative power stays the same as staking levels drop equally across the board, the actual inflation level does not matter with regards to any of the reasoning that was given for the proposal of this SIMD.

In other words, if we believe that a 50% reduction in inflation rate would result in 50% less SOL being staked, that is irrelevant, as everyone is equally motivated to reduce their stake by that 50%; and after this equal reduction across all stakeholders, the relative stake held in validators is the same as before, so validator voting power does not effectively change.  And with no change in voting power, there is no change in security properties of the network.  And thus, no reason to try to target a specific inflation rate for security purposes.  So I think this SIMD is not needed.

mertimus: > So while the overall staking rate may drop, the relative rate between stakeholders should stay the same; there is no reason for different stakeholders to be more or less likely to stake because they are all subject to the same rewards rate.

this assumes everyone is equally rational or has the same circumstances, which is obviously not the case

not to mention that it *does* change incentives differently across different types of stakeholders, this is not a homogenous set. inflation being lower means CEXes offering "staking yield" are less attractive as one big counter example

michaelh-laine: Yield-oriented stakers are equally disincentivised by a 50% reduction, the issue is that with a 50% reduction in total stake, the cost to attack the network decreases significantly as well. Yes, the distribution of stake may be the same if we only have 20% of total supply staked, but it means someone would only need to purchase and stake 10% of total supply to halt the network.

Of course there's more nuance to such an attack vector but I believe this is what the proposal speaks to in reference to the 33% figure, though I'm not sure the proposal's conclusion that 33% is "safe" is correct.

proposals/simd0221
t-nelson: it was always retarded

proposals/simd0221
bji: > > So while the overall staking rate may drop, the relative rate between stakeholders should stay the same; there is no reason for different stakeholders to be more or less likely to stake because they are all subject to the same rewards rate.
> 
> this assumes everyone is equally rational or has the same circumstances, which is obviously not the case
> 
> not to mention that it _does_ change incentives differently across different types of stakeholders, this is not a homogenous set. inflation being lower means CEXes offering "staking yield" are less attractive as one big counter example

So you're looking to pick defi winners and losers with this proposal?

bji: > Yield-oriented stakers are equally disincentivised by a 50% reduction, the issue is that with a 50% reduction in total stake, the cost to attack the network decreases significantly as well. Yes, the distribution of stake may be the same if we only have 20% of total supply staked, but it means someone would only need to purchase and stake 10% of total supply to halt the network.
> 
> Of course there's more nuance to such an attack vector but I believe this is what the proposal speaks to in reference to the 33% figure, though I'm not sure the proposal's conclusion that 33% is "safe" is correct.

There are already safeguards in place.  Large stake movements invoke tampering down of the amount of stake that activates/deactivates per epoch.

witaliy12 (approved):

proposals/simd0221
kevinheavey: I am not saying that inflation is the only factor affecting nominal rates. I am saying inflation does not affect *real* rates

kevinheavey: That said, if inflation was 0 then the nominal rate would indeed equal the real rate, by definition

fplp782:
This proposal assumes that staking % is positively correlated with economic activity but provides no evidence for it. Tbh, I don't think it's positively correlated at all. Over the past year, Solana staking APRs have grown from 6.5% to almost 10%, and staking % has actually gone down (69% → 65%).

mertimus:
> This proposal assumes that staking % is positively correlated with economic activity but provides no evidence for it. Tbh, I don't think it's positively correlated at all. Over the past year, Solana staking APRs have grown from 6.5% to almost 10%, and staking % has actually gone down (69% → 65%).

Now talk to me about all the other variables that changed in the past year

This is also not the point of the proposal — it simply makes no sense that the issuance is a hardcoded value independent of anything on the chain but time

fplp782:
I posted this article on twitter which outlines a few problems with this SIMD, alongside the point I made in here earlier regarding staking % as a poor metric to determine inflation.

https://x.com/FPLParana/status/1880334708225216931

To Mert's comment above, I don't view the current model as ideal and I do believe revisions need to be made. But this SIMD is not a sensible solution IMO

davidcameron1533:
In DeFi protocols, inflation typically arises from the issuance of new tokens distributed as rewards. While these rewards attract users, excessive token issuance can lead to value depreciation. Similarly, the Solana network mints a set amount of SOL tokens annually to ensure network security and distribute staking rewards. However, this inflation rate, initially high, follows a declining mechanism over time. Excessive inflation can undermine token value and trust in the network, making a balanced inflation rate essential for maintaining attractive rewards and the long-term sustainability of the token's value.

proposals/simd0221
davidcameron1533: In the crypto world, when interest rates are zero, the appeal of lending decreases, but borrowers can take out loans at zero cost. This situation is not sustainable for maintaining economic balance. In short, having a nominal rate of 0 is a very illogical scenario for the Solana system.

bsparango:
@mertimus I think the thing to be addressed in @fplp782's pushback is that there needs to at least be a body of evidence proving that the market-based inflation mechanism is going to produce the expected outcome as laid out in this proposal. 

We already know that crypto markets are not efficient markets. I'm not yet convinced that a market for SOL inflation would be any more rational than the rest of this industry.

mertimus:
> @mertimus I think the thing to be addressed in @fplp782's pushback is that there needs to at least be a body of evidence proving that the market-based inflation mechanism is going to produce the expected outcome as laid out in this proposal.
> 
> We already know that crypto markets are not efficient markets. I'm not yet convinced that a market for SOL inflation would be any more rational than the rest of this industry.

I think it's a valid point but I'm not sure that's the thing to prove in this context. It's more of a nice to have. We can't show or prove the non-linear effects of any mechanism that goes live beforehand — we just need to agree on one thing first: 

Does the current inflation of SOL make sense? Is it optimal for the network?

If the answer is no, then we must change it. So how can we change it? We could set another arbitrary curve like the existing one and just reduce it in half. But then we are again doing something arbitrarily with no basis. That might've been OK during the release, but now at this stage of maturity, we can not be the arbiters of truth of the network anymore. 

So what's the alternative? We let the market set it. I don't see how it could be any other way. That's what this proposal is about — should inflation be purely time-based, based on copying Cosmos 4 years ago, or market-based? My vote is market-based. Market-based doesn't mean the results will make sense on a 1-year time horizon with several variables, it's that it will eventually lead to a reasonable equilibrium over time. We can't control the output, but we can give people incentives. Do we disagree that a market-set variable is better than an arbitrary one that takes nothing but time into account?

The alternative, of course, is that we *do* think the current inflation of SOL makes sense and/or that it is optimal for the network. But if that's the position, then where is the evidence for it? Especially with SIMD96 going live, the 50% burn is now removed and will increase inflation even more — while stakeholders in the network have repeatedly expressed that they believe inflation is too high to begin with. This will cause even more people to stake just to be protected from issuance. And I say this as a validator who will benefit negatively from lowering issuance and hence the staking rate.

Let me repeat this again, very large numbers of stakeholders in the network are *already* complaining about the inflationary mechanics of SOL (and with good reason), and it is actually about to get **much worse** with SIMD96 (removing the burn). I don't think it's good for the future of the network to ignore something requested by so many users. We don't even have to agree on what it should be set to yet; we just have to agree on whether market-based is less incorrect than the existing setup.

fplp782:
@mertimus basing inflation on staking % is not market based. they have nothing to do with one another. there are instances where mev and prio fees going up leads to staking % going down, and vice versa. If we want market based, we need to use a different unit of measure, not staking %. that is what I'm saying. Also, consider the fact that the proposed constant k = .05 is also an arbitrary number the proposers came up with. How do we know that that's the right number, or did we just come up with it the same way we came up with the current schedule?

From the article I posted, I say the following:
Adjusting Inflation Based on Staking % Is Not “Market-Based”
The proposal aims to reduce inflation when validators don’t need it and increase inflation when validators do need it in order to find the Minimum Necessary Amount (MNA) of inflation. It does this by adjusting inflation to staking %, banking on the assumption that staking % is positively correlated with economic activity.
But that doesn’t seem to be the case. Over the past year, Solana staking APRs have grown from 6.5% to almost 10%, and staking % has actually gone down (69% → 65%). If this trend continues, then issuance will increase when it should decrease. On the flip side, it could very well be the case that when economic activity slows down, where fees and MEV decrease, staking % goes up.
On the surface this might not make sense, but economies work in unpredictable ways. Maybe slow economic periods means users are less attracted to the liquidity and composability of unstaked sol and are more incentivized, not less, to stake. Or maybe economic activity goes to Layer 2s that use USDC for gas, so there’s less of a reason to hold unstaked SOL.
My point isn’t to predict how staking % may change over time. My point is that we can’t predict how staking % may change or how it will correlate with economic activity. And if it turns out to not be positively correlated to economic activity in the absolute sense, then this model fails to adjust with market dynamics to find the Minimum Necessary Amount (MNA).
I kindly ask the proposers to present reliable evidence that suggests a high positive correlation between economic activity and staking %.

I also propose these potential alternatives:
If we want a “market-based” endgame type of model right now, we need to find a provably correlated, ungameable statistic that the model can reliably base its equation on. Staking % is not provably correlated. The idea to burn 50% of MEV and/or priority fees is gameable due to dark pools.
Unfortunately I don’t have the research or implementation details to suggest anything seriously, however, I think data propagated or node count could make sense. Before going through with either, there would have to be clear evidence of strong correlation, where the constant k would be determined by that rate of correlation. There would also have to be a way to reliably measure either of these statistics and implement it into the inflation model.

fplp782:
Paraphrasing Tushar, we want to create an economic policy that emulates an algorithmic "fed funds" rate. One that's trustless as its independent of human decision making. That requires research and deliberation. The model can't just be spun out of thin air, but needs statistical backing that it will adjust the inflation rate in the way we want it to. I don't believe this proposal will do that, nor has there been any evidence to suggest that it will. If we want to do this, we should do it right

proposals/simd0221
tjain-mcc: Kevin, it seems like you are saying only real rates matter. I disagree with that claim. Nominal rates also matter for the reasons we laid out in the proposal.

proposals/simd0221
tjain-mcc: Yes 10,000% inflation rate is obviously extreme. It is a thought experiment to illustrate the costs of inflation. Inflation has negative externalities which are easier to explain with really big numbers, though those externalities are still there with smaller numbers too.

tjain-mcc:
> From the article I posted, I say the following: Adjusting Inflation Based on Staking % Is Not “Market-Based” The proposal aims to reduce inflation when validators don’t need it and increase inflation when validators do need it in order to find the Minimum Necessary Amount (MNA) of inflation. It does this by adjusting inflation to staking %, banking on the assumption that staking % is positively correlated with economic activity. But that doesn’t seem to be the case. Over the past year, Solana staking APRs have grown from 6.5% to almost 10%, and staking % has actually gone down (69% → 65%). If this trend continues, then issuance will increase when it should decrease. On the flip side, it could very well be the case that when economic activity slows down, where fees and MEV decrease, staking % goes up. On the surface this might not make sense, but economies work in unpredictable ways. Maybe slow economic periods means users are less attracted to the liquidity and composability of unstaked sol and are more incentivized, not less, to stake. Or maybe economic activity goes to Layer 2s that use USDC for gas, so there’s less of a reason to hold unstaked SOL. My point isn’t to predict how staking % may change over time. My point is that we can’t predict how staking % may change or how it will correlate with economic activity. And if it turns out to not be positively correlated to economic activity in the absolute sense, then this model fails to adjust with market dynamics to find the Minimum Necessary Amount (MNA). I kindly ask the proposers to present reliable evidence that suggests a high positive correlation between economic activity and staking %.

The only purpose of inflation is to incentivize staking. It doesn't require some complex statistical analysis to demonstrate that all else equal, higher inflation leads to higher incentive to stake. The data from the past year you reference is not useful because everything else is not equal, there was a lot that happened in the ecosystem. Trying to prove the relationship between two variables (staking rate and inflation rate) when there are hundreds of other variables in the data is a fools errand. Rather than relying on data which is contaminated with all those other variables, we can just use simple logic: higher inflation leads to higher incentive to stake.

tjain-mcc:
> @mertimus I think the thing to be addressed in @fplp782's pushback is that there needs to at least be a body of evidence proving that the market-based inflation mechanism is going to produce the expected outcome as laid out in this proposal.
> 
> We already know that crypto markets are not efficient markets. I'm not yet convinced that a market for SOL inflation would be any more rational than the rest of this industry.

Trying to prove the relationship between 2 variables in a system with hundreds of variables is not feasible. Nor is it necessary. We can use the very simple logic that higher inflation rate leads to higher incentive to stake. Imagine a simple example to illustrate this point: there are 2 chains which are identical in every way (including MEV) but one has 1% inflation and the other has 50% inflation. It is obvious that the second chain will have higher staking participation.

tjain-mcc:
> Thank you Tushar and Vishal for this proposal.
> 
> I appreciate the depth of thought that has gone into the reasoning for this proposal, though I still wish it were possible (maybe it is, maybe not) to develop empirical data showing the impacts of the current and proposed inflation schedules.
> 
> This being said, I support a long-term sustainable dynamic inflation schedule, but if we make a change it needs to be well thought out and sustainable. The current inflation schedule, while simple, has worked, where decreasing stake leads to higher real yields, resulting in staking yield (net of MEV) remaining relatively stable over the years in the 6.5-7.5% range. It lends itself to a gradual release of staked capital into DeFi with the equilibrium rate being around 7%, as nominal inflation drops more stake is released into the open market.
> 
> My other point here is that I see this being interlinked with a need for block reward distribution (#123) - in the long-term simple inflationary yield and MEV alone will likely become insufficient to attract stake, or put anothe way, with dynamic inflation the inflation rate may need to become sufficiently high to remain economically attractive, while we could see zero or negative nominal rates, as you mentioned in an earlier comment, if staking rewards are supplemented by block rewards, i.e. stakers pay to delegate to a certain validator, where they then receive a share of block rewards. That seems a bit of an extreme suggestion but theoretically the market forces could lead to this.

I absolutely agree with the interlinked need for a block reward distribution. Validators should have an in protocol way to distribute block rewards to stakers.

fplp782:
> 
> The only purpose of inflation is to incentivize staking. It doesn't require some complex statistical analysis to demonstrate that all else equal, higher inflation leads to higher incentive to stake. The data from the past year you reference is not useful because everything else is not equal, there was a lot that happened in the ecosystem. Trying to prove the relationship between two variables (staking rate and inflation rate) when there are hundreds of other variables in the data is a fools errand. Rather than relying on data which is contaminated with all those other variables, we can just use simple logic: higher inflation leads to higher incentive to stake.

The purpose of inflation is to help validators be profitable when fees + mev aren't enough.

Let's say your proposal ends up reducing inflation to 2%. Then, for whatever reason, economic activity slows down such that mev and prio fees are virtually 0%. There could absolutely be the case that staking % does not go down, because of "hundreds of other variables" as you mentioned. This would lead to many validators, particularly the smaller ones, being unprofitable. If those smaller validators close down, the node count goes down. Meanwhile staking % might remain sticky as most staked sol is delegated and can be redelagated to other, larger validators that are able to remain profitable. This increases centralization.

fplp782:
I would just like to reiterate that I am not against a market-based algorithm to determine inflation. I'm against one based on staking %

mertimus:
> consider the fact that the proposed constant k = .05 is also an arbitrary number the proposers came up with. How do we know that that's the right number, or did we just come up with it the same way we came up with the current schedule?

i) The authors clearly tell you how 0.5 was determined: https://github.com/solana-foundation/solana-improvement-documents/pull/228/files#diff-4c9f52adaf6cb3c34374dfbcdddbeae747adaad976b844aa00935e00a2576940R291

>  @mertimus basing inflation on staking % is not market based. they have nothing to do with one another. there are instances where mev and prio fees going up leads to staking % going down, and vice versa. If we want market based, we need to use a different unit of measure, not staking %. that is what I'm saying. From the article I posted, I say the following: Adjusting Inflation Based on Staking % Is Not “Market-Based”. Over the past year, Solana staking APRs have grown from 6.5% to almost 10%, and staking % has actually gone down (69% → 65%). If this trend continues, then issuance will increase when it should decrease. On the flip side, it could very well be the case that when economic activity slows down, where fees and MEV decrease, staking % goes up. On the surface this might not make sense, but economies work in unpredictable ways. Maybe slow economic periods means users are less attracted to the liquidity and composability of unstaked sol and are more incentivized, not less, to stake. I kindly ask the proposers to present reliable evidence that suggests a high positive correlation between economic activity and staking %.

This is my problem with this line of thinking. You say something has "nothing to do with one another" and ask for "reliable evidence" while fallaciously citing **yourself**. You can't disprove that 1+1 equals 2 by saying that 6+5 does not equal 0.

The proposal is simple: make the incentives market-based instead of divorced from reality. It is of course more market-based than the status quo since it takes into account MEV and commissions which lead into staking rate rather than simply ignoring all of it. Saying "they have nothing to do with another" because you looked at an 8-month time slice with 500 other variable changes is incorrect.

As I've already said above, Market-based doesn't mean the results will make sense on a 1-year time horizon with several variables; it's that it will eventually lead to a reasonable equilibrium over time. We can't control the output, but we can give people incentives. Blockchains run entirely on incentives.

If you think looking at the staking rate is incorrect, then please provide robust reasoning as to why instead of citing data that is unrelated or using hypothetical anecdotes (like the one below). I am happy to engage if one makes an actual argument.

> Or maybe economic activity goes to Layer 2s that use USDC for gas, so there’s less of a reason to hold unstaked SOL.

?
 
> Unfortunately I don’t have the research or implementation details to suggest anything seriously

Then what are we doing here? You are saying you don't think a basic incentive change should be implemented without some mathematical proof while simultaneously citing flawed data to back up your own claim which doesn't even engage with the premise and then saying you have no alternatives. This is not helpful or productive. 

It is simple: with SIMD96, issuance will get much worse. We need some way of tying it back to reality instead of a fixed time curve that ignores everything about the chain except time. SOL has a higher $ amount staked than Ethereum already, even before SIMD96! We can't have users literally being forced to stake so that they don't have the value of their assets melt. Nor can any human provide controlled modeling in a non-linear complex system by looking at an 8-month data slice, and luckily, this is not necessary; incentives are how blockchains work.

kevinheavey:
if this is going to pass and become a reference document it would be nice to fix the formatting:
- missing newlines (Markdown requires two newlines to display a line break)
- Some lint that you can see here if you scroll down: https://github.com/solana-foundation/solana-improvement-documents/pull/228/files
- Whatever went wrong with the yaml at the start:

image

fplp782:
> 
> i) The authors clearly tell you how 0.5 was determined: https://github.com/solana-foundation/solana-improvement-documents/pull/228/files#diff-4c9f52adaf6cb3c34374dfbcdddbeae747adaad976b844aa00935e00a2576940R291

Clicked the link, still don't see it

>
> It is of course more market-based than the status quo since it takes into account MEV and commissions which lead into staking rate rather than simply ignoring all of it.

There is nothing to show that staking rate takes MEV or commissions into account

> 
> If you think looking at the staking rate is incorrect, then please provide robust reasoning as to why instead of citing data that is unrelated or using hypothetical anecdotes (like the one below). I am happy to engage if one makes an actual argument.

My goal is for my points to start this discussion. I proposed basing the model on data propagation or node count. Laine Stake has proposed basing it on priority fees, but without taking any burn as to disincentivize switching to tips. 

> 
> Then what are we doing here? You are saying you don't think a basic incentive change should be implemented without some mathematical proof while simultaneously citing flawed data to back up your own claim which doesn't even engage with the premise and then saying you have no alternatives. This is not helpful or productive.

I'm saying that this proposal is not good enough. By saying its not good enough, we're forced to create a better one. Instead of just settling on what is being proposed, take the feedback and let's improve it. This is how governance works

> 
> It is simple: with SIMD96, issuance will get much worse. We need some way of tying it back to reality instead of a fixed time curve that ignores everything about the chain except time. SOL has a higher $ amount staked than Ethereum already, even before SIMD96! We can't have users literally being forced to stake so that they don't have the value of their assets melt. Nor can any human provide controlled modeling in a non-linear complex system by looking at an 8-month data slice, and luckily, this is not necessary; incentives are how blockchains work.

I agree with this. My question in the article regarding why now has been answered so I better understand the urgency for a market-based solution. That being said this proposal fails to do that.

In the coming week, I'll dive into the source of our different perspectives - the purpose of inflation. I'll also focus more on alternatives as that seems to be the next step.

https://x.com/FPLParana/status/1880861764256141556

proposals/simd0221
kevinheavey: This paragraph is the only part of the proposal that discusses interest rates (real or nominal)

tjain-mcc:
> There is nothing to show that staking rate takes MEV or commissions into account

This is simple logic. People respond to incentives. If you don't think that is true, how do you design anything forward looking?

> My goal is for my points to start this discussion. I proposed basing the model on data propagation or node count. Laine Stake has proposed basing it on priority fees, but without taking any burn as to disincentivize switching to tips.

Neither of those ideas are viable because they are easy to game. You could see validators spinning up more or fewer nodes to change inflation. You could see validators paying themselves massive priority fees to game inflation. It is critical to pick something which cannot be gamed. Staking participation rate cannot be gamed.

> I'm saying that this proposal is not good enough. By saying its not good enough, we're forced to create a better one. Instead of just settling on what is being proposed, take the feedback and let's improve it. This is how governance works

Happy to hear some actual suggestions on what to change.

fplp782:
> 
> This is simple logic. People respond to incentives. If you don't think that is true, how do you design anything forward looking?

They also respond to opportunity cost. High priority fees and MEV exist during times of ample opportunity, where the opportunity cost of staking is high

Both of you and mert have mentioned that there are "hundreds of variables", not just yield, that impact the incentive to stake. That's my point tbh

> 
> Neither of those ideas are viable because they are easy to game. You could see validators spinning up more or fewer nodes to change inflation. You could see validators paying themselves massive priority fees to game inflation. It is critical to pick something which cannot be gamed. Staking participation rate cannot be gamed.

Why would validators spin up more nodes as that would reduce inflation? Why would validators pay themselves massive priority fees as that would also reduce inflation? Validators are not incentivized to do either of those things

> > Happy to hear some actual suggestions on what to change.

Sure! I made two suggestions above that I think should be considered. And as I mentioned, they're not gameable in the way you outlined. That being said, more feedback on them would be appreciated. Poke holes wherever you can please. It'll help me create a robust proposal. The feedback I've already gotten has helped me understand your argument better and where we differ. It's also helping me come up with an alternative I think might work for all of us

proposals/simd0221
kankanivishal1: @kevinheavey  not sure i exactly follow the commentary on real rates. If I stake my SOL, and get 6-8% without taking any risk, I would expect to get paid meaningfully more than that to a) experiment with new protocols, b) provide liquidity to DEXes or borrow/lend. Reducing the emissions rate will encourage people to look for riskier sources of yield, i.e. encourage people to participate in the DeFi economy atop Solana.  Would love to hear if you are seeing holes in this rationale?

kankanivishal1:
> I appreciate the depth of thought that has gone into the reasoning for this proposal, though I still wish it were possible (maybe it is, maybe not) to develop empirical data showing the impacts of the current and proposed inflation schedules.

we definitely thought about it. we can assume declining/increasing staking participation rates, different values of 'k', etc. but eventually decided that running numbers on this is mostly garbage in, garbage out scenario because of lack of empirical data. we did not want to give a false sense of precision.


> It lends itself to a gradual release of staked capital into DeFi with the equilibrium rate being around 7%, as nominal inflation drops more stake is released into the open market.

while this may happen, its not a guarantee. and i see reasons to accelerate the transition.

kankanivishal1:
> in the long-term simple inflationary yield and MEV alone will likely become insufficient to attract stake, or put anothe way, with dynamic inflation the inflation rate may need to become sufficiently high to remain economically attractive

as any economy matures and develops, the risk free rate generally lowers. if there is substantial economic activity happening on solana, i believe stakers would be happier to stake for meaningfully lower rates than today.

kankanivishal1:
> ###
> **[fplp782](/fplp782) ** commented [Jan 17, 2025](#issuecomment-2599102475)



> This proposal assumes that staking % is positively correlated with economic activity but provides no evidence for it. Tbh, I don't think it's positively correlated at all. Over the past year, Solana staking APRs have grown from 6.5% to almost 10%, and staking % has actually gone down (69% → 65%).

I address that concern and others in detail here
https://x.com/kankanivishal/status/1880816073282318734

proposals/simd0221
nihar-dz: My understanding is that inflation affects activities _differentially_ in Solana — stakers face zero inflation, non-stakers face actual inflation — and that is the problem. Every percentage point of inflation raises the appeal of staking automatically relative to other activities. If the proceeds of inflation were distributed evenly, i.e. the 5% inflation was distributed pro-rata to all SOL holders (stakers and non-stakers alike), then real yields would be all that matter (aside from the standard frictions associated with inflation.) But they aren't, and that raises the hurdle rate for non-staking DeFi activities to be viable.

kankanivishal1:
> I posted this article on twitter which outlines a few problems with this SIMD, alongside the point I made in here earlier regarding staking % as a poor metric to determine inflation.
> 
> https://x.com/FPLParana/status/1880334708225216931
> 
> To Mert's comment above, I don't view the current model as ideal and I do believe revisions need to be made. But this SIMD is not a sensible solution IMO

Thank you for that article. And I have shared clear responses to it here: https://x.com/kankanivishal/status/1880816073282318734 .

proposals/simd0221
kevinheavey: @kankanivishal1 for simplicity's sake let's assume stakers get a nominal yield of 6% paid for entirely by emissions and the expected annual decline in the SOL price from inflation is 6%.

You are only considering the lender's side. The borrower is happy to pay a higher nominal rate because inflation reduces the real rate that they pay.

If nominal rates did not move linearly with inflation expectations, there would be an arbitrage. You could get outsized returns by just borrowing/lending at rates that were made unfair by inflation

nihar-dz: @kevinheavey, if the decline in price was 6% because of aggregate inflation, and let's say 60% of SOL was staked, then the staking yield would be 6%/60% = 10%. So while a borrower would be happy to pay up to 6% (if real rates are zero), the non-staking lender would need a nominal yield of 10% to compensate for lending. This is the issue of inflation affecting stakers and non-stakers differentially.

kevinheavey: @nihar-dz USD inflation is not evenly distributed among USD holders, do you reject the distinction between real and nominal rates there too?

If inflation expectations go up 2% and nominal rates go up 2%, has the real hurdle rate changed?

bji: @nihar-dz It's not clear to me if you think the "issue of inflation affecting stakers and non-stakers differentially" is a problem or not.  But to be clear, it's not a problem.  Not staking SOL has an opportunity cost of not making stake rewards; but it gives the utility of allowing the SOL to be transferred or used for paying tx fees or funding accounts.   Staking and nonstaking yields cannot be the same, otherwise there would be no incentive to stake and Proof of Stake would not work.

nihar-dz: @bji I don't think it's a problem. It's simply the reason that actually inflation matters, in terms of moving value from non-staking applications to staking. To be clear, I am fully onboard that we want to subsidize staking to some degree, because it provides a shared public good of security.

kevinheavey: > if the decline in price was 6% because of aggregate inflation, and let's say 60% of SOL was staked, then the staking yield would be 6%/60% = 10%. So while a borrower would be happy to pay up to 6% (if real rates are zero), the non-staking lender would need a nominal yield of 10% to compensate for lending.

@nihar-dz the borrower would be happy to pay up to 10%, because he could stake his borrowed SOL. Anything less than that would be an arb for him

nihar-dz: @kevinheavey Actually I really like the case of USD inflation, because its proceeds are not distributed evenly. For most assets, indeed as inflation goes up, the real yield stays the same, the nominal yield rises, and nothing really happens (again, holding aside the frictions of inflation).

But suppose there was one magical asset out there that effectively faced zero inflation. That is, as inflation rose, the lender's yield would automatically rise. This would cause a market imbalance, and raise the nominal threshold that any other borrower would have to beat.

And such an asset does actually exist! It's the ability to buy TIPS at par value (not at the market price). Now, the US government limits purchases of TIPS at par value to $10,000 annually per person, so it isn't a serious threat to the rest of the financial system. But if that limit was removed, then you could easily imagine that any other borrower would severely struggle to raise funds in high inflation environments.

nihar-dz: > @nihar-dz the borrower would be happy to pay up to 10%, because he could stake his borrowed SOL. Anything less than that would be an arb for him

@kevinheavey But isn’t this precisely the problem? Anything that the borrower wants to do with his SOL that doesn’t involve staking would require a hurdle rate of at least 10%, in an environment when SOL is expected to depreciate by 6% only. So there are a bunch of projects with positive real returns that are made uneconomic, solely by the existence of (differential proceeds of) inflation.

kankanivishal1: @kevinheavey

1) i do not think you can compare usd inflation to solana emissions because usd inflation is not distributed to usd holders via airdrops if they parked the money in the savings accounts. also, buying treasury bills is a way to get access to some of it, but you are taking sovereign risk. in solana emissions, via staking, you are basically able to park your assets such that you'll be airdropped (helicopter money) new tokens proportionally.

2) I still don't understand the argument of real rates not changing and it not helping DeFi. When I can make nominal 8% without taking any risk, I want other DeFi protocols to give me more than nominal 8% for doing whatever they want me to do.

kevinheavey: I am trying to find the exact words to describe how staking yield doesn't make borrowing and lending worse, but does disincentivise holding spot SOL, as intended.

I think you are looking for examples of defi activities where frictions make it hard to lend out your SOL while using it for something else. For example, an order book protocol can put the SOL collateral in your open orders into a borrow/lend (Mango basically did this), but there are big costs to doing this so it's not a widespread practice.

I don't think there's anything special about the TIPS example really. Inflation protection is priced into regular treasuries, it's just a rougher estimate since it's forward-looking.

nihar-dz: @kevinheavey in looking at this thread, I think the detours into TIPS, sovereign risk, Mango, etc obscure the simple point.

If inflation is x (e.g. 6%) and that is given to fraction y of the SOL (i.e. the stakers, e.g. 60%), then the real rate for any project that wants to get financing is automatically x/y - x. (As a reminder, the lender could earn x/y by staking so they must earn at least that with the external project; and x is of course just inflation.) The higher the inflation value x, therefore, the more projects that are excluded.

Notice that this stops being a problem if, theoretically, y is 100%, which is why the subtlety about inflation having differential impacts is key.

kevinheavey: @nihar-dz unless we want to assume a chronic mispricing, in USD terms we have to assume that the real yield from holding staked SOL is just the USD risk-free rate plus some risk premium for holding SOL. So changing SOL emissions won't change the dollar yield from holding staked SOL.

And projects are financed in dollars, more or less

nihar-dz: I think there is a long history of uncovered interest parity, which is what you're referring to (i.e. future price of SOL/USD adjusting to offset SOL and USD nominal interest gaps so that real yields, net of risk premiums, equalize), not working out in practice.

kevinheavey: There are lots of times where arbitrage persists but you can't use that to tell whatever story you like. If you just assume that there is free money to be made staking SOL then we should raise emissions.

On a more practical note, most projects that I know of are directly financed in dollars or USDC

nihar-dz: More to the point, nothing in this story breaks down if it is financed in dollars.

> If inflation is x (e.g. 6%) and that is given to fraction y of the SOL (i.e. the stakers, e.g. 60%), then the real rate for any project that wants to get financing is automatically x/y - x. (As a reminder, the lender could earn x/y by staking so they must earn at least that with the external project; and x is of course just inflation.) The higher the inflation value x, therefore, the more projects that are excluded.

Inflation raises the real rate, in SOL terms, that a project must surpass to get financing. Even if you wave aside the real rate in SOL terms, it is quite the stretch to argue it is totally uncorrelated to the real rate in dollars. Unless you assume uncovered interest rate parity, which I believe is a bad assumption.

kevinheavey: Let's avoid confusion about the extraordinary claim here. Your model assumes that there is persistent alpha in holding staked SOL. That's going to need more support than "there are people who do FX carry trades"

nihar-dz: I don't get the sense you are amenable to being persuaded, which is fine. So we can maybe just agree to disagree and move on. You have in mind a Physics-esque model of international finance, where the laws of interest parity are immutable and hold perfectly. This differs strongly from empirical reality. In the spirit of good faith, I'll suggest one last article, which nicely assesses uncovered interest parity over two centuries and generally finds support for it: https://www.sciencedirect.com/science/article/abs/pii/S0261560611000155

But more generally, even if the answer is "sometimes it does hold, sometimes it doesn't" -- the point stands that Solana inflation drives up the SOL-based real rate, which is positively correlated if not perfectly with the dollar-based real rate on Solana. Which is the problem we are trying to fix here.

The only way to break this is (borrowing your phrase) to accept your extraordinary claim that the exchange rate always depreciates by exactly the right amount to offset deviations in the SOL- and USD-based real rate. This does not happen in currency markets nor has it happened in the last five years of the SOLUSD price, but you seem to believe it will happen now.

kevinheavey: > You have in mind a Physics-esque model of international finance, where the laws of interest parity are immutable and hold perfectly. This differs strongly from empirical reality. 

You are making a false dichotomy. The alternative to "all assets are correctly priced" is not "this asset is chronically mispriced in the direction I want". If we were talking about the degree to which assets become mispriced one way or another that would be one thing, but the extraordinary claim here is that there is a chronic *under*pricing of SOL staking yield.

> Solana inflation drives up the SOL-based real rate, which is positively correlated if not perfectly with the dollar-based real rate on Solana

That there is a correlation between these two does not mean `d(SOL-based real rate)/d(dollar-based real rate)` is non-zero. This argument is again just assuming an arbitrary and chronic mispricing. Why not assume a mispricing in the other direction?

kevinheavey: The other problem with the mispricing assumption is it would suggest that we can increase the excess returns to SOL staking by increasing emissions. If so, why would anyone here support lowering emissions, given we all know how to stake?

t-nelson:
> > This proposal assumes that staking % is positively correlated with economic activity but provides no evidence for it. Tbh, I don't think it's positively correlated at all. Over the past year, Solana staking APRs have grown from 6.5% to almost 10%, and staking % has actually gone down (69% → 65%).
> 
> Now talk to me about all the other variables that changed in the past year
> 
> This is also not the point of the proposal — it simply makes no sense that the issuance is a hardcoded value independent of anything on the chain but time

as an equally flippant reply, study chesterton's fence

mertimus:
> chesterton's fence

it's there arbitrarily from copy/pasting what existed on cosmos to minimize introducing new variables at launch from what toly told me, is there another reason?

michaelh-laine:
Thanks Tushar & Vishal for joining today's community call. To follow up on the conversation there, some comments:

- Regarding block reward sharing (#123 ) Tushar mentioned this being orthogonal, however one of the motivations in this SIMD is the incentive for capital to flow from the risk free rate of staking rewards into DeFi and thereby stimulate DeFi, if block rewards revenue increases the risk free rate sufficiently this incentive won't be met.
- 
- The formula offered for stake changes with a speed co-efficient of k=0.05 would result in a "shock decrease" of inflation if this were implemented today, where within 3-4 epochs we would see inflation drop from currently 4.7% to 1.5%, causing potential turmoil with mass unstaking (recall this is limited to 9% per epoch) and unbalanced destaking potentially adversely affecting some validators more than others. With a smaller delta from actual to target stake rate this co-efficient might be acceptable but I believe it is too extreme for an initial starting point where we are now, I would either suggest a reduction to k=0.01 or a reformulation of k as a function of the actual v target stake rate delta (d), e.g. (1/d/5), for a 20% delta (actual stake rate 70% v target 50%) this results in k=0.01, while with a delta of 5%  it would be k=0.04.

MariaEverstake:
Hello everyone, and thank you for putting forward this proposal. As a validator, we are committed to fostering positive changes and ensuring a brighter future for the Solana network.

We’d like to share some thoughts regarding the proposed changes:

- With a lower inflation rate, this will have minimal impact on stake decentralization, as large wallets will continue to claim the majority of the rewards. Currently, the vast majority (98%) of wallets stake amounts up to 500 SOL, collectively representing less than 5% of the total stake.

- Lowering emissions could have unintended consequences. Reduced rewards might incentivize users to move liquidity to ecosystems offering higher APYs. Moreover, the proposed 50% staking level target appears misaligned with Solana’s dynamics. Historical data suggests no consistent correlation between staking APR and the percentage of tokens staked. For example, over the past year, Solana’s staking APR increased, yet the staking percentage declined. This indicates that targeting 50% may not effectively balance rewards and network security.

We appreciate the efforts of teams dedicated to implementing the best mechanisms for Solana, but some aspects of this proposal warrant further consideration. Also, it would be helpful to have some modeling, as this proposal seems a bit uncertain without it. We value the opportunity to share our feedback!

MaxResnick:
I'm sympathetic to the idea of lowering issuance. My two main comments on this proposal:

1. I think the controller is probably not ideal for what the authors are trying to achieve. A static curve would be simpler to think about and less gameable. 
2. I don't think we should ever get to an inflation rate of 0 because issuance-based rewards incentivize voting and having some positive reward for voting is a good thing. If we were to set issuance close to 0 it might result in a deprioritization of validator reliability since a few missed votes here and there don't really matter. 

With those two concerns out of the way I want to emphasize that I very much agree with the spirit of the proposal, with the idea that issuance is currently higher than necessary, and with the idea that high issuance is not "just accounting" and actually imposes a concrete and quantifiable cost on the protocol due to extremely unfavorable tax treatment towards issuance in certain jurisdictions. 

**Issuance Is a Leaky Bucket**

The first point I want to make which I feel hasn't been discussed enough is that on its face, issuance should be thought of as an internal transfer of wealth from Solana holders to Solana stakers. This, in a frictionless world without taxes might be thought of as "just accounting" and indeed the math works out that without taxes issuance shouldn't matter that much (even for DeFi applications!). But in the real world we do have frictions and in particular we have an extremely unfavorable tax treatment for emissions. The problem arises because emissions are seen as taxable income by the US government rather than as capital gains. For US individuals this can be a difference of 37%-20% = 17% for individuals in the highest income bracket. The tax treatment is similarly unfavorable for US companies who, although they face the same tax rate for cap gains and income, cannot offset with losses on capital in the same way they can with capital gains. This means that for US companies, the inflation is causing expected capital losses (through emissions) which cannot be offset by the expected capital gains (vote rewards funded by emissions). 

You will notice in the above paragraph I have talked about the _difference_ in tax treatment between issuance and standard capital gains because in some sense emissions are an alternative to capital gains. This is because we are really trading off between these two tax treatments. If we reduce emissions by 1%, in expectation we should expect capital gains to increase by 1% so what really matters is the difference in tax treatment between these two not the absolute tax rate since we cannot avoid at least paying capital gains. 

**Proposed Alternative Controller**

The current controller resembles 1559 a lot. This means if staking participation is slightly below target for a prolonged period of time inflation will shoot up to current levels quickly. It is very rigid and is the type of controller you would want to implement if your utility function valued nearly 50% of stake on average. 

I would argue that this doesn't really make sense as an objective function. From the perspective of the network, is it really all that different to have 49% of the total supply staked than 50%? If the answer is no, it suggests we should use an alternative controller. 

With that in mind, and the fact that I don't think we should ever be at 0 emissions I think we should do something like the following: 
\begin{equation}
I(x) = r \left(1-\ \sqrt{x}\right)
\end{equation}
where $r$ is the current inflation rate and $x$ is the fraction of the total supply staked. 

This would yield a reward rate for validators with good performance of 
\begin{equation}
R(x) = \frac{r \left(1-\ \sqrt{x}\right)}{x}
\end{equation}

Screenshot 2025-01-27 at 10 14 56 PM

Over time we could let $r$ continue on its current trajectory resulting in the following reward curves over time for performant validators

Screenshot 2025-01-27 at 10 15 32 PM

**Setting the Right Parameters**
Of course there are plenty of variations of this so how should we choose the right parameters so that we aren't shooting in the dark?

The typical approach to these types of questions is to understand the tradeoffs. On the one hand, higher issuance gives us more security budget (because more SOL ends up staked) and gives us better incentives for operating a reliable validator (because vote rewards are higher) but it costs us money through the leaky bucket of less favorable tax treatment. The right way to make this decision then is to figure out how much we value security and vote reliability in $ terms and then estimate the tax treatment difference and from there we can simply solve a first order condition. I'm open to feedback about what our utility function should look like as a network but I think we should commit to not making issuance changes lightly without at least doing the work to understand our utility function otherwise we may be in the same position on issuance again in the future with a more ossified network that is more resistant to changes.

aeyakovenko:
Tend to agree with Max that a controller that's a continuous function of stake % is better than a hard switch at 50%. 

It's hard to tell what the minimum inflation should be.  Dumb fermi approximation would be the cost to insure gold or some other asset.  AFAIK that's roughly 0.5% or fifty basis points. Is worth considering that not all the inflation is going to validators.

tjain-mcc:
I agree with Max that a smooth controller is better than a hard switch at 50%. I discussed with him offline and he will submit a pull request to incorporate his idea.

michaelh-laine:
> This would yield a reward rate for validators with good performance of
> 
> ```
> R(x) =r (1-sqrt(x))/x
> ```

Thanks for this amendment Max, I greatly prefer this approach, with one caveat: The change is immediate. If implemented today the reward rate for validators would be approximately 1.4% if my maths is correct (4.77% * (1 - 0.8) / 0.64) - as with the previous proposal my main gripe here is the "shock" value of the change. 

I think a tempered rate of change would be better, where we use the above formula but limit the rate of change in the rewards rate to `r/10`, so if r=4.7 the rewards rate can change by a maximum of 0.47 per epoch, thus it would take approximately 7 epochs for the rate to drop from 4.7% to 1.4%

nihar-dz:
Is there a reason the formula is r(1 - sqrt(x)) rather than r(1 - x)? They achieve similar goals, but there are two salient differences:

1. The r(1 - x) controller is systematically more generous (e.g. today, given the same level of stake, it will offer stakers a yield of 2.5% rather than 1.4%). This is a double-edged sword, in that higher inflation has all the harms outlined above -- which I am deeply sympathetic to personally -- but it might have the practical advantage of getting a broader community onboard the proposal.
2. The linear controller is more sensitive to unstaking actions that could threaten the network's security. That is, when stake falls, the yield jumps more aggressively under the linear formulation than the square root one. For instance, if stake fall from 64% to 60%, the yield jumps by 47 basis points under r(1 - x) controller but only 28 basis points under the r(1 - sqrt(x)) controller. This is of course a fairly natural byproduct of the increased generosity, but it does have this nice security-based rationale too.

Here's a sketch of the two controllers compared for today's inflation level, over the 20% - 100% range of staking participation.
returns

mertimus:
> I'm sympathetic to the idea of lowering issuance. My two main comments on this proposal:
> 
> 1. I think the controller is probably not ideal for what the authors are trying to achieve. A static curve would be simpler to think about and less gameable.
> 2. I don't think we should ever get to an inflation rate of 0 because issuance-based rewards incentivize voting and having some positive reward for voting is a good thing. If we were to set issuance close to 0 it might result in a deprioritization of validator reliability since a few missed votes here and there don't really matter.
> 
> With those two concerns out of the way I want to emphasize that I very much agree with the spirit of the proposal, with the idea that issuance is currently higher than necessary, and with the idea that high issuance is not "just accounting" and actually imposes a concrete and quantifiable cost on the protocol due to extremely unfavorable tax treatment towards issuance in certain jurisdictions.
> 
> **Issuance Is a Leaky Bucket**
> 
> The first point I want to make which I feel hasn't been discussed enough is that on its face, issuance should be thought of as an internal transfer of wealth from Solana holders to Solana stakers. This, in a frictionless world without taxes might be thought of as "just accounting" and indeed the math works out that without taxes issuance shouldn't matter that much (even for DeFi applications!). But in the real world we do have frictions and in particular we have an extremely unfavorable tax treatment for emissions. The problem arises because emissions are seen as taxable income by the US government rather than as capital gains. For US individuals this can be a difference of 37%-20% = 17% for individuals in the highest income bracket. The tax treatment is similarly unfavorable for US companies who, although they face the same tax rate for cap gains and income, cannot offset with losses on capital in the same way they can with capital gains. This means that for US companies, the inflation is causing expected capital losses (through emissions) which cannot be offset by the expected capital gains (vote rewards funded by emissions).
> 
> You will notice in the above paragraph I have talked about the _difference_ in tax treatment between issuance and standard capital gains because in some sense emissions are an alternative to capital gains. This is because we are really trading off between these two tax treatments. If we reduce emissions by 1%, in expectation we should expect capital gains to increase by 1% so what really matters is the difference in tax treatment between these two not the absolute tax rate since we cannot avoid at least paying capital gains.
> 
> **Proposed Alternative Controller**
> 
> The current controller resembles 1559 a lot. This means if staking participation is slightly below target for a prolonged period of time inflation will shoot up to current levels quickly. It is very rigid and is the type of controller you would want to implement if your utility function valued nearly 50% of stake on average.
> 
> I would argue that this doesn't really make sense as an objective function. From the perspective of the network, is it really all that different to have 49% of the total supply staked than 50%? If the answer is no, it suggests we should use an alternative controller.
> 
> With that in mind, and the fact that I don't think we should ever be at 0 emissions I think we should do something like the following:
> 
> ```
> I(x) = r(1 - sqrt(x))
> ```
> 
> where `r` is the current inflation rate and `x` is the fraction of the total supply staked.
> 
> This would yield a reward rate for validators with good performance of
> 
> ```
> R(x) =r (1-sqrt(x))/x
> ```
> 
> Screenshot 2025-01-27 at 10 14 56 PM
> Over time we could let r continue on its current trajectory resulting in the following reward curves over time for performant validators
> 
> Screenshot 2025-01-27 at 10 15 32 PM
> **Setting the Right Parameters** Of course there are plenty of variations of this so how should we choose the right parameters so that we aren't shooting in the dark?
> 
> The typical approach to these types of questions is to understand the tradeoffs. On the one hand, higher issuance gives us more security budget (because more SOL ends up staked) and gives us better incentives for operating a reliable validator (because vote rewards are higher) but it costs us money through the leaky bucket of less favorable tax treatment. The right way to make this decision then is to figure out how much we value security and vote reliability in $ terms and then estimate the tax treatment difference and from there we can simply solve a first order condition. I'm open to feedback about what our utility function should look like as a network but I think we should commit to not making issuance changes lightly without at least doing the work to understand our utility function otherwise we may be in the same position on issuance again in the future with a more ossified network that is more resistant to changes.

I agree with ethereum double agent Max

cfl0ws:
> The typical approach to these types of questions is to understand the tradeoffs. On the one hand, higher issuance gives us more security budget (because more SOL ends up staked) and gives us better incentives for operating a reliable validator (because vote rewards are higher) but it costs us money through the leaky bucket of less favorable tax treatment. The right way to make this decision then is to figure out how much we value security and vote reliability in $ terms and then estimate the tax treatment difference and from there we can simply solve a first order condition. I'm open to feedback about what our utility function should look like as a network but I think we should commit to not making issuance changes lightly without at least doing the work to understand our utility function otherwise we may be in the same position on issuance again in the future with a more ossified network that is more resistant to changes.

@MaxResnick one of my first questions when reviewing this proposal was why the price of SOL wasn't incorporated into the calculation, as it seems like economic security is ultimately calculated in a dollar value. Since then, I understand a bit more about how the % of tokens staked can be a proxy to estimate economic security, i.e. the amount of stake that has voted to finalize a block. However, seeing you mention security and vote reliability in $ terms brings me back to this question, i.e. would it be more accurate to include current SOL price in the calculation to determine the desired inflation rate, rather than exclude it?

Benhawkins18:
@cfl0ws The core protocol does not and should not have any concept of price. It would be very problematic to try to introduce it here.

MaxResnick:
> Is there a reason the formula is r(1 - sqrt(x)) rather than r(1 - x)? They achieve similar goals, but there are two salient differences:
> 
> 1. The r(1 - x) controller is systematically more generous (e.g. today, given the same level of stake, it will offer stakers a yield of 2.5% rather than 1.4%). This is a double-edged sword, in that higher inflation has all the harms outlined above -- which I am deeply sympathetic to personally -- but it might have the practical advantage of getting a broader community onboard the proposal.
> 2. The linear controller is more sensitive to unstaking actions that could threaten the network's security. That is, when stake falls, the yield jumps more aggressively under the linear formulation than the square root one. For instance, if stake fall from 64% to 60%, the yield jumps by 47 basis points under r(1 - x) controller but only 28 basis points under the r(1 - sqrt(x)) controller. This is of course a fairly natural byproduct of the increased generosity, but it does have this nice security-based rationale too.
> 
> Here's a sketch of the two controllers compared for today's inflation level, over the 20% - 100% range of staking participation. returns

This is a great question. I think there are 2 points here. The first is about how generous the controller is and the other is about how responsive to changes in stake the amount of issuance is. On the generosity point, I think looking at the what the reward level at today's staking rate is not really the right comparison to make. In the updated proposal I have some analysis showing that the protocol rewards today would be achieved at around 30% of the network staked under the new curve. Depending on the shape of the demand curve we can expect that the new staking rate will settle somewhere between 30-65% I would expect stake rate to settle down somewhere around 45-50% particularly after accounting for the additional MEV rewards stakers will earn when other stakers exit. Again the goal is the minimum amount of issuance such that we maintain security and incentives for good behavior. 

On the second point which is about responsiveness. The thing to be careful about is "discouragement attacks" If issuance level is too responsive to changes in stake then it may be advantageous, particularly for large stakers to strategically censor stake transactions or make side deals with other stakers to get them to unstake. For this reason I think we should choose the less responive curve over the more responsive one.

nihar-dz:
> This is a great question. I think there are 2 points here. The first is about how generous the controller is and the other is about how responsive to changes in stake the amount of issuance is. On the generosity point, I think looking at the what the reward level at today's staking rate is not really the right comparison to make. In the updated proposal I have some analysis showing that the protocol rewards today would be achieved at around 30% of the network staked under the new curve. Depending on the shape of the demand curve we can expect that the new staking rate will settle somewhere between 30-65% I would expect stake rate to settle down somewhere around 45-50% particularly after accounting for the additional MEV rewards stakers will earn when other stakers exit. Again the goal is the minimum amount of issuance such that we maintain security and incentives for good behavior.
> 
> On the second point which is about responsiveness. The thing to be careful about is "discouragement attacks" If issuance level is too responsive to changes in stake then it may be advantageous, particularly for large stakers to strategically censor stake transactions or make side deals with other stakers to get them to unstake. For this reason I think we should choose the less responive curve over the more responsive one.

I agree on the first point that there is just some tradeoff between over-paying and under-paying for security — a tradeoff between protocol efficiency and validator community support. But the second point — on the potential attack — is much more interesting and I’ve dug a bit deeper. **TLDR is that I think the attack is not a practical concern; and so it is outweighed by the benefit of the linear controller being more responsive.**

Attached is a plot seeded with today’s level of stake (65%). It answers: for a given level of stake controlled by a single entity (x-axis), what percentage should be optimally deployed (y-axis), for the two types of proposals? We see that if a single entity has less than 20% directly or indirectly, they should just deploy 100% of it. The monopoly power they could potentially exert is outweighed by the loss of rewards. It is only if a single entity has over 20% should they begin to exert the monopoly power.

This condition — an adversary has more than 20% of SOL (total SOL, not staked SOL) — strikes me as highly unlikely on a network as decentralized as Solana. Put another way, the attack — while fascinating on a theoretical basis — can be likely ignored on a practical basis. So on net, I think the increased responsiveness of the r * (1 - x) controller over the r * (1 - sqrt(x)) controller is just a pure positive.

Fwiw, I actually think the attack is slightly more problematic on the square root controller over the linear controller, though I’m not entirely sure why. Here’s a quick gut check: suppose an entity has 22% of stake. Under the linear controller, (1 - 0.65) * 0.22/0.65 > (1 - 0.64) * (0.21/0.64), i.e. no attack; but under the square root controller, (1 - sqrt(0.65)) * 0.22/0.65 < (1 - sqrt(0.64)) * (0.21/0.64), i.e. an attack is possible.

plot

MaxResnick:
> > This is a great question. I think there are 2 points here. The first is about how generous the controller is and the other is about how responsive to changes in stake the amount of issuance is. On the generosity point, I think looking at the what the reward level at today's staking rate is not really the right comparison to make. In the updated proposal I have some analysis showing that the protocol rewards today would be achieved at around 30% of the network staked under the new curve. Depending on the shape of the demand curve we can expect that the new staking rate will settle somewhere between 30-65% I would expect stake rate to settle down somewhere around 45-50% particularly after accounting for the additional MEV rewards stakers will earn when other stakers exit. Again the goal is the minimum amount of issuance such that we maintain security and incentives for good behavior.
> > On the second point which is about responsiveness. The thing to be careful about is "discouragement attacks" If issuance level is too responsive to changes in stake then it may be advantageous, particularly for large stakers to strategically censor stake transactions or make side deals with other stakers to get them to unstake. For this reason I think we should choose the less responive curve over the more responsive one.
> 
> I agree on the first point that there is just some tradeoff between over-paying and under-paying for security — a tradeoff between protocol efficiency and validator community support. But the second point — on the potential attack — is much more interesting and I’ve dug a bit deeper. **TLDR is that I think the attack is not a practical concern; and so it is outweighed by the benefit of the linear controller being more responsive.**
> 
> Attached is a plot seeded with today’s level of stake (65%). It answers: for a given level of stake controlled by a single entity (x-axis), what percentage should be optimally deployed (y-axis), for the two types of proposals? We see that if a single entity has less than 20% directly or indirectly, they should just deploy 100% of it. The monopoly power they could potentially exert is outweighed by the loss of rewards. It is only if a single entity has over 20% should they begin to exert the monopoly power.
> 
> This condition — an adversary has more than 20% of SOL (total SOL, not staked SOL) — strikes me as highly unlikely on a network as decentralized as Solana. Put another way, the attack — while fascinating on a theoretical basis — can be likely ignored on a practical basis. So on net, I think the increased responsiveness of the r * (1 - x) controller over the r * (1 - sqrt(x)) controller is just a pure positive.
> 
> Fwiw, I actually think the attack is slightly more problematic on the square root controller over the linear controller, though I’m not entirely sure why. Here’s a quick gut check: suppose an entity has 22% of stake. Under the linear controller, (1 - 0.65) * 0.22/0.65 > (1 - 0.64) * (0.21/0.64), i.e. no attack; but under the square root controller, (1 - sqrt(0.65)) * 0.22/0.65 < (1 - sqrt(0.64)) * (0.21/0.64), i.e. an attack is possible.
> 
> plot

The derivative of $1-\sqrt(x)$ is $-1/2sqrt(x)$ while the derivative of $1-\sqrt(x)$ is -1. This means that when stake rate $x> .25$ meaning more than  1/4 of the sol is staked, the marginal disutility of adding stake to existing stakers is lower under the square root controller than the linear controller. 

Discouragement attacks need not take the form of explicit collusion. you can also choose not to include new stake messages in your blocks as the leader.

nihar-dz:
> Discouragement attacks need not take the form of explicit collusion. you can also choose not to include new stake messages in your blocks as the leader.

But this will always be true, right? Regardless of the controller, any leader will want to discourage the competition by censoring blocks that include new stake messages (including today, when the inflation rate is fixed). I agree this attack vector theoretically exists, but I don't get the sense the controller is the linchpin to this issue.

> The derivative of 1 − x is − 1 2 x while the derivative of 1 − x is -1. This means that when stake rate x > .25 meaning more than 1/4 of the sol is staked, the marginal disutility of adding stake to existing stakers is lower under the square root controller than the linear controller.

That tells you the system-wide disincentive, but isn't the correct way to think about it from the perspective of an individual staker who imposes a negative externality on the system? The staker stakes more, their marginal stake takes some of those rewards from others, and their marginal stake lowers the aggregate rewards too (of which they bear some of those losses and the other staker bears some of those losses). And that's where you need a single staker today to hold more than 20% of the SOL supply for the first channel to be dominated by the second channel, which seems unrealistic today.

brianlong:
**Devil's Advocate...**
> the network is now overpaying [for] security
"Overpaying" is subjective. Also, the emissions rate is trending down. If you don't like the high rate today, wait for an epoch, and emissions will be lower.

> earning SOL from other sources (e.g. MEV)
This is a massive assumption about out-of-protocol revenue. Jito's existence is not guaranteed. Why should out-of-protocol factors influence inflation rates?

> Reducing inflation spurs SOL usage in DeFi
Not really. This assumption is unproven. Stake moves around, but that doesn't mean it moves into DeFi. Most of my stakeholders, myself included, are passive and too busy to trade defi seriously. Furthermore, SOL lending rates on Drift, for example, are currently below staking returns, so their app has a sufficient supply of SOL. Free markets have already solved this problem. It doesn't need to be solved again.

> Minimum Necessary Amount (MNA) to secure the network.
MNA seems important. I don't see much discussion above about what people think might be a reasonable MNA. How will the market determine MNA with the new formula?

> New Emission Rate Formula
The existing formula remains and represents the ceiling for the new modified formula. Finding MNA will be difficult if emissions are only down.

**That Being Said...**
Thanks for your proposal! I'm happy you all took the time to think about this -- it got me thinking about the failure case for security at the lower bounds.

Imagine years from now if the reference emissions rate is 1.5% and the global risk-free rate is high. Or imagine a reduction in out-of-protocol revenue, with low emission rates not enough to incentivize & retain active stake. I see a scenario where the percentage of SOL staked is much lower. So, my concern is the opposite of your primary concern: I'm worried about underpaying for security.

If the future market-clearing rate is 7.5% yield:
- The current curve suggests about 20% will stake when emissions are 1.5%
- The modified curve suggests about 12.5% staked with emissions of about 0.97% (using the 1.5% long-term value of `r` in the proposal).

:shrug: Which feels more secure?

As I noted above, many stakeholders are passive and not especially sensitive to inflation rewards. Validators also share other sources of revenue, so the total active stake may never get as low as 12.5%, but the fact that it could is troubling.

MNA matters at the lower bound. I don't see a mechanism for the market to determine MNA.

I like the idea that inflation should be dynamic. The inflation rate should go **up** as we approach MNA. Going down only will lead to failure at the lower bound.

I like the direction of this proposal, but it needs more work to protect the protocol at the lower bounds.

MaxResnick:
> Imagine years from now if the reference emissions rate is 1.5% and the global risk-free rate is high. Or imagine a reduction in out-of-protocol revenue, with low emission rates not enough to incentivize & retain active stake. I see a scenario where the percentage of SOL staked is much lower. So, my concern is the opposite of your primary concern: I'm worried about underpaying for security.
> 
> If the future market-clearing rate is 7.5% yield:
> 
> * The current curve suggests about 20% will stake when emissions are 1.5%
> * The modified curve suggests about 12.5% staked with emissions of about 0.97% (using the 1.5% long-term value of `r` in the proposal).
> 
> 🤷 Which feels more secure?
> 
> As I noted above, many stakeholders are passive and not especially sensitive to inflation rewards. Validators also share other sources of revenue, so the total active stake may never get as low as 12.5%, but the fact that it could is troubling.
> 
> I like the idea that inflation should be dynamic. The inflation rate should go **up** as we approach MNA. Going down only will lead to failure.
> 
> I like the direction of this proposal, but it needs more work to protect the protocol at the lower bounds.

Appreciate these comments Brian. I have to say I very much disagree with the idea that Solana staking rate is at all effected by the global risk free rate. **Solana staking rate is based on the outside option of Solana returns not global returns of USD or any other currency.** I suppose the intuitive story that leads to this kind of reasoning is that if global rates are high people will choose to earn yield in other currencies but there is a conservation of matter property here. If I sell SOL to go earn high USD yields then somebody must have bought the Sol from me. Then the buyer of SOL must also choose whether to stake or not. 

Classical asset pricing theory would suggest that a high risk free rate on USD relative to the risk free return on SOL (both in real terms) would result in USD appreciating against SOL but not a reduction in the staking rate of SOL. 

Now if we replace your concern with an outside rate of Sol at 7.5% per annum we are now in a closed system. It is unlikely that any opportunity on SOL can support 50% of total SOL supply without diluting rewards substantially.

brianlong:
The **why** was just an example; there could be other reasons. Maybe defi lending rates are high enough to tempt my passive friends and me off the sidelines. Instead, think about outcomes.

The proposal starts with the assertion we are overpaying for security, implying the network is secure but expensive. I'm exploring the other end of the spectrum. Under what circumstances could we underpay for security? Implying the network will not be secure. A good solution must consider the lower bounds.

Assuming SOL moves more freely, the low long-term inflation rates in both the original and modified curves could be problematic.

The emissions are supposed to compensate stakeholders for keeping the network secure. What if 1.5% inflation isn't enough? Should emissions go up as active stake percentage goes down?

cfl0ws:
Is the author going to start the formal vote process for this SIMD @tjain-mcc?

I think that's what @michaelh-laine suggested was the next step during last week's foundation validator call, however I don't see a post announcing the vote here https://forum.solana.com/c/gov/11

As far as I'm aware this is still the recommended process and timeline to follow https://sg.laine.one/exploration/how#lifecycle-and-timeline

michaelh-laine:
> Is the author going to start the formal vote process for this SIMD @tjain-mcc?
> 
> 
> 
> I think that's what @michaelh-laine suggested was the next step during last week's foundation validator call, however I don't see a post announcing the vote here https://forum.solana.com/c/gov/11
> 
> 
> 
> As far as I'm aware this is still the recommended process and timeline to follow https://sg.laine.one/exploration/how#lifecycle-and-timeline

Yes this was delayed a bit due to some changes to account for downside risk on the proposed curve. We anticipate the proposal being published early next week now

dimandotsol:
Motivation is very weak. There are questions about many statements.

I see other issues that concern me more, such as the percentage inflation rate.

You start the motivation by mentioning MEV payouts, which are not part of the protocol at all. Validators themselves are customers/clients in this case, meaning they depend on a commercial project rather than the protocol, or they can act independently (i.e., validators extract MEV themselves, and it is distributed to stakers without intermediaries), or there is real competition among MEV proxy providers in the market. For me, the way MEV currently works on Solana is already a huge problem. So making protocol changes based on the MEV issue seems like the wrong move to me. The MEV issue needs to be solved first.

And so on. Brian mentioned several points above and did not receive answers to them. I won’t repeat.

Also, for example, I see a completely opposite conclusion from the four points in the proposal.

To me, it is obvious that reducing the APY of native staking will lead to an outflow of small stakers from native staking. This means that the share of large stakers in active staking will increase. And active native staking controls the PoS network. So the network control by large stakers will grow.

Technically, finding the balance point involves delegating and undelegating. But the problem is that fees in liquid pools and skipping one epoch in native staking make this process less active. Due to restaking (not to be confused with the currently popular marketing term), up to a month of rewards can be lost. All of this combined will cause small stakes to shift under the control of liquid pools with additional yield.

So, I see why this is bad for the Solana network.

My conclusion: perhaps this proposal is ahead of its time.

autocatalytic:
I like how the proposal forces me to think and work harder to build something valuable for Solana. Also thanks for slowing the rollout to 50 epochs, and to Ben for building the inflation curve/formula vercel visualizer. 

As a small validator I'm likely voting against my individual self-interest for what I think is the good of the network, as continuing to pay large and historically poor performing CEX's (and more importantly, future bank custodians) lots of SOL really grinds my gears. 

Hope I don't get fired as a validator after finally finding a way into the community (via SFDP), but IMO, there are too many moving pieces to predict with certainty how this plays out long-term.

dwdwone (approved):

dwdwone:
What's not to like?

PR #230, opened 2025-01-21
Amend SIMD-0175: Exclude rent_epoch field changes

jstarry:
- No longer require changes to the `rent_epoch` field. Simply disabling partitioned rent is enough.
- Add feature issue

### DIFF
--- a/proposals/0175-disable-partitioned-rent-collection.md
+++ b/proposals/0175-disable-partitioned-rent-collection.md
@@ -6,7 +6,7 @@ category: Standard
 type: Core
 status: Accepted
 created: 2024-09-25
-feature: (fill in with feature tracking issues once accepted)
+feature: https://github.com/anza-xyz/agave/issues/4562
 ---
 
 ## Summary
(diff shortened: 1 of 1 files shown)


jstarry:
Once rent collection is disabled, the runtime won't care what value the `rent_epoch` is actually set to. Therefore there's no need to attempt to set the `rent_epoch` field of all on-chain accounts to a certain value. It's still important that all validators set this field consistently to avoid consensus divergence, but this field doesn't necessarily have to be set to `u64::MAX` as we were previously intending to do.

HaoranYi (approved):
lgtm.

topointon-jump (approved):

jstarry:
Sorry for the thrashing @topointon-jump but we might go with a slightly different approach which will allow us to clean up more rent collection code from the runtime: https://github.com/solana-foundation/solana-improvement-documents/pull/231

HaoranYi (approved):

jstarry:
@Benhawkins18 we have consensus here

Benhawkins18 (approved):

PR #231, opened 2025-01-23
SIMD-0231: Disable account rent epoch field updates

jstarry:


HaoranYi:
> After this proposal has been applied and activated, new accounts will be
> distinguishable from old accounts because they will have a different rent epoch
> field. This will be observable both onchain and offchain until the rent epoch
> field is fully deprecated in a future proposal.

I had a little concern about this. Does this mean "all" new accounts created after the feature will have rent_epoch = 0? 
Old account will still have rent_epoch = max even if it written again?

topointon-jump (approved):

jstarry:
>> After this proposal has been applied and activated, new accounts will be
distinguishable from old accounts because they will have a different rent epoch
field. This will be observable both onchain and offchain until the rent epoch
field is fully deprecated in a future proposal.

> I had a little concern about this. Does this mean "all" new accounts created after the feature activated will have rent_epoch = 0?
Old account will still have rent_epoch = max even if it written again?

I'm personally not too concerned with this but I'm fine with being more conservative here. Let's stick with https://github.com/solana-foundation/solana-improvement-documents/pull/230 as a simple approach to disable partitioned rent collection. Before making the proposed change in this SIMD, we can introduce a new SIMD which completely ignores the stored `rent_epoch` in accounts db and overwrites with `u64::MAX`.

PR #234, opened 2025-01-24
Fix header linter

jstarry:
The header linter doesn't go through all the proposal headers when checking that the required headers are present. So if a proposal adds a custom header in the middle of the required headers, the linter will fail: https://github.com/solana-foundation/solana-improvement-documents/actions/runs/12943366763/job/36102577418


Benhawkins18 (approved):
Thanks for the fix

PR #236, opened 2025-01-29
update SIMD status and feature gate key for SIMD-0170

tao-stones:
SIMD-170 has been implemented. Update its status, and attach its feature gate key


Benhawkins18 (commented):
Thank Tao :)

Benhawkins18 (approved):

PR #237, opened 2025-01-29
update SIMD status and feature gate key

tao-stones:
SIMD-181 has been implemented. Update its status, and attach its feature gate key


Benhawkins18 (approved):
Thanks for updating :)

PR #246, opened 2025-02-13
SIMD-0096 is now activated

tigarcia:
Changed simd96 implemented -> activated


Benhawkins18 (approved):
Thanks for putting up the PR

PR #248, opened 2025-02-17
SIMD-0248: TPU feedback

lijunwangs:
This document describes the protocol for sending and receiving TPU feedback in solana network.

### DIFF
--- a/proposals/0248-tpu-feedback.md
+++ b/proposals/0248-tpu-feedback.md
@@ -7,7 +7,7 @@ category: Standard
 type: Core
 status: Review
 created: 2025-02-14
-development: 
+development:
   - Anza - WIP
   - Firedancer - Not started
 ---
@@ -62,7 +62,7 @@ datagrams to clients which are not interested in receiving feedback or clients
 which are not upgraded to have such capabilities.
 
 The client and server must enable the support of datagrams QUIC frame types
-via transport parameter `max_datagram_frame_size`. 
+via transport parameter `max_datagram_frame_size`.
 
 The server sends the feedback datagrams to interested clients periodically.
 The frequency is decided by the server and consideration should be given to
--- a/proposals/0248-tpu-feedback.md
+++ b/proposals/0248-tpu-feedback.md
@@ -32,7 +32,7 @@ transactions and help reduce congestion condition in the network.
 ## Alternatives Considered
 
 There are different transport mechanism to send the feedback such as
-bidrectional streams, reverse unistream from the server to the client and
+bidirectional streams, reverse unistream from the server to the client and
 QUIC datagrams. We choose datagrams due to its simplicity. One reason we have
 used stream over datagrams is the expectation that transaction size can be
 increased in the future which can not be carried in one datagram. Sending
--- a/proposals/0248-tpu-feedback.md
+++ b/proposals/0248-tpu-feedback.md
@@ -1,5 +1,5 @@
 ---
-simd: '0224'
+simd: '0248'
 title: TPU Feedback
 authors:
   - Lijun Wang <lijun@anza.xyz>
@@ -27,7 +27,14 @@ are too many transactions with higher paying fees than this transaction, the
 transaction is not included in the block. Providing further information on the
 transaction state will help the client to make more informed decision on
 handling transaction failures and reduce the incentive to blindly resubmitting
-transactions and help reduce congestion condition in the network.
+transactions and help reduce congestion condition in the network. The QUIC
+client library can be enhanced to use this information to not send transactions
+with fees too low.
+
+In addition, on the server side, the TPU server could use the feedback
+information to more accurately perform QOS. For example, if a peer is
+persistently submitting high percentage pf transactions which are rejected,
+its permitted throughput might be reduced.
 
 ## Alternatives Considered
 
@@ -74,16 +81,16 @@ can be sent without waiting for the feedback interval.
 ### The format of the TPU feedback
 
 struct TpuFeedback {
-  version: u32,
-  timestamp: u32,
+  version: u8,
+  timestamp: u64,
   transaction_state: TransactionState,
   priority_fee_info: PriorityFeeInfo
 }
 
-The version is a 32 bit unsinged integer, it is set to 1 for first version. It
+The version is a 8 bit unsinged integer, it is set to 1 for first version. It
 is designed to allow the protocol to extend for future extension.
 
-The timestamp is the 32 bit unsigned integer representing the number of seconds
+The timestamp is the 64 bit unsigned integer representing the number of seconds
 since midnight 01/01/1970.
 
 TransactionState is defined as the following:
@@ -105,6 +112,7 @@ enum TransactionStateValue {
   Packed = 0,
   FeeTooLow = 1,
   FeePayerBalanceTooLow = 2,
+  NotPackedDueToUnspecifiedReason = 2^32 -1,
 }
 
 Newer state value can be defined as needed in the future with its own unique
--- a/proposals/0248-tpu-feedback.md
+++ b/proposals/0248-tpu-feedback.md
@@ -88,7 +88,10 @@ struct TpuFeedback {
 }
 
 The version is a 8 bit unsinged integer, it is set to 1 for first version. It
-is designed to allow the protocol to extend for future extension.
+is designed to allow the protocol to support for future extension. This should
+have enough space for the versions in the relatively long future. It could also
+be wrapped around in the future to differentiate as we do not expect to support
+many current different versions.
 
 The timestamp is the 64 bit unsigned integer representing the number of seconds
 since midnight 01/01/1970.
@@ -128,7 +131,7 @@ struct PriorityFeeInfo {
   }
 }
 
-These ratio are represented as f64 numbers using 8 bytes.
+These ratio are represented as u64 numbers using 8 bytes.
 
 ## Impact
 
--- a/proposals/0248-tpu-feedback.md
+++ b/proposals/0248-tpu-feedback.md
@@ -68,7 +68,7 @@ The server sends the feedback datagrams to interested clients periodically.
 The frequency is decided by the server and consideration should be given to
 coalesce feedback on more transactions to reduce datagram counts and short
 enough feedback latency. For example 100ms can be considered. If there are
-more transactions to be acknowldged than the one in one datagram, more datagrams
+more transactions to be acknowledged in one datagram, more datagrams
 can be sent without waiting for the feedback interval.
 
 ### The format of the TPU feedback
--- a/proposals/0224-tpu-feedback.md
+++ b/proposals/0224-tpu-feedback.md
@@ -1,0 +1,128 @@
+---
+simd: '0195'
+title: TPU Feedback
+authors:
+  - Lijun Wang <lijun@anza.xyz>
+category: Standard
+type: Core
+status: Review
+created: 2025-02-14
+development: 
+  - Anza - WIP
+  - Firedancer - Not started
+---
+
+## Summary
+
+This document defines the protocol to send feedback on TPU processing from the
+servers to the clients.
+
+## Motivation
+
+A client sends TPU transactions to the leader as QUIC uni-streams. Currently we
+only send transport level acknowledememt to the clients. I.e we only tell if a
+transaction has been received by the server. A transaction received by the
+server may not be processed due to resource contention. For example, if there
+are too many transactions with higher paying fees than this transaction, the
+transaction is not included in the block. Providing further information on the
+transaction state will help the client to make more informed decision on
+handling transaction failures and reduce the incentive to blindly resubmitting
+transactions and help reduce congestion condition in the network.
+
+## Alternatives Considered
+
+There are different transport mechanism to send the feedback such as
+bidrectional streams, reverse unistream from the server to the client and
+QUIC datagrams. We choose datagrams due to its simplicity. One reason we have
+used stream over datagrams is the expectation that transaction size can be
+increased in the future which can not be carried in one datagram. Sending
+feedback has no such concern as we can send multiple datagrams if there are
+too many transactions to acknowledge.
+
+## New Terminology
+
+In this document we define the following,
+
+TPU Feedback -- information sent by the server about transaction processing.
+Server -- the validator receiving the TPU transactions and sending the feedback.
+Client -- the validator or application using QUIC TPU client sending the TPU
+          transactions and receiving the feedback.
+
+## Detailed Design
+
+When the client sends transactions to the server, it must first open a QUIC
+connection to the server. The feedback is sent by the server using the same
+connection. The feedback is sent using a QUIC datagram. The client wishing to
+receive feedback from the server must be ready to receive application datagrams
+from its connections.
+
+A client connection notifies the server of its interest to receive feedback by
+sending a datagram soliciting feedback. This helps server to avoid sending
+datagrams to clients which are not interested in receiving feedback or clients
+which are not upgraded to have such capabilities.
+
+The client and server must enable the support of datagrams QUIC frame types
+via transport parameter `max_datagram_frame_size`. 
+
+The server sends the feedback datagrams to interested clients periodically.
+The frequency is decided by the server and consideration should be given to
+coalesce feedback on more transactions to reduce datagram counts and short
+enough feedback latency. For example 100ms can be considered. If there are
+more transactions to be acknowldged than the one in one datagram, more datagrams
+can be sent without waiting for the feedback interval.
+
+### The format of the TPU feedback
+
+struct TpuFeedback {
+  version: u32,
+  timestamp: u32,
+  transaction_state: TransactionState,
+  priority_fee_info: PriorityFeeInfo
+}
+
+The version is a 32 bit unsinged integer, it is set to 1 for first version.
+The timestamp is the 32 bit unsigned integer representing the number of seconds
+since midnight 01/01/1970.
+
+TransactionState is defined as the following:
+
+struct TransactionState {
+  transactions_count: u8
+  (transaction, state) pairs
+}
+
+The transactions_count is 8 bit unsgined integer representing the count of
+transactions in this feedback. It is followed by the count of
+(transaction, state) pairs. Each transaction is represented by the first 8 bytes
+of the transaction signature and each state is represented by an 32 bit unsigned
+integer.
+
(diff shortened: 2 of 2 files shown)


proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,178 @@
+---
+simd: '0224'
```
Benhawkins18, line 2: Please update the SIMD # to 248

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,178 @@
...
+
+A client sends TPU transactions to the leader as QUIC uni-streams. Currently we
+only send transport level acknowledememt to the clients. I.e we only tell if a
+transaction has been received by the server. A transaction received by the
+server may not be processed due to resource contention. For example, if there
+are too many transactions with higher paying fees than this transaction, the
+transaction is not included in the block. Providing further information on the
+transaction state will help the client to make more informed decision on
+handling transaction failures and reduce the incentive to blindly resubmitting
+transactions and help reduce congestion condition in the network.
```
tao-stones, line 30: > reduce the incentive to blindly resubmitting transactions and help reduce congestion condition in the network

imo, by providing feedbacks itself does not necessarily disincentivize spamming; but Server could use `TransactionState` to throttle connections with excessively high failure rate, or use `PriorityFeeInfo` to drop transactions have priority below `min`. Are these reasonable use cases for feedbacks for this SIMD?

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,178 @@
...
+can be sent without waiting for the feedback interval.
+
+### The format of the TPU feedback
+
+struct TpuFeedback {
+  version: u32,
+  timestamp: u32,
+  transaction_state: TransactionState,
+  priority_fee_info: PriorityFeeInfo
+}
```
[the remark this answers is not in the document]
lijunwangs: This allow the same datagram to be able to carry both type of information or priority_fee_info -- so we could reduce datagram count. I think priority_fee_info is always good to have to understand better why it failed. In the case only priority_fee_info is transmitted, the TransactionState's transaction count is set to 0.

If we have a separate message -- we would need another flag field to indicate the message type.

proposals/0248-tpu-feedback.md
lijunwangs: The client can use the information to 

1. modify the retry behavior of transactions
2. proactively not sending transactions which are seriously below the water mark
3. percolate the information and set higher fee for transactions to maximize the chance to be included.

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,178 @@
...
+
+The version is a 32 bit unsinged integer, it is set to 1 for first version. It
+is designed to allow the protocol to extend for future extension.
```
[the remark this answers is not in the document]
lijunwangs: I think for foreseeable future u8 should be enough. If we can devise sub version extensions in the future.

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,178 @@
...
+
+The timestamp is the 32 bit unsigned integer representing the number of seconds
+since midnight 01/01/1970.
```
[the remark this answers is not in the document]
lijunwangs: Changed to 64.

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,178 @@
...
+(transaction, state) pairs. Each transaction is represented by the first 8 bytes
+of the transaction signature and each state is represented by an 32 bit unsigned
+integer.
+
+The transaction state value is defined as the below enum:
+
+enum TransactionStateValue {
+  Packed = 0,
+  FeeTooLow = 1,
+  FeePayerBalanceTooLow = 2,
```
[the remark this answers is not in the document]
lijunwangs: Great idea

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,178 @@
...
+
+struct PriorityFeeInfo {
+  priority_fee_stats: {
+    min_priority_fee/compute unit ratio
+    median_priority_fee/compute unit ratio
+    max_priority_fee/compute unit ratio
+  }
+}
+
+These ratio are represented as f64 numbers using 8 bytes.
```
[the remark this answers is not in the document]
lijunwangs: What is the value we use at the scheduler? u64? If so yes.

proposals/0248-tpu-feedback.md
lijunwangs: will do

proposals/0248-tpu-feedback.md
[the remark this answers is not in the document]
tao-stones: Scheduler uses `u64` for priority [here](https://github.com/anza-xyz/agave/blob/master/core/src/banking_stage/transaction_scheduler/receive_and_buffer.rs#L627-L647).

proposals/0248-tpu-feedback.md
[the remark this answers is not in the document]
tao-stones: I see your intention to reduce datagram count. 

Per our chat offline, it's possible to add 3rd type message to indicate the lowest tx priority Scheduler is accepting. Adding a `message_type: u8` to header might not be bad idea; at least different type messages don't have to always go out in same datagram, especially some messages don't have new updates.

> priority_fee_info is always good to have to understand better why it failed.

If transaction failed due to prio fee, then the reason should be clear by `TransactionStateValue::FeeTooLow`. With `priority_fee_info`, user would know "how low".

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,186 @@
...
+When the client sends transactions to the server, it must first open a QUIC
+connection to the server. The feedback is sent by the server using the same
+connection. The feedback is sent using a QUIC datagram. The client wishing to
+receive feedback from the server must be ready to receive application datagrams
+from its connections.
+
+A client connection notifies the server of its interest to receive feedback by
+sending a datagram soliciting feedback. This helps server to avoid sending
+datagrams to clients which are not interested in receiving feedback or clients
+which are not upgraded to have such capabilities.
```
alexpyattaev, line 69: I think it is best to assume we ask for feedback in the transaction request itself. Sending the feedback request in a separate datagram is adding statful behavior where it is not needed.

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,186 @@
...
+
+The client and server must enable the support of datagrams QUIC frame types
+via transport parameter `max_datagram_frame_size`.
+
+The server sends the feedback datagrams to interested clients periodically.
+The frequency is decided by the server and consideration should be given to
+coalesce feedback on more transactions to reduce datagram counts and short
+enough feedback latency. For example 100ms can be considered. If there are
```
alexpyattaev, line 77: I would suggest coalesce based on slot number. This way all the transactions that have landed in the block would be confirmed at once, not 100ms after the block is completed.

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,186 @@
...
+since midnight 01/01/1970.
+
+TransactionState is defined as the following:
+
+struct TransactionState {
+  transactions_count: u8
+  (transaction, state) pairs
+}
+
+The transactions_count is 8 bit unsgined integer representing the count of
```
alexpyattaev, line 103: Suggest we use TLV encoding here rather than a fixed format. Typically, TLV encodings provide far better ability for upgrade. Also with TLV we do not need to include total count at all.

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,186 @@
...
+per 100ms the network cost is:
+
+2500*1232*10 bytes /s ~= 30 MB/s
+
+For sending only priority info to idle connections, the size of the feedback
+datagram is smaller -- 33 bytes.
+
+The cost is:
+
+2500*21*10 bytes /s = 0.8 MB
```
alexpyattaev, line 147: you are not counting the headers. IP + UDP+QUIC are already over 30 bytes, so you get some 12 Mbps.

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,186 @@
...
+
+As not all connections are actively sending transactions, optimization can be
+done to reduce the cost to send feedback to clients only having sent
+transactions in the last report period.
```
alexpyattaev, line 151: Validators should be able to charge fees for this service I think.

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,186 @@
...
+
+A client connection notifies the server of its interest to receive feedback by
+sending a datagram soliciting feedback. This helps server to avoid sending
+datagrams to clients which are not interested in receiving feedback or clients
+which are not upgraded to have such capabilities.
+
+The client and server must enable the support of datagrams QUIC frame types
+via transport parameter `max_datagram_frame_size`.
+
+The server sends the feedback datagrams to interested clients periodically.
```
alexpyattaev, line 74: I believe this info should only be provided if a client has submitted a valid transaction in the last 3-4 blocks. Keeping this active for ALL transactions sounds like overkill.

proposals/0248-tpu-feedback.md
lijunwangs: There is a problem with that -- the transaction format need to be modified  and changes that protocol-- I think we run out of free flags.  I think this is a way to show the interest with minimal performance and protocol impact.

lijunwangs:
Tagging @apfitzge for comments on this SIMD.

proposals/0248-tpu-feedback.md
alexpyattaev: do we not have some legacy variants we could reuse?

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,189 @@
...
+server may not be processed due to resource contention. For example, if there
+are too many transactions with higher paying fees than this transaction, the
+transaction is not included in the block. Providing further information on the
+transaction state will help the client to make more informed decision on
+handling transaction failures and reduce the incentive to blindly resubmitting
+transactions and help reduce congestion condition in the network. The QUIC
+client library can be enhanced to use this information to not send transactions
+with fees too low.
+
+In addition, on the server side, the TPU server could use the feedback
```
mmcgee-jump, line 34: I don't really understand this part. How does sending feedback to clients help the server to perform QoS?

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,189 @@
...
+information to more accurately perform QOS. For example, if a peer is
+persistently submitting high percentage pf transactions which are rejected,
+its permitted throughput might be reduced.
+
+## Alternatives Considered
+
+There are different transport mechanism to send the feedback such as
+bidirectional streams, reverse unistream from the server to the client and
+QUIC datagrams. We choose datagrams due to its simplicity. One reason we have
```
mmcgee-jump, line 43: QUIC datagrams are unreliable, from what I can tell. This seems like a non-starter. Also having the client open a bidirectional stream for feedback as opposed to a unidirectional stream just seems like a completely obvious choice, and eliminates the handshake you describe below in detailed design.

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,189 @@
...
+used stream over datagrams is the expectation that transaction size can be
+increased in the future which can not be carried in one datagram. Sending
+feedback has no such concern as we can send multiple datagrams if there are
```
mmcgee-jump, line 46: Sending feedback obviously does have similar issues to transactions: how do you know the size of a feedback is always fit in a packet? You indicate you can fragment if there are "too many transactions" but you cannot fragment feedback for an individual transaction.

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,189 @@
...
+more transactions to be acknowledged in one datagram, more datagrams
+can be sent without waiting for the feedback interval.
+
+### The format of the TPU feedback
+
+struct TpuFeedback {
+  version: u8,
+  timestamp: u64,
+  transaction_state: TransactionState,
+  priority_fee_info: PriorityFeeInfo
```
mmcgee-jump, line 87: This is generic RPC information and QUIC is not the right place to disseminate it.

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,189 @@
...
+}
+
+The version is a 8 bit unsinged integer, it is set to 1 for first version. It
+is designed to allow the protocol to support for future extension. This should
```
mmcgee-jump, line 91: This completely custom binary protocol is neither forward nor backward compatible, is not extensible, and does not reserve space for application specific values (for example, Firedancer specific transaction states). I do not think this is the correct way to specify this.

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,189 @@
...
+have enough space for the versions in the relatively long future. It could also
+be wrapped around in the future to differentiate as we do not expect to support
+many current different versions.
+
+The timestamp is the 64 bit unsigned integer representing the number of seconds
```
mmcgee-jump, line 96: New protocol should always use nanoseconds

proposals/0248-tpu-feedback.md
mmcgee-jump: Also tying the frequency of "priority fee info" dissemination to the publishing of transaction states doesn't make sense, if it was going to be published over this protocol (which is a bad idea), it should be a separate message.

Note that adding a "message_type: u8" to this protocol after the fact would break every single consumer of it.

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,189 @@
...
+since midnight 01/01/1970.
+
+TransactionState is defined as the following:
+
+struct TransactionState {
+  transactions_count: u8
```
mmcgee-jump, line 102: Coalescing messages within the message like this doesn't make sense. Why not just use a stream of states? No state updates published by Firedancer will ever share a timestamp, so we would not coalesce in this way.

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,189 @@
...
+
+The transactions_count is 8 bit unsgined integer representing the count of
+transactions in this feedback. It is followed by the count of
+(transaction, state) pairs. Each transaction is represented by the first 8 bytes
+of the transaction signature and each state is represented by an 32 bit unsigned
+integer.
+
+The transaction state value is defined as the below enum:
+
+enum TransactionStateValue {
```
mmcgee-jump, line 114: This enum misses a set of obvious "states" (invalid signature, duplicate, blockhash expired) -- Firedancer knows at least 24 reasons to drop a transaction currently, and this is always increasing.

The design of the enum is also deeply flawed. The failure states cannot be enumerated, and they need to be categorized so a caller can group them even if the don't understand the specific error. For example, look at HTTP 2xx vs. 4xx vs. 5xx. You also need to pre-reserve space for application specific errors within these groupings.

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,189 @@
...
+  Packed = 0,
+  FeeTooLow = 1,
+  FeePayerBalanceTooLow = 2,
+  NotPackedDueToUnspecifiedReason = 2^32 -1,
+}
+
+Newer state value can be defined as needed in the future with its own unique
+value. Client should disregard the value if it does not understand its meaning.
```
mmcgee-jump, line 122: > Client should disregard the value if it does not understand its meaning.
This is not a suitable design, and hasn't been since the 80s. The protocol should enable a client to handle new values even if they don't specifically know what it is.

proposals/0248-tpu-feedback.md
tao-stones: I think this is about TPU server receives feedbacks from TPU; TPU server could throttle a sender if it's transaction has very low landing rate, or other poor performance.

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,189 @@
...
+increased in the future which can not be carried in one datagram. Sending
+feedback has no such concern as we can send multiple datagrams if there are
+too many transactions to acknowledge.
+
+## New Terminology
+
+In this document we define the following,
+
+TPU Feedback -- information sent by the server about transaction processing.
+Server -- the validator receiving the TPU transactions and sending the feedback.
```
godmodegalactus, line 54: These are old terminology no need to add.

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,189 @@
...
+of the transaction signature and each state is represented by an 32 bit unsigned
+integer.
+
+The transaction state value is defined as the below enum:
+
+enum TransactionStateValue {
+  Packed = 0,
+  FeeTooLow = 1,
+  FeePayerBalanceTooLow = 2,
+  NotPackedDueToUnspecifiedReason = 2^32 -1,
```
godmodegalactus, line 118: Should also included info like dropped by reasons like, too many unistreams, invalid sigverify, dropped by  dedup stage,

proposals/0248-tpu-feedback.md
```diff
@@ -0,0 +1,189 @@
...
+    median_priority_fee/compute unit ratio
+    max_priority_fee/compute unit ratio
+  }
+}
+
+These ratio are represented as u64 numbers using 8 bytes.
+
+## Impact
+
+There will be higher cost for the server to send the feedback to all connected
```
godmodegalactus, line 138: Dont see why we need an SIMD as it does not affect concenseus.

proposals/0248-tpu-feedback.md
lijunwangs: Transaction status is just Transaction Key + Status. There is NO fragmentation for a single transaction, we can send another feedback diagram for transactions not fit into the previous diagram

proposals/0248-tpu-feedback.md
lijunwangs: You cannot RPC to each leader to get this info.

proposals/0248-tpu-feedback.md
lijunwangs: I dont't understand what coalescing you are talking about.

lijunwangs:
Closing due to lack of community interest in this feature. Will revisit if needed in the future.

PR #267, opened 2025-03-26
SIMD-0267: Sets rent_epoch to a constant in the VM

brooksprumo:


proposals/0267-rent-epoch-in-vm-is-a-constant.md
```diff
@@ -0,0 +1,53 @@
...
+
+## New Terminology
+
+None.
+
+## Detailed Design
+
+The value of the `rent_epoch` field serialized into the VM per account shall be
+set to the constant used to denote a rent-exempt account.  This value is
+`u64::MAX`, aka `0xFFFF_FFFF_FFFF_FFFF`.
```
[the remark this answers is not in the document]
jstarry: The account that block producers collect fees into could be a new account and in that case we don't set the rent epoch to u64 MAX either until partitioned rent runs.

topointon-jump (approved):

proposals/0267-rent-epoch-in-vm-is-a-constant.md
```diff
@@ -0,0 +1,53 @@
...
+## Motivation
+
+The `rent_epoch` field on an account is no longer meaningful.  This is because
+all accounts must be rent exempt, and rent fees collection has been disabled
+(see SIMD-84).  With this in mind, it would be beneficial to remove
+`rent_epoch` from the computation of a single account's hash.
+
+However, before we can remove the `rent_epoch` from the account hash
+computation, we must set the value of `rent_epoch`—as serialized into the VM
+for transaction processing—to a constant.  If we did not do this, different
```
HaoranYi, line 27: Does this sentence refer to the excluding rent_epoch from lthash? Should we include a reference to lthash simd for clarity?

HaoranYi: Ok. To be clear, do we expect any other types of accounts hash besides lthash after lthash is activated?

HaoranYi: Right, I am talking about account's hash, which is the blake3 hash.  And other
types of "accounts hash", either lthash or merkel tree, depends on this hash.  

If we change this blake3 hash to exclude "rent-epoch", it will impact all
other types of "accounts hash" - i.e. they have to be recomputed from scratch.
This means that for merkel tree hash we can't use old cached value. And for
lt-hash we can't incrementally update the lthash. 

We have address this concern for lthash - by rekeying the feature recently, so
that lthash will exclude rent-epoch from start.

But how about the merkle tree with hash cache? I don't think we have thought
about it. 

If we keep merkel tree the same and use the old hash with `rent_epoch`, then
we have two ways to hash an account. And one of them depends on the rent_epoch
of the account. This will contradict to "no multiple ways to compute account
hash".

If we switch merkle tree hash to exclude rent-epoch, then we need to make sure
to invalidate hash cache at the feature activation.

From reading SIMD-0224, it not clear whether merkel tree hash will also switch to
exclude rent-epoch. 

I wonder if we have a plan for merkle tree hash when we exclude rent epoch
from account hash?

topointon-jump: I think after SIMD-0215, SIMD-0223 and SIMD-0220 have been activated we will no longer have a merkle hash, so we can effectively close SIMD-0224.

topointon-jump: Yes

HaoranYi (approved):

HaoranYi (approved):

topointon-jump (approved):

Benhawkins18 (approved):

PR #282, opened 2025-05-01
SIMD-0186: Define LoaderV4

2501babe:
im drafting the impl of this now (https://github.com/anza-xyz/agave/pull/6063), and loaderv4 is on the horizon, but previously its handling was undefined. my understanding now is it uses a single account for a program so we dont have to do anything special at all

also clarify two things:
* transactions can actually have multiple lookup tables, so specify we count them all
* we dont abort loading if a loaderv3 programdata account we load _implicitly_ (ie, it isnt specified in `account_keys()`) doesnt exist, for instance if the program was closed

### DIFF
--- a/proposals/0186-loaded-transaction-data-size-specification.md
+++ b/proposals/0186-loaded-transaction-data-size-specification.md
@@ -79,9 +79,9 @@ irrespective of how they are used.
 accounts specified on the transaction.
 2. Each account's size is defined as the byte length of its data prior to
 transaction execution plus 64 bytes to account for metadata.
-3. There is an additional flat 8248 byte cost for transactions that use an
-address lookup table, accounting for the 8192 bytes for the maximum size of such
-a table plus 56 bytes for metadata.
+3. There is an additional flat 8248 byte cost for each address lookup table used
+by a transaction, accounting for the 8192 bytes for the maximum size of such a
+table plus 56 bytes for metadata.
 4. The total transaction loaded account data size is the sum of these sizes.
 
 Transactions may include a
@@ -105,14 +105,12 @@ mapping is enabled, this SIMD may be amended to count them differently.
 We include programdata size for LoaderV3 programs because using the program
 account on a transaction forces an unconditional load of programdata to compile
 the program for execution. We always count it, even when the program account is
-not a transaction program ID, because the program must be available for CPI.
+not a transaction program ID, because the program must be available for CPI. If
+the programdata account does not exist, then its size is 0, and transaction
+loading continues as normal.
 
 There is no special handling for any account owned by the native loader,
-LoaderV1, or LoaderV2.
-
-Account size for programs owned by LoaderV4 is left undefined. This SIMD should
-be amended to define the required semantics before LoaderV4 is enabled on any
-network.
+LoaderV1, LoaderV2, or LoaderV4.
 
 ## Alternatives Considered
 


proposals/0186-loaded-transaction-data-size-specification.md
```diff
@@ -105,14 +105,12 @@ mapping is enabled, this SIMD may be amended to count them differently.
...
+the programdata account does not exist, then its size is 0, and transaction
+loading continues as normal.
 
 There is no special handling for any account owned by the native loader,
-LoaderV1, or LoaderV2.
-
-Account size for programs owned by LoaderV4 is left undefined. This SIMD should
-be amended to define the required semantics before LoaderV4 is enabled on any
-network.
+LoaderV1, LoaderV2, or LoaderV4.
```
2501babe, line 113: @Lichtso if you could confirm my understanding is correct

Lichtso: Yes, only loader-v3 had an indirection. All other loaders (including v4) store the executable directly in the program account.

apfitzge (approved):
Appreciate you adding the clarification on ALT use

mjain-jump (approved):
Looks good to me!

2501babe:
@jacobcreech

2501babe:
@jacobcreech could this be approved? the code is already in monorepo, and we might be making some fresh changes after. im not sure what the repo permissions are but assume this requires foundation

Lichtso:
Has https://github.com/anza-xyz/agave/pull/6063#discussion_r2164128450 been added here yet?

2501babe:
i wanted to land this pr now because the confusion in both https://github.com/anza-xyz/agave/pull/6063#discussion_r2161666186 and https://github.com/anza-xyz/agave/pull/6063#discussion_r2161690342 were because @jstarry was probably looking at the language as it exists in master and i was looking at the language in this pr which i implemented knowing it was approved by both teams

im going to open a separate pr with a sidebar section describing program/loader validation and we can debate if its appropriate to go with that or if it should be its own simd. we cant just add a one-liner because this simd as-is doesnt talk about program/loader validation at all. and the changes in the 186 code essentially move forward changes in the 162 code which are similarly underdescribed

2501babe:
adding it in https://github.com/solana-foundation/solana-improvement-documents/pull/311

proposals/0186-loaded-transaction-data-size-specification.md
```diff
@@ -77,11 +77,11 @@ the unique intersection of:
...
 accounts specified on the transaction.
-2. Each account's size is defined as the byte length of its data prior to
+2. Each loaded account's size is defined as the byte length of its data prior to
 transaction execution plus 64 bytes to account for metadata.
-3. There is an additional flat 8248 byte cost for transactions that use an
-address lookup table, accounting for the 8192 bytes for the maximum size of such
-a table plus 56 bytes for metadata.
+3. There is an additional flat 8248 byte cost for each address lookup table used
+by a transaction, accounting for the 8192 bytes for the maximum size of such a
+table plus 56 bytes for metadata.
```
jstarry: I wouldn't mind a clarifying comment stating that the 64 bytes for account metadata is not added here. I think as written the SIMD is probably clear enough though.

2501babe:
@jstarry @mjain-jump if i could have your approvals on this again, so foundation can merge

jstarry (approved):

mjain-jump (approved):
looks good to me!

topointon-jump (approved):

joncinque (approved):
We've got approval from Anza and Firedancer, merging!

PR #285, opened 2025-05-16
Rename 0250 -> 0256 to match the PR number and simd header

willhickey:
The md file for SIMD-0256 was named 0250 instead. This just renames the file for consistency


Benhawkins18 (approved):
LGTM

PR #291, opened 2025-05-29
SIMD-0291: Commission Rate in Basis Points

jstarry:


joncinque (dismissed):
Looks good to me!

dimandotsol:
Personally, I don’t like the use of the term “basis points” - I would prefer to stick to mathematical terminology.

topointon-jump (approved):

buffalojoec (approved):

PR #297, opened 2025-06-05
SIMD-029x: Relax Invalid Nonce Transaction Constraint

tao-stones:


proposals/0297-relax-invalid-nonce.md
```diff
@@ -0,0 +1,79 @@
...
+  successfully.
+
+## Alternatives Considered
+
+- Only relax a subset of errors (e.g., missing nonce account but not mismatched
+  blockhash). Adds complexity and partial gains.
+
+## Impact
+
+- Invalid nonced transactions will be included in blocks and will be charged
```
0x0ece: nit: will **not** be charged

tao-stones: thank you, fixed!

ptaffet-jump:
Doesn't this introduce a new DoS vector from a malicious leader?

proposals/0297-relax-invalid-nonce.md
```diff
@@ -0,0 +1,79 @@
...
+- The transaction fails to advance the nonce.
+
+
+### Proposed Change
+
+Update the replay logic to treat invalid nonce transactions as
+non-state-modifying, non-fee-charging failures, as follows:
+
+- The transaction is not executed.
+- The transaction is metered by CU limits.
```
topointon-jump: What specifically does this mean?

proposals/0297-relax-invalid-nonce.md
```diff
@@ -0,0 +1,79 @@
...
+- The nonce account is not a statically included account.
+- The nonce account does not exist.
+- The nonce account is not properly initialized.
+- The stored nonce does not match the transaction's recent blockhash.
+- The transaction fails to advance the nonce.
+
+
+### Proposed Change
+
+Update the replay logic to treat invalid nonce transactions as
```
topointon-jump: Could we make it clear that an "invalid nonce transaction" is a transaction that doesn't pass the checks listed above?

topointon-jump (commented):
Looks good, could we just address my comments adding some clarity around some points? Thanks!

proposals/0297-relax-invalid-nonce.md
tao-stones: This means the CU cost of such a transaction still counts toward the block’s CU limit, even if it isn’t executed or charged a fee. I'll update it

tao-stones:
> Looks good, could we just address my comments adding some clarity around some points? Thanks!

Thanks for feedback, updates in 57b1fb295e0cdd8e85cc6ea461477c5dc4d3e950

proposals/0297-relax-invalid-nonce.md
```diff
@@ -0,0 +1,79 @@
...
+checks above — as non-state-modifying and non-fee-charging failures, as follows:
+
+- The transaction is not executed.
+- The transaction’s CU cost still applies to the block limit.
+- The transaction is not charged with fee.
+- The transaction is recorded in the block (marked as failed).
+- No account state is modified, including the nonce account (i.e., nonce is
+  not advanced).
+- The block is not rejected as long as all other transactions replay
+  successfully.
```
sakridge: not included in status cache as well?

tao-stones: My initial thought was to exclude these transactions from the status cache, since that aligns better with their non-state-modifying nature. The downside, however, is that invalid transactions might be re-executed multiple times, and their status may not be immediately available via RPC.

ptaffet-jump: Good catch. This definitely needs to be mentioned in the SIMD.

tao-stones: If no objection of excluding such transactions from status cache, I'll add it to spec.

tao-stones: in c613fef1138ab7be123db75817224337478d2db4

proposals/0297-relax-invalid-nonce.md
```diff
@@ -0,0 +1,80 @@
...
+
+
+### Proposed Change
+
+Update replay logic to treat invalid nonce transactions — those failing the
+checks above — as non-state-modifying and non-fee-charging failures, as follows:
+
+- The transaction is not executed.
+- The transaction is not included in status cache.
+- The transaction’s CU cost still applies to the block limit.
```
ptaffet-jump: How many CUs? Just the non-execution CUs? The full requested CUs?

proposals/0297-relax-invalid-nonce.md
```diff
@@ -0,0 +1,80 @@
...
+- The transaction is not charged with fee.
```
ptaffet-jump: I think I know what you mean, but this sounds a bit clunky. How about `The fee payer is not charged the transaction's fee` or something like that? It's also implied by `No account state is modified` below.

tao-stones: combined them in 184854e85789d49b3157bea9ada9539df059f5ed

proposals/0297-relax-invalid-nonce.md
tao-stones: It's non-execution CUs since "transaction is not executed". Clarified in 110d4200466c022d67cc12ca914a3f15c06156d1. Transaction's static part of CU (signature, write locks etc) are small, this requirement is mainly to cover cost of loading accounts.

sakridge (approved):

tao-stones:
@Benhawkins18 this simd is ready for merge, thanks.

proposals/0297-relax-invalid-nonce.md
```diff
@@ -0,0 +1,81 @@
...
+- The nonce account does not exist.
+- The nonce account is not properly initialized.
+- The stored nonce does not match the transaction's recent blockhash.
+- The transaction fails to advance the nonce.
+
+
+### Proposed Change
+
+Update replay logic to treat invalid nonce transactions — those failing the
+checks above — as non-state-modifying and non-fee-charging failures, as follows:
```
apfitzge: I don't think we should relax the first condition:

> The nonce account is not a statically included account.

This can be easily checked very early in the pipeline. Has no value to async to relax this since it can be verified without state.

proposals/0297-relax-invalid-nonce.md
```diff
@@ -0,0 +1,81 @@
...
+
+- The transaction is not executed.
+- The transaction is not included in status cache.
+- The transaction’s non-execution CU cost (i.e., transaction's static CUs, plus
+  CUs for loading accounts) still applies to the block limit.
```
apfitzge: We should clarify how this applies. Loading accounts is just the actually loaded accounts (fee-payer + nonce) or requested amount of data?

proposals/0297-relax-invalid-nonce.md
```diff
@@ -0,0 +1,81 @@
...
+- The transaction fails to advance the nonce.
+
+
+### Proposed Change
+
+Update replay logic to treat invalid nonce transactions — those failing the
+checks above — as non-state-modifying and non-fee-charging failures, as follows:
+
+- The transaction is not executed.
+- The transaction is not included in status cache.
```
apfitzge: we should make this change for nonces in general 😉 (if we're not going to delete them)

proposals/0297-relax-invalid-nonce.md
```diff
@@ -0,0 +1,81 @@
...
+## Alternatives Considered
+
+- Only relax a subset of errors (e.g., missing nonce account but not mismatched
+  blockhash). Adds complexity and partial gains.
+
+## Impact
+
+- Invalid nonced transactions will be included in blocks but will not be charged
+  fees.
+
```
apfitzge, line 90: We should highlight the impact of downstream effects of this.
Nonced transactions can appear multiple times in the blockchain's history now. This could have a confusing impact on RPCs.

When a user queries a signature, how do I report what slot it's in? Maybe only one that it succeeded in advancing the nonce, if any?

apfitzge (changes requested):

proposals/0297-relax-invalid-nonce.md
tao-stones: Good point about the RPC use case where the same nonced transaction can appear in multiple blocks. Since any invalid instance would be marked as failed, the RPC has flexibility in whether to return the succeeded one or the failed occurrences. In any case, this situation stems from non–state-modifying transactions being included in blocks (as stated), so I don’t think it’s necessary to enumerate specific cases.

proposals/0297-relax-invalid-nonce.md
tao-stones: Previous comments on this line said it is not the "requested" account bytes. The main point is "non-execution cost". How about just adding word "actual" to clarification: 
```suggestion
  actual CUs for loading accounts) still applies to the block limit.
```

proposals/0297-relax-invalid-nonce.md
tao-stones: Agree the first check is quick and not state related. Left it here for simplicity as mentioned in "alternatives" section.

apfitzge: I don't think it really adds much complexity though to not relax it though. I'd think relaxing this adds more complexity, which was the whole justification for recently adding this condition in #242

proposals/0297-relax-invalid-nonce.md
apfitzge: Well I think it does differ from other non-state-changing transactions because it can have multiple instances (many that change no state, and up to one that can).

afaict the default behavior in our code rn would be to return the first one that appears, though this is probably misleading since that may not be the one that changes state. if rpcs had to check that makes their job harder.

proposals/0297-relax-invalid-nonce.md
tao-stones: The complexity, if we must distinguish between different kinds of "invalid-ness", lies in reasoning about which types of invalid nonced transactions should fail the block entirely, and which should simply result in non-state-changing behavior. 

Thanks for pointing out #242 — note relaxing first condition still preserves #242 function and purpose. If there aren't downsides to including this case among other conditions that are already considered relaxable. I'd lean towards avoid defining different "invalid-ness," as that adds unnecessary confusion for users and complexity for implementers and code maintainers alike.

proposals/0297-relax-invalid-nonce.md
tao-stones: Yes, this case is different from other relaxed SIMDs because the same nonced transaction can appear in multiple blocks, as noted in the "Impact". My point was that this impact likely goes beyond just RPC queries, and I didn’t think it was necessary to enumerate every special case. 

However, it might be helpful to explicitly highlight this RPC example. How about to add a sub-bullet like:
```
   - The inclusion of the same nonced transaction in multiple blocks may affect RPC behavior, such as querying a transaction by signature. The exact handling of this is outside the scope of this proposal.
```

proposals/0297-relax-invalid-nonce.md
tao-stones: per our discussion, updated to exclude stateless failure from being relaxed. at 3ae6737d21bee572c4608dc0b16051f7501a4aeb

proposals/0297-relax-invalid-nonce.md
tao-stones: @ 6b825d62dc807f676e65cd264647fc13fc1e272b

proposals/0297-relax-invalid-nonce.md
```diff
@@ -81,8 +81,12 @@ N/A
 
 ## Impact
 
-- Invalid nonced transactions will be included in blocks but will not be charged
-  fees.
+- Invalid nonce transactions will be included in multiple blocks but will not be
```
apfitzge: nit:
please update to "may" be included, there's not a requirement that it is.

PR #298, opened 2025-06-06
SIMD 0298: Add Parent Bank Hash to Block Header

MaxResnick:
This proposal adds the parent bank hash to the block header and introduces validity constraints to ensure the parent bank hash matches the expected value. This change is part of the broader set of SIMDs neccesary for async execution.


proposals/0298-bank-hash-in-block-footer.md
```diff
@@ -0,0 +1,107 @@
...
+2. The `bank_hash` is missing (for blocks after the feature is activated)
+
+### Implementation Details
+
+1. The `bank_hash` will be computed at the end of packing each block.
+The validator proposing the block will include it in the block footer.
+2. Replaying validators will verify the `bank_hash` in the block footer
+matches their computed hash of the state after executing the block.
+This is similar to what they do with the `bank_hash`
+contained in votes today.
```
lidatong, line 180: can you specify what validators should do if the bank hash doesn't match? eg. should they vote "skip" in alpenglow?

AshwinSekar: Yeah I think explicitly we should call out that the bank will be marked as dead and alpenglow validators should vote skip for this block (and any descendants)

proposals/0298-bank-hash-in-block-footer.md
```diff
@@ -0,0 +1,107 @@
...
+
+1. **Remove Execution from Consensus Entirely**: We could simply remove the
+   `bank_hash` from votes and not include it anywhere in the block structure. This
+   would mean that execution bugs could go undetected for longer periods,
+   making it harder to identify and resolve runtime issues. Eventually we want
+   to remove this completely but for now it seems useful to keep training
+   wheels on, particularly as we rollout Firedancer's VM implementation.
+
+2. **XOR bank_hash with block_id**: Instead of adding a separate
+   bank_hash field to the footer, we could XOR the `bank_hash` with the `block_id`
```
AshwinSekar, line 210: Not clear from the description but i'm assuming you mean putting this XOR in the vote not the footer?

AshwinSekar (commented):
Overall looks good, just a couple of comments

proposals/0298-bank-hash-in-block-footer.md
MaxResnick: yeah this was something Joachim suggested at NYC core dev day. It saves the space from storing both block_id and bank_hash.

proposals/0298-bank-hash-in-block-footer.md
MaxResnick: Yes the block is invalid and should be treated the same way as we treat a protocol violating error.

MaxResnick: Will update to clarify this behavior.

AshwinSekar (approved):

proposals/0298-bank-hash-in-block-footer.md
```diff
@@ -0,0 +1,113 @@
...
+
+In addition to the current validity constraints, a block will be considered
+invalid if:
+
+1. The `bank_hash` does not match the post execution state of the block
+2. The `bank_hash` is missing (for blocks after the feature is activated)
+
+### Implementation Details
+
+1. The `bank_hash` will be computed at the end of packing each block.
```
bw-solana, line 175: maybe state explicitly that this will be a new version of Block Marker Footer Variant?

proposals/0298-bank-hash-in-block-footer.md
```diff
@@ -0,0 +1,113 @@
...
+The validator proposing the block will include it in the block footer.
+2. Replaying validators will verify the `bank_hash` in the block footer
+exists and matches their computed hash of the state after executing the block.
+This is similar to what they do with the `bank_hash`
```
bw-solana, line 179: I know we just plan on using the same `bank_hash` that is included in votes today, but might not be a bad idea to spell out how the bank hash is generated in this SIMD

proposals/0298-bank-hash-in-block-footer.md
```diff
@@ -0,0 +1,113 @@
...
+
+If the `bank_hash` does not match the post execution `bank_hash` then the block
+is marked invalid. Validators will treat this the same way they treat a
+protocol violatiing error. In particular, the block will be marked dead and
+the validator will not vote on it or any of its children. Post alpenglow
+the validator will vote skip on this block and any children.
+
+### Feature Activation
+
+The feature can be activated at any time before or after the Alpenglow rollout.
```
bw-solana: We need this for Alpenglow, right? So I think it must be activated either before or as part of the overall Alpenglow feature activation

proposals/0298-bank-hash-in-block-footer.md
MaxResnick: added to New Terminology section

proposals/0298-bank-hash-in-block-footer.md
MaxResnick: adjusted.

proposals/0298-bank-hash-in-block-footer.md
```diff
@@ -0,0 +1,230 @@
...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+| footer_version=2          (16 bits)   |  ←UPDATED 1->2
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+| block_producer_time_nanos   (64 bits) |  
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+| block_user_agent_len         (8 bits) |  
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+| block_user_agent        (0-255 bytes) |  
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+| bank_hash                  (32 bytes) |  ←NEW
```
bw-solana, line 161: I wonder if we want to sneak this before the agent/agent length fields so that it's at a fixed offset and doesn't require reading prior fields to extract

bw-solana:
CC @topointon-jump

proposals/0298-bank-hash-in-block-footer.md
```diff
@@ -0,0 +1,230 @@
...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+
+```
+
+**Bank Hash Components:**
+
+- `parent_bank_hash` - Hash of the parent bank's state (32 bytes)
+- `signature_count` - Total number of signatures processed in the bank (64 bits)
+- `last_blockhash` - The blockhash used for transaction processing (32 bytes)
+- `accounts_lt_hash_checksum` - Checksum of the accounts ledger tree (32 bytes)
```
brooksprumo, line 122: Hey folks, sorry for not seeing this SIMD until now.

The accounts lt hash is meant to be a *private* implementation detail of how the bank hash is calculated. We should avoid making it public/standalone where possible, so that we retain the ability to change it more easily. (For background, read SIMD 223 on removing the Accounts Delta Hash). 

If the intent to have the hash of the final state for this bank, why not include the bank hash itself? Even if we have the accounts lt hash here, that may be insufficient to compute the bank hash, since the bank hash also includes hard forks.

(And then as a nit, this value, as stated, is the accounts lt hash. It is not the accounts lt hash *checksum*. The checksum is 8 bytes, and should only be used for logging purposes. So the intent is correct, just the word "checksum" should be removed.)

Edit: Re-reading, I think my original reading was wrong. We're not adding the lt hash to the block footer, right? This is just outlining what makes up the bank hash? Only the bank hash is being added to the block footer. Sorry for the noise!

bw-solana: that's right. I think I was the one that asked Max to include Bank Hash components just for background context. The Bank Hash is the only thing being put into the footer

PR #300, opened 2025-06-09
Amend simd-0123

jstarry:
> In the case of additional rewards, we move amounts from the vote account to the stake accounts. By the way, this will lead to writing all vote accounts every block (assuming the partition reward distribution is uniformly distributed), right?

> I think it’s better to follow the approach of SIMD-0118, but we’ll have to use a sysvar for that (since capitalization is not increased). That is, in the first block of the epoch, record the reward distribution amounts by vote accounts into the sysvar data, and move that amount to the sysvar.

_Originally posted by @dimandotsol in https://github.com/solana-foundation/solana-improvement-documents/pull/123#discussion_r2125735097_

### DIFF
--- a/proposals/0123-block-revenue-distribution.md
+++ b/proposals/0123-block-revenue-distribution.md
@@ -151,9 +151,7 @@ block of an epoch `N` and any time the node is restarted during the partitioned
 rewards distribution period.
 
 For each vote account, get its total active stake delegation
-during the reward epoch `N - 1`. Let this value be `A`. If `A` is zero, skip
-this vote account (any pending delegator rewards will be distributed in a future
-epoch when the vote account has active stake.)
+during the reward epoch `N - 1`. Let this value be `A`.
 
 Then for each vote account, get its pending delegator rewards from the
 `pending_delegator_rewards` field in the vote state at the end of reward epoch
@@ -161,10 +159,12 @@ Then for each vote account, get its pending delegator rewards from the
 is nothing to be distributed.
 
 Lastly, if this is the first block of epoch `N`, the vote state's
-`pending_delegator_rewards` field MUST be reset to `0` and `P` lamports MUST be
-deducted from the vote account's lamport balance and credited to the epoch
-rewards sysvar account's lamport balance before any transactions or votes are
-processed.
+`pending_delegator_rewards` field MUST be reset to `0`. Then, if `A` (active
+stake) is zero, the delegator rewards will effectively returned to the voter
+after the `pending_delegator_rewards` field is reset to `0`. Otherwise, if `A`
+is non-zero, `P` lamports MUST be debited from the vote account's lamport
+balance and credited to the epoch rewards sysvar account's lamport balance
+before any transactions or votes are processed.
 
 Note that unlike inflation rewards distribution, block revenue distribution will
 not impact any internal epoch rewards sysvar state fields like `total_rewards`
--- a/proposals/0123-block-revenue-distribution.md
+++ b/proposals/0123-block-revenue-distribution.md
@@ -146,18 +146,29 @@ distribution mechanism described in [SIMD-0118].
 
 #### Delegator Rewards Calculation
 
-During the beginning of the first block of an epoch `N`, all pending delegator
-rewards from all vote accounts with a non-zero active stake delegation during
-the reward epoch `N - 1` MUST be transferred by the runtime to the epoch rewards
-sysvar account for distribution.
+Delegator rewards MUST be calculated during both the beginning of the first
+block of an epoch `N` and after restarting during partitioned rewards
+distribution.
 
-For each actively staked vote account, retrieve its state at the end of the
-reward epoch and check the `pending_delegator_rewards` field in its vote state.
-Let this value be `P`. If `P` is non-zero, set the `pending_delegator_rewards`
-field to `0` and deduct `P` from the vote account's lamport balance and credit
-it to the epoch rewards sysvar account's lamport balance.  Then if `P` is
-non-zero, sum all active stake delegated to the vote account during the reward
-epoch epoch `N - 1`. Let this total be `A`.
+For each vote account, get its total active stake delegation
+during the reward epoch `N - 1`. Let this value be `A`. If `A` is zero, skip
+this vote account (any pending delegator rewards will be distributed in a future
+epoch when the vote account has active stake.)
+
+Then for each vote account, get its pending delegator rewards from the
+`pending_delegator_rewards` field in the vote state at the end of reward epoch
+`N - 1`. Let this value be `P`. If `P` is zero, skip this vote account as there
+is nothing to be distributed.
+
+Lastly, if this is the first block of epoch `N`, the vote state's
+`pending_delegator_rewards` field MUST be reset to `0` and `P` lamports MUST be
+deducted from the vote account's lamport balance and credited to the epoch
+rewards sysvar account's lamport balance.
+
+Note that unlike inflation rewards distribution, block revenue distribution will
+not impact any internal epoch rewards sysvar state fields like `total_rewards`
+or `distributed_rewards` since block revenue will instead be tracked via the
+epoch rewards sysvar lamport balance.
 
 #### Individual Delegator Reward
 
@@ -178,6 +189,11 @@ lamport balance MUST be debited by the block revenue reward distribution amount
 to keep capitalization consistent.
 
 [SIMD-0118]: https://github.com/solana-foundation/solana-improvement-documents/pull/118
+
+#### Delegator Rewards Completion
+
+After distributing all partitioned delegator rewards, the epoch rewards sysvar balance
+MUST be reset to its rent exemption balance and any surplus lamports are burned.
 
 ### Vote Program
 
--- a/proposals/0123-block-revenue-distribution.md
+++ b/proposals/0123-block-revenue-distribution.md
@@ -137,43 +137,45 @@ vote state field `pending_delegator_rewards` and added to the balance of vote
 account. If either of these conditions is violated, the delegator rewards amount
 MUST be burned.
 
-### Runtime: Delegator Rewards Distribution
+### Runtime: Block Revenue Distribution
 
-When calculating stake delegation rewards for a particular completed reward
-epoch, construct a list of all vote accounts that were initialized at the
-beginning of the reward epoch and had a non-zero active stake delegation. For
-each vote account, retrieve its state at the end of the reward epoch and check
-the `pending_delegator_rewards` field in its vote state. Let this value be `P`.
-If `P` is non-zero, use it to calculate rewards for each of the stake accounts
-delegated to the vote account as follows: 
+At the beginning of each epoch, pending delegator rewards will be collected from
+actively staked vote accounts into the epoch rewards sysvar account and then
+distributed to stake accounts utilizing the existing partitioned rewards
+distribution mechanism described in [SIMD-0118].
 
-1. Sum all active stake delegated to the vote account during the reward epoch
-epoch. Let this total be `A`.
+#### Delegator Rewards Calculation
 
-2. For each individual stake account, multiply its active stake from the
-reward epoch by `P`, and divide the result by `A` using integer division.
-Discard any fractional lamports.
+During the beginning of the first block of an epoch `N`, all pending delegator
+rewards from all vote accounts with a non-zero active stake delegation during
+the reward epoch `N - 1` MUST be transferred by the runtime to the epoch rewards
+sysvar account for distribution.
 
-After calculating all individual stake rewards, sum them to obtain `D`, the
-total distribution amount. Because of integer division, the full amount `P` may
-not be distributed so compute the amount to be burned, `B`, as the difference
-between `P` and `D`.
-
-If no blocks in the epoch following the completed reward epoch have been
-processed yet, subtract `B` from both the vote account’s lamport balance and its
-`pending_delegator_rewards` field and store the updated vote account. Finally,
-the burn amount `B` should also be deducted from the cluster capitalization.
+For each actively staked vote account, retrieve its state at the end of the
+reward epoch and check the `pending_delegator_rewards` field in its vote state.
+Let this value be `P`. If `P` is non-zero, set the `pending_delegator_rewards`
+field to `0` and deduct `P` from the vote account's lamport balance and credit
+it to the epoch rewards sysvar account's lamport balance.  Then if `P` is
+non-zero, sum all active stake delegated to the vote account during the reward
+epoch epoch `N - 1`. Let this total be `A`.
 
 #### Individual Delegator Reward
 
-The stake reward distribution amounts for each stake account calculated above
-can then be used to construct a list of stake reward entries which MUST be
-partitioned and distributed according to [SIMD-0118].
+For each individual stake account with an active non-zero delegation, multiply
+its active stake in reward epoch `N - 1` by `P`, and divide the result by
+`A` using integer division to get the individual stake account's block revenue
+reward distribution amount.
 
-When reward entries are used to distribute rewards pool funds during partitioned
-rewards distribution, the delegated vote account for each rewarded stake account
-must have its `pending_delegator_rewards` field and its balance deducted with
-the amount of rewards distributed to keep capitalization consistent.
+#### Delegator Rewards Distribution
+
+The reward distribution amounts for each stake account can then be used to
+construct a list of stake reward entries which MUST be partitioned and
+distributed according to [SIMD-0118].
+
+During partitioned reward distribution in a given slot, when reward entries are
+used to distribute block revenue rewards, the epoch rewards sysvar account's
+lamport balance MUST be debited by the block revenue reward distribution amount
+to keep capitalization consistent.
 
 [SIMD-0118]: https://github.com/solana-foundation/solana-improvement-documents/pull/118
 
@@ -246,5 +248,5 @@ NA
 
 ## Backwards Compatibility
 
-A feature gate will be used to enable block reward distribution at an epoch
-boundary.
+A feature gate will be used to enable block revenue reward distribution at an
+epoch boundary.
--- a/proposals/0123-block-revenue-distribution.md
+++ b/proposals/0123-block-revenue-distribution.md
@@ -146,9 +146,10 @@ distribution mechanism described in [SIMD-0118].
 
 #### Delegator Rewards Calculation
 
-Delegator rewards MUST be calculated during both the beginning of the first
-block of an epoch `N` and any time the node is restarted during the partitioned
-rewards distribution period.
+Delegator rewards MUST be calculated during the beginning of the first block of
+an epoch `N` where the first block is any block with a parent in the previous
+epoch. Rewards MUST be recalculated if a node is restarted during the partitioned
+rewards distribution period as described in [SIMD-0118].
 
 For each vote account, get its total active stake delegation
 during the reward epoch `N - 1`. Let this value be `A`.
@@ -181,10 +182,11 @@ reward distribution amount, truncating any sub-lamport portion.
 #### Delegator Rewards Distribution
 
 The reward distribution amounts for each stake account can then be used to
-construct a list of stake reward entries which MUST be partitioned and
-distributed according to [SIMD-0118].
+construct a list of stake reward entries which MUST be partitioned according to
+[SIMD-0118] which describes which stake rewards are distributed in each block
+during the partitioned rewards distribution period.
 
-During partitioned reward distribution in a given slot, when reward entries are
+During partitioned reward distribution for a given slot, when reward entries are
 used to distribute block revenue rewards, the epoch rewards sysvar account's
 lamport balance MUST be debited by the block revenue reward distribution amount
 to keep capitalization consistent.
--- a/proposals/0123-block-revenue-distribution.md
+++ b/proposals/0123-block-revenue-distribution.md
@@ -147,8 +147,8 @@ distribution mechanism described in [SIMD-0118].
 #### Delegator Rewards Calculation
 
 Delegator rewards MUST be calculated during both the beginning of the first
-block of an epoch `N` and after restarting during partitioned rewards
-distribution.
+block of an epoch `N` and any time the node is restarted during the partitioned
+rewards distribution period.
 
 For each vote account, get its total active stake delegation
 during the reward epoch `N - 1`. Let this value be `A`. If `A` is zero, skip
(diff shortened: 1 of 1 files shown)


jstarry:
@dimandotsol can you take a look now and see if the distribution section is easier to understand?

dimandotsol:
> @dimandotsol can you take a look now and see if the distribution section is easier to understand?

Looks great! Thank you!

dimandotsol:
I noticed that lamports will remain on the sysvar balance due to integer division. Maybe it’s worth explicitly mentioning this?

In general, it’s fine, and how to handle it can be addressed later. But while reading, I had the following idea:
```
SysvarData {
  total_rewards: u64, // sum of all P
  total_distributed: u64,
  validators: [
    ValidatorRewardParams {
      vote_account: Pubkey,
      rewards: u64,       // P
      active_stake: u64,  // A
    }
  ]
}
```

And on each distribution block, increase `total_distributed` by the actual amount deducted.
At the end of the distribution, burn `total_rewards - total_distributed`.
(Technically, `total_rewards` isn’t necessary here, since it’s the sum of `rewards` in the validators array.)

HaoranYi:
The idea sounds good to me intuitively. However, there are a few more details that need to be fleshed out. 

1. If we clear pending_rewards from the vote accounts and only store the sum in sysvar in the 1st block, how do we recover the payout for each stake delegation later when distribution? 
2. How do we handle restart during reward distribution? We need to recompute the payout for each delegation but the orignal pending_rewards on the vote accounts is not available? 
3. If we store ValidatorRewardParams in sysvar, it will be too big ~hundreds of Mega bytes. That will be too big for a sysvar data. It seems like this is not a good option.
4. Have we considered the impact of computing all the stake delegation at the first block. Will this make the already-busy first bock even more busy? 
5. If we do in the original way, pending_rewards in the vote account may change before we distribute the payout, because new block are played and rewards between first_epoch block till distribution block will be accumulated into pending_rewards.  When we distribute, we will include the new rewards for current epoch?

dimandotsol:
1.	We’re adding this to the partitioned inflation rewards distribution algorithm that’s already working (SIMD-118):
`active_stake_from_stake_account * this_validator_reward_from_sysvar / this_validator_active_stake_from_sysvar`
(where `active_stake_from_stake_account / this_validator_active_stake_from_sysvar` is the share of a specific stake account in the additional rewards of the given validator)

2.	We don’t need to do anything directly in this SIMD. SIMD-118 will continue to function, and so will we alongside it.

3.	Only for ~2000 validators: `2000 * (32 + 8 + 8) + 8 (array length)` (+ 8 or 16 with my idea above). So it's about 100 kilobytes.

4.	Only a “transfer” of the amounts for distribution from vote accounts to the sysvar.

5.	I’ve already forgotten, but I think the second field was proposed to be used there.

jstarry:
> I noticed that lamports will remain on the sysvar balance due to integer division. Maybe it’s worth explicitly mentioning this? [..] At the end of the distribution, burn `total_rewards - total_distributed`.

@dimandotsol Thanks for pointing this out, agreed on burning the excess, just amended!

> If we clear `pending_rewards` from the vote accounts and only store the sum in sysvar in the 1st block, how do we recover the payout for each stake delegation later when distribution?

@HaoranYi  We recover this value from the epoch stakes cache. We already use the epoch stakes cache for recalculating partitioned inflation rewards after restart. In the future we can store the last few epochs of delegator rewards in the vote account state so that we don't need to rely on the epoch stakes cache.

> Have we considered the impact of computing all the stake delegation at the first block. Will this make the already-busy first bock even more busy?

Yes, have considered it. Basically it's just a few extra arithmetic operations to calculate the extra rewards.

> If we store ValidatorRewardParams in sysvar, it will be too big ~hundreds of Mega bytes. That will be too big for a sysvar data. It seems like this is not a good option.

@HaoranYi agreed. I don't think it's a good option and not necessary as I explained above. @dimandotsol please note that I didn't include this part in the amendment.

dimandotsol:
> @dimandotsol please note that I didn't include this part in the amendment.

@jstarry Wasn’t the validators array supposed to be stored in the sysvar data? Those are just the `P` and `A` needed for the calculations
```
validators: [
    ValidatorRewardParams {
      vote_account: Pubkey,
      rewards: u64,       // P
      active_stake: u64,  // A
    }
  ]
 ```
 And that’s less than 100 KB for 2000 validators.

File: proposals/0123-block-revenue-distribution.md [hunk 1]
```diff
@@ -179,6 +190,11 @@ to keep capitalization consistent.
 
 [SIMD-0118]: https://github.com/solana-foundation/solana-improvement-documents/pull/118
 
+#### Delegator Rewards Completion
+
+After distributing all partitioned delegator rewards, the epoch rewards sysvar balance
+MUST be reset to its rent exemption balance and any surplus lamports are burned.
```
dimandotsol, line 199: This approach will allow burning any amount of lamports that can be transferred to the sysvar address.

By simply additionally storing values in only two u64 fields (in fact, one is enough, because the second is just the sum of `P` over the array and computing it once isn’t a problem), I’m proposing to prevent burning of arbitrary amounts - only rounding burns would remain.

jstarry: You can't transfer to the sysvar in a transaction, it's reserved

dimandotsol: Ah, okay, I didn’t know that.

jstarry:
> Wasn’t the validators array supposed to be stored in the sysvar data? Those are just the P and A needed for the calculations

Nope, the calculated values do not to be stored onchain, this SIMD is designed so that those values can always be recalculated after restart.

HaoranYi:
It seems to me that the complexity of introducing a new sysvar for block
revenue payout in this spec may not be worth it for the following reasons.

1. From what I understand, the main motivation of introducing the sysavar is to
avoid update vote accounts for reward paying blocks. However, most of of those
vote accounts are likely to be updated anyways per block due to voting
transactions. And the payout reward update to vote accounts will happens in
write-cache. Extra update in write cache is not expensive. 

2. If we avoid creating this sysvar, we can avoid computing the delegation
rewards at the first block of the epoch, which will be nice.

HaoranYi:
I was confused by the comments and the actual SIMD. 

The SIMD didn't introduce a new sysvar. So my last comments should be like this.

~~It seems to me that the complexity of introducing a new sysvar for block revenue payout in this spec may not be worth it for the following reasons.~~

t seems to me that the complexity of collecting rewards into epoch reward sysvar for block revenue payout in this spec may not be worth it for the following reasons.

HaoranYi (approved):
Lgtm

proposals/0123-block-revenue-distribution.md
```diff
@@ -137,46 +137,64 @@ vote state field `pending_delegator_rewards` and added to the balance of vote
...
+
+Lastly, if this is the first block of epoch `N`, the vote state's
+`pending_delegator_rewards` field MUST be reset to `0` and `P` lamports MUST be
+deducted from the vote account's lamport balance and credited to the epoch
+rewards sysvar account's lamport balance.
+
+Note that unlike inflation rewards distribution, block revenue distribution will
+not impact any internal epoch rewards sysvar state fields like `total_rewards`
+or `distributed_rewards` since block revenue will instead be tracked via the
+epoch rewards sysvar lamport balance.
```
topointon-jump, line 173: This feels hacky and could lead to edge cases - how do we ensure that the sysvar account is rent-exempt, for example? Is it that bad to just add a field to the sysvar? Overloading the lamport balance feels like it could be confusing, is there a good reason for doing this?

jstarry: Since the reward lamports already exist onchain they need to live somewhere, unlike inflation rewards which are newly created for distribution. Adding another field might lead to more edge cases since we would need to track both the lamport balance and the field balance. The sysvar account already always exists so no need to worry about rent exempt balance here.

topointon-jump: The capatialization thing makes sense, thanks! It would be great to get a bit more clarification on how this interacts with the code for making sysvars always rent exempt.

When sysvars are saved, `update_sysvar_account` calls `adjust_sysvar_balance_for_rent` which ensures the balance is always rent-exempt. Are you proposing that the lamport balance of the epoch reward sysvar would be `minimum_balance_for_rent_exemption` + the reward lamports?

jstarry: > Are you proposing that the lamport balance of the epoch reward sysvar would be minimum_balance_for_rent_exemption + the reward lamports?

Yes, that's correct. On all clusters the sysvar already exists with the rent exempt balance, so this SIMD doesn't detail anything about topping up the sysvar account balance.

Btw I talk about the sysvar balance a bit more a little lower in the SIMD

> After distributing all partitioned delegator rewards, the epoch rewards sysvar balance MUST be reset to its rent exemption balance and any surplus lamports are burned.

topointon-jump (approved):

proposals/0123-block-revenue-distribution.md
```diff
@@ -137,46 +137,64 @@ vote state field `pending_delegator_rewards` and added to the balance of vote
...
+### Runtime: Block Revenue Distribution
 
-When calculating stake delegation rewards for a particular completed reward
-epoch, construct a list of all vote accounts that were initialized at the
-beginning of the reward epoch and had a non-zero active stake delegation. For
-each vote account, retrieve its state at the end of the reward epoch and check
-the `pending_delegator_rewards` field in its vote state. Let this value be `P`.
-If `P` is non-zero, use it to calculate rewards for each of the stake accounts
-delegated to the vote account as follows: 
+At the beginning of each epoch, pending delegator rewards will be collected from
```
t-nelson, line 142: > At the beginning of each epoch

"beginning" is any block with a parent in the previous epoch, as per inflation rewards calculations, right?

also this needs to happen before any transactions/votes (post ag) are processed

proposals/0123-block-revenue-distribution.md
```diff
@@ -137,46 +137,64 @@ vote state field `pending_delegator_rewards` and added to the balance of vote
...
-1. Sum all active stake delegated to the vote account during the reward epoch
-epoch. Let this total be `A`.
+#### Delegator Rewards Calculation
 
-2. For each individual stake account, multiply its active stake from the
-reward epoch by `P`, and divide the result by `A` using integer division.
-Discard any fractional lamports.
+Delegator rewards MUST be calculated during both the beginning of the first
+block of an epoch `N` and after restarting during partitioned rewards
+distribution.
```
t-nelson:
```suggestion
block of an epoch `N` and any time the node is restarted during the
partitioned rewards distribution period.
```

proposals/0123-block-revenue-distribution.md
```diff
@@ -137,46 +137,64 @@ vote state field `pending_delegator_rewards` and added to the balance of vote
...
 
-After calculating all individual stake rewards, sum them to obtain `D`, the
-total distribution amount. Because of integer division, the full amount `P` may
-not be distributed so compute the amount to be burned, `B`, as the difference
-between `P` and `D`.
+For each vote account, get its total active stake delegation
+during the reward epoch `N - 1`. Let this value be `A`. If `A` is zero, skip
+this vote account (any pending delegator rewards will be distributed in a future
+epoch when the vote account has active stake.)
```
t-nelson: what happens to pending delegator rewards in a vote account that has been intentionally destaked with the intention of shutting down the validator? so long as leader schedule lags stake, there's potential for delegator reward lamports to end up in a vote account with no stake

proposals/0123-block-revenue-distribution.md
```diff
@@ -137,46 +137,64 @@ vote state field `pending_delegator_rewards` and added to the balance of vote
...
 
 #### Individual Delegator Reward
 
-The stake reward distribution amounts for each stake account calculated above
-can then be used to construct a list of stake reward entries which MUST be
-partitioned and distributed according to [SIMD-0118].
+For each individual stake account with an active non-zero delegation, multiply
+its active stake in reward epoch `N - 1` by `P`, and divide the result by
+`A` using integer division to get the individual stake account's block revenue
+reward distribution amount.
```
t-nelson:
```suggestion
reward distribution amount, truncating any sub-lamport portion.
```

proposals/0123-block-revenue-distribution.md
```diff
@@ -137,46 +137,64 @@ vote state field `pending_delegator_rewards` and added to the balance of vote
...
-`pending_delegator_rewards` field and store the updated vote account. Finally,
-the burn amount `B` should also be deducted from the cluster capitalization.
+Then for each vote account, get its pending delegator rewards from the
+`pending_delegator_rewards` field in the vote state at the end of reward epoch
+`N - 1`. Let this value be `P`. If `P` is zero, skip this vote account as there
+is nothing to be distributed.
+
+Lastly, if this is the first block of epoch `N`, the vote state's
+`pending_delegator_rewards` field MUST be reset to `0` and `P` lamports MUST be
+deducted from the vote account's lamport balance and credited to the epoch
```
t-nelson: how will these debits be recorded on chain?

proposals/0123-block-revenue-distribution.md
```diff
@@ -137,46 +137,64 @@ vote state field `pending_delegator_rewards` and added to the balance of vote
...
+reward distribution amount.
+
+#### Delegator Rewards Distribution
 
-When reward entries are used to distribute rewards pool funds during partitioned
-rewards distribution, the delegated vote account for each rewarded stake account
-must have its `pending_delegator_rewards` field and its balance deducted with
-the amount of rewards distributed to keep capitalization consistent.
+The reward distribution amounts for each stake account can then be used to
+construct a list of stake reward entries which MUST be partitioned and
```
t-nelson:
```suggestion
construct a list of stake reward entries which MUST be partitioned, stored and
```

"in the bank"

proposals/0123-block-revenue-distribution.md
jstarry: Good point. In the future, I would like to fix the issue where leader schedule lags stake which would eliminate this issue. For now, any pending rewards for unstaked vote accounts should be directly credited to the vote account.

proposals/0123-block-revenue-distribution.md
jstarry: Good point. Though I don't think they should be recorded onchain, we should record these debits in the block rewards field as we do for other balance changes (rewards, fee revenue, rent due, etc). I don't think this needs to be described in the SIMD

proposals/0123-block-revenue-distribution.md
jstarry: this is an implementation detail

proposals/0123-block-revenue-distribution.md
jstarry: Correct

proposals/0123-block-revenue-distribution.md
jstarry: Revised here: https://github.com/solana-foundation/solana-improvement-documents/pull/300/commits/dbc39ce766dcdf0cd84e77715a15d01b12b60ca5

proposals/0123-block-revenue-distribution.md
t-nelson: right, 'cause rewards are reconstructed from, rather than stored directly in, snapshots

proposals/0123-block-revenue-distribution.md
t-nelson: this needs to be precise to ensure work is committed in the correct blocks, no? maybe we should just add a "be very familiar with simd0118" at the beginning of the document to be sure the reader is cognitively primed with the right assumptions about how this works. atm the simd0118 references are more of an aside, but we're really leaning on it pretty hard throughout

proposals/0123-block-revenue-distribution.md
t-nelson: are those fields only under consensus implicitly due to their denoting state updates? for some reason i thought that they were explicit

jstarry: Which fields are you talking about? If you mean the bank rewards, yeah they are not in consensus right now. They are just a convenient way to observe changes made to the account state. The bank hash only includes `parent_hash`, `blockhash`, `signature_count`, `accounts_lt_hash`, and rarely a `hard_fork` count.

proposals/0123-block-revenue-distribution.md
jstarry: Yeah agreed. I made the simd-0118 references more explicit below, lmk what you think: https://github.com/solana-foundation/solana-improvement-documents/pull/300/commits/ae7456744a45107ecbf8f0cc7812e7065aef9c78

proposals/0123-block-revenue-distribution.md
t-nelson: yeah the rewards field. pretty wrecked to leave these optional. i wasted 3hrs the other night trying to recover an unrecorded debit that ended up being rent collection

a problem for another simd i suppose, but we should probably at least mention that these transfers should be recorded in rewards field here

jstarry: Yeah let's follow up with a new SIMD for this. Until it's in consensus, it shouldn't be in the SIMD. We didn't detail the block rewards field record keeping in SIMD-0118 either

t-nelson: ah the glory of being primary implementer

t-nelson (approved):

topointon-jump (approved):

jstarry:
@t-nelson @topointon-jump @buffalojoec @Benhawkins18 

I think I spent too much energy thinking about the distribution mechanism and not enough about what we're actually sharing. As written, SIMD-0123 will allow validators to set a commission for ALL block revenue and the rest goes to delegators. This means that EVEN base fees will be shared with delegators. This feels wrong to me. We should only support sharing priority fees right? Base fees in theory pay for the base cost of tx compute?

Benhawkins18:
I lean towards keeping this simple: all block revenue (both base fees and priority fees) should flow through the same commission mechanism. Validators already have multiple potential revenue streams, and they can decide what to share (or not share) via their overall commission.

Introducing a separate carve-out just for base fees adds complexity for both validators and delegators. Delegators would need to understand that “commission” applies differently depending on the source of the revenue, which feels like unnecessary nuance. If everything is bundled, the mental model is much cleaner: delegators know they’re sharing in all block rewards minus validator commission, and validators retain the flexibility to share other income streams however they see fit.

So my preference is that SIMD-0123’s “block reward commission” applies uniformly across both base fees and prioritization fees.

tao-stones:
> I think I spent too much energy thinking about the distribution mechanism and not enough about what we're actually sharing. As written, SIMD-0123 will allow validators to set a commission for ALL block revenue and the rest goes to delegators. This means that EVEN base fees will be shared with delegators. This feels wrong to me. We should only support sharing priority fees right? Base fees in theory pay for the base cost of tx compute?

Conceptually I like the separation of base fee vs. priority fee from the revenue-sharing mechanism.

- If we treat the base fee as covering operational costs, it feels more right for validators to share “profit” rather than “revenue”

- That said, delegators do help secure leader slots, which in turn increases base fee collection, so it’s not entirely wrong to share base fees too.

- Personally I lean toward sharing priority fees only (thinking alone the lines of profit sharing), it doesn’t seem to introduce too much complexity.

tigarcia:
Isn't another motivation for SIMD-123 to allow operators to send fees to other keys besides the identity if they choose? It seems wrong to force them into sending it to the identity if they don't want to.

t-nelson:
do not discern priority fees. they need to die and we should make that as easy as possible

jstarry:
> Isn't another motivation for SIMD-123 to allow operators to send fees to other keys besides the identity if they choose? It seems wrong to force them into sending it to the identity if they don't want to.

Yes, that is something we are doing but it's technically part of https://github.com/solana-foundation/solana-improvement-documents/pull/232. That SIMD does refer to the new `block_revenue_collector` account as a "commission" account but the base fees would all get collected there even if they were not part of delegator rewards distribution. So that shouldn't be a concern.

> If we treat the base fee as covering operational costs, it feels more right for validators to share “profit” rather than “revenue”

This is what I was thinking as well. But now I'm realizing this doesn't make sense. Each transaction is processed by all the validators and if base fees were truly meant to cover cluster operational costs, they would get divided evenly (regardless of stake) to all validators rather than just getting paid to the current leader.

So now I'm thinking @Benhawkins18 and @t-nelson have the right idea that we shouldn't differentiate between base and priority fees here. It's all block revenue. We can always differentiate later if we really need to but I think it makes most sense to combine them for block revenue sharing with delegators. The whole point of a commission on block revenue (and collection to a specific collector address for bespoke solutions) is for validators to adjust how much revenue makes sense for their operational strategy rather than rely on base fees.

buffalojoec:
Just throwing my two cents in here since I was tagged!

I think the drawback of including both the base fees and the priority fees in the block commission is that ecosystem "standards" for evaluating a validator's commission could give larger validators a competitive edge.

With the fees combined, imagine a scenario where potential stakers are evaluating which node to stake with: a small node and a large node.
* The small node relies on some block revenue to survive, so they can't set a commission below N%.
* The large node has other revenue streams (ie. MEV) it can lean on to allow them to present a block commission of 0%.

To the uninformed retail staker, this might be grounds to stake with the larger node, to save that N%. Thus, the larger node can more effectively leverage low block commission rates to attract stake and increase their MEV profit.

I think keeping base fees separate could help level the playing field a bit. The small node could safely rely on base fee revenue, while still advertising a 0% block commission just like the large node.

jstarry:
Yeah but small stake nodes have small base fee revenue because you only earn base fees for blocks that you produce. Sure you will get compensated for executing transactions in your block, but that's not going to be enough to cover operational costs unless you set non-zero commission or have other revenue streams

buffalojoec:
> Yeah but small stake nodes have small base fee revenue because you only earn base fees for blocks that you produce. Sure you will get compensated for executing transactions in your block, but that's not going to be enough to cover operational costs unless you set non-zero commission or have other revenue streams

That's until you attract enough stake though. Base fee compensation would increase if you're competitively attracting new stake. It could be acceptable to operate at a diminishing loss for a bit as you grow your stake.

To be fair, this may not be a strong enough argument for splitting them, since it would mean we'd have less flexibility to redesign block revenue characteristics and we'd be stuck with another lengthy protocol change if enough operators request the ability to distribute all block revenue, but it's worth at least considering IMO.

joncinque:
I'm a big fan of this amendment for implementing SIMD-123. Considering it received approvals from Anza and Firedancer, are there any issues or concerns with re-opening and merging it?

PR #301, opened 2025-06-10
SIMD XXXX: Async Execution

MaxResnick:
This proposal represents the final feature flag for asynchronous execution on Solana. It will remove BankHash from votes, leaving behind only the BlockID and therefore allow validators to vote on blocks before they have finished executing them. the feature flag is intended to be activated after Alpenglow activates.


MaxResnick:
@Benhawkins18 the linter is complaining about line length in my tables and diagrams. Can you relax it.

lidatong:
i'm guessing the title is a typo and it means to say BankHash?

PR #302, opened 2025-06-12
SIMD-0302: BN254 G2 Arithmetic Syscalls

blockiosaurus:


Benhawkins18:
Previous discussion here: https://github.com/solana-foundation/solana-improvement-documents/discussions/293

samkim-crypto:
As previously discussed in https://github.com/solana-foundation/solana-improvement-documents/discussions/293, I am in support of this. Maybe we can also mention the fact that the same changes of [simd-284](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0284-alt-bn128-little-endian.md) should be applied for G2 as well in the simd.

 @0x0ece do you want to comment with an acknowledgement here too?

0x0ece:
Same here. Agree on little endian support.
Let’s also make sure deserialization is unambiguous (and we can fix G1 in parallel).

blockiosaurus:
Apologies for the delay. I threw in a section for endianness to be comprehensive if you can re-review.

0x0ece (approved):

samkim-crypto (approved):

Benhawkins18 (approved):
I see approvals from the right folks. Merging

PR #308, opened 2025-06-18
Withdrawal of SIMD-0163

Lichtso:
explains why in the "Security Considerations" section


buffalojoec (approved):
I don't think the details added to the security section are very clear, but I'm in favor of withdrawing the SIMD, on the basis of the low protocol gain to ruggability attack spread tradeoff.

PR #311, opened 2025-06-26
SIMD-0186: Clarify program/loader validation

2501babe:
the text is straightforward but where this lives and how its added is a bit tricky. we made these changes in the 186 agave impl, but they mirror changes in the 162 agave impl, and neither simd actually describes them. they are highly desirable changes because they drastically simplify the code and there is essentially no difference between them aside from whether the errors happen during execution or during loading

tbh we dont want to rename and rewrite the whole 186 text to expand the scope of the simd to fully describe account loading, so the simplest solutions are to either add it as an independent section (as done here) or make a new short "hey this is how program and loader validation works btw" simd. i think adding the section here is better because these changes are associated with the 186 code (although they reflect changes made in the 162 code) and because it is less likely for someone to miss them, because a new simd would be a random document tied to nothing

these changes were made in 186 because it was a fortuitous place to be able to delete complex code and replace it with simple code. the main concern is just that other validator clients replicate the changes


proposals/0186-loaded-transaction-data-size-specification.md
```diff
@@ -114,6 +114,35 @@ Account size for programs owned by LoaderV4 is left undefined. This SIMD should
 be amended to define the required semantics before LoaderV4 is enabled on any
 network.
 
+### Sidebar: Program ID and Loader Validation
+
+After all accounts are loaded, if the transaction's size comes at or under its
+loaded accounts data size limit, the program IDs of each instruction, along with
+their program owners, are validated. This process is unrelated to the rest of
+this SIMD, but is not defined elsewhere, so we define it here.
```
jstarry: I think it's fair to say that it *is* related to this SIMD because now that we don't need to load the owner account to track its size, we are able to make this simplification.

proposals/0186-loaded-transaction-data-size-specification.md
```diff
@@ -114,6 +114,35 @@ Account size for programs owned by LoaderV4 is left undefined. This SIMD should
...
+    * `LoaderV411111111111111111111111111111111111`
+
+If any of these conditions are violated, loading is aborted and the transaction
+pays fees.
+
+Previously, these checks were skipped if the program ID was
+`NativeLoader1111111111111111111111111111111` itself. This special case has been
+removed, and `NativeLoader1111111111111111111111111111111` behaves like any
+other account.
+
```
jstarry, line 149: I think we should also say that previously the program id's owner was validated by checking if it was equal to the native loader or, if not, was loaded and validated by checking that it was executable and that the owner's owner == native loader.

proposals/0186-loaded-transaction-data-size-specification.md
2501babe: ill reword it, im trying to adhere to the philosophy that the simd document is primary and the solana validators are concrete implementations of the simd. its unrelated in the sense that this simd is exclusively about how sizes are determined so the validation is sort of a tangent

2501babe:
@jstarry @mjain-jump if i could have your approvals on this, so foundation can merge

proposals/0186-loaded-transaction-data-size-specification.md
```diff
@@ -114,6 +114,41 @@ Account size for programs owned by LoaderV4 is left undefined. This SIMD should
...
+loaded transaction data size, but it is part of account loading (and consensus),
+and it is not explicitly defined elsewhere. So we define it here.
+
+The process is as follows; for each instruction's program ID:
+
+* Verify the account exists.
+* If [SIMD-0162](
+https://github.com/solana-foundation/solana-improvement-documents/pull/162)
+has not been activated, verify the program account is marked executable. When
+SIMD-0162 is active, skip this step.
```
jstarry: nit: SIMD-0162 is activated on all clusters so we can remove this now

proposals/0186-loaded-transaction-data-size-specification.md
```diff
@@ -114,6 +114,41 @@ Account size for programs owned by LoaderV4 is left undefined. This SIMD should
...
+
+After all accounts are loaded, if the transaction's size comes at or under its
+loaded accounts data size limit, the program IDs of each instruction, along with
+their program owners, are validated. This process is not part of determining
+loaded transaction data size, but it is part of account loading (and consensus),
+and it is not explicitly defined elsewhere. So we define it here.
+
+The process is as follows; for each instruction's program ID:
+
+* Verify the account exists.
```
jstarry: Can we add the expected error for this check? (`ProgramAccountNotFound`)

proposals/0186-loaded-transaction-data-size-specification.md
```diff
@@ -114,6 +114,41 @@ Account size for programs owned by LoaderV4 is left undefined. This SIMD should
...
+* If [SIMD-0162](
+https://github.com/solana-foundation/solana-improvement-documents/pull/162)
+has not been activated, verify the program account is marked executable. When
+SIMD-0162 is active, skip this step.
+* Verify the program account's owner is one of:
```
jstarry, line 128: Same here, can you add the expected error if this check fails? (`InvalidProgramForExecution`)

proposals/0186-loaded-transaction-data-size-specification.md
2501babe: [52c4a8e](https://github.com/solana-foundation/solana-improvement-documents/pull/311/commits/52c4a8ebc8319cb810113e23a2352defa32f9302)

proposals/0186-loaded-transaction-data-size-specification.md
2501babe: [52c4a8e](https://github.com/solana-foundation/solana-improvement-documents/pull/311/commits/52c4a8ebc8319cb810113e23a2352defa32f9302)

proposals/0186-loaded-transaction-data-size-specification.md
2501babe: [52c4a8e](https://github.com/solana-foundation/solana-improvement-documents/pull/311/commits/52c4a8ebc8319cb810113e23a2352defa32f9302)

PR #312, opened 2025-06-28
SIMD create-account-prefunded

rustopian:


proposals/0312-create-account-prefunded.md
```diff
@@ -0,0 +1,103 @@
...
+The existing `CreateAccount` wrapper instruction creates accounts within a
+program by use of `transfer`, `allocate`, and `assign`. However,
+`CreateAccount` fails if `lamports > 0`. This was intended as a protection
+measure, preventing developers from accidentally passing in some normal
+wallet address and permanently locking its lamports after `allocate` and
+`assign`.
+
+However, it is common practice to provide rent lamports to accounts prior to
+the actual creation (allocation and assigning) of the account space, rather
+than forcing the payer of the ATA account creation transaction to provide
```
joncinque: nit: the ATA program hasn't been mentioned until this point. Do you want to call it out more explicitly before this sentence or after?
```suggestion
than forcing the fee payer of the transaction to provide
```

proposals/0312-create-account-prefunded.md
```diff
@@ -0,0 +1,103 @@
...
+all of the required lamports. In this and similar instances, developers
+currently must manually construct a patched `CreateAccount` call of their
+own with 2-3 CPI calls: sometimes `Transfer`, then `Allocate`, and then
+`Assign`. While these actions themselves are minimally expensive, the overhead
+incurred with every Cross-Program Invocation - depth check, signer check,
+account copy, etc. - can make up the bulk of the computation done in the
+transaction.
```
joncinque: Can you also call out the rough number of CUs for this overhead? A minimum of 1k plus a variable amount based on instruction length and account length?

proposals/0312-create-account-prefunded.md
```diff
@@ -0,0 +1,103 @@
+---
+simd: '0312'
+title: CreateAccountPrefunded
```
joncinque: nit: I lean towards `CreatePrefundedAccount` as a name that sounds a bit more natural for the new instruction

proposals/0312-create-account-prefunded.md
```diff
@@ -0,0 +1,103 @@
...
+status: Review
+created: 2025-06-27
+feature: create-account-prefunded
+---
+
+## Summary
+
+A `CreateAccountPrefunded` instruction reduces network overhead for
+applications that need to fund account rent (in whole or in part) in advance
+of account creation.
```
joncinque: nit: can you specify that the new instruction is being added to the system program?

proposals/0312-create-account-prefunded.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Motivation
+
+The existing `CreateAccount` wrapper instruction creates accounts within a
```
joncinque: nit: same here, add that it's in the system program

proposals/0312-create-account-prefunded.md
```diff
@@ -0,0 +1,103 @@
...
+## New Terminology
+
+* Prefunded: an account which receives lamports to pay for its rent in whole
+or in part before its space is `allocate`d and its owner is `assign`ed.
+
+## Detailed Design
+
+`create_account_prefunded()` takes the same arguments as `create_account()`
+and performs the same operations; however, it does not fail if the new
+account's `lamports > 0`.
```
joncinque: This section could use a little more detailed information for other teams implementing validators, like Firedancer, including the following:

* exact instruction data format, ie the discriminant and expected value
* exact accounts in order, and their role (readonly, writable, signer, writable-signer)
* special cases. For example, if the transfer amount is 0, is a payer account even required?  If not, then we don't perform any checks on the payer account.

proposals/0312-create-account-prefunded.md
```diff
@@ -0,0 +1,103 @@
...
+This is a stopgap measure, as it extends the undesired interface sprawl of
+helper instructions such as `CreateAccount`. However, any redesign of
+Cross-Program Invocations to safely reduce overhead would be an extensive
+protocol upgrade and slow to land. In the meantime, the network will benefit
+from `CreateAccountPrefunded`.
+
+## New Terminology
+
+* Prefunded: an account which receives lamports to pay for its rent in whole
+or in part before its space is `allocate`d and its owner is `assign`ed.
```
joncinque: nit: the term would be "prefunded account" then. And isn't a prefunded account one that *may* receive lamports?

rustopian: Updated to Prefunded Account in 145e947. However, no, if it hasn't received lamports, it is no different, and is treated no differently, than a non-prefunded account. Or how do you mean?

proposals/0312-create-account-prefunded.md
rustopian: I came around to agree as we use same pattern in e.g. InitializeImmutableOwner. Changed throughout in 0a790cf.

proposals/0312-create-account-prefunded.md
rustopian: Of course, added

proposals/0312-create-account-prefunded.md
rustopian: Done, also added mention of p-ata for benchmarks (`CreatePrefundedAccount` saves 2,500 CUs) in 145e947

proposals/0312-create-account-prefunded.md
joncinque: I mean that in the instruction, the prefunded account *may* receive lamports. The possibilities are:
* the account has 0 lamports, it's not prefunded, so it *does* receive lamports
* the account is partially funded, but not enough for rent-exemption, so it's prefunded *and* it receives lamports
* the account has enough for rent-exemption, so it's prefunded *but* it receives no lamports

So what I mean: the optionality of receiving lamports is important

rustopian: Ah, I see what you mean. I'm defining it as an account that has received lamports prior to the create transaction/instruction, but the original definition sounds like it could refer the lamports `transfer` which occurs in the `CreatePrefundedAccount` call itself.

By "which receives lamports," I meant to refer to the funding in advance of creation (perhaps months before, for example).

I've clarified to:

```
* Prefunded Account: an account which has received lamports to pay for its rent
in whole or in part in any prior instruction or prior transaction, before the
instruction in which its space is `allocate`d and its owner is `assign`ed.
```

proposals/0312-create-account-prefunded.md
```diff
@@ -12,68 +12,118 @@ feature: create-account-prefunded
...
+| `space` | `u64` | Number of bytes of memory to allocate. |
+| `owner` | `Pubkey` | Address of program that will own the new account. |
+
+### Accounts
+
+The instruction requires the following accounts:
+
+| Index | Role | Description |
+|---|---|---|
+| 0 | `[WRITE, SIGNER]` | **Funding account**: The account that will pay
```
joncinque: What do you think about making this account optional? This would make it different from `CreateAccount`, but that should be ok

joncinque (commented):
Looks good to me! Just the last question that we've been kicking around a bit for the payer

proposals/0312-create-account-prefunded.md
rustopian: Updated in 25fddea

joncinque (commented):
This looks good to me!

rustopian:
@t-nelson or @jstarry, requesting a review at your convenience.

proposals/0312-create-account-prefunded.md
```diff
@@ -0,0 +1,168 @@
...
+category: Standard
+type: Core
+status: Review
+created: 2025-06-27
+feature: create-account-prefunded
+---
+
+## Summary
+
+A `CreatePrefundedAccount` instruction added to the system program reduces
```
jstarry: Might be too pedantic but to me this name contradicts itself. If an account is "prefunded," it has already been "created." And this instruction succeeds even if the account is not actually prefunded. What about `CreateAccountRelaxed` or `CreateOrModifyAccount`?

proposals/0160-static-instruction-limit.md
```diff
@@ -5,7 +5,7 @@ authors:
   - Andrew Fitzgerald (anza)
 category: Standard
 type: Core
-status: Draft
+status: Accepted
```
jstarry, line 8: can we remove this change from this PR?

rustopian: Branch just needed rebase, and it's gone 👍

rustopian:
> I assume the code path will be pretty much identical to the current `CreateAccount` instruction (minus the account ordering and lamports check)?

Yes, exactly.

proposals/0312-create-account-prefunded.md
rustopian: Thanks, after discussion and thought, `CreateAccountRelaxed` is generally liked.

What do you think about `CreateAccountAllowPrefund`?

jstarry: Like it!

rustopian:
> Let's give one more week for objections before merging.

Following up here.

joncinque:
Thanks for the feedback everyone -- merging!

PR #315, opened 2025-07-01
SIMD-0315: Loader-v3 to loader-v4 migration

Lichtso:
As requested in https://github.com/solana-foundation/solana-improvement-documents/pull/167#discussion_r2176598286

### DIFF
--- a/proposals/0315-loader-v3-v4-migration.md
+++ b/proposals/0315-loader-v3-v4-migration.md
@@ -7,7 +7,7 @@ category: Standard
 type: Core
 status: Review
 created: 2024-08-15
-feature: 2aQJYqER2aKyb3cZw22v4SL2xMX7vwXBRWfvS4pTrtED
+feature: TBD
 extends: SIMD-0167
 ---
 
@@ -23,7 +23,33 @@ loader-v4.
 
 ## Alternatives Considered
 
-None.
+### Loader-v1 programs
+
+Programs of loader-v1 have ABI v0, which is incompatible with ABI v1 used
+by programs of loader-v2, v3 and v4. Thus there is no way to migrate these.
+
+### Loader-v2 programs
+
+The two loader-v3 accounts per program in their sum are always larger than
+the one resulting loader-v4 account. Thus there is no need for additional
+funding. This would not be the case when migrating from loader-v2. Meaning
+that expanding this SIMD to cover it as well would require a funding source.
+
+### Global Migration: Coordinated in valiator or out of validator
+
+The global migration could be implemented in the validator, however:
+
+- If the global migration mechanism is inside the validator, the risk of it
+being detrimental to block production outweights any possible benefits.
+- It would have to be coordinated across all validator implementations,
+tested, fuzzed, etc. simply a whole lot more work for something which is only
+used once.
+- It being triggered manually per program or once (via a feature gate) for all
+programs changes nothing about it being controlled by a single key.
+- The only difference is in having more fine granular control over the
+timing in when a specific programs migration is triggered.
+- Doing it outside of the validator allows for the process to be aborted or
+patched quickly in case things start going sideways.
 
 ## New Terminology
 
@@ -31,12 +57,7 @@ None.
 
 ## Detailed Design
 
-The feature gate must:
-
-- enable loader-v4 `LoaderV411111111111111111111111111111111111` program
-management and execution (see SIMD-0167).
-- enable the loader-v3 `BPFLoaderUpgradeab1e11111111111111111111111`
-instruction `UpgradeableLoaderInstruction::Migrate`.
+The feature gate must enable the new loader-v3 instruction.
 
 ### Loader-v3 Instruction: Migrate
 
@@ -51,8 +72,9 @@ instruction `UpgradeableLoaderInstruction::Migrate`.
   otherwise throw `NotEnoughAccountKeys`
   - Check that the program data account is writable,
   otherwise throw `InvalidArgument`
-  - Check that the program data was last modified before the current slot
-  if the program data has the state `ProgramData`,
+  - Check that the last modified slot (stored in the program data accounts
+  header) is less than the current slot if the program data has the state
+  `ProgramData`,
   otherwise throw `InvalidArgument`
   - Check that the provided authority is either:
     - the migration authority
@@ -67,26 +89,26 @@ instruction `UpgradeableLoaderInstruction::Migrate`.
   otherwise throw `InvalidArgument`
   - Check that the program account is owned by loader-v3,
   otherwise throw `IncorrectProgramId`
-  - Check that the program account has the state `Program`,
-  otherwise throw `InvalidAccountData`
-  - Check that the program account points to the program data account,
-  otherwise throw `InvalidArgument`
-  - Clear the program account (setting its size to zero)
-  - Transfer all funds from the program data account to the program account
-  - Assign ownership of the program account to loader-v4
-  - If the program data account was not closed / empty or uninitialized:
-    - CPI loader-v4 `SetProgramLength` the program account to the size of the
-    program data account minus the loader-v3 header size and use the migration
-    authority.
+  - If the program account has the state `Program` and
+  the referenced program data account is owned by loader-v3:
+    - Set the length of the program account to 0
+    - Transfer all funds from the program data account to the program account
+    - Assign ownership of the program account to loader-v4
+    - CPI loader-v4 `SetProgramLength` the program account to the program data
+    account size minus the loader-v3 header size (45 bytes) and use the
+    provided authority.
     - CPI loader-v4 `Copy` the program data account into the program account
     - CPI loader-v4 `Deploy` the program account
     - If the program data account was finalized (upgrade authority is `None`):
-      - CPI loader-v4 `Finalize` without a next version forwarding
+      - CPI loader-v4 `Finalize`
     - otherwise, if the program data account was not finalized and the
     migration authority (as opposed to the upgrade authority) was provided:
       - CPI loader-v4 `TransferAuthority` to the upgrade authority
-  - Clear the program data account (setting its size to zero)
-  - Assign ownership of the program data account to the system program
+    - Set the length of the program data account to 0 (removing the header too)
+  - otherwise, if the program account is empty, has the state `Buffer` or has
+  the state `Program` but the program data account is not owned by loader-v3:
+    - Set the length of the program account to 0
+  - Set the `is_executable` flag of the program account to `false`
 
 ## Impact
 
--- a/proposals/0315-loader-v3-v4-migration.md
+++ b/proposals/0315-loader-v3-v4-migration.md
@@ -1,0 +1,121 @@
+---
+simd: '0315'
+title: Loader-v3 to loader-v4 migration
+authors:
+  - Alexander Meißner
+category: Standard
+type: Core
+status: Review
+created: 2024-08-15
+feature: 2aQJYqER2aKyb3cZw22v4SL2xMX7vwXBRWfvS4pTrtED
+extends: SIMD-0167
+---
+
+## Summary
+
+Migration of loader-v3 programs to loader-v4.
+
+## Motivation
+
+In order to remove the issues of loader-v3 (mentioned in SIMD-0167) from
+validator implementations, all remaining loader-v3 must be migrated to
+loader-v4.
+
+## Alternatives Considered
+
+None.
+
+## New Terminology
+
+None.
+
+## Detailed Design
+
+The feature gate must:
+
+- enable loader-v4 `LoaderV411111111111111111111111111111111111` program
+management and execution (see SIMD-0167).
+- enable the loader-v3 `BPFLoaderUpgradeab1e11111111111111111111111`
+instruction `UpgradeableLoaderInstruction::Migrate`.
+
+### Loader-v3 Instruction: Migrate
+
+- Instruction accounts:
+  - `[writable]` The program data account.
+  - `[writable]` The program account.
+  - `[signer]` The migration authority.
+- Instruction data:
+  - Enum variant `8u32`
+- Behavior:
+  - Check that there are at least three instruction accounts,
+  otherwise throw `NotEnoughAccountKeys`
+  - Check that the program data account is writable,
+  otherwise throw `InvalidArgument`
+  - Check that the program data was last modified before the current slot
+  if the program data has the state `ProgramData`,
+  otherwise throw `InvalidArgument`
+  - Check that the provided authority is either:
+    - the migration authority
+    (pubkey is `3Scf35jMNk2xXBD6areNjgMtXgp5ZspDhms8vdcbzC42`)
+    - or the upgrade authority stored in the program data account
+    - or the program signer if the program is finalized, closed or
+    uninitialized
+  otherwise throw `IncorrectAuthority`
+  - Check that the provided authority is a signer,
+  otherwise throw `MissingRequiredSignature`
+  - Check that the program account is writable,
+  otherwise throw `InvalidArgument`
+  - Check that the program account is owned by loader-v3,
+  otherwise throw `IncorrectProgramId`
+  - Check that the program account has the state `Program`,
+  otherwise throw `InvalidAccountData`
+  - Check that the program account points to the program data account,
+  otherwise throw `InvalidArgument`
+  - Clear the program account (setting its size to zero)
+  - Transfer all funds from the program data account to the program account
+  - Assign ownership of the program account to loader-v4
+  - If the program data account was not closed / empty or uninitialized:
+    - CPI loader-v4 `SetProgramLength` the program account to the size of the
+    program data account minus the loader-v3 header size and use the migration
+    authority.
+    - CPI loader-v4 `Copy` the program data account into the program account
+    - CPI loader-v4 `Deploy` the program account
+    - If the program data account was finalized (upgrade authority is `None`):
+      - CPI loader-v4 `Finalize` without a next version forwarding
+    - otherwise, if the program data account was not finalized and the
+    migration authority (as opposed to the upgrade authority) was provided:
+      - CPI loader-v4 `TransferAuthority` to the upgrade authority
+  - Clear the program data account (setting its size to zero)
+  - Assign ownership of the program data account to the system program
+
+## Impact
+
+This changes enables the migration of programs from loader-v3 to loader-v4
+without changing their program address via a new loader-v3 instruction. This
+will count as a redeployment and thus render the program unavailable for the
+rest of the slot (delay visibility).
+
+Once new programs can not be deployed on loader-v3 anymore, the list of all
+loader-v3 programs becomes fixed and can be extracted from a snapshot. Using
+the added loader-v3 migration instruction and the global migration authority,
+the core protocol developers will then migrate all loader-v3 programs to
+loader-v4 programs, which once completed:
+
+- removes the need to copy ELFs during program loading to align them.
+- allows transaction account loading to be simplified, because every program
+would load exactly one account, no need to load the proxy account to get to
+the actual program data (which is not listed in the transaction accounts).
+- allows the removal of the write lock demotion exception if loader-v3 is
+present in a transaction.
+- allows dApp devs to resuscitate closed loader-v3 programs if they still
+control the program authority. This allows redeployment at the same address
+or completely closing the program account in order to retrieve the locked
+funds.
+
+## Security Considerations
+
+None.
+
+## Backwards Compatibility
+
+None.


Semen2322 (approved):

mjain-jump (approved):

proposals/0315-loader-v3-v4-migration.md
```diff
@@ -0,0 +1,121 @@
...
+  - Check that there are at least three instruction accounts,
+  otherwise throw `NotEnoughAccountKeys`
+  - Check that the program data account is writable,
+  otherwise throw `InvalidArgument`
+  - Check that the program data was last modified before the current slot
+  if the program data has the state `ProgramData`,
+  otherwise throw `InvalidArgument`
+  - Check that the provided authority is either:
+    - the migration authority
+    (pubkey is `3Scf35jMNk2xXBD6areNjgMtXgp5ZspDhms8vdcbzC42`)
```
jstarry, line 81: I don't think it's a good idea for anyone to have the unilateral authority to migrate programs even if it's a relatively harmless operation. I prefer the migration to be done by the runtime, maybe automatically at an epoch boundary.

proposals/0315-loader-v3-v4-migration.md
```diff
@@ -0,0 +1,121 @@
...
+    - CPI loader-v4 `Deploy` the program account
+    - If the program data account was finalized (upgrade authority is `None`):
+      - CPI loader-v4 `Finalize` without a next version forwarding
+    - otherwise, if the program data account was not finalized and the
+    migration authority (as opposed to the upgrade authority) was provided:
+      - CPI loader-v4 `TransferAuthority` to the upgrade authority
+  - Clear the program data account (setting its size to zero)
+  - Assign ownership of the program data account to the system program
+
+## Impact
```
jstarry, line 113: It looks like the migrate instruction was not added to the list of bpf loader instructions that can be invoked via cpi. So a multisig owned program would need to first transfer the authority to a keypair based signer to do the migration, right?

proposals/0315-loader-v3-v4-migration.md
```diff
@@ -0,0 +1,121 @@
...
+  otherwise throw `InvalidArgument`
+  - Check that the program data was last modified before the current slot
+  if the program data has the state `ProgramData`,
+  otherwise throw `InvalidArgument`
+  - Check that the provided authority is either:
+    - the migration authority
+    (pubkey is `3Scf35jMNk2xXBD6areNjgMtXgp5ZspDhms8vdcbzC42`)
+    - or the upgrade authority stored in the program data account
+    - or the program signer if the program is finalized, closed or
+    uninitialized
```
jstarry, line 84: Can you explain what you mean by an uninitialized program here?

proposals/0315-loader-v3-v4-migration.md
```diff
@@ -0,0 +1,121 @@
...
+This changes enables the migration of programs from loader-v3 to loader-v4
+without changing their program address via a new loader-v3 instruction. This
+will count as a redeployment and thus render the program unavailable for the
+rest of the slot (delay visibility).
+
+Once new programs can not be deployed on loader-v3 anymore, the list of all
+loader-v3 programs becomes fixed and can be extracted from a snapshot. Using
+the added loader-v3 migration instruction and the global migration authority,
+the core protocol developers will then migrate all loader-v3 programs to
+loader-v4 programs, which once completed:
```
jstarry, line 124: I would prefer if the mass migration was defined and activated in a separate SIMD + feature gate but could be convinced otherwise.
1. This SIMD is useful already without global migration
2. This SIMD is already big and complex enough without global migration
3. Global migration could use some more discussion about alternative approaches to the global migration authority

proposals/0315-loader-v3-v4-migration.md
```diff
@@ -0,0 +1,121 @@
...
+  - Check that the program data account is writable,
+  otherwise throw `InvalidArgument`
+  - Check that the program data was last modified before the current slot
+  if the program data has the state `ProgramData`,
+  otherwise throw `InvalidArgument`
+  - Check that the provided authority is either:
+    - the migration authority
+    (pubkey is `3Scf35jMNk2xXBD6areNjgMtXgp5ZspDhms8vdcbzC42`)
+    - or the upgrade authority stored in the program data account
+    - or the program signer if the program is finalized, closed or
```
jstarry, line 83: I'm in favor of migrating closed programs to allow resuscitation but I think migrating finalized programs should be done as part of the mass migration because they are more sensitive because they are presumed to be in use, never unavailable (delayed visibility), and there is no guarantee that the program keypair is even stored or secured.

proposals/0315-loader-v3-v4-migration.md
```diff
@@ -0,0 +1,121 @@
+---
+simd: '0315'
+title: Loader-v3 to loader-v4 migration
+authors:
+  - Alexander Meißner
+category: Standard
+type: Core
+status: Review
+created: 2024-08-15
+feature: 2aQJYqER2aKyb3cZw22v4SL2xMX7vwXBRWfvS4pTrtED
```
jstarry: This should have a different feature gate from the loader-v4 simd. Is there any reason this needs to be the same feature gate?

proposals/0315-loader-v3-v4-migration.md
```diff
@@ -0,0 +1,121 @@
...
+      - CPI loader-v4 `TransferAuthority` to the upgrade authority
+  - Clear the program data account (setting its size to zero)
+  - Assign ownership of the program data account to the system program
+
+## Impact
+
+This changes enables the migration of programs from loader-v3 to loader-v4
+without changing their program address via a new loader-v3 instruction. This
+will count as a redeployment and thus render the program unavailable for the
+rest of the slot (delay visibility).
```
jstarry, line 118: Is it possible to not have the migration incur the delay visibility? When finalized programs are migrated it would be best to not make them temporarily unavailable even for one slot. And similarly, depending on how non-upgrade-authority migrations are performed (eg migration authority or some other approach), it would be best to not make them unavailable as it could fail transactions later in the slot.

proposals/0315-loader-v3-v4-migration.md
```diff
@@ -0,0 +1,121 @@
...
+    authority.
+    - CPI loader-v4 `Copy` the program data account into the program account
+    - CPI loader-v4 `Deploy` the program account
+    - If the program data account was finalized (upgrade authority is `None`):
+      - CPI loader-v4 `Finalize` without a next version forwarding
+    - otherwise, if the program data account was not finalized and the
+    migration authority (as opposed to the upgrade authority) was provided:
+      - CPI loader-v4 `TransferAuthority` to the upgrade authority
+  - Clear the program data account (setting its size to zero)
+  - Assign ownership of the program data account to the system program
```
jstarry: I don't see this in the agave implementation FYI

proposals/0315-loader-v3-v4-migration.md
```diff
@@ -0,0 +1,121 @@
...
+  - Check that the program account is writable,
+  otherwise throw `InvalidArgument`
+  - Check that the program account is owned by loader-v3,
+  otherwise throw `IncorrectProgramId`
+  - Check that the program account has the state `Program`,
+  otherwise throw `InvalidAccountData`
+  - Check that the program account points to the program data account,
+  otherwise throw `InvalidArgument`
+  - Clear the program account (setting its size to zero)
+  - Transfer all funds from the program data account to the program account
```
jstarry: This should be more explicitly specified. There are two cases:
1. The programdata account is still owned by the loader-v3 program. In this case, we can directly debit lamports from the programdata account and credit the program account.
2. The programdata account is owned by the system program. In that case, we cannot directly debit the account. I think we could either skip the transfer (my preference) or we would have to do a native cpi invocation to the system program to do the transfer.

proposals/0315-loader-v3-v4-migration.md
```diff
@@ -0,0 +1,121 @@
...
+  - Assign ownership of the program account to loader-v4
+  - If the program data account was not closed / empty or uninitialized:
+    - CPI loader-v4 `SetProgramLength` the program account to the size of the
+    program data account minus the loader-v3 header size and use the migration
+    authority.
```
jstarry: nit: I think this should say the "provided authority" because "migration authority" in this SIMD refers to the static core dev one, right?

proposals/0315-loader-v3-v4-migration.md
```diff
@@ -0,0 +1,121 @@
...
+  - Check that the program account has the state `Program`,
+  otherwise throw `InvalidAccountData`
+  - Check that the program account points to the program data account,
+  otherwise throw `InvalidArgument`
+  - Clear the program account (setting its size to zero)
+  - Transfer all funds from the program data account to the program account
+  - Assign ownership of the program account to loader-v4
+  - If the program data account was not closed / empty or uninitialized:
+    - CPI loader-v4 `SetProgramLength` the program account to the size of the
+    program data account minus the loader-v3 header size and use the migration
```
jstarry: nit: should specify that it's the programdata header and also call out the exact byte length here: `45`

proposals/0315-loader-v3-v4-migration.md
```diff
@@ -0,0 +1,121 @@
...
+  otherwise throw `MissingRequiredSignature`
+  - Check that the program account is writable,
+  otherwise throw `InvalidArgument`
+  - Check that the program account is owned by loader-v3,
+  otherwise throw `IncorrectProgramId`
+  - Check that the program account has the state `Program`,
+  otherwise throw `InvalidAccountData`
+  - Check that the program account points to the program data account,
+  otherwise throw `InvalidArgument`
+  - Clear the program account (setting its size to zero)
```
jstarry: nit: clear the program account data (setting its data length to zero)

proposals/0315-loader-v3-v4-migration.md
```diff
@@ -0,0 +1,121 @@
...
+- Instruction data:
+  - Enum variant `8u32`
+- Behavior:
+  - Check that there are at least three instruction accounts,
+  otherwise throw `NotEnoughAccountKeys`
+  - Check that the program data account is writable,
+  otherwise throw `InvalidArgument`
+  - Check that the program data was last modified before the current slot
+  if the program data has the state `ProgramData`,
+  otherwise throw `InvalidArgument`
```
jstarry: nit: should explicitly specify that the last modification slot comes from the programdata state

proposals/0315-loader-v3-v4-migration.md
```diff
@@ -0,0 +1,121 @@
...
+    program data account minus the loader-v3 header size and use the migration
+    authority.
+    - CPI loader-v4 `Copy` the program data account into the program account
+    - CPI loader-v4 `Deploy` the program account
+    - If the program data account was finalized (upgrade authority is `None`):
+      - CPI loader-v4 `Finalize` without a next version forwarding
+    - otherwise, if the program data account was not finalized and the
+    migration authority (as opposed to the upgrade authority) was provided:
+      - CPI loader-v4 `TransferAuthority` to the upgrade authority
+  - Clear the program data account (setting its size to zero)
```
jstarry: Similarly we need to handle the programdata account differently depending on whether it's system or loader-v3 owned

proposals/0315-loader-v3-v4-migration.md
Lichtso: programdata account is always owned by loader-v3. Otherwise the program was invalid to begin with and would not be executable.

proposals/0315-loader-v3-v4-migration.md
Lichtso: Again, programdata account is always owned by loader-v3 for valid programs. Edit: Closed programs can get in this situation.

proposals/0315-loader-v3-v4-migration.md
Lichtso: That is when you allocate an account, fund it and assign it to loader-v3 but don't do anything further. Loader-v3 never initialized it.

proposals/0315-loader-v3-v4-migration.md
Lichtso: This SIMD was already split off from the loader-v4 one and is not even that long anymore. I think the global migration very much belongs in here as well as the program local one. Both use the same logic / mechanism, only the trigger is different.

proposals/0315-loader-v3-v4-migration.md
Lichtso: Delay visibility is necessary because the index of the global program cache assumes that no two versions of a program coexist within the same slot.

proposals/0315-loader-v3-v4-migration.md
jstarry: If a program was closed, its programdata account would be deleted and could then receive lamports to become a system owned account.

proposals/0315-loader-v3-v4-migration.md
Lichtso: Explained the reasons against doing so in the alternatives section.

proposals/0315-loader-v3-v4-migration.md
Lichtso: Think @buffalojoec wanted to split the feature gate as well

proposals/0315-loader-v3-v4-migration.md
jstarry: Ok that's what I thought but in the agave implementation we only migrate accounts that deserialize to an initialized program account so uninitialized programs aren't handled. I think we can leave it out from the proposal and match the implementation.

proposals/0315-loader-v3-v4-migration.md
jstarry: I'm not very familiar with the inner workings of the program cache but could we make it work? Programs are executed exactly the same way (excluding validator perf differences) no matter if they are loader-v3 or loader-v4 owned right?

proposals/0315-loader-v3-v4-migration.md
Lichtso: Yes, you are right. It could be used to prevent closed programs from being resuscitated. But skipping the transfer might not work because loader-v4 accounts are slightly larger (by 3 bytes) than loader-v3 program accounts (to reach 16 byte alignment).

proposals/0315-loader-v3-v4-migration.md
jstarry: Thanks for writing that up. Is there a reason we don't just allow the migration instruction to be run permissionlessly by anyone?

proposals/0315-loader-v3-v4-migration.md
Lichtso: Well, that would be a bug IMO. Rather change the implementation to support that too. After all people might have funds stuck this way.

proposals/0315-loader-v3-v4-migration.md
jstarry: If our goal is to migrate ALL loader-v3 programs, that would include closed programs too right? So we shouldn't leave a way to block the migration. If we can't skip the transfer, seems like we'll need an optional system program instruction account to facilitate a system program cpi

proposals/0315-loader-v3-v4-migration.md
jstarry: I'm not sure they would be stuck, couldn't they initialize as a buffer and then withdraw?

proposals/0315-loader-v3-v4-migration.md
Lichtso: No the issue is literally a collision in the map `(Pubkey, Slot) => LoadedProgram`. You can't have two `LoadedProgram` entries for the same `(Pubkey, Slot)` key.

proposals/0315-loader-v3-v4-migration.md
Lichtso: Could be done too fast (too many at once). I would rather have them be done one at a time and slowly work though a fixed list of loader-v3 programs.

proposals/0315-loader-v3-v4-migration.md
Lichtso: Yes, maybe. Would still add extra steps for the global migration later.

proposals/0315-loader-v3-v4-migration.md
Lichtso: How about we approach this the other way around: Closed programs are closed after all, same goes for uninitialized. Why bother migrating them to loader-v4, if we could directly give them to the system program for users to withdraw from (provided they still have the keys). Only issue with that is resetting the `is_executable` flag.

proposals/0315-loader-v3-v4-migration.md
jstarry: Why would there be extra steps? I don't think I'm following now. Seems like we can just ignore any uninitialized accounts because they wouldn't be able to become initialized after the loader-v3 deploys are disabled

Lichtso: The goals are:

- We get rid of all loader-v3 owned accounts
- Users get their locked funds back
- The global migration only has to call `Migrate` on every program account once

A mentioned in the other thread, we can also approach this the other way around and not migrate closed and uninitialized programs.

Closing this thread and let's continue over here: https://github.com/solana-foundation/solana-improvement-documents/pull/315#discussion_r2182457202

proposals/0315-loader-v3-v4-migration.md
jstarry: Yeah that sounds worth exploring to me. For closed programs, we currently need to load their programdata account to know if they are closed or not. But after the migration, execution of loader-v3 programs will be disabled and we will know that only closed programs are left and no longer need to load their programdata accounts. Recovering funds could happen well after loader v3 deploys / execution is disabled.

proposals/0315-loader-v3-v4-migration.md
buffalojoec: Yeah, I think it should be separate. I also was under the impression separate SIMDs meant separate feature gates as well. Not sure though.

proposals/0315-loader-v3-v4-migration.md
```diff
@@ -0,0 +1,142 @@
...
+loader-v4.
+
+## Alternatives Considered
+
+### Loader-v1 and loader-v2 programs
+
+The two loader-v3 accounts per program in their sum are always larger than
+the one resulting loader-v4 account. Thus there is no need for additional
+funding. This would not be the case when migrating from older loaders. Meaning
+that expanding this SIMD to cover these as well would require a funding source.
```
buffalojoec: To be honest, why not include it? If we just include the caveat that it requires pre-funding, then anyone who _wants_ the benefits of v4 can choose to pay for it. Lots of the SPL programs come to mind as well.

proposals/0315-loader-v3-v4-migration.md
buffalojoec: Heh, we could deploy an on-chain program to manage rate-limiting on this. So, it would be effectively permissionless, but the program could track how many have been submitted per-slot and deny past a certain limit.

I have to agree I'm also apprehensive about a keypair that can single-handedly change even a frozen program's state, although I recognize the similarity to a feature keypair.

proposals/0315-loader-v3-v4-migration.md
```diff
@@ -0,0 +1,142 @@
...
+None.
+
+## Detailed Design
+
+The feature gate must:
+
+- enable loader-v4 `LoaderV411111111111111111111111111111111111` program
+management and execution (see SIMD-0167).
+- enable the loader-v3 `BPFLoaderUpgradeab1e11111111111111111111111`
+instruction `UpgradeableLoaderInstruction::Migrate`.
```
buffalojoec: Two feature gates might not be bad, just to let a bunch of first-time v4 deployments go out and bake before we enable mass migrations of v3 programs. What do you think?

proposals/0315-loader-v3-v4-migration.md
buffalojoec: > When finalized programs are migrated it would be best to not make them temporarily unavailable even for one slot.

I would even extend this idea to non-finalized programs as well. If the ELF didn't change, there's no reason the program should have to pay with a slot.

@Lichtso since the ELF isn't changing, and we are guaranteeing that in the loader v3 instruction, could we add a codepath to the program cache that handles in-place mutation of a loaded entry? It sounds sketchy typing it out, but it might be worth exploring.

proposals/0315-loader-v3-v4-migration.md
buffalojoec: Have we ever done anything like this before? What kind of coordination goes into a global migration like the one proposed? Just curious if we have any precedents.

proposals/0315-loader-v3-v4-migration.md
```diff
@@ -0,0 +1,148 @@
...
+
+Programs of loader-v1 have ABI v0, which is incompatible with ABI v1 used
+by programs of loader-v2, v3 and v4. Thus there is no way to migrate these.
+
+### Loader-v2 programs
+
+The two loader-v3 accounts per program in their sum are always larger than
+the one resulting loader-v4 account. Thus there is no need for additional
+funding. This would not be the case when migrating from loader-v2. Meaning
+that expanding this SIMD to cover it as well would require a funding source.
```
buffalojoec, line 36: Any reason why we couldn't just make the instruction support v2 -> v4 migrations, even if we don't plan to crank those ourselves?

proposals/0315-loader-v3-v4-migration.md
buffalojoec: I still vote for two separate features!

proposals/0315-loader-v3-v4-migration.md
```diff
@@ -0,0 +1,148 @@
...
+  - Enum variant `8u32`
+- Behavior:
+  - Check that there are at least three instruction accounts,
+  otherwise throw `NotEnoughAccountKeys`
+  - Check that the program data account is writable,
+  otherwise throw `InvalidArgument`
+  - Check that the last modified slot (stored in the program data accounts
+  header) is less than the current slot if the program data has the state
+  `ProgramData`,
+  otherwise throw `InvalidArgument`
```
buffalojoec, line 78: Maybe verify the header is `ProgramData` first?

proposals/0315-loader-v3-v4-migration.md
buffalojoec: @Lichtso any thoughts on doing this through an on-chain program? It could give us some better security guarantees and relax any concerns over a god mode key.

proposals/0315-loader-v3-v4-migration.md
buffalojoec: I don't think the trigger matters for the delayed visibility, does it? Seems like no matter who calls the migration, there will be a delayed visibility slot.

> they are more sensitive because they are presumed to be in use, never unavailable (delayed visibility)

This is a really salient point. We should see if there's a way to avoid halting execution on finalized programs.

proposals/0315-loader-v3-v4-migration.md
```diff
@@ -0,0 +1,148 @@
...
+    - Set the length of the program account to 0
+    - Transfer all funds from the program data account to the program account
+    - Assign ownership of the program account to loader-v4
+    - CPI loader-v4 `SetProgramLength` the program account to the program data
+    account size minus the loader-v3 header size (45 bytes) and use the
+    provided authority.
+    - CPI loader-v4 `Copy` the program data account into the program account
+    - CPI loader-v4 `Deploy` the program account
+    - If the program data account was finalized (upgrade authority is `None`):
+      - CPI loader-v4 `Finalize` without a next version forwarding
```
buffalojoec:
```suggestion
      - CPI loader-v4 `Finalize`
```

proposals/0315-loader-v3-v4-migration.md
```diff
@@ -0,0 +1,148 @@
...
+    - otherwise, if the program data account was not finalized and the
+    migration authority (as opposed to the upgrade authority) was provided:
+      - CPI loader-v4 `TransferAuthority` to the upgrade authority
+    - Set the length of the program data account to 0 (removing the header too)
+  - otherwise, if the program account is empty, has the state `Buffer` or has
+  the state `Program` but the program data account is not owned by loader-v3:
+    - Set the length of the program account to 0
+  - Set the `is_executable` flag of the program account to `false`
```
buffalojoec, line 111: Should we just error if someone presents a v3 buffer account here? It can still be used to deploy a v4 program, since `Copy` allows it, so there's no need to support it here. If we just need a way to close them, maybe we keep the `Close` instruction active on loader v3?

proposals/0315-loader-v3-v4-migration.md
Lichtso: You mean giving the migration authority to an on-chain program? A program only runs if somebody sends an transaction. It is still the same controlling of the timing, with extra steps.

> god mode key

Again, the migration authority has almost no say in anything, but the timing.

nazreen: @Lichtso I think you should at least include what @jstarry said under Alternatives Considered

> Thanks for writing that up. Is there a reason we don't just allow the migration instruction to be run permissionlessly by anyone?

apfitzge: > Could be done too fast (too many at once). I would rather have them be done one at a time and slowly work though a fixed list of loader-v3 programs.

Can you explain what too fast means? Why is that an issue, unless the CU cost is incorrect? If txs to migrate programs are winning bids to get executed in a block, then that's what the block should be filled with, right?

Lichtso: Yes, as long as the migration is triggered from the outside via TXs it should not be a problem. But, if it is done runtime internally then the CU budget of the block would be off.

apfitzge: Right, so why not just make the instruction permissionless to upgrade? Then you can also do what @buffalojoec said, and allow upgrade of v2 programs to v4. As long as their willing to pay the rent-difference for size.

Downside would be that permissionless upgrade -> user could censor any program for one slot by calling the migration. But I think from dev perspective it's not really that different from US censoring them for one slot?

Overall it seems better than having a new authority key, at least to me. Please let me know if there's something I'm missing.

Lichtso:
Was thinking about the suggestion of @deanmlittle in Rome: Having a union of the ELF and the enum discriminator at the beginning of the (loader-v3) program account. But that would also break existing programs trying to read the program data account because they typically don't derive the PDA, they read the pubkey from the program account.

However, only moving the ELF from the program data account to the end of the program account (just concatenation, not trying to overlap the structures) could work, at least for programs which only read the loaders metadata and not the ELF. This still solves two performance issues of loader-v3: Misalignment of the ELF in the account and following the indirection of the program data account in every TX.

We would still offer the loader-v3 to v4 migration path and disable all other loader-v3 program management instructions, so that we can have loader-v4 as a core program on-chain and eliminate all built-in programs. We would not rewrite all existing loader-v3 programs to v4, but instead move the ELF from the program data account to the program account. This way finalized loader-v3 programs which read the loaders metadata would stay functional.

buffalojoec:
> Was thinking about the suggestion of @deanmlittle in Rome

I'm in favor of this. We should consider introducing a new SIMD to address the change to Loader v3 standalone. Then, this one can remove all mentions of forced migration and the god mode key. 🙌🏼

buffalojoec:
> But that would also break existing programs trying to read the program data account because they typically don't derive the PDA, they read the pubkey from the program account.

I don't think we should split hairs here. We're already going to break stuff by moving the ELF from one account to the other, and if we try to mitigate that damage by introducing a different discriminator to the PDA (formerly programdata account), then we might as well do the same thing in the program account and make that discriminator the ELF magic, as stated in the original suggestion.

Lichtso:
> introducing a different discriminator to the PDA (formerly programdata account)

I think that was never any suggestion by anyone, or at least that is news to me.

> program account and make that discriminator the ELF magic

As stated, that does not work because the program account does not only contain the enum discriminator, but also the Pubkey (https://github.com/anza-xyz/solana-sdk/blob/6b7745f8467365c8fd6829d5a2c98d689cc6a7fe/loader-v3-interface/src/state.rs#L21C5-L24C7) thus the `programdata_address: Pubkey` would collide with the ELF if the ELF was at the start at the program account. And again, we would have to move the ELF behind the existing struct in the program account to make it work.

buffalojoec:
> > introducing a different discriminator to the PDA (formerly programdata account)
> 
> I think that was never any suggestion by anyone, or at least that is news to me.
> 
> > program account and make that discriminator the ELF magic
> 
> As stated, that does not work because the program account does not only contain the enum discriminator, but also the Pubkey (https://github.com/anza-xyz/solana-sdk/blob/6b7745f8467365c8fd6829d5a2c98d689cc6a7fe/loader-v3-interface/src/state.rs#L21C5-L24C7) thus the `programdata_address: Pubkey` would collide with the ELF if the ELF was at the start at the program account. And again, we would have to move the ELF behind the existing struct in the program account to make it work.

If you move the ELF out of the `ProgramData` account without changing the discriminator, that's a pretty egregious breaking change. If you do it while introducing a new discriminator, say `UpgradeableLoaderState::ProgramDataV2`, then you can at least make the claim that developers should be keying on discriminators anyway, in order to validate they are working with the account state they expect. This cushions the blow on the sudden change in state, but it's breaking nonetheless.

I'm suggesting we apply the same method of thinking to the `Program` account, and just omit any additional payload (such as the PDA pubkey) from the new `ProgramV2` state variant. You either break both account states by introducing new discriminators or you break neither, but it makes no sense to permit the removal of the entire ELF in one account but forbid the removal of a public key in another.

Lichtso:
> it makes no sense to permit the removal of the entire ELF in one account but forbid the removal of a public key in another.

That is not necessarily true, the two have very different use cases and usage patterns. Every program that inspects a loader-v3 program needs to lookup that pubkey pointing to the program-data account. Many programs only look for metadata fields, typically the update authority. Few programs ever inspect or look for the ELF.

> that's a pretty egregious breaking change.

It is and the reason why I was initially opposed to it, but it does have some merit.

> If you do it while introducing a new discriminator, say `UpgradeableLoaderState::ProgramDataV2`

Which is essentially loader-v4, plus / minus the changes to the instruction interface.

buffalojoec:
> That is not necessarily true, the two have very different use cases and usage patterns. Every program that inspects a loader-v3 program needs to lookup that pubkey pointing to the program-data account. Many programs only look for metadata fields, typically the update authority. Few programs ever inspect or look for the ELF.

A good rule of thumb for on-chain programs that rely on values contained in account state is to first verify ownership, then a discriminator, if there is one. If they weren't doing this, and they got unexpected state, they could be reading values they don't expect.

Even if the ELF bytes themselves are not part of the `UpgradeableLoaderState::ProgramData` enum fields, their presence after the metadata is part of the account state specification. To change that account state in any way should be done under a new variant. This is the least disruptive way to introduce such a change, so the aforementioned checks can fail fast.

That being said, I also consider the _addition_ of an ELF to either state layout as yet another change to the specification, so the new program account containing the ELF should also get its own new variant. If we're going to redefine the two accounts' state specifications, why not remove fields we don't need anymore?

While it's true that changing the owner of these accounts to introduce a new state specification would also respect my logic above, determining the best way to perform such a migration has proven highly contentious. By retaining the same owner, we gain more flexibility in migration implementation, and avoid centralized control over frozen programs by core contributors.

The Loader v3 program is the entrypoint for all BPF program invocations on the network. It's completely within Solana's program-ownership ruleset to perform such an account migration (`Program` -> `ProgramV2` & `ProgramData` -> `ProgramDataV2`) during program execution. All you would need to do to trigger a migration is invoke the program.

Lichtso:
It is true that all the contents in the program account as well as the program data account are part of the protocol, not just the enum and structure. That is not what I was debating.

The entire point of that idea (moving ELF inside loader-v3) was to "trick" programs into thinking that everything else stays the same, so they stay compatible. When you introduce new enum variants in loader-v3 you lose that advantage completely and there is no more reason to prefer this over loader-v4, which changes enum variant, owner and also the instruction interface.

> By retaining the same owner, we gain more flexibility in migration implementation, and avoid centralized control over frozen programs by core contributors.

Changing the owner is not more or less special that any other change to existing accounts. All will require a feature gate or migration authority: The feared "centralized control". Why ever that is so scary here but not in any other SIMD is beyond me.

buffalojoec:
> The entire point of that idea (moving ELF inside loader-v3) was to "trick" programs into thinking that everything else stays the same, so they stay compatible.

That's not the way I interpreted the proposed change. To me, it seemed the intention was twofold:
* Retain the same account ownership while accomplishing the goals of Loader v4, such as aligned ELFs in the program account, not the secondary account.
* Store nothing but the ELF in the program account, simplifying the dumping and inspection of on-chain program accounts.

It was also mentioned that a bonus benefit is the ability for the PDA (effectively the "metadata" account) to be more flexible as a result, since changing its state would no longer involve ELFs.

Perhaps @deanmlittle can ameliorate my argument, since it was his suggestion.

deanmlittle:
> > The entire point of that idea (moving ELF inside loader-v3) was to "trick" programs into thinking that everything else stays the same, so they stay compatible.
> 
> 
> 
> That's not the way I interpreted the proposed change. To me, it seemed the intention was twofold:
> 
> * Retain the same account ownership while accomplishing the goals of Loader v4, such as aligned ELFs in the program account, not the secondary account.
> 
> * Store nothing but the ELF in the program account, simplifying the dumping and inspection of on-chain program accounts.
> 
> 
> 
> It was also mentioned that a bonus benefit is the ability for the PDA (effectively the "metadata" account) to be more flexible as a result, since changing its state would no longer involve ELFs.
> 
> 
> 
> Perhaps @deanmlittle can ameliorate my argument, since it was his suggestion.

I can think of many benefits, but the top few that came to mind that led to me making this suggestion include:

1) Nothing but the ELF file would be present in the program account, solving the indirection problem and giving us the same behavior as every other loader
2) Modifying program metadata to allow more granular controls or support new use cases becomes extremely low stakes
3) It becomes much cheaper to check/modify metadata such as upgrade authority of a program in runtime
4) Migration path can be made permissionless as owner remains unchanged

proposals/0315-loader-v3-v4-migration.md
febo: I also agree that it would be better to have permissionless upgrades and avoid a "master" authority.

febo:
> > The entire point of that idea (moving ELF inside loader-v3) was to "trick" programs into thinking that everything else stays the same, so they stay compatible.
> 
> That's not the way I interpreted the proposed change. To me, it seemed the intention was twofold:
> 
> * Retain the same account ownership while accomplishing the goals of Loader v4, such as aligned ELFs in the program account, not the secondary account.
> * Store nothing but the ELF in the program account, simplifying the dumping and inspection of on-chain program accounts.
> 
> It was also mentioned that a bonus benefit is the ability for the PDA (effectively the "metadata" account) to be more flexible as a result, since changing its state would no longer involve ELFs.
> 
> Perhaps @deanmlittle can ameliorate my argument, since it was his suggestion.

This is a good (short) summary of the ideas we discussed in Rome, which sound very interesting. I think there is a bit of overlap between what is being discussed here and Loader v4, in the sense that some of the changes here (alignment of the ELF) is one of the benefits of Loader v4 but it seems that we could achieve the same result by improving Loader v3.

Lichtso:
There are multiple aspects here which are conflated all the time, but they are independent:

- The layout of the loader-v3 program accounts (which is where the performance issues comes from)
- The instruction interface of loader-v3 (which limits some use cases and exposes security risks for dApp devs)
- The implementation of loader-v3 (which causes an excessive amount of security issues for the validator)

loader-v3 is not like the other built-ins. It is deeply intertwined with all parts of the general runtime (even outside the program runtime) and that makes it extremely hard to change without all sorts of unintended side effects and risking of breaking the validator. IMO the least risky way is to first make sure that there are no more loader-v3 programs and then carefully cutting it out of the code base. Attempting to fundamentally change it while it is also Atlas carrying the entire chain on its shoulders is insane.

> Migration path can be made permissionless as owner remains unchanged

Again, and again: The owner field influences the instruction interface for the most part. It has nothing to do with the permissions of a migration. If we only have dApp devs manually migrate (which won't work for finalized programs and others which have lost their authority key), then we can do without a global authority. Otherwise, if we migrate all programs we need a global authority or feature gate. Sticking with loader-v3 or going with loader-v4 (owner field) does not change anything about this.

buffalojoec:
> > Migration path can be made permissionless as owner remains unchanged
> 
> Again, and again: The owner field influences the instruction interface for the most part. It has nothing to do with the permissions of a migration. If we only have dApp devs manually migrate (which won't work for finalized programs and others which have lost their authority key), then we can do without a global authority. Otherwise, if we migrate all programs we need a global authority or feature gate. Sticking with loader-v3 or going with loader-v4 (owner field) does not change anything about this.

I think you're maybe missing what I'm suggesting and why it's enabled by the owner being unchanged.

In the Vote program, we migrate between state versions as certain instructions (like `InitializeAccount`, `Authorize`, etc.) are invoked. We can do this because the Vote program owns the account involved and the lamport accounting checks out.

I'm saying in the Loader v3 program, we could do the exact same thing. The awesome part with the loader is that it just so happens that Loader v3 is the program being invoked by the runtime when the underlying BPF program is being invoked [here](https://github.com/anza-xyz/agave/blob/30a9e90372edebe95031ef18f11b9a300827ea72/program-runtime/src/invoke_context.rs#L516-L525). So, it can legally mutate account state without having to be deliberately invoked for such a job.

Lichtso:
> The awesome part with the loader is that it just so happens that Loader v3 is the program being invoked by the runtime when the underlying BPF program is being invoked [here](https://github.com/anza-xyz/agave/blob/30a9e90372edebe95031ef18f11b9a300827ea72/program-runtime/src/invoke_context.rs#L516-L525). So, it can legally mutate account state without having to be deliberately invoked for such a job.

Um, no! When a program is being executed / invoked it (the program account) has no write lock, it is always read-only. Further more you would be invoked as a loader-v3 program, not as loader-v3 itself. The code you linked clearly reads `owner_id` not `program_id`.

> why it's enabled by the owner being unchanged.

Again, complete nonsense. For the 7th or so time (I am a broken record at this point) the owner field has nothing to do with anything here. Please, everybody take a look at the code that is confusing all of you: https://github.com/anza-xyz/agave/blob/30a9e90372edebe95031ef18f11b9a300827ea72/transaction-context/src/instruction_accounts.rs#L357 

For a modification of a loader-v3 account you need to be invoked as loader-v3 (not as a program owned by it), yes, but that does not matter. It is the same regardless of what kind of modification you are doing, be it changing the structure inside loader-v3 or gifting the account to loader-v4. The global authority or feature gate suggested in this SIMD would only be used as a trigger to kick off the migration. It is not going to change or circumvent any of the existing program runtime rules.

PR #317, opened 2025-07-08
SIMD-0317: Enforce 32 data + 32 coding shreds

0x0ece:
All clients currently send 32 data + 32 coding shreds, however Turbine still allows to send and receive a variable number of shreds. With this change we will enforce sending and receiving only 32 data + 32 coding shreds.

### DIFF
--- a/proposals/0317-enforce-32-data-shreds.md
+++ b/proposals/0317-enforce-32-data-shreds.md
@@ -1,0 +1,55 @@
+---
+simd: '0317'
+title: Enforce 32 data + 32 coding shreds
+authors:
+  - Emanuele Cesena
+category: Standard
+type: Core
+status: Review
+created: 2025-07-08
+feature:
+---
+
+## Summary
+
+All clients currently send 32 data + 32 coding shreds, however Turbine still allows
+to send and receive a variable number of shreds. With this change we will enforce
+sending and receiving only 32 data + 32 coding shreds.
+
+## Motivation
+
+It is inconvinient to support many combinations of data + coding shreds.
+There's no security benefit, and in fact security is reduced when the number of
+shreds is low.
+
+## New Terminology
+
+N/A
+
+## Detailed Design
+
+A sender should always produce 32 data shreds + 32 coding shreds per FEC set
+(this is currently already happening).
+
+Receivers currently accept FEC sets with variable number of shreds.
+
+If `enforce_32_data_shreds: <PUBKEY>`
+is active, then any FEC set with a number of shreds different than 32 data + 32 coding
+will be dropped on ingest.
+
+## Alternatives Considered
+
+Leave as is.
+
+## Impact
+
+Clients will no longer accept FEC sets (thus blocks) with any number of shreds
+different than 32 data + 32 coding.
+
+## Security Considerations
+
+Security is improved since the (minimum) number of shreds is now 32 + 32.
+
+## Backwards Compatibility
+
+This feature is not backwards compatible.
--- a/proposals/0317-enforce-32-data-shreds.md
+++ b/proposals/0317-enforce-32-data-shreds.md
@@ -19,8 +19,17 @@ sending and receiving only 32 data + 32 coding shreds.
 ## Motivation
 
 It is inconvinient to support many combinations of data + coding shreds.
-There's no security benefit, and in fact security is reduced when the number of
-shreds is low.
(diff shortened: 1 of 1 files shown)


ptaffet-jump:
Definitely in favor! We should clarify if these shreds are still valid for a duplicate block proof.

proposals/0317-enforce-32-data-shreds.md
```diff
@@ -0,0 +1,55 @@
...
+
+All clients currently send 32 data + 32 coding shreds, however Turbine still allows
+to send and receive a variable number of shreds. With this change we will enforce
+sending and receiving only 32 data + 32 coding shreds.
+
+## Motivation
+
+It is inconvinient to support many combinations of data + coding shreds.
+There's no security benefit, and in fact security is reduced when the number of
+shreds is low.
```
lidatong: maybe elaborate on the security benefit? can reference https://github.com/solana-foundation/solana-improvement-documents/pull/204/files#diff-e9a92f44cc9e8d1789705b2a4a51bba175df745484c6638e475514d14eefe716R146

basically equivocation detection is simplified, because you have to see the same shred idx to see the equivocation (vs. before you could have overlapping windows)

proposals/0317-enforce-32-data-shreds.md
```diff
@@ -0,0 +1,55 @@
...
+## New Terminology
+
+N/A
+
+## Detailed Design
+
+A sender should always produce 32 data shreds + 32 coding shreds per FEC set
+(this is currently already happening).
+
+Receivers currently accept FEC sets with variable number of shreds.
```
lidatong, line 43: can we also add a note here that the FEC set payload must be exactly equal to 31840 bytes, with 995 bytes of payload per data shred?

0x0ece:
Thanks for the feedback! I'm going to collect comments from Anza (and rest of the world) then apply all changes at once.

proposals/0317-enforce-32-data-shreds.md
qkniep: yes, the only check required to detect equivocation after this change should be: did I see any two shreds in the same FEC set, with different Merkle roots and valid signatures

alexpyattaev: 32:32 makes a lot of checks around whether a particular shred index is valid or not trivial. Current logic to verify all that is very involved since we do not get a clear picture of where each FEC set starts unless we have seen shreds from all/most of them.

alexpyattaev:
Would it be appropriate to provision some mechanism to evolve the protocol to support larger FEC set size later, or do we expect to redesign shred format by then?

0x0ece:
> Would it be appropriate to provision some mechanism to evolve the protocol to support larger FEC set size later, or do we expect to redesign shred format by then?

In another SIMD maybe?

proposals/0317-enforce-32-data-shreds.md
0x0ece: thanks a lot, I've included all 3 comments

proposals/0317-enforce-32-data-shreds.md
0x0ece: done!

lidatong (approved):

bw-solana:
I think I'm good with the proposal and direction.

The only question I have is around implementation and what happens for a malformed block. Do we do anything to actively prevent repair from trying to get the missing shreds?

bw-solana (approved):

alexpyattaev:
> The only question I have is around implementation and what happens for a malformed block. Do we do anything to actively prevent repair from trying to get the missing shreds?

This would have to be a feature gate => everyone stops admitting oversized FEC sets at once => no problem with repair (since invalid shreds would be dropped by everybody for new blocks). Repair should be able to work with any shred configs for a while after transition.

0x0ece:
The way I'd do it is that if feat gate gets activated at epoch E, slot N, any shred with slot >= N must be 32+32 (or it gets dropped). Any shred with slot < N can be as it is today (turbine, repair, anything...). Starting from epoch E+1 we can remove all old code. Like if this was a parsing rule for the shred.

topointon-jump:
@Benhawkins18 can we merge this? 🙏

topointon-jump:
@jacobcreech could we merge this? 🙏

PR #320, opened 2025-07-10
Update Access Controls - SIMD-0007

Benhawkins18:
We are updating SIMD-0007 to reflect that Foundation no longer needs to approve every PR; simplifying the access policy removes the Foundation as a bottleneck.

### Summary of changes
**CI:**
Add github CI requirement to have at least one approval from Jump and at least one approval from Anza 

**SIMD-0007:**
Collapse three access tiers into two: Write and Maintain.

State that every PR requires one approval from jump and one from anza (enforced by CI).

Clarify privileges:

Write – open PRs, self-merge after checks pass.

Maintain – merge any PR, manage repo settings, grant/revoke access.

Update vouching and removal rules to fit the two-tier model.

### DIFF
--- a/.github/workflows/require-jump-anza.yml
+++ b/.github/workflows/require-jump-anza.yml
@@ -17,6 +17,10 @@ permissions:
 
 jobs:
   check:
+    # Skip the job on pure "commented" review submissions to avoid noisy failures
+    if: >
+      github.event_name != 'pull_request_review' ||
+      (github.event.review.state != 'commented')
     runs-on: ubuntu-latest
     steps:
       - name: Verify Jump + Anza approvals
@@ -25,15 +29,17 @@ jobs:
           script: |
             // github, context, core are injected by github-script
 
-            // ──────────── 1. MANUAL ROSTERS ────────────
+            // ──────────── 1) MANUAL ROSTERS (update as needed) ────────────
             const jumpApprovers = ['ptaffet-jump', 'topointon-jump', '0x0ece', 'lidatong', 'ripatel-fd', 'benhawkins18', 'jacobcreech'];
             const anzaApprovers = ['t-nelson', 'sakridge', 'bw-solana', 'benhawkins18', 'jacobcreech'];
-            // ───────────────────────────────────────────
+            // If desired, exclude foundation names from error hints:
+            const hidden = new Set(['benhawkins18','jacobcreech']);
+            // ───────────────────────────────────────────────────────────────
 
             const pr = context.payload.pull_request;
             if (!pr) { core.setFailed('No pull_request context'); return; }
 
-            // 2. fetch *all* reviews on this PR
+            // 2) fetch all reviews
             const { data: reviews } = await github.rest.pulls.listReviews({
               owner: context.repo.owner,
               repo: context.repo.repo,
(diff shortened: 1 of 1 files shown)


buffalojoec (commented):
Why not just use the built-in support for GitHub teams on the platform itself? This script seems like a major burden to keep up-to-date, and it's also missing a few reviewers.

It would be a lot better to have two teams:
* Anza core devs
* Firedancer core devs

And then we could go a step further and use some teams for SME-related stuff, depending how annoying this gets:
* Fees/tx pipeline
* Alpenglow
* VM
* etc.

You could add the two core dev team reviewers automatically as required, and then just overhaul the `Type` field to use keywords than can bring in other SME-based reviewers (ie. `Type: VM`).

jstarry:
> You could add the two core dev team reviewers automatically as required, and then just overhaul the Type field to use keywords than can bring in other SME-based reviewers (ie. Type: VM).

I think defining SME (subject matter expert) teams is a good idea but sometimes SIMDs don't fall into specific categories so I don't think we need to try to tie the SIMD "type" to a certain SME team just yet. Getting the appropriate SME approval should be a responsibility of either the SIMD author and core team approvers

buffalojoec:
> > You could add the two core dev team reviewers automatically as required, and then just overhaul the Type field to use keywords than can bring in other SME-based reviewers (ie. Type: VM).
> 
> I think defining SME (subject matter expert) teams is a good idea but sometimes SIMDs don't fall into specific categories so I don't think we need to try to tie the SIMD "type" to a certain SME team just yet. Getting the appropriate SME approval should be a responsibility of either the SIMD author and core team approvers

Agreed, so we could just have that field be omitted if there isn't a focus. Maybe just label it "SME".

Benhawkins18:
> This script seems like a major burden to keep up-to-date, and it's also missing a few reviewers.

The intention is to keep the list of reviewers to the same folks that are doing backport reviews at Anza and similarly small list at Jump. They will be responsible for making sure the appropriate SME's have signed off. I do not think there will be any significant update burden in this case. Trivial to update a purposely small list if needed. 

> Why not just use the built-in support for GitHub teams on the platform itself? 

GitHub's built-in codeowners and branch protection rules don't currently support enforcing the logic we need: specifically, ensuring there's at least one approval from Jump Crypto and one approval from Anza. They only let you specify how many total approvals you need not which teams or groups those approvals must come from.

If GitHub ever adds the ability to enforce "at least one approval from each of these separate teams," I'd happily drop this script entirely. Until then, this is the simplest way I found to enforce the desired policy and avoid the Foundation being a merge bottleneck.

If you're aware of any simpler native GitHub method I might have missed, let me know

buffalojoec:
> GitHub's built-in codeowners and branch protection rules don't currently support enforcing the logic we need: specifically, ensuring there's at least one approval from Jump Crypto and one approval from Anza. They only let you specify how many total approvals you need not which teams or groups those approvals must come from.

We have this feature in Agave. It looks like we just combine the "Require approvals from codeowners" setting with a folder-based codeowners setup: https://github.com/anza-xyz/agave/blob/master/.github/CODEOWNERS

Benhawkins18:
> We have this feature in Agave. It looks like we just combine the "Require approvals from codeowners" setting with a folder-based codeowners setup: https://github.com/anza-xyz/agave/blob/master/.github/CODEOWNERS

Yes, this is the first approach I looked at but as I mentioned they only let you specify how many total approvals you need not which teams or groups those approvals must come from.

.github/workflows/require-jump-anza.yml
```diff
@@ -0,0 +1,116 @@
...
+                  // COMMENTED etc – ignore
+              }
+            }
+
+            // DEBUG 2: effective map
+            core.info('=== Effective state per reviewer ===');
+            Object.entries(status).forEach(([u, s]) =>
+              core.info(`${u}: approved=${s.approved}, blocked=${s.blocked}`));
+
+            // 4. final approved list (approved && not blocked)
```
jstarry: I think a single block from a group member should override all approvers from that group. Right now you could have a situation where a group member approves and another member of the same group requests changes and the result is that the group check passes.

jstarry:
Also I just noticed that after I comment on this PR, I get a notification that the approval checks are failing. Pretty annoying so we should fix that.

Benhawkins18:
@jstarry I addressed your comments. Good suggestions.

.github/workflows/require-jump-anza.yml
```diff
@@ -17,6 +17,10 @@ permissions:
 
 jobs:
   check:
+    # Skip the job on pure "commented" review submissions to avoid noisy failures
+    if: >
+      github.event_name != 'pull_request_review' ||
+      (github.event.review.state != 'commented')
```
jstarry, line 23: nit: I think this is hard to read, can we change to the following?

```
  !(github.event_name == 'pull_request_review' &&
    github.event.review.state == 'commented')
```

jstarry (approved):
small suggestion, otherwise looks good!

Separately, we should make sure that all approvals are dismissed when a new commit is pushed.

jstarry:
nvm the notification thing is still borked. And there are two instances of the approval check.. one for `pull_request_review` and one for `pull_request`

jstarry:
I think we should use a bot which detects approval conditions and then once the bot approves, the PR can be merged.

Benhawkins18:
> I think we should use a bot which detects approval conditions and then once the bot approves, the PR can be merged.

Yeah, upon reflecting I think that's the better way to do it. I'm going to look into doing it this way. I also don't like that the current form is a CI block and that all PRs will show a red CI X even though they may just be missing approvals.

Benhawkins18:
> Separately, we should make sure that all approvals are dismissed when a new commit is pushed.

I considered this but it's not uncommon that people make small changes like spelling error fixes or title tweaks. I'm hesitant to make approvers continuously re approve. 

@jacobcreech any thoughts on this one?

jstarry:
> I considered this but it's not uncommon that people make small changes like spelling error fixes or title tweaks. I'm hesitant to make approvers continuously re approve.

It's a small sacrifice to uphold the integrity of the review process (ie that last minute changes can't get added without an approver explicitly acknowledging them)

PR #322, opened 2025-07-17
SIMD-0322: Introduce Serial Execution CU Tracking and Constraint System

igor56D:


igor56D:
Review of this SIMD was postponed due to broader discussions around asynchronous execution and the cost model. Some of those upstream topics have since been resolved, likely unblocking review for this SIMD.

bw-solana:
@apfitzge who is the right person to SME from Anza side?
@ptaffet-jump are you the right person from FD side?

apfitzge:
> @apfitzge who is the right person to SME from Anza side?

@tao-stones and myself.

PR #332, opened 2025-08-06
cha cha 8

bw-solana:


ripatel-fd:
I just wrote a simple implementation of ChaCha20RNG ChaCha8RNG in AVX-512. Here are my results.

This is on EPYC 9B45 3.7GHz Zen 5.

Does anyone know how many RNG values are sampled for, let's say, a node with 1% of the network's stake at 100M CU block limit, per second? I'm curious how many milliseconds per slot we'd spend pulling out RNG samples with an optimized implementation such as the one I wrote. (fd_chacha20rng_ulong is the actual RNG sampling, and fd_chacha20rng_refill_avx512 is the raw throughput) 

```
Benchmarking fd_chacha20rng_ulong
  ~ 48.540 Gbps            / core
  ~758.436 Mulong / second / core
  ~  1.319 ns / ulong
Benchmarking fd_chacha20rng_refill_avx512
  ~ 60.388 Gbps / core
Benchmarking fd_chacha20rng_refill_avx
  ~ 30.631 Gbps / core
Benchmarking fd_chacha20rng_refill_seq
  ~  3.895 Gbps / core

Benchmarking fd_chacha8rng_ulong
  ~  88.133 Gbps            / core
  ~1377.086 Mulong / second / core
  ~   0.726 ns / ulong
Benchmarking fd_chacha8rng_refill_avx512
  ~139.071 Gbps / core
```

https://github.com/firedancer-io/firedancer/pull/5974

ripatel-fd:
Ignoring above benchmark, I think this is an obviously good proposal that various people have raised. I think we should do it. The only thing that speaks against it is the core dev work of plumbing through feature activation logic to Turbine and the associated implementation risk. Which is manageable.

sakridge (approved):

ripatel-fd:
```c
/* Solana uses different mechanisms of mapping a ulong to an unbiased
   integer in [0, n) in different parts of the code.  In particular,
   leader schedule generation uses MODE_MOD and Turbine uses MODE_SHIFT.
   See the note in fd_chacha20_rng_ulong_roll for more details. */
#define FD_CHACHA_RNG_MODE_MOD   1
#define FD_CHACHA_RNG_MODE_SHIFT 2
```

Unrelated to this proposal but still relevant: Turbine RNG uses a different rejection method than leader schedule RNG.
The modulo rejection method uses fewer RNG samples but uses a more expensive rejection method (division).
Since we can generate 1.3 billion raw RNG samples per second now, we should definitely use the shift method, so Turbine is good as-is.

In case we change the leader schedule derivation in the future to use ChaCha8-RNG, we should also make it use the shift rejection method.

PR #313, opened 2025-06-30
SIMD-0313: Drop unchained merkle shreds

AshwinSekar:
Backfilling the SIMD for this feature https://github.com/anza-xyz/feature-gate-tracker/issues/80

PR #333, opened 2025-08-06
SIMD-0327: Serialize block ID in bank into snapshot

emwang-jump:


proposals/0333-snapshot-include-block-id.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Summary
+
+Currently the block ID (the merkle root of the last FEC set) of the
+snapshot slot is not included in the snapshot Bank Fields. This would
+enforce the snapshot producer to include the block ID in the serialized
+bank.
+
+## Motivation
+
```
AshwinSekar, line 21: Additionally it would be good to specify that in Alpenglow votes / certificates and related tracking is only performed on (Slot, block_id), so it's imperative that we have block_id available in snapshots.

EDIT: also alpenglow repair uses block id, parent linking, duplicate block resolution etc. Basically we want this very much :)

proposals/0333-snapshot-include-block-id.md
```diff
@@ -0,0 +1,57 @@
...
+## New Terminology
+
+N/A
+
+## Detailed Design
+
+The block_id already lives in the bank. Snapshot producers for all
+clients must include the block ID of the snapshot slot in the serialized
+bank of the snapshot. This can be implemented by adding a `block_id`
+field to `BankFieldsToSerialize` in Agave.
```
lidatong:
```suggestion
bank of the snapshot. This can be implemented by adding a `block_id`
field to the snapshot serialization schema (eg. `BankFieldsToSerialize` in Agave or `fd_snapshot_manifest_t` in Firedancer).
```

lidatong (dismissed):

proposals/0333-snapshot-include-block-id.md
```diff
@@ -0,0 +1,62 @@
...
+resolution uses block ID. Having the block ID available in snapshots
+enables this.
+
+## New Terminology
+
+N/A
+
+## Detailed Design
+
+The block_id already lives in the bank. Snapshot producers for all
```
brooksprumo, line 36: > The block_id already lives in the bank.

Is that the `Bank.block_id` field? I see its definition currently is:

```rust
    block_id: RwLock<Option<Hash>>,
```

from:
https://github.com/anza-xyz/agave/blob/45d2f9da3bc2c8cb6932e86d17809a7dd967bbda/runtime/src/bank.rs#L896

What should happen if this value is `None` when a snapshot is being generated? Is that valid, or must block_id always be `Some`?

I see below it says that clients must include the block ID, so mostly wanted to confirm that block_id in the Bank will always be Some.

proposals/0333-snapshot-include-block-id.md
```diff
@@ -0,0 +1,62 @@
...
+clients must include the block ID of the snapshot slot in the serialized
+bank of the snapshot. This can be implemented by adding a `block_id`
+field to the snapshot serialization schema (eg. `BankFieldsToSerialize`
+in Agave or `fd_snapshot_manifest_t` in Firedancer).
+
+Snapshot consumers may optionally to read the block ID in the
```
brooksprumo: Seems like there's a word missing or one too many? Maybe removing the "to" is the intention? Or maybe s/optionally/opt/?

proposals/0333-snapshot-include-block-id.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Security Considerations
+
+Security is improved because this now allows inter-slot verification
+between the snapshot slot and its child without repairing the block ID
+of the snapshot slot.
+
+## Backwards Compatibility
+
+This feature is backwards compatible.
```
brooksprumo: Changes to the snapshot format are unfortunately backwards incompatible, but we have known ways to support snapshot format changes to mitigate this.

Since snapshots must be loadable by adjacent versions, then if version `X` is the version we write `block_id`, then version `X-1` must be able to read—and likely ignore—this field.

We do this by added the new field to the end of the snapshot manifest as an Option and with a "default if eof". (I know many already know this, but wanted to be explicit here in the SIMD for more folks that didn't have the same background.)

proposals/0333-snapshot-include-block-id.md
AshwinSekar: This is a good point, `block_id` is an `Option` for two reasons:
- The `block_id` cannot be filled in until the last data shred for the block is ingested. 
For non-leader banks it is set by replay once the bank is frozen [1]
For leader banks it is set by broadcast stage when shredding is complete. [2]
The broadcast stage change is not yet implemented in Agave, only in the Alpenglow repo at the moment - it will need to be upstreamed to support this SIMD.
- Until recently https://github.com/solana-foundation/solana-improvement-documents/pull/313, we still supported shreds of the legacy format which did not specify the block id.

When implementing block ID serialization into the snapshot, we must guarantee that any bank that is being snapshotted has the block id filled in as Some. 

For non-leader banks I believe this is guaranteed, as the bank must be frozen before snapshotting. 

For leader banks, we could have a scenario where the bank is frozen, however shredding and dissemination is not yet complete. If we wish to snapshot a leader block we must wait until shredding is complete and the block id is filled in. 
EDIT: Correct me if I'm wrong, but I believe this is also a non-issue, since we only snapshot rooted banks. For a bank to reach rooted status, it must have been voted on by the cluster, meaning that it must have been fully shredded and thus the block_id should have been set.

[1] https://github.com/anza-xyz/agave/blob/74de4eca3696114c84f1beb5a381ba8efe55a80d/core/src/replay_stage.rs#L3196
[2] https://github.com/anza-xyz/alpenglow/blob/2c817f583208d312ee6f3939cceeff2845401533/turbine/src/broadcast_stage/standard_broadcast_run.rs#L341

brooksprumo: > I believe this is also a non-issue, since we only snapshot rooted banks. For a bank to reach rooted status, it must have been voted on by the cluster, meaning that it must have been fully shredded and thus the block_id should have been set.

Correct, we only snapshot rooted banks. For the validator, this is straight-forward.

When creating a snapshot with ledger-tool, we may create a new bank to add/remove accounts/feature-gates/hard-forks/etc, and then than bank gets rooted as part of taking the snapshot. Will `block_id` still be `Some` in this case? We won't be doing any replay/voting on it.

AshwinSekar: If we create an artificial bank without actual shreds, then the block id cannot be set.

If for example the snapshot is being used to restart a cluster:
1) Cluster goes down, restart slot is X
2) Perform some ledger tool operations to modify state and create a snapshot for X + 1
3) Restart the cluster on X + 1

We'll need the snapshot to specify a block_id for X + 1 and have it be deterministic for all operators that create the snapshot independently.

I would lean towards ledger tool always filling in a block_id based on the modifications it performs, but open to suggestions.

brooksprumo: > I would lean towards ledger tool always filling in a block_id based on the modifications it performs, but open to suggestions.

Works for me!

lidatong: > I would lean towards ledger tool always filling in a block_id based on the modifications it performs, but open to suggestions.

agreed -- this will also be much better for FD's conformance work

lidatong:
@brooksprumo looks like @emwang-jump addressed your comments, could you take another look?

proposals/0333-snapshot-include-block-id.md
brooksprumo: Resolved.

proposals/0333-snapshot-include-block-id.md
brooksprumo: Resolved.

proposals/0333-snapshot-include-block-id.md
brooksprumo: Resolved.

lidatong:
> One last question: Can the Block ID live on-chain instead of in the snapshot?

do you mean whether it should be part of a block?

lidatong:
the benefit of putting it in the snapshot is it makes the snapshot verifiable with the blocks you are received over repair while catching up (ie. the first FEC after the snapshot slot has a chained merkle root that matches what's inside the snapshot). i think if it's only inside a block you can no longer verify that

brooksprumo:
> do you mean whether it should be part of a block?

No, in an account, e.g. new sysvar.

lidatong:
> > do you mean whether it should be part of a block?
> 
> No, in an account, e.g. new sysvar.

oh that's an interesting idea. the vote program could also do verification of the tower_sync ixs in that case. though trade-off is it would probably be a more invasive change?

so you would also move it out of the bank in that case? would this also have benefits for alpenglow?

brooksprumo:
> oh that's an interesting idea. the vote program could also do verification of the tower_sync ixs in that case. though trade-off is it would probably be a more invasive change?
> 
> so you would also move it out of the bank in that case? would this also have benefits for alpenglow?

Other Firedancer folks wanted to remove as much from the snapshot manifest as possible. To do that, fields in the manifest would be put on chain into sysvars (as an example). Then those fields could be removed from the manifest, as they'd be present within snapshots implicitly via the sysvar account. And moreso they'd be under consensus explicitly. And snapshot fowards/backwards compatibility is moot.

So I just wanted to double check here, since adding another field to the snapshot manifest is moving in the opposite direction.

I'm not sure how hard/invasive it would be to go the sysvar route vs the snapshot manifest route though.

I don't see how this would be related to Alpenglow.

AshwinSekar:
In terms of the leader workflow we would:
1) Schedule + replay txs in the bank
2) Shred these entries
3) Fill in the BlockID sysvar in the bank based on the last data shred
4) Freeze the bank

So most importantly (2) must happen before (4), while currently (2) and (4) happen independently. The leader cannot begin building the next bank until (4) has happened.
This doesn't seem unreasonable to me, but it does introduce this dependency on shredding.

brooksprumo:
To be clear, I'm fine with either way y'all wanna do it.

lidatong:
> To be clear, I'm fine with either way y'all wanna do it.

sorry for the delay -- chatted with a few of the folks who wanted to remove to the manifest, and i think where we landed is this is important enough for the initial security check on snapshot load (combined with https://github.com/solana-foundation/solana-improvement-documents/pull/340), and we can remove later if we decide to pursue no manifest down the line

if still ok with you @brooksprumo, then @jacobcreech can we merge this?

brooksprumo:
> [..] we can remove later if we decide to pursue no manifest down the line

Nothing can be removed from the current snapshot format. We *can* stop writing the value though. Just a clarification; I don't think it impacts any decision here.


> if still ok with you @brooksprumo 

Yep, still good!

proposals/0333-snapshot-include-block-id.md
t-nelson: > I would lean towards ledger tool always filling in a block_id based on the modifications it performs, but open to suggestions.

seems that there is agreement on this path, but no specification? i don't think we can merge this without a plan for how to restart the cluster

AshwinSekar: Fair point. I'm in favor of taking a coarse approach here:
When creating an artificial bank, set the block id to `hash(parent_block_id, bank_hash)` where `parent_block_id` is the block id of the parent bank, and `bank_hash` is the bank hash of the artificial bank.

Lets us stay agnostic of ledger-tool implementation, any new bank should result in a new bank hash and appropriate block id.

lidatong: yeah this is a good idea. @emwang-jump could you update the SIMD with spec of above?

lidatong:
so i discussed with emily and in lieu of the recent back-and-forth on double merkle for alpenglow vs. chained merkle, this field could get replaced with the double merkle hash of the snapshot slot (and we can separately SIMD / feature activate that as well)

proposals/0333-snapshot-include-block-id.md
emwang-jump: Sure -- to clarify in this example w/ the artificial bank
> Cluster goes down, restart slot is X

 X in this case is the artificial bank, and in this case the serialized bank in the snapshot would be for slot X and we would fill in `hash(parent_block_id, bank_hash)` for block_id?
 
 edit: got it, X+1 is the artificial bank, and the serialized bank in the snapshot is for slot X+1 and hash(id of X, bank_hash of X+1)

t-nelson: > set the block id to hash(parent_block_id, bank_hash) 

does bank hash cover the hard_forks field? for a restart snapshot, that may be the only change

emwang-jump: It looks like in agave `bank.rs` hard_forks is included if it exists, but maybe someone else can confirm

AshwinSekar: yep `hard_forks` is hashed as part of the bank hash
https://github.com/anza-xyz/agave/blob/21d88de4b6ef80c64b78a828416d1137e17f32c0/runtime/src/bank.rs#L4404-L4413

emwang-jump:
Updated the SIMD to address concerns with artificial banks, if could approvers take a second pass (@t-nelson ?)

bw-solana (approved):
Approving from Anza side after SME review by Ashwin & Brooks

lidatong (dismissed):

proposals/0333-snapshot-include-block-id.md
```diff
@@ -0,0 +1,68 @@
...
+## New Terminology
+
+N/A
+
+## Detailed Design
+
+The block_id already lives in the bank. Snapshot producers for all
+clients must include the block ID of the snapshot slot in the serialized
+bank of the snapshot. This can be implemented by adding a `block_id`
+field to the snapshot serialization schema (eg. `BankFieldsToSerialize`
```
t-nelson: surely we need to more specifically declare _where_ the field should be added?

proposals/0333-snapshot-include-block-id.md
```diff
@@ -0,0 +1,68 @@
...
+in Agave or `fd_snapshot_manifest_t` in Firedancer).
+
+Snapshot consumers may optionally read the block ID in the deserialized
+bank, and use it to validate as seen fit.
+
+In cases where an artificial bank is created (ex. via ledger tool during
+development or for a cluster restart), the block_id must also be
+populated.  If the artificial bank with bank hash `bank_hash` builds on
+slot S with block ID `parent_block_id`, the block ID value for the
+serialized bank will be a SHA256 `hash(parent_block_id, bank_hash)`
```
t-nelson: afaik sha256 does not declare a representation. we need to here for interoperability

proposals/0333-snapshot-include-block-id.md
emwang-jump: specified 👍

emwang-jump:
@bw-solana @lidatong addressed the nits, can we get a reapproval?

lidatong (approved):

PR #337, opened 2025-08-14
SIMD-0337: Markers for Alpenglow Fast Leader Handover

ksn6:
We propose augmenting `BlockMarkerV1`, introduced in SIMD-0307, to include `BlockHeader` and `UpdateParent` variants. `BlockHeader`, placed at the beginning of a block, indicates the parent of the block. `UpdateParent` signals that the intended parent is different from what was initially indicated, paving the way for Fast Leader Handover support in Alpenglow (section 2.7 of https://www.anza.xyz/alpenglow-1-1).

Fast Leader Handover will minimize the synchronization delay between consecutive leaders, increasing throughput.

This SIMD purely introduces the types required for Fast Leader Handover - the algorithmic details of Fast Leader Handover itself are outlined in [SIMD-0326](https://github.com/solana-foundation/solana-improvement-documents/pull/326).

NOTE: Fast Leader Handover may also be referred to as "optimistic block production."

### DIFF
--- a/proposals/0337-parent-ready-update-marker.md
+++ b/proposals/0337-parent-ready-update-marker.md
@@ -21,6 +21,9 @@ signals that the intended parent is different from what was initially indicated,
 paving the way for Fast Leader Handover support in Alpenglow (section 2.7 of
 https://www.anza.xyz/alpenglow-1-1).
 
+Fast Leader Handover will minimize the synchronization delay between consecutive
+leaders, increasing throughput.
+
 ## Motivation
 
 SIMD-0326 proposes Alpenglow, a consensus protocol specified in


rogerANZA:
Can we please not call it ParentReadyUpdate? ParentReady is an internal event, so it might be confusing to use the same name here. We just communicate the parent again... so the same as the regular parent announcement in the first slice? Also, why *block*marker?

ksn6:
> We just communicate the parent again... so the same as the regular parent announcement in the first slice?

Right now, the initial parent specification for a block is disseminated in the first shred of a block. In contrast, we will be using the recently introduced BlockComponent type to disseminate the special marker required for a parent switch in optimistic block packing. In a separate SIMD, post-Alpenglow launch, we'll eventually disseminate the initial block parent via a BlockComponent type as well.

This being said, I'm strongly in favor of having separate marker types for (1) the initial parent and (2) the parent switch for optimistic block packing, despite the high likelihood of these two types essentially doing the same thing.

This is because maintaining code is far easier (e.g., we can easily upgrade each type separately) + reading code is much easier and less error-prone.

> Can we please not call it ParentReadyUpdate? ParentReady is an internal event, so it might be confusing to use the same name here.

How about `UpdateParent`?

> Also, why _block_marker?

What do you mean?

rogerANZA:
I was only worried about language. UpdateParent I like.

nit:
> first shred of a block

first slice?

kobi-lizard:
I think we have to change how the parent is initially indicated for correctness. I mean before Alpenglow.

proposals/simd-0335-parent-ready-update-marker.md
```diff
@@ -0,0 +1,229 @@
...
+- **Optimistic Block Packing**: A technique where leaders begin constructing
+  blocks on a tentative parent before the actual parent is finalized
+- **ParentReady Event**: A consensus event indicating that a block has met
+  all requirements to serve as a parent for new block production
+- **Parent Switch**: The action of changing from one parent block to another
+  during optimistic block production
+- **ParentReadyUpdate**: A block marker that records when and to which parent
+  a switch occurred during optimistic packing
+
+## Detailed Design
```
AshwinSekar: might have missed it but good to call out the failure cases:
- multiple ParentReadyUpdate's within the same block
- erroneous values (e.g. parent slot is >= current slot)
and what we do in such cases - do not attempt to continue replay, mark the block as dead

ksn6: Originally had this section, although - the SIMD proposes to introduce the types rather than optimistic block packing itself. Given this, not sure it's* worth tossing in failure cases (which should instead go in the Alpenglow SIMD). That being said, I'm happy to add this in if you'd really like.

ksn6:
> I think we have to change how the parent is initially indicated for correctness. I mean before Alpenglow.

This doesn't seem necessary for correctness. Agreed that the current design isn't optimal, though we can look into changing this after Alpenglow launches.

rogerANZA:
why is it called "Alpenglow Optimistic Block Packing". To me this sounds a bit like greedily pack Txs into a block. something like "Alpenglow Fast Leader Handover" would make more sense.

rogerANZA:
maybe add a reference to the section 2.7 in the whitepaper in the SIMD?

ksn6:
> why is it called "Alpenglow Optimistic Block Packing". To me this sounds a bit like greedily pack Txs into a block. something like "Alpenglow Fast Leader Handover" would make more sense.

it's because the next leader is optimistically packing a block, in the hopes that ParentReady doesn't switch to an earlier parent; "optimistic" != "optimal" which is what greedy typically refers to in the context of approximation algorithms for packing

that being said - don't feel too strongly about this; let me know if you'd really like to change this

> maybe add a reference to the section 2.7 in the whitepaper in the SIMD?

we make the reference to the paper here in the SIMD:

https://github.com/solana-foundation/solana-improvement-documents/pull/337/files#diff-631937531eeb51a129767f8899e4398615b51d0c411e73c05885dcd190fea4e0R38-R40

rogerANZA:
> > why is it called "Alpenglow Optimistic Block Packing". To me this sounds a bit like greedily pack Txs into a block. something like "Alpenglow Fast Leader Handover" would make more sense.
> 
> it's because the next leader is optimistically packing a block, in the hopes that ParentReady doesn't switch to an earlier parent; "optimistic" != "optimal" which is what greedy typically refers to in the context of approximation algorithms for packing

Mh. I probably would call it "Alpenglow Fast [Leader] Handover". Packing (e.g., bin packing) means something very different in CS and Math. And since since packing a block actually makes a lot of sense even for Solana, it's just very misleading in this context.

> that being said - don't feel too strongly about this; let me know if you'd really like to change this

Yes, I would. Otherwise we have another thing where the naming is not optimal in the future.

bw-solana:
> > > why is it called "Alpenglow Optimistic Block Packing". To me this sounds a bit like greedily pack Txs into a block. something like "Alpenglow Fast Leader Handover" would make more sense.
> > 
> > 
> > it's because the next leader is optimistically packing a block, in the hopes that ParentReady doesn't switch to an earlier parent; "optimistic" != "optimal" which is what greedy typically refers to in the context of approximation algorithms for packing
> 
> Mh. I probably would call it "Alpenglow Fast [Leader] Handover". Packing (e.g., bin packing) means something very different in CS and Math. And since since packing a block actually makes a lot of sense even for Solana, it's just very misleading in this context.
> 
> > that being said - don't feel too strongly about this; let me know if you'd really like to change this
> 
> Yes, I would. Otherwise we have another thing where the naming is not optimal in the future.

Another alternative: "Alpenglow Optimistic Parent Selection" ?

We're being optimistic that we know the right parent block to build off of.

Agree that "Optimistic Block Packing" could easily confuse a casual reader.

Edit: Just noticed title was already changed (weeks ago, ha!). I'm fine with it as is

rogerANZA:
> Another alternative: "Alpenglow Optimistic Parent Selection" ?

Also good. I would still say that handover is a bit more explicit since this problem only appears during a leader handover.

proposals/simd-0335-parent-ready-update-marker.md
```diff
@@ -0,0 +1,347 @@
...
+| Length = 41                 (2 bytes) |
++---------------------------------------+
+| Version = 1                 (1 byte)  |
++---------------------------------------+
+| Parent Slot                 (8 bytes) |
++---------------------------------------+
+| Parent Block ID            (32 bytes) |
++---------------------------------------+
+
+Total overhead: 54 bytes per marker
```
bw-solana: Maybe call this `Total size`? Overhead makes me think of non-payload bytes

proposals/simd-0335-parent-ready-update-marker.md
```diff
@@ -0,0 +1,347 @@
...
+  - Enables full implementation of Alpenglow's fast leader handover
+  - Fast leader handover, upon implementation, reduces block production
+    latency
+  - Minimal space overhead (54 bytes per switch event)
+  - Independent versioning allows evolution without affecting other variants
+
+- **Negative**:
+  - Slight increase in block size when parent switches occur
+
+## Security Considerations
```
bw-solana: not sure if we care to call this out since it is more related to the actions we will take upon observing `ParentReadyUpdate`, but one security thing that will require special care is ensuring any transaction cost budgets  are not exceeded for a block even in the presence of a parent update resetting state

proposals/simd-0335-parent-ready-update-marker.md
wen-coding: But we won't have another SIMD on Fast Leader Handover algorithm later right? I think it helps to specify at the protocol level:
- the format of new data structure
- the expected checks associated with the new data structure

We can add more checks later for sure, but it does help the reader understand how the new data is going to be verified. And you do have a section "Parent Switch Validation" later, so maybe we can specify all checks there?

proposals/simd-0335-parent-ready-update-marker.md
ksn6: Done.

proposals/simd-0335-parent-ready-update-marker.md
ksn6: Done.

proposals/simd-0335-parent-ready-update-marker.md
```diff
@@ -0,0 +1,347 @@
...
+  rejected because:
+  - Higher overhead due to transaction processing requirements
+  - Having a transaction "undo" the state transitions induced by other
+    transactions is ill-advised
+
+- **Extended Block Footer**.
+  Extending the block footer to include parent switch information was
+  considered but is inadequate because:
+  - Parent switches occur during block production, not at the end
+  - The footer appears only at block completion
```
wen-coding: Theoretically footer can go anywhere, but I think what you intended to say is that this Marker is time-sensitive, so it must happen and be processed asap, putting it on block footer will limit location where we can put block footer.

Also, this is a extremely rare event right?

proposals/simd-0335-parent-ready-update-marker.md
```diff
@@ -0,0 +1,347 @@
...
+The introduction of this marker does not introduce any new security risks
+until Alpenglow itself is fully implemented and launched. Upon launching
+Alpenglow, we make the following considerations:
+
+### Parent Switch Validation
+
+When a `ParentReadyUpdate` marker is present, validators MUST perform rigorous
+validation:
+
+- Validators MUST verify that the referenced parent block exists and is valid
```
wen-coding: nit: "exists and is valid" here is a bit vague, do you mean they have the block locally? What do they do if they don't? Does this require successful replay of the parent block?

proposals/simd-0335-parent-ready-update-marker.md
```diff
@@ -0,0 +1,347 @@
...
+- The certificate pool must have issued the event `ParentReady(s, hash(b))`,
+  where `s` denotes the first slot of the leader window and `hash(b)` denotes
+  the hash (Block ID) of the new parent
+- The new parent MUST have either a `NotarizeFallback` or `Notarize` certificate,
+  followed by skip certificates until slot `s`
```
wen-coding: I'm confused, how is this different from ParentReady?

proposals/simd-0335-parent-ready-update-marker.md
```diff
@@ -0,0 +1,347 @@
...
+- The switch MUST occur only when a `ParentReady` event is triggered according
+  to Algorithm 3, Line 7-8 of the [Alpenglow whitepaper v1.1](https://www.anza.xyz/alpenglow-1-1)
```
wen-coding: We can maybe add "only one switch allowed and only for the first block of the window" somewhere around here maybe?

proposals/simd-0335-parent-ready-update-marker.md
```diff
@@ -0,0 +1,347 @@
...
+
+**State Transition Rules**:
+
+- State changes from transactions before the `ParentReadyUpdate` marker are discarded
+- The `ParentReadyUpdate` marker itself does not modify state
+- Post-`ParentReadyUpdate` marker transactions are built on the new parent's state
+- Notably, despite pre-marker transaction state changes being discarded, the
+transactions themselves WILL be included in the constructed block.
```
wen-coding: Does the shred index change?

proposals/simd-0335-parent-ready-update-marker.md
ksn6: Ah, forgot to hit resolve on this conversation; I addressed this comment a couple of days ago.

By popular demand, I outlined a number of failure cases under the "Attack Vectors and Mitigations" section. The checks that need to occur are present under "Parent Switch Validation" as you indicated.

Resolving this conversation now, although - feel free to re-open if there are further thoughts here.

proposals/simd-0335-parent-ready-update-marker.md
ksn6: It's not that the `ParentReadyUpdate` marker is "time-sensitive" per-se. Rather, the marker must occur in the middle of a block, since there are going to be txs before the marker (which will not affect state) and txs after the marker (which will affect state.

This is in contrast with the footer, which must actually go at the end of the block, since e.g., we eventually plan on storing the bank hash there, and the hash is calculated after all txs are processed.

The point of refuting the alternative of including `ParentReadyUpdate` in the extended block footer section was to articulate the point that `ParentReadyUpdate` must be in the middle while the footer must be at the end.

I'll add color to the SIMD to explain the details above.

ksn6: > Also, this is a extremely rare event right?

Yes, `ParentReadyUpdate` markers appearing in a block will likely be rare. The occurrence of this event implies that e.g.:

- Leader at slot 99 disseminated a block
- Leader at slot 100 received the above block, and optimistically starts constructing a new block on top of the block in slot 99.
- For some reason, the block at slot 99 ends up getting skipped.
- Leader at slot 100 issues a parent switch

I imagine that step (3) here is what makes the occurrence very rare; I don't think it's common for a block to be disseminated + then skipped.

proposals/simd-0335-parent-ready-update-marker.md
ksn6: Nice catch; I've updated this to read as:

```
- Validators MUST verify that the reference parent block (1) exists and (2) has been replayed. For
  avoidance of doubt - condition (2) ensures that the state after processing the parent block is
  available for use in optimistically constructing and disseminating transactions.
```

I don't think we need to mention repair here.

proposals/simd-0335-parent-ready-update-marker.md
ksn6: `ParentReady` is defined in Definition 15 of the [Alpenglow whitepaper](https://github.com/rogerANZA/Alpenglow-White-Paper/blob/main/Alpenglow-v1.1.pdf) on page 21.

In particular, this is an event type issued by the certificate pool. Every leader window's first slot requires a `ParentReady` event to be present to commit to a parent block. As such, given liveness, a `ParentReady` event is guaranteed on the start of every leader window.

In contrast, `ParentReadyUpdate` is a block marker disseminated indicating that  the parent of a block is switching. This should be a relatively rare occurrence as you indicated previously above.

proposals/simd-0335-parent-ready-update-marker.md
ksn6: Done; see "Double-Parent Attack" for the "only one switch" piece. Added in another attack vector re. the "first block of the window" piece.

proposals/simd-0335-parent-ready-update-marker.md
ksn6: Not sure what you're referring to here - could you expand?

proposals/simd-0335-parent-ready-update-marker.md
wen-coding: I mean this sentence:
```
- The new parent MUST have either a `NotarizeFallback` or `Notarize` certificate,
  followed by skip certificates until slot `s`
```
This seems to be similar to `ParentReady(s, hash(b))` definition in the line above?

proposals/simd-0335-parent-ready-update-marker.md
wen-coding: I mean, maybe we should specify how we switch, if you already sent out data shred 0 to 7, and now you do a switch, is the next data shred 0 or 8 or something else?

proposals/simd-0335-parent-ready-update-marker.md
```diff
@@ -0,0 +1,363 @@
...
+
+### Integration with BlockComponent
+
+The `ParentReadyUpdate` marker integrates with the existing `BlockComponent` system:
+
+```rust
+/// Version 2 block marker with extended functionality.
+///
+/// Supports all V1 features plus ParentReadyUpdate for optimistic block
+/// packing in Alpenglow.
```
rogerANZA: it's still here.

proposals/simd-0335-parent-ready-update-marker.md
```diff
@@ -0,0 +1,363 @@
+---
+simd: '0337'
+title: ParentReadyUpdate Block Marker
+authors:
+  - ksn6 (Anza)
```
rogerANZA, line 5: kobi is your co-author?

proposals/simd-0335-parent-ready-update-marker.md
ksn6: In particular - the parent that `ParentReadyUpdate` indicates as the new parent is exactly `b` within `ParentReady(s, hash(b))`.

proposals/simd-0335-parent-ready-update-marker.md
ksn6: Gotcha - the next shred should be 8 in your example. Basically, you don't want to start building a new block - you want to keep building the block.

The key sentence here:

```
Notably, despite pre-marker transaction state changes being discarded, the transactions themselves WILL be included in the constructed block.
```

Resolving - feel free to re-open with more questions / comments.

lidatong (commented):
sorry for the delay -- lgtm seems my questions were already addressed by others in PR comments

proposals/0337-parent-ready-update-marker.md
```diff
@@ -0,0 +1,364 @@
...
++---------------------------------------+
+| Entry Count = 0             (8 bytes) |
++---------------------------------------+
+| Marker Version = 2          (2 bytes) |
++---------------------------------------+
+| Variant ID = 1              (1 byte)  |
++---------------------------------------+
+| Length = 41                 (2 bytes) |
++---------------------------------------+
+| Version = 1                 (1 byte)  |
```
jherrera-jump, line 172: I would make the `UpdateParentV1` version field 16 bits, so it's consistent with the footer.

bw-solana:
Holding off on approving from Anza side.

We've discussed potential issues w/ observing the parent switch in replay execution that are triggering new ideas on how to communicate this information

proposals/0337-parent-ready-update-marker.md
ksn6: This isn't necessary - it's very unlikely that we ever saturate 256 separate versions for this field, given that this field's goal is to simply just update a parent switch.

The footer is a bit different in this regard, in that it's a bit unknown.

kobi-lizard:
My demands have been satisfied and I support the current form of this SIMD :frog:

ksn6:
> Holding off on approving from Anza side.
>
> We've discussed potential issues w/ observing the parent switch in replay execution that are triggering new ideas on how to communicate this information



> My demands have been satisfied and I support the current form of this SIMD 🐸

For more context, a few folks came up with a scenario that leads to liveness issues if we are *only* able to detect `UpdateParent` messages during replay.

To address this, we enable `UpdateParent` detection on shred ingest as discussed in the "`DATA_COMPLETE_SHRED` Placement Rules" section.

proposals/0337-parent-ready-update-marker.md
```diff
@@ -0,0 +1,325 @@
...
+### Serialization Format
+
+Both `BlockHeader` and `UpdateParent` (variant 1 and 2) have the following
+structure while serialized within a `BlockComponent` in the first version:
+
+```
++---------------------------------------+
+| Entry Count = 0             (8 bytes) |
++---------------------------------------+
+| Marker Version = 2          (2 bytes) |
```
lidatong: so i looked and i dont think you actually need to version bump the marker here -- the schema format here seems compatible with v1 as described in #307 ?

proposals/0337-parent-ready-update-marker.md
```diff
@@ -0,0 +1,325 @@
...
+
+- **BlockHeader**: Block marker variant indicating the block's parent. Placed in
+  the beginning of block data.
+- **UpdateParent**: Block marker variant indicating a changed parent.
+
+## Detailed Design
+
+### Block Marker Version 2
+
+We introduce `BlockMarkerV2` as an extension of the original `BlockMarkerV1`:
```
lidatong: see comment about marker versioning below

ksn6: Done.

proposals/0337-parent-ready-update-marker.md
ksn6: Agreed - done.

lidatong (approved):

PR #340, opened 2025-08-20
SIMD-0340: Validate chained block id

lidatong:


proposals/0340-validate-chained-block-id.md
```diff
@@ -0,0 +1,67 @@
...
+correctly across slot boundaries. Consensus can converge on a block even
+if the first FEC set's chained merkle root is invalid ie. does not chain
+off the parent block's last FEC set merkle root (the block id).
+
+This is a problem because chained merkle roots should validate an entire
+ancestry, so that you have a canonical linear chain (all the way back to
+the snapshot slot - see also SIMD-0333 proposal for including a block_id
+in the snapshot manifest). Otherwise, you don't know if your parent slot
+based on `slot - parent_off` is in fact your actual parent block because
+slot numbers do not key blocks uniquely when there is equivocation.
```
AshwinSekar, line 29: would also specify that this is critical for Alpenglow repair / equivocation handling

proposals/0340-validate-chained-block-id.md
```diff
@@ -0,0 +1,67 @@
...
+
+N/A
+
+## Detailed Design
+
+Verify the chained merkle root of the first FEC set in a block correctly
+matches the merkle root of the last FEC set of the parent slot (based on
+the slot and parent_off field in every shred in the FEC set).
+
+If they do not match, mark the child block (ie. block with the first FEC
```
AshwinSekar: to be extra clear I would say "do not attempt to replay a child block off a parent block unless BOTH the parent slot # and parent block id match"

Then in the section below we can say that for TowerBFT we mark it as dead (poh ticks already does this) so that repair can be notified. In Alpenglow we don't need any special handling, certificate conditions will notify repair.

AshwinSekar (commented):
Thanks for writing this up! just some minor nits

proposals/0340-validate-chained-block-id.md
lidatong: @AshwinSekar thanks -- rephrased based on your feedback. take another look?

proposals/0340-validate-chained-block-id.md
```diff
@@ -0,0 +1,75 @@
...
+## Motivation
+
+Currently it is not required to verify that a FEC set merkle root chains
+correctly across slot boundaries. Consensus can converge on a block even
+if the first FEC set's chained merkle root is invalid ie. does not chain
+off the parent block's last FEC set merkle root (the block id).
+
+This is a problem because chained merkle roots should validate an entire
+ancestry, so that you have a canonical linear chain (all the way back to
+the snapshot slot - see also SIMD-0333 proposal for including a block_id
```
AshwinSekar, line 26: I believe there's a section for dependencies it would be good to call out SIMD-0333 and SIMD-0313 are necessary to do this SIMD without special casing

lidatong: good idea, updated

kobi-lizard:
Hi humans. Forgive the invasion. We had this big discussion of Chained Merkle vs Double Merkle back in the day, so I'm here to remind that for Alpenglow we want a different block id (Double Merkle). (Irrespectively of Alpenglow I have a negative opinion about the Chained Merkle effort, but whatevs.) All Chained Merkle-related stuff will ultimately be tech debt. If this is clear to all of us and we still want to do this, all goody. Otherwise let me know this is not clear and argue with me.

lidatong:
@jacobcreech we discussed this in-person this week, and I think Anza and FD reached consensus that we want to move forward this validation (though for Alpenglow V1 we'll be using double merkle instead). could we merge this?

topointon-jump (dismissed):

bw-solana:
@AshwinSekar do we still want to move forward with this? Makes sense why this is useful for Tower, but the optimistic case seems like we would have this extra protection for 1 release cycle before Alpenglow/Double Merkle.

Does this design still make sense with Double Merkle?

AshwinSekar:
> Does this design still make sense with Double Merkle?

For vanilla alpenglow this is not needed. This specific check outlined in the SIMD

 >Do not attempt to replay a child block off a parent block unless both
the parent slot (ie. `slot - parent_off`) matches the first FEC's slot
and the parent block_id matches the first FEC's chained merkle root.

Already exists, (instead of matching the chained merkle root, we match to the parent slot / DMR in the block header). 

In that sense this SIMD is not needed for Agave, however for FD:
- They use CMR for replay chaining purposes
- They don't implement ancestor hashes service 🥳 (they use the CMR to fix duplicates instead)

Given this, we agreed to keep chained merkle shreds for now and reassess removing it later. It's critical then that agave does not finalize blocks with invalid parent child chained merkle roots as that would mess up FD regardless of tower bft / alpenglow.


So the short answer is yes, although agave will no longer consume CMR with alpenglow, we should do this (super easy) check to keep compatibility for FD.

proposals/0340-validate-chained-block-id.md
```diff
@@ -0,0 +1,93 @@
...
+
+Clients will mark as blocks as dead they previously may have replayed.
+
+## Security Considerations
+
+Security is improved because of enhanced equivocation protection.
+
+## Backwards Compatibility
+
+This feature is not backwards compatible.
```
bw-solana: Should be backwards compatible as far as parsing/processing shreds, but newer code might mark some blocks dead, which obviously diverges consensus.

proposals/0340-validate-chained-block-id.md
```diff
@@ -0,0 +1,93 @@
...
+equivocation) then the certificate will notify repair which retrieves
+the alternate block.
+
+## Alternatives Considered
+
+N/A
+
+## Impact
+
+Clients will mark as blocks as dead they previously may have replayed.
```
bw-solana: nit:
Clients will mark blocks that don't properly chain to their parent as dead that they previously would have replayed.

proposals/0340-validate-chained-block-id.md
```diff
@@ -0,0 +1,93 @@
...
+
+In Alpenglow, it is not necessary to mark the block as dead. Instead,
+the validator will vote "skip". In case the rest of the cluster
+generates a certificate for an alternate version of the block (due to
+equivocation) then the certificate will notify repair which retrieves
+the alternate block.
+
+## Alternatives Considered
+
+N/A
```
bw-solana, line 93: Might be worthwhile mentioning the Double Merkle / Alpenglow strategy here

proposals/0340-validate-chained-block-id.md
```diff
@@ -0,0 +1,93 @@
...
+## Dependencies
+
+This proposal depends on the following accepted proposal:
+
+- **[SIMD-0313]: Drop unchained merkle roots**
+
+    As the new shred format is being sent by all clients, this feature
+    deprecates the old shred format.
+
+As well as the following proposal in review:
```
bw-solana: this has now merged

proposals/0340-validate-chained-block-id.md
```diff
@@ -0,0 +1,93 @@
...
+
+N/A
+
+## Detailed Design
+
+Verify the chained merkle root of the first FEC set in a block correctly
+matches the merkle root of the last FEC set of the parent slot (based on
+the slot and parent_off field in every shred in the FEC set).
+
+Do not attempt to replay a child block off a parent block unless both
```
bw-solana, line 61: Might be nice to call out if we should propagate shreds through turbine for blocks that fail to chain (I think the answer is yes)

bw-solana (commented):
Concept ACK and overall looks good.

Left a few small comments

proposals/0340-validate-chained-block-id.md
lidatong: clarified this

proposals/0340-validate-chained-block-id.md
lidatong: added some clarification

lidatong:
thanks for the review @bw-solana! addressed your feedback above -- take another look?

PR #341, opened 2025-08-22
v0 Account Compression

igor56D:


proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,275 @@
...
+**Purpose**: Compresses an existing account by replacing its data with a
+cryptographic commitment.
+
+**Parameters**:
+
+- `account_pubkey`: pubkey of the account to be compressed
+- `combined_hash`: 32-byte hash of concatenated account pubkey and current
+  account data
+
+**Behavior**:
```
jacobcreech, line 73: What authorization is required to compress/decompress an account?

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,275 @@
...
+
+#### `sol_decompress_account(original_data: *const u8, data_len: u64)`
+
+**Purpose**: Recovers a compressed account by restoring its original data.
+
+**Parameters**:
+
+- `original_data`: Pointer to the bincode serialization of the original
+  `AccountSharedData` object, which includes the pubkey
+- `data_len`: Length of the original data in bytes
```
jacobcreech: Is it expected that this information would be stored offchain for recovery?

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,275 @@
...
+compression can reduce the global state size sufficiently.
+
+### Conclusion
+
+The proposed hash-based approach provides the optimal balance of storage
+savings, performance predictability, and implementation complexity.
+
+## Impact
+
+### DApp and Wallet Developers
```
jacobcreech, line 327: If an app or wallet developer wanted to view the data of an account, would this mean that they would have to simulate a transaction with a decompress call to `compression system program`?

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,275 @@
...
+- include checks for account compression status in program interaction
+  workflow
+- add instructions to transactions for account recovery when appropriate
+- collisions on the `pubkey_hash` are extremely unlikely but not impossible
+  due to using 10-bytes. In this case, the user or developer will need to use
+  a different address.
+
+### Validators
+
+- **Memory/Storage savings**: If enough accounts are compressed, a
```
jacobcreech, line 339: What is the expected storage savings given an account being compressed with X bytes?

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,275 @@
...
+
+### Snapshot Format
+
+- **Impact**: New snapshot format with compressed accounts section
+- **Mitigation**: Version-aware snapshot loading with backward compatibility
+  for old snapshots
+
+### Account Creation Behavior  
+
+- **Impact**: Account creation fails if `SHA_256(pubkey)[0:10]` exists in
```
jacobcreech: Will this account creation check require account inits to consume more CUs?

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,275 @@
...
+  compressed set
+- **Mitigation**: if the account corresponding to the pubkey was previously
+  compressed it must be recovered rather than recreated. if a collision has
+  occured a different pubkey must be used. RPCs and good errors can provide
```
jacobcreech: > RPCs and good errors can provide the relevant info

This is incredibly important to get right. It we don't outline exactly all the edgecases and how this works for developers, it'll be a nightmare for anyone to integrate with. We should consider putting more detail either in this SIMD on how these edge cases and errors will be displayed, or have a separate document outlining how we can properly setup the right infrastructure to handle the expected behavior with compressed state.

wjthieme:
Would it be low-hanging fruit to allow programs to compress/decompress their own state? Either by allowing to syscall directly or to cpi into compression_system_program with the account_to_compress as signer?

wjthieme:
Will an account appear as uninitialized if it is compressed? It would be useful to have programs be able to distinguish between uninitialized vs compressed. Or will the program a compressed account be assigned to be compressed_system_program vs normal system_program?

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,275 @@
...
+significant storage costs and blockchain state bloat. Rent/storage cost is
+already a significant complaint among app developers, and without reducing
+the state size or growth, rent cannot be safely lowered and optimizations
+like fully in-memory account state are infeasible in the medium to long
+term. To solve this in an enduring way, we need:
+
+1. an economic mechanism to limit state growth
+2. a predicate to determine which accounts can be compressed
+3. a compression scheme that removes accounts from the global state while
+   allowing for safe and simple recovery.
```
brooksprumo, line 33: Worthwhile to include lowering rent cost in this list too? It's a main motivator in the above paragraph. (Will be a separate SIMD, of course.)

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,275 @@
...
+#### `sol_compress_account(account_pubkey: &[u8; 32])`
+
+**Purpose**: Compresses an existing account by replacing its data with a
+cryptographic commitment.
+
+**Parameters**:
+
+- `account_pubkey`: pubkey of the account to be compressed
+- `combined_hash`: 32-byte hash of concatenated account pubkey and current
+  account data
```
brooksprumo: The function signature above only lists one parameter. Is this `combined_hash` meant to be the output/return value?

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,275 @@
...
+
+- **Compression**: replacing arbitrary account data with a fixed size
+  commitment. Not to be confused with traditional compression; the data
+  cannot be directly recovered from compressed state.
+- **Compression Condition**: a predicate determining whether or not an
+  existing account can be compressed. A specific compression condition isn't
+  provided in this SIMD -- it is assumed to always be false, meaning no
+  account is eligible for compression.
+- `account_hash = sha_256(account_pubkey)[0..10]`
+- `data_hash = sha_256(bincode_serialize(AccountSharedData))`
```
brooksprumo: Note, `AccountSharedData` is an agave-specific implementation detail.

In the Accounts Lattice Hash SIMD, we enumerated the fields explicitly:
https://github.com/solana-foundation/solana-improvement-documents/blob/3e323637e5cbdf46a87a491777aa6c86aa96f566/proposals/0215-accounts-lattice-hash.md?plain=1#L94-L108

Maybe for this SIMD we just define what an `Account` is first?

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,275 @@
...
+  been decompressed.
+
+While marking the account as compressed must be done synchronously, the
+actual compression (ie the full replacement of the account with its
+compressed form stored in the `CAM`) can be done asynchronously for
+performance:
+
+- compute the 10-byte `account_hash`
+- compute the 32-byte `data_hash`
+- assign `data_hash` to `CAM[account_hash]`
```
brooksprumo: I'm on board with `data_hash` here, but I think this is veering into implementation detail, and explicitly details that I think are unideal. I think the mapping of the CAM from address to account can be left out.

We already store/index all accounts via their address (i.e. `Pubkey`). If we need to grab a 10-byte subslice, then that information needs to be stored somewhere too. I'm not seeing a benefit to this.

We can say (1) the client implementations need to provide a mapping from account address to the `data_hash`, and (2), the decompression syscall can be adjusted to include the address as a parameter.

Or even don't add the address to the decompression syscall! The hash of an account is fully unique. If two unique address are both USDC token accounts with 10 tokens, those *accounts* will hash to the same value. That's totally fine! Decompression doesn't need to care. The account data is passed in, the hashes match, the account data is put back in place _for that specific account's address being decompressed_.

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,275 @@
...
+
+- MUST verify the caller is the hardcoded compression system program
+- MUST verify the account is currently in compressed state by computing the
+  `account_hash` from the provided `original_data` and checking the `CAM`
+- MUST verify the `data_hash` computed from the provided `original_data`
+  matches `CAM[account_hash]`
+- if verification succeeds, the original account MUST be restored to the active set
+    - this is treated exactly like a new account allocation so rent
+      requirements, load limits, etc all apply
+- if verification succeeds, the `pubkey_hash` must be removed from the `CAM`
```
brooksprumo: I think an implementation _will_, and should, do this, but doesn't seem required. Also, there'd be no way to know assuming everything else is implemented correctly. Just need to ensure that if an account is decompressed, that later we still see it as decompressed (again assuming it wasn't recompressed in the meantime).

```suggestion
```

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,275 @@
...
+
+### Database and Snapshot Extensions
+
+#### Compressed Accounts Storage
+
+The data structure used for the `CAM` in the runtime is left as an
+implementation detail, but it MUST be stored in the snapshots as a
+`Vec<CompressedAccountEntry>`:
```
brooksprumo: I disagree. We already have account storage, I'd like to reuse that existing model.

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,275 @@
...
+    pub account_hash: [u8; 10],
+    pub data_hash: [u8; 32],
+}
+```
+
+The `CAM` MUST be:
+
+- **Persisted**: Included in snapshots and incremental snapshots
+- **Versioned**: Track compression/decompression operations across slots
+- **Fork Aware**: Track compressed account data across different forks.
```
brooksprumo: I'd prefer we just leverage the existing AccountsDb model, which handles all this. We then augment the Account with a state enum of compressed or not.

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,275 @@
...
+- **New behavior**: A new lattice hash, `compressed_lattice_hash_bytes`,
+  committing to the state of the `CAM` is maintained and also rolled into
+  the bank hash.
+- **Hash calculation (simplified)**: 
+    - current: `bank_hash = H(H(parent_hash, signature_count, last_blockhash),
+      accounts_lattice_hash_bytes)`
+    - new: `bank_hash = H(H(parent_hash, signature_count, last_blockhash),
+      accounts_lattice_hash_bytes, compressed_lattice_hash_bytes)`
+    - the important distinction here is that the compressed lattice hash
+      immediately follows the uncompressed lattice hash
```
brooksprumo: I think we can remove all this too. An account is compressed, and then the account is stored as part of normal transaction processing. We already handle accounts lattice hash updates for transaction processing.

I agree that calculating the hash of an account needs to be updated though!

Calculating the hash of an account will need the discriminant of compressed or not, and if compressed, its compressed hash value.

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,275 @@
...
+    lthash.init()
+    lthash.append( entry.pubkey_hash )
+    lthash.append( entry.data_hash )
+    return lthash.fini()
+```
+
+#### Snapshot Format Changes
+
+Snapshots will include a new section `compressed_accounts` containing the
+full compressed account state as a `Vec<CompressedAccountEntry>`
```
brooksprumo: Can be removed.

```suggestion
```

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,275 @@
...
+
+### Account Creation Validation
+
+When creating new accounts, the runtime MUST verify the target pubkey is not
+in the `CAM`. Because the `CAM` will likely be stored on disk to limit memory
+usage, an efficient in-memory data structure should be used to verify
+non-membership. A counting bloom filter, cuckoo filter, or similar data
+structure that supports deletions can be used but must be machine-diversified
+to prevent a cluster-wide slow disk read.
```
brooksprumo: I agree we can create a side-band structure for the compressed accounts, if we want. For v0 I think we can leverage the existing AccountsDb storage and index to handle this.

So "yes" to this:
> When creating new accounts, the runtime MUST verify the target pubkey is not
in the `CAM`.

and a "no, unless necessary" to this:
> an efficient in-memory data structure should be used to verify
non-membership. 

(And that's only a "no" for adding to the SIMD! We can still implement this in the client as we see fit.)

aeyakovenko:
My preference is that the data is stored in a secondary db that can be fetched by rpcs just like a snapshot.  But it’s not in the hot path for a validator snapshot. 

Rpc provides and indexing protocols can plug it into their systems

igor56D:
> Would it be low-hanging fruit to allow programs to compress/decompress their own state? Either by allowing to syscall directly or to cpi into compression_system_program with the account_to_compress as signer?

I wouldn't expect this to be too much of a lift -- would just require modifying the compression condition to allow owners/authorities to compress their accounts. The compression condition is intentionally made to be `false` in this SIMD (meaning no accounts are compressible) so the scope of discussion can be narrowed to just the compression mechanism itself. Modifications to the condition will be in separate SIMDs.


> Will an account appear as uninitialized if it is compressed? It would be useful to have programs be able to distinguish between uninitialized vs compressed. Or will the program a compressed account be assigned to be compressed_system_program vs normal system_program?

The intention is that "compressed account : decompression" is functionally very similar to "uninitialized account : account creation". So, unless a compressed account is being decompressed, it should appear the same as an uninitialized account to users and programs. Main rationale is that this is just much simpler and less error prone implementation-wise.

I'm curious why being able to distinguish them within the runtime would be useful. If there's a good enough reason I'll think about how best to expose the distinction in the runtime.

proposals/0341-v0-account-compression.md
igor56D: forgot to remove that. artifact from an older design.

proposals/0341-v0-account-compression.md
igor56D: 1. the idea is the full pubkey of a compressed account would *not* be stored anywhere in the snapshot. It's replaced by the 10-byte `pubkey_hash` to save space. Going from 32 bytes to 10 is a reduction in snapshot size of 22 GB for a compressed account set containing 1B accounts. The `pubkey_hash` is used for indexing compressed data and collision checks on new account creations.

3. this isn't an implementation detail because it determines which accounts can be created. Two validators cannot equivocate on whether a new account creation collides with a compressed account.

proposals/0341-v0-account-compression.md
igor56D: Isn't this required by all client implementations if the `CAM` is the authority on the compressed account set and is included in the snapshot? validators may disagree on if a new account creation is valid if they don't have an equivalent `CAM` locally.

proposals/0341-v0-account-compression.md
igor56D: The reason for a new data structure is to save 22-bytes by not requiring storage of the full pubkey. I suppose we can use the `pubkey_hash` as the pubkey and reuse the existing model. Is that preferable? or do you think the 22-byte space savings isn't worth the extra hassle?

proposals/0341-v0-account-compression.md
igor56D: I agree, if the `CAM` doesn't need to exist as an independent data structure then this can be replaced with an update to the lattice hash calculation. See other comments for why I wanted to keep a separate data structure.

proposals/0341-v0-account-compression.md
igor56D: agreed on the bloom filter part being an implementation detail. just thought it'd be useful to mention so people don't assume they'll be a lot more disk reads. I'll remove it.

proposals/0341-v0-account-compression.md
topointon-jump: Second this, we should define the hash function without reference to Agave implementation details.

proposals/0341-v0-account-compression.md
topointon-jump: > Going from 32 bytes to 10 is a reduction in snapshot size of 22 GB for a compressed account set containing 1B accounts.

Using 10-byte subslices adds some complexity, are we sure this is worth it for a saving of 22GB?

zfedoran:
`sha_256(account_pubkey)[0..10]`

Where did the design decision for a 10byte slice come from? Eg: why not 16 bytes? 

Also, not sure it’s worth the trouble, using the full 32 byte pubkey is easier and reduces the trouble of having to check if an account can be created

proposals/0341-v0-account-compression.md
igor56D: > Using 10-byte subslices adds some complexity, are we sure this is worth it for a saving of 22GB?

after discussing this offline, the answer to that question is decidedly no. I'll remove `pubkey_hash` from the proposal.

igor56D:
> `sha_256(account_pubkey)[0..10]`
> 
> Where did the design decision for a 10byte slice come from? Eg: why not 16 bytes?
> 
> Also, not sure it’s worth the trouble, using the full 32 byte hash is easier and reduces the trouble of having to check if an account can be created

The intention was to use the minimum number of bytes to reduce the snapshot size as much as possible. Given what the `pubkey_hash` is used for, cryptographically strong collision resistance isn't necessary. That said, I discussed this with FD/Agave devs and the consensus is that the complexity isn't worth the extra space savings so I'll remove `pubkey_hash` and use the full pubkey instead.

zfedoran:
I’m curious, the document seems to imply that only accounts that meet some TBD criteria can be compressed. Why not allow any account to be compressed as long as the signature or signing seeds are properly provided. 

Some protocols want to be able to compress their program accounts, on demand, to recover account fees (as is done [here](https://github.com/code-payments/code-vm/blob/main/program/src/instruction/compress.rs#L5)).

igor56D:
> I’m curious, the document seems to imply that only accounts that meet some TBD criteria can be compressed. Why not allow any account to be compressed as long as the signature or signing seeds are properly provided.
> 
> Some protocols want to be able to compress their program accounts, on demand, to recover account fees (as is done [here](https://github.com/code-payments/code-vm/blob/main/program/src/instruction/compress.rs#L5)).

I don't see any reason we can't eventually support something like this. It'd be a fairly straightforward extension to the compression condition, but I intentionally left that and similar things out of this SIMD to minimize controversy and get alignment on the basics first.

zfedoran:
> > I’m curious, the document seems to imply that only accounts that meet some TBD criteria can be compressed. Why not allow any account to be compressed as long as the signature or signing seeds are properly provided.
> > Some protocols want to be able to compress their program accounts, on demand, to recover account fees (as is done [here](https://github.com/code-payments/code-vm/blob/main/program/src/instruction/compress.rs#L5)).
> 
> I don't see any reason we can't eventually support something like this. It'd be a fairly straightforward extension to the compression condition, but I intentionally left that and similar things out of this SIMD to minimize controversy and get alignment on the basics first.

Awesome, fully understand and appreciate that. Just wanted to make sure it was seen as a potentially useful thing to do instead of hard-coding a “last_seen_epoch > 10” or something like that.

wjthieme:
> I'm curious why being able to distinguish them within the runtime would be useful. If there's a good enough reason I'll think about how best to expose the distinction in the runtime.

Imo there is functionally a difference between `uninitialized` and `compressed`. I think there is an important distinction:
* From client-side code it is important to be able to know the difference. How can I otherwise know to init or decompress an account.
* From program-runtime being able to distinguish might be more of a nice-to-have / future-proofing. I can imagine a scenario where you might want to decompress/init conditionally through a cpi or something. Could also just be over-optimization from my end tho

proposals/0341-v0-account-compression.md
ripatel-fd: `data_hash` should be BLAKE3, see Brooks comment above. 

The main way to hash Solana accounts is LtHash, which is BLAKE3 with 2048 bytes extended output. The only functional difference is that the existing account hash includes the pubkey in its preimage.

Account hashing is rather expensive. Reusing the existing hash function also allows validator implementations to reduce computations (The nice thing about BLAKE3 is that the 32 byte mode outputs the same first 32 bytes as the LtHash mode). SHA256 and BLAKE3 also have a high per-hash fixed cost because of scheduling overhead (assigning multiple hash states to SIMD lanes), so it would be nice if we could avoid having two extra hashes.

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,275 @@
...
+## New Terminology
+
+- **Compression**: replacing arbitrary account data with a fixed size
+  commitment. Not to be confused with traditional compression; the data
+  cannot be directly recovered from compressed state.
+- **Compression Condition**: a predicate determining whether or not an
+  existing account can be compressed. A specific compression condition isn't
+  provided in this SIMD -- it is assumed to always be false, meaning no
+  account is eligible for compression.
+- `account_hash = sha_256(account_pubkey)[0..10]`
```
ripatel-fd: I don't understand the motivation of truncating this hash down. 80 bits really isn't a lot. 
Bitcoin's network hashrate is like 900 exahash/s right now, so it can find a 69 bit collision once per second. So, if the entire Bitcoin network switched over, they could do a second preimage attack, thereby steer the reupload to a different account, in ~34 seconds. Which would prevent the account at the real address from being restored. 

Of course this is not a realistic attack but just illustrating the problem is not just theoretical.

IMO, just use the `account_pubkey` as the key at the protocol level, and let implementations choose how they want to compress the index.

ripatel-fd: sha256 is also very expensive for this set-lookup use case

proposals/0341-v0-account-compression.md
ripatel-fd: Yeah, this use of sha256 seems expensive

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,275 @@
...
+- `data_hash = sha_256(bincode_serialize(AccountSharedData))`
+- **Compressed Accounts Map (CAM)**: for each compressed account, maps the
+  `account_hash` to the `data_hash`.
```
ripatel-fd: I thought the usages of CAM meant content addressable memory at first :) 
Kinda works out though

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,275 @@
...
+- **Atomic operations**: Compression/recovery operations are atomic to
+  ensure consistent state across the cluster
+
+### Attack Vectors
+
+- **Hash collision attacks**: collisions on the `data_hash` would allow for
+  introducing arbitrary data into the account state. SHA-256 provides
+  sufficient collision resistance. Collisions on the `pubkey_hash` are not
+  concerning since they're extremely rare and only prevent the new account
+  from being created.
```
ripatel-fd: This is still concerning on paper. A deliberate collision (doable with today's existing hardware in under 1 minute, see my post above) on `pubkey_hash` causes the CAM entry to be removed, thus preventing the original state of the account being restored.

Basically, highly sophisticated attackers could now permanently render expired accounts unusable within minutes (by changing the pubkey to something random). This would discourage people from holding valuable token accounts. 

So, I suggest just getting rid of `pubkey_hash` and directly keying by pubkey.

igor56D: I'll be updating the SIMD to replace `pubkey_hash` with the full pubkey sometime today, primarily because the general consensus is that the extra space savings aren't worth the increase in complexity.

I do want to clarify that a collision on `pubkey_hash` would not allow decompressing an account to an address different than the original. The `data_hash` pre-image includes the account data, metadata, and pubkey, so you'd need to produce a collision on that 32-byte hash as well for decompression to be valid.

proposals/0341-v0-account-compression.md
igor56D: That depends on the "compression condition", which is intentionally set to `false` in this SIMD to focus discussion on the mechanism. [SIMD-0344](https://github.com/solana-foundation/solana-improvement-documents/pull/344) extends the condition to include rent delinquency but other SIMDs can add account authority/owner checks to the condition as well. In any case, it's out of scope for this SIMD.

proposals/0341-v0-account-compression.md
ripatel-fd: I see, sorry, I was not aware that AccountSharedData includes the pubkey

brooksprumo: AccountSharedData does not include the account's address. (It does include the *owner's* address though, but that's different.)

igor56D:
Updates:

- removed `CAM` and `pubkey_hash`. Compressed accounts will instead be tracked in the regular accounts database.
- elaborated on updates to RPCs to give devs more clarity on how account management can be implemented to handle potentially compressed accounts.
- replaced SHA256 with BLAKE3 for `data_hash` calculation
- removed unnecessary implementation details

cc @brooksprumo @ripatel-fd @jacobcreech

proposals/0341-v0-account-compression.md
igor56D: yes, it will be off-chain but who stores the data and how isn't specified here. RPC providers and archival nodes are the more obvious options but wallets can cache account data for users, apps can subscribe to account update events and store the data, etc. It's pretty open ended and the best choice will likely depend on the account and its usage.

proposals/0341-v0-account-compression.md
igor56D: If the account is compressed, the data isn't available on-chain. If that account needs to be used then it must be decompressed, which requires providing the data. The off-chain source of that data can vary, like I mentioned in a previous comment.

proposals/0341-v0-account-compression.md
igor56D: Compressed accounts will take up 64 bytes of space (pubkey and data hash), so the space savings will be `X - 64`. Every active account must at the very least store the pubkey + owner pubkey + lamports + etc so space savings will always be positive. I believe token accounts are the most common and they're about 165 bytes, for reference.

proposals/0341-v0-account-compression.md
igor56D: update: replaced the hash prefix with the pubkey

proposals/0341-v0-account-compression.md
igor56D: I added more to the SIMD regarding errors and RPC updates that will hopefully mitigate some of the concern. A separate doc on recommended infra setup + workflow changes is a good idea. I'll reach out to some of the RPC providers to see if we can put something like that together.

proposals/0341-v0-account-compression.md
igor56D: fixed: 944752816a6f06ddb6e712abc90cbef80557f80c

proposals/0341-v0-account-compression.md
igor56D: fixed: 944752816a6f06ddb6e712abc90cbef80557f80c

proposals/0341-v0-account-compression.md
igor56D: removed all the pubkey_hash and CAM complexity in favor of the simpler approach: 944752816a6f06ddb6e712abc90cbef80557f80c

proposals/0341-v0-account-compression.md
igor56D: fixed: 944752816a6f06ddb6e712abc90cbef80557f80c

Updated to use `lthash.out()`, which I assume is what you meant by 32-byte mode but lmk if you had something else in mind. I'm not too familiar with the perf details here.

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,389 @@
...
+**Purpose**: Compresses an existing account by replacing its data with a
+cryptographic commitment.
+
+**Parameters**:
+
+- `account_pubkey`: pubkey of the account to be compressed
+
+**Behavior**:
+
+- MUST verify that the caller is the hardcoded compression system program
```
ripatel-fd, line 75: Could you specify whether this check happens at program deploy-time or run-time?

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,389 @@
...
+  been decompressed.
+
+While marking the account as compressed must be done synchronously, the
+actual compression (ie the full replacement of the account with its
+compressed form in the account database) can be done asynchronously for
+performance:
+
+- compute the 32-byte `data_hash`
+- replace the account in account databse with a compressed account entry
+- (optional) emit a compression event for off-chain data archival
```
ripatel-fd, line 89: `sol_compress_account` needs CU pricing

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,389 @@
...
+
+- MUST verify the caller is the hardcoded compression system program
+- MUST verify the account is currently in compressed state
+- MUST compute `lthash.out(Account)` from the provided parameters and verify
+  it matches the stored compressed account's data_hash
+- if verification succeeds, the original account MUST be restored to the active set
+    - this is treated exactly like a new account allocation so rent
+      requirements, load limits, etc all apply
+- if verification succeeds, the compressed account entry must be replaced
+  with the full account data
```
ripatel-fd, line 115: `sol_decompress_account` needs CU pricing

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,389 @@
...
+
+**Parameters**:
+
+- `account_pubkey`: 32-byte public key of the account to decompress
+- `lamports`: The lamport balance of the original account
+- `data`: Pointer to the original account data bytes
+- `data_len`: Length of the account data in bytes
+- `owner`: 32-byte public key of the account owner
+- `executable`: Whether the account is executable
+- `rent_epoch`: The rent epoch of the original account
```
ripatel-fd, line 103: Syscalls unfortunately can only take 5 arguments (iirc). 
Maybe we can use one of the account structs in the program SDK, such as the ones from the CPI syscalls.

ripatel-fd, line 103: I'm pretty sure that `rent_epoch` is not part of the lthash on any clusters anymore. I don't know which SIMD removed rent epoch from the hash though (cc @brooksprumo)

It's fine to remove from this syscall

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,389 @@
...
+compressed form in the account database) can be done asynchronously for
+performance:
+
+- compute the 32-byte `data_hash`
+- replace the account in account databse with a compressed account entry
+- (optional) emit a compression event for off-chain data archival
+
+#### `sol_decompress_account(..)`
+
+**Purpose**: Recovers a compressed account by restoring its original data.
```
ripatel-fd, line 93: I wonder if this syscall interface should be changed to improve user UX.
Correct me if I'm wrong but the implied flow here is as follows (for large accounts):

- User creates a dummy account that will hold a copy of the original data
- User uploads fragments of original account data to target account using multiple transactions
- User calls invokes the compression program, which invokes this syscall. Under the hood, the runtime allocates the target account and copies over data
- User cleans up the dummy account

I feel like we could combine some of these steps and skip the  syscall copy

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,389 @@
...
+
+#### Compressed Accounts Storage
+
+Compressed accounts are stored directly in the account database like regular
+accounts, but with a special compressed account structure:
+
+```rust
+pub struct CompressedAccount {
+    pub pubkey: Pubkey,
+    pub data_hash: [u8; 32],
```
ripatel-fd, line 127: Please note that the 'lattice hash' is 2048 bytes long, so the spec should mention how to go from 2048 bytes to 32 bytes hash size. Lattice hash is simply BLAKE3 with extended output (XOF). We can simply take the first 32 bytes here since BLAKE3 has the convenient property that, for a constant input, the low bytes of the hash output don't change as the XOF output length parameter is increased.

EDIT: `blake3_2048(lthash_preimage)[:32] == blake3_32(lthash_preimage)`

ripatel-fd: (What we should NOT do is something like `blake3_32(blake3_2048(lthash_preimage))`)

proposals/0341-v0-account-compression.md
igor56D: Semantically, the compression program itself can delete the dummy account and transfer rent/balance/metadata to the new account. In that case, the copy vs move becomes an implementation detail. Is that sort of what you have in mind?

proposals/0341-v0-account-compression.md
igor56D: I was using `lthash.out` as described in SIMD-0215:

> out(a): blake3.fini( blake3.append( blake3.init(), a ) ), i.e. the 32-byte blake3 of the 2048-byte data.

^^ I'm confused by this but I assumed it was what you meant by 32-byte mode. I'll change the doc to explicitly use `blake3_32` to avoid confusion.

proposals/0341-v0-account-compression.md
igor56D: rather, we can make it so the decompression syscall does the delete. If the program does it then move vs copy isn't an implementation detail.

ripatel-fd: > the copy vs move becomes an implementation detail. Is that sort of what you have in mind?

@igor56D I feel like it would be very tricky to implement an account move in a syscall handler. Since all the sorrounding code (transaction executor, VM) assumes that accounts are pinned in memory.

EDIT: Would probably be easier if the move happens in the post-transaction cleanup after the program has finished executing.

proposals/0341-v0-account-compression.md
ripatel-fd: Haha yeah we don't have definitions of these blake3 pseudocode functions

proposals/0341-v0-account-compression.md
igor56D: good call, I'm being pretty hand-wavy because I'm not as familiar with these low-level details. Conceptually, it seems feasible to take `tmp_pubkey -> data`, "alias" `original_pubkey -> tmp_pubkey -> data` at decompression time, then resolve/collapse that whenever it makes the most sense.

I'll need to do some digging into the execution env to formulate something more concrete.

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,389 @@
...
+3. a compression scheme that removes accounts from the global state while
+   allowing for safe and simple recovery.
+
+This proposal focuses on (3), leaving (1) and (2) for other SIMDs.
+
+## New Terminology
+
+- **Compression**: replacing arbitrary account data with a fixed size
+  commitment. Not to be confused with traditional compression; the data
+  cannot be directly recovered from compressed state.
```
topointon-jump, line 41: "Compression" is overloaded, especially in this space (see: compressed nfts) and as mentioned here, account compression isn't analogous to traditional compression. Using a completely new term might be clearer, maybe "Compacting"? Just a suggestion, feel free to ignore.

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,389 @@
...
+- **Compression Condition**: a predicate determining whether or not an
+  existing account can be compressed. A specific compression condition isn't
+  provided in this SIMD -- it is assumed to always be false, meaning no
+  account is eligible for compression.
+- `data_hash = lthash.out(Account)` where Account includes the pubkey, 
+  lamports, data, owner, executable, and rent_epoch fields. See
```
topointon-jump, line 47: `rent_epoch` is not an input to the hash function.

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,389 @@
...
+
+## Detailed Design
+
+### Syscalls for Compression Operations
+
+The following new syscalls will be introduced to support account compression
+operations. The compression system program will support two instructions that
+just wrap these syscalls. For the time being, these syscalls can only be used
+from the compression system program but that constraint may be relaxed in the
+future.
```
topointon-jump, line 62: What is the reason for doing compression/uncompression in syscalls, instead of adding two new system program instructions for doing this? Is there an advantage to these operations being syscalls rather than native program instructions?

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,389 @@
...
+
+#### `sol_decompress_account(..)`
+
+**Purpose**: Recovers a compressed account by restoring its original data.
+
+**Parameters**:
+
+- `account_pubkey`: 32-byte public key of the account to decompress
+- `lamports`: The lamport balance of the original account
+- `data`: Pointer to the original account data bytes
```
topointon-jump, line 99: What if the original account data is larger than the max transaction size? There probably needs to be a mechanism for buffering the account data somewhere before uncompressing it.

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,389 @@
...
+    lthash.append( account.data_hash )
+    return lthash.fini()
+```
+
+### Account Creation Validation
+
+When creating new accounts, the runtime MUST verify the target pubkey does not
+already exist as a compressed account, just like with uncompressed accounts.
+
+#### Execution Error
```
topointon-jump, line 155: What happens if a user attempts to access a compressed account in the vm? Is this an execution error? I think it's simpler if it's required that all accounts that an instruction accesses (including through nested cpi calls) have already been decompressed before the vm invocation.

proposals/0341-v0-account-compression.md
topointon-jump: Similarly, why a new program? Compression/un-compression feels like it fits within the system programs purview.

proposals/0341-v0-account-compression.md
brooksprumo: I'd also love to not call it compression. I've been in favor of freeze/frozen/thaw/thawed personally.

proposals/0341-v0-account-compression.md
brooksprumo: Yeah, we should do something akin to program deployment.

proposals/0341-v0-account-compression.md
igor56D: I also don't like the name but figured it had enough history that renaming would introduce more confusion.

proposals/0341-v0-account-compression.md
igor56D: During VM execution, attempts to access a compressed account are equivalent to attempts to access a non-existing account, and returns the same error

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,389 @@
...
+- (optional) emit a compression event for off-chain data archival
+
+#### `sol_decompress_account(..)`
+
+**Purpose**: Recovers a compressed account by restoring its original data.
+
+**Parameters**:
+
+- `account_pubkey`: 32-byte public key of the account to decompress
+- `lamports`: The lamport balance of the original account
```
roryharr, line 98: What will happen to the lamports when the account is compressed from a global tracking perspective? 
Currently (I think) if you sum up all the lamports in all the accounts it should be equal to the bank capitalization. With this change that will no longer work without decompressing all the accounts.

Maybe the get transferred to some kind of system account?

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,389 @@
...
+**Behavior**:
+
+- MUST verify that the caller is the hardcoded compression system program
+- MUST verify the provided account satisfies the compression condition
+- MUST mark the account as compressed if verification succeeds
+- if the transaction containing the compression request succeeds, all
+  subsequent attempts to access the account MUST fail unless the account has
+  been decompressed.
+
+While marking the account as compressed must be done synchronously, the
```
roryharr, line 82: I'm not sure what this paragraph is meant to address, or if I'm missing something. This is an implementation detail for the backing database that is irrelevant to this SIMD.

A validator could never compress the account in the database (this may be useful for an RPC provider) as long as they calculate the lt_hash correctly.

igor56D: I agree, this is bordering on an implementation detail. I'll remove it to avoid confusion.

proposals/0341-v0-account-compression.md
brooksprumo: Yep, rent epoch was removed with the Accounts Lt Hash in SIMD-215. Let's remove it here as well.

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,389 @@
...
+      requirements, load limits, etc all apply
+- if verification succeeds, the compressed account entry must be replaced
+  with the full account data
+
+### Database and Snapshot Extensions
+
+#### Compressed Accounts Storage
+
+Compressed accounts are stored directly in the account database like regular
+accounts, but with a special compressed account structure:
```
brooksprumo, line 122: Seems like implementation details? We may or may not do *exactly* this. I agree we need the account's pubkey and `data_hash`, but I envision us doing it as the `CompressedAccount` struct shows below.

proposals/0341-v0-account-compression.md
brooksprumo: > (What we should NOT do is something like `blake3_32(blake3_2048(lthash_preimage))`)

Oops 🙃 I do `blake3_32(blake3_2048(..))` for the checksum currently:
https://github.com/anza-xyz/agave/blob/7df820d832d2ae1152e7cdc13e495797600f4b3d/lattice-hash/src/lt_hash.rs#L52-L56

But, that's only used for logging purposes and can be changed. We should do the right thing for compression stuff.

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,389 @@
...
+### Bank Hash Changes
+
+- **Impact**: Bank hash calculation includes compressed accounts, affecting
+  consensus
+- **Mitigation**: Feature gate activation ensures all validators adopt
+  simultaneously
+
+### Snapshot Format
+
+- **Impact**: New snapshot format including compressed accounts
```
brooksprumo, line 379: I don't think we need (nor want) to change the snapshot format.

We *do* need to modify the account storage file format though.

proposals/0341-v0-account-compression.md
ultd: Agreed with the buffering but thinking out loud, would it actually be beneficial to limit the size of compressed accounts to the max transaction size? If an account is bigger than the tx size limit, maybe it should live uncompressed onchain to prevent tpu/tx bloat?

igor56D: this shouldn't materially impact tx bloat. If someone wants a large account with specific data on-chain, they need to  upload the data in parts across multiple transactions. The same applies to a large account getting decompressed. When an account is decompressed, it persists on-chain like any other active account unless it gets compressed again so subsequent accesses don't need to re-upload.

proposals/0341-v0-account-compression.md
```diff
@@ -0,0 +1,389 @@
...
+compression event. This ensures that off-chain services have the opportunity
+to archive the original account data before it is permanently removed.
+
+This approach enables:
+
+- **Archive services**: Third-party services can maintain comprehensive
+  compressed data archives using existing subscription infrastructure
+- **Application-specific storage**: DApps can store their own compressed
+  account data through established patterns
+- **Redundancy**: Multiple parties can maintain copies for data availability
```
Sobeston, line 220: Could we end up in a situations where:
a) the original data of a compressed account is entirely lost
b) the data can be found by some validators, but not others (e.g. some validators are hooked up to a paid 3rd party for storage and some aren't)

What would you expect/want to happen in these scenarios? Can redundancy be truly guaranteed, and can it be fairly distributed to all validators (and quickly enough for a slot, globally)? 

Is the expectation that a solana project would ensure that it does a good job at storing its data off-chain?

Let's say there's some trading happening which requires the decompression of an account. If that project turned their archive service off, and validators can't access that data, could that risk assets becoming un-tradeable? This feels like we could risk things becoming centralised. Let me know if I've misunderstood something.

proposals/0341-v0-account-compression.md
Sobeston: Some alternative naming suggestions: "stubbed" accounts, "digest" accounts, "proxy" accounts

proposals/0341-v0-account-compression.md
igor56D: There is a theoretical possibility that compressed account data is completely lost, though it's extremely low. Even if every indexer, RPC provider, wallet provider, etc somehow loses the data (or refuses to release it), you can always fall back to replaying the ledger history from some snapshot before the account was compressed.

The protocol provides the guarantee that you can always recover a compressed account *if* you can provide the original data. Users/apps must decide for themselves how strong the data availability guarantee must be for their use case, and plan accordingly. The best guarantee comes from running your own full node that's synced with the network. Others may prefer to subscribe to a stream of compression events from an RPC provider and store the compressed data themselves. Less sophisticated users are likely to rely on wallet backups or indexers to keep the data available.

In any case, the complexity is moved out of the protocol and into off-chain infra.

> If that project turned their archive service off, and validators can't access that data, could that risk assets becoming un-tradeable?

In reality, most full node indexers will have this data stored, though this isn't backed by clearcut incentives for now. Worst case scenario is needing to replay the ledger. There's also the option of storing the data in some distributed storage protocol like filecoin or IPFS. The availability concern is purely theoretical until the compressed state size grows dramatically, in which case we'll need to consider a more complete data availability system.

ultd:
Could we add a transaction-scoped (ephemeral) decompression syscall so a tx can read (and optionally commit a write to) a `CompressedAccount` without rehydrating it in AccountsDB? This would cut AccountsDB churn and still let programs operate over the account state. 

We would intro a a cpl other new syscalls:

```rs
sol_transient_decompress_account(
    compressed_acct_pubkey: Pubkey, // aka index into accounts array
    current_state: &[u8]            // full account struct bytes 
);

sol_transient_commit_compressed_account(
    compressed_acct_pubkey: Pubkey, 
    new_state: &[u8],         
);
```
 
The runtime just verifies `hash(current_state) == data_hash` from the `CompressedAccount`. No modification/rehydration of AccountsDB. If succeeds, converts `CompressedAccount` to an ephemeral, read-only view of the account that is addressable during the tx (maybe stored directly in `TransactionContext` for other ix to use). The account must be listed in the tx’s account metas per usual, locks still ensure consistent access ordering if multiple txs target it. At the tx's end, ephemeral memory is discarded.

Example pseudo-code of how used in a tx:  
```python
compressed_account_state = [...]
compressed_account_pubkey = compressedAccount1A3b5D7f9GhKm2pQ4rStVxYz6uL8

TX(a):
    accounts: [owner, compressed_account_pubkey, new_owner, SYSTEM_PROGRAM]
    signers: [owner]
    instructions:
         0: sol_transient_decompress_account(compressed_account_pubkey, compressed_account_state)
         1: transfer_asset_ownership(compressed_account_pubkey, new_owner); sol_transient_commit_compressed_account(compressed_account_pubkey, new_state)
```

The other `sol_transient_commit_compressed_account` syscall would allow one to optionally commit new account state to a `CompressedAccount`. This would work with `sol_transient_decompress_account` syscall which would need to be called first to acquire the compressed account. On commit, the `sol_transient_commit_compressed_account` syscall would update the `CompressAccount` entry’s `data_hash` field with new computed hash without ever storing full bytes in AccountsDB.

This covers the common pattern where an authority needs to read or locally authorize effects (NFTs/RWAs/hotspots, etc.) but doesn’t need the account to remain "active" after the tx. It avoids touching AccountsDB entirely in the read-only case, and avoids the rehydrate/recompress loop in the write-through case. It also keeps snapshots small since the compressed entry never expands.

This wouldn't work well with Accounts meant to be "global", shared state (DEX orderbooks, AMMs) that needs a consistent, persistent view across txs. Those should just remain in active set. Also, this probably only works well for small account sizes. Probably should limit to Accounts that are max N bytes as you don't want each tx submitting excess ix bytes causing TPU bloat. 

### **EDIT:** 

Maybe we don't need `sol_transient_commit_compressed_account` as the runtime should probably commit the new hash to a `CompressedAccount`'s `data_hash` field upon tx success implicitly. In fact, it's probably necessary.

igor56D:
@ultd IIUC this isn't incompatible with the existing proposal -- no new syscalls would need to be introduced. This proposal intentionally leaves the compression condition itself stubbed out to focus discussion on the compression mechanism rather than all the potential ways to use it (ephemeral compressed accounts, rent eviction, etc).

For example, if there was a subsequent SIMD that expanded that compression condition to include something like `... || caller == account.owner` it would give the owner of an account the authority to compress it. With this they can use the existing syscalls to compress/decompress at will.

proposals/0341-v0-account-compression.md
brianlong: Anything but "compression"! Frozen/thawed sounds better.

proposals/0341-v0-account-compression.md
brianlong: > There is a theoretical possibility that compressed account data is completely lost, though it's extremely low.

Guarantees of future availability are essential. This doesn't go far enough.

> fall back to replaying the ledger history from some snapshot before the account was compressed

This is not as easy as it sounds (according to people who have done it and told me it was tough).

> full node indexers will have this data stored

This is a weak economic assumption. There's not much incentive to store terabytes of cold data in a separate data store for the outside chance that someone will pay $0.000025 to read it. Keep in mind the accounts are primarily cold because people aren't using/reading them.

However, some of the discussion further above suggests a solution that comes closer to a guarantee of future availability with a data store that will always exist and does not require replay. The process for thawing a frozen account, which is similar to deploying a program using a buffer account, might also be used to freeze the account: write the account state to the ledger in one or more TXs before freezing it.

1. The theoretical possibility of losing the ledger history is almost nil as long as Solana exists.
2. The ledger history is replicated across many independent players who have a financial incentive to keep it. Also, the raw data is publicly available at no cost today, so anyone can load historical data if they prefer to roll their own. The economic incentives are aligned.
3. The process to thaw a frozen account will still require an off-chain actor to read the account data from the ledger and then use that as an input to the thawing process. However, you won't need to replay TXs; just read the ledger history.

Footnotes:
- The validators are disconnected from the full ledger history, so the entire process can't be done in-protocol. The external actor will be required to read from the ledger history & execute the thawing process.
- Using the ledger history might also give some extra cryptographic peace of mind since the archived account state can be verified within the block by using the hash. In comparison, data stored in an off-chain SQL database might become corrupted or incomplete.

A parting question: Since the validators are not connected to the ledger history, does that mean the ledger history is technically an "off-chain database"? (I'll see myself out) ;-)

aeyakovenko: So first of all, a solana "user" is assumed to have a full node, or is paying someone else to run a full node directly or indirectly. There is no way for the protocol to make any guarantees about users data without this assumption. The best any protocol can do is that if the users one full node stays consistent then the user is not at risk of any loss. 


If a user has access to a full node they can always replicate just part of the compressed state that user cares about.

We need tools to automate the compressed state recovery from the ledger. 

If you want weak assumption that there exists at least one full node in the network that will keep the user state, that user just needs to pay enough to keep it from being compressed. But this is also a weak economic incentive.

brianlong: > We need tools to automate the compressed state recovery from the ledger.

The elegance of storing the frozen account state in the ledger is that the DB storage & retrieval tools already exist. getSignaturesForAddress() + getTransaction() or an Old Faithful gRPC stream will return the account state without a specialized SQL index or API. The new tooling will be related to the freeze/thaw processes.

igor56D: Interesting suggestion. I'm not totally opposed but it does make compressing larger accounts much trickier. Not only will it need to be done over multiple transactions, but we'll need to add an intermediate "frozen but not compressed" state that doesn't allow modifications to the account so that whoever is providing the data doesn't get rugged.

This gets even trickier w/ incentives. The compressing needs to be attributable so rewards can be distributed but a reasonable amount of time needs to be provided for the compressor to provide all the data. This time can't be indefinite because the account would stay in the "frozen but not compressed" state forever. Overall, it's messy.

Is there maybe other existing offchain infra besides raw ledger storage that we can piggyback on that doesn't require adding complexity to the protocol?

brianlong: None that can offer a guarantee of completeness or future availability. For example, streaming accounts through Geyser to a database won't work because there's no guarantee that messages will be received or that the database will be complete. Losing a single message could have catastrophic consequences for the account holder.

I think all of your points about the tricky implementation are valid. Nonetheless, the guarantee is more important than the implementation challenge. Saving data in the ledger checks the boxes for storage & retrieval guarantees, universal availability, & economic alignment for the infrastructure providers.

igor56D: I can't figure out a way to make this work well enough. Even if we introduce the complexity I mentioned above, compression requests will still be prohibitively expensive. It's very important that compression is cheap -- the network needs to be able to promptly and sufficiently respond to growing state/rent prices by evicting delinquent state, which may require a lot of compression requests. Requiring all the compressed state to flow through the blocks will be a challenge for throughput.

You mentioned that replaying the ledger is difficult now. Would addressing that problem make more sense? Ledger history can tell you the slot the account was compressed at, so you'll know how far to replay. Not sure how much we can rely on historical snapshot retention, but reliable replay + snapshot fetching should make account reconstruction relatively doable.

igor56D: if that doesn't seem sufficient, I'd prefer @aeyakovenko's previous idea of maintaining a separate compressed state snapshot that's replicated on all validators over making compression really difficult.

brianlong: > maintaining a separate compressed state snapshot that's replicated on all validators

This also works, and was my original suggestion at some earlier meetups. Hot & Cold account sets with separate snapshots & hashes. NVMe storage is affordable, and drive sizes keep getting bigger. JWash mentioned a concern about validator start times when the validator needs to validate two account DBs. That could be a more solvable problem and provide the guarantees.

brianlong: P.S. Validator startup times will be slower, but all good validators run hot spares, so there's minimal downtime during failover. Slow startup times are manageable.

JSTONE1111 (approved):

Benhawkins18:
stale and limited alignment on direction

PR #357, opened 2025-09-12
SIMD-0357: Alpenglow VAT implementation

wen-coding:


proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,179 @@
...
+
+4. If the number of filtered accounts exceeds 2,000, then sort according to the
+following rules and pick the top 2,000, otherwise return the whole list:
+
+  - Sort by descending order of stake (largest to smallest)
+
+  - If several validators have exactly the same amount of stake and including
+  all of them would exceed the 2,000 limit, then all of them are excluded
+
+5. Subtract 1.6 SOL from the corresponding vote identity account in the list of
```
jstarry: what happens if the vote identity doesn't have 1.6 SOL? We should also probably restrict vote identities to being a system owned account so that we don't unintentionally subtract lamports from a locked stake account or something

wen-coding: The proposal right now is to exclude those accounts without sufficient funds from staked validators in the new Epoch, so they can't vote or be leaders.

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,179 @@
...
+White Paper) is that Alpenglow uses BLS to aggregate signatures. So a future
+Alpenglow vote account must contain a valid BLS pubkey deducted from its
+current vote authority ed25519 private key.
+
+In summary, a valid vote account in an Alpenglow epoch must contain:
+
+- a BLS pubkey
+
+- at least 1.6 SOL VAT fee plus the necessary storage rent amount for a new
+epoch in its corresponding identity account
```
AshwinSekar: any reason we choose identity vs vote account?
Implementation wise it seems more complex as multiple vote accounts can specify the same identity

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,179 @@
...
+
+When the staked validators for a new epoch is calculated, the leader will
```
AshwinSekar: this affects all uses of staked validators - turbine, leader schedule etc correct? would be good to specify.
Also any behavior regarding staked validator overrides - if I have it set does it create a 2001 spot or replace a validator.

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,179 @@
...
+
+  - Sort by descending order of stake (largest to smallest)
+
+  - If several validators have exactly the same amount of stake and including
+  all of them would exceed the 2,000 limit, then all of them are excluded
+
+5. Subtract 1.6 SOL from the corresponding vote identity account in the list of
+accepted validators in the previous step
+
+6. Record the VAT fee subtraction in the bank to calculate inflation. Step 6
```
AshwinSekar: how is this used to calculate inflation? would be good to expand

proposals/0357-alpenglow_validator_admission_ticket.md
0x0ece: No please, I'm in favor of the current design. All rewards go into the validator's identity and no sol goes into vote account (unless there's commissions). It'll be a pain to keep 1.6 sol/epoch in vote account.

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,179 @@
...
+1. When a new bank crosses the epoch boundary (bank.epoch() >
+parent_bank.epoch()), we calculate the participating staked validators for the
+next epoch (bank.epoch() + 1)
+
+2. Perform stake activation and deactivation, so we will use the intended stake
+for the new epoch
+
+3. The calculation goes through all vote accounts and filter the following:
+
+  - Contains at least 1.6 SOLs as balance in its corresponding identity account
```
0x0ece: Currently, the same identity could be used on multiple vote accounts. 😭

My strong preference is to enforce that, for every epoch, there's only 1 vote account per identity (if there's more, we can pick the vote account with higher stake). This simplifies a ton of other things.

Alternatively, we'd have to handle the case where the identity has enough sol for some vote accounts but not for all, and we'd have to agree on which ones we'd include and exclude.

proposals/0357-alpenglow_validator_admission_ticket.md
jstarry: We already have a precedent that validator's identity account pays vote tx fees so I agree we should keep it as the identity. But @AshwinSekar makes a great point that we need to figure out what to do if multiple vote accounts specify the same identity account and the account doesn't have a high enough balance to pay for all of them.

wen-coding: Yeah the identity account was mostly chosen because currently all rewards go there, so if someone has enough rewards to cover VAT they don't ever need to worry about transferring money into vote accounts.

Per the current algorithm, if multiple accounts use the same identity account, the fees will be deducted in descending stake order (then descending pubkey). If you have 3 vote accounts using the same identity but your fee can only cover 2, then the two higher staked accounts get in while the third one is out.

If people are okay with that behavior, I'll specify it in the SIMD.

0x0ece: @jstarry I added the same comment below :) I'd fix 1-1 mapping at epoch boundary.

proposals/0357-alpenglow_validator_admission_ticket.md
0x0ece: Today this is not the case. The turbine tree is generated with a list sorted by (stake, identity), while the leader schedule is generated from the same list sorted by (stake, vote). So the lists are currently different. I'm of course highly in favor to unify, but that's prob a separated turbine-specific simd?

wen-coding: Hmm do we have staked validator overrides for mainnet or testnet? I suppose staked validator overrides should be considered before we do the sorting, so it doesn't create the 2001th spot, it kicks one validator in the 2000 list out.

Err, I think currently Turbine gets node stake from the same list we are calculating here, so yes Turbine will be affected, unless we intentionally give Turbine a different list. Is there a reason we don't want unity here?

That said, theoretically we should never hit this 2000 limit, so we should be fine...

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Would be ideal if we can enforce 1-1 mapping of vote account to identity account, less confusion. Let me chat with Foundation folks to see whether there is any legitimate use of multiple vote accounts for one identity account.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Asked @tigarcia and currently we don't know any good reason for multiple vote accounts per identity account. So maybe let's put 1:1 mapping requirement in this SIMD or some previous SIMD and see if the community can accept it.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Any fee burned will reduce total capitalization of the bank. In current implementation capitalization does change epoch inflation reward. But I re-read SIMD 326 and I'm not sure that will be true in Alpenglow inflation calculation.

Anyways, I think it's out of the scope for this SIMD to discuss the impact of reduced capitalization on inflation rewards, so I reworded this bullet point to just say with the burned fee the bank capitalization will be reduced accordingly.

proposals/0357-alpenglow_validator_admission_ticket.md
0x0ece: The core issue is just that SIMD-0180 was done only for leader schedule and not for turbine tree. So now we seed the two randomizations with different lists. What I wanted to say is that I'm in favor of unifying, but maybe we should just do an independent simd to re-align turbine tree with the new list.

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,179 @@
...
+- The VAT is a temporary measure before we implement a better mechanism
+
+- There is a strict limit of the 2,000 highest staked validators
+
+- Before being admitted to an epoch, the 1.6 SOL per epoch VAT fee is deducted
+  from each validator and burned
+
+Another requirement (not explicitly mentioned in SIMD 326 but in the Alpenglow
+White Paper) is that Alpenglow uses BLS to aggregate signatures. So a future
+Alpenglow vote account must contain a valid BLS pubkey deducted from its
```
lidatong: "deducted"? should this mean "deduced" / "derived"?

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,179 @@
...
+
+7. The epoch stakes for epoch (bank.epoch() + 1) is generated from the list
+generated in step 4. The epoch stakes are saved in bank snapshots, so if we
+restart from a snapshot from any bank in the same epoch as bank.epoch(), the
+1.6 SOLs fee will not be charged again
+
+## Operation Considerations
+
+- To be included in epoch e+1, validator operators must ensure the vote
+identity account has at least 1.6 SOL before epoch e-1 ends
```
lidatong: do you mean `epoch e` or there actually is 1 epoch delay?

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Oops, thanks for catching this, fixed.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Yes, there is actually 1 epoch delay. At end of epoch e-1 we calculate e+1.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: For the record, we discussed this at the SLC dev meeting, I believe the current consensus is that it's okay to unify the list, since we should almost never hit the 2000 limit, this should be fine and unifying simplifies the reasoning.

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,179 @@
...
+status: Review
+created: 2025-09-11
+feature: (fill in with feature key and github tracking issues once accepted)
+---
+
+## Summary
+
+This SIMD describes how the validator admission ticket (VAT) collection
+described in SIMD 326 will be implemented. Specifically, how it affects
+validator operation procedures.
```
bw-solana: we should add a very brief explanation of what the Validator Admission Ticket is and maybe why it is needed here (can expand on this in Motivation section)

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,179 @@
...
+## Dependencies
+
+The Validator Admission Ticket is specified in SIMD 326 (Alpenglow).
+
+The compressed BLS Pubkey in Vote Account is specified in SIMD 185 (Vote
+Account v4)
+
+## Motivation
+
+The VAT discussion in SIMD 326 proposes:
```
bw-solana: All of this seems like it belongs in the description or design portions of the SIMD.

Motivation should cover how there needs to be some cost (the VAT) associated w/ everyone doing work on your behalf (primarily processing your votes and rewards calculations)

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,179 @@
...
+  - If this validator ends up not being selected, this VAT fee needs to be
+returned, which means we have to implement some type of accounting and ensure
+the VAT fee returning properly lands on the chain.
+
+  - It does deduct an additional transaction fee from the vote account
+
+  - If somehow validators make a mistake by sending multiple transactions,
+we also need to return the fee collected.
+
+## Impact
```
bw-solana, line 132: Would be good to mention the nodes selected by this process are the only ones that votes/certs will be sent to be default

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Added a few sentences, does this look better?

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Reorganized it a bit, how does this look?

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: done.

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,207 @@
...
+
+The compressed BLS Pubkey in Vote Account is specified in SIMD 185 (Vote
+Account v4)
+
+## Motivation
+
+Adding more voting validators to a blockchain does come with costs. At the
+very least, the votes and corresponding rewards need to be processed by
+every voting validator. Therefore, if an attacker can cheaply start a lot
+of voting validators, it will put more pressure on the chain.
```
bw-solana: doesn't necessarily have to be an attacker. Even a well meaning full node exerts a cost on all of the other nodes

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,207 @@
...
+
+Right now every voting validator pays voting fee on any vote transaction
```
bw-solana, line 48: staked* validator

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,207 @@
...
+it sends. The voting fee adds up to ~2 SOLs per epoch if the validator votes
```
bw-solana: s/it sends/included in a block/

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,207 @@
...
+most of the time. It is a burden, yet at the same time it's an economic
+barrier to the above attacks.
+
+Of course, VAT is only a temporary solution to maintain the current economic
+equilibrium. Because Alpenglow is a revolutionary consensus redesign, we prefer
+to keep other things like economics unchanged for now. There will be future
+proposals for how to discourage the attackers while not putting too much burden
+on validator operators.
```
bw-solana: I would reword this to something simple like:
"This proposal intentionally strives to keep voting validator protocol costs similar to pre-Alpenglow consensus."

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,207 @@
...
+type: Core
+status: Review
+created: 2025-09-11
+feature: (fill in with feature key and github tracking issues once accepted)
+---
+
+## Summary
+
+This SIMD describes how the validator admission ticket (VAT) collection
+described in SIMD 326 will be implemented. Specifically, how it affects
```
bw-solana: would be nice to link out when referring to other SIMDs

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Done.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Hmm, actually if they don't vote they don't pay any voting fee even if they are staked I believe? I changed a few voting validator to staked validator

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Done.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Done.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Added.

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,206 @@
...
+
+This SIMD describes how the validator admission ticket (VAT) collection
+described in [SIMD 326](https://github.com/solana-foundation/solana-improvement-documents/pull/326)
+will be implemented. Specifically, how it affects validator operation
+procedures.
+
+Validator admission ticket is a mechanism translating the current cost of
+voting into a similar economic equilibrium for Alpenglow. By charging every
+voting validator 1.6 SOL per epoch, it replaces the current voting fee at ~2
+SOLs per epoch, and reduces the likelyhood there are too many voting validators
```
qkniep: nit: typo
```suggestion
SOLs per epoch, and reduces the likelihood there are too many voting validators
```

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,206 @@
...
+validator eligible to participate in the next epoch
+
+## Different Vote Related Accounts
+
+Before Alpenglow, we have the following accounts related to voting (the
+identity account and vote authority account can be the same):
+
+- Vote account for saving all the vote states
+
+- Identity account for receiving block rewards and commission
```
AshwinSekar: nit: commission is paid to the vote account

wen-coding: Oops, thanks for catching that.

proposals/0357-alpenglow_validator_admission_ticket.md
AshwinSekar: so just to clarify this affects all usages of validator stake / total stake in the protocol? Or could you specify exactly when we use this filtering vs the "normal" stake view?

wen-coding: This will affect all usages of validator stake / total stake in the protocol.

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,206 @@
...
+---
+
+## Summary
+
+This SIMD describes how the validator admission ticket (VAT) collection
+described in [SIMD 326](https://github.com/solana-foundation/solana-improvement-documents/pull/326)
+will be implemented. Specifically, how it affects validator operation
+procedures.
+
+Validator admission ticket is a mechanism translating the current cost of
```
rogerANZA: The ...

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,206 @@
...
+voting into a similar economic equilibrium for Alpenglow. By charging every
+voting validator 1.6 SOL per epoch, it replaces the current voting fee at ~2
+SOLs per epoch, and reduces the likelyhood there are too many voting validators
+immediately after Alpenglow launches.
+
+Everything specified below are protocol-level changes, they need to be
+implemented by all Solana clients. The client specific data structure
+changes are omitted.
```
rogerANZA, line 32: The general VAT concept has already been accepted with the governance vote on SIMD 326. (at the top of this paragraph)

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,206 @@
...
+
+The compressed BLS Pubkey in Vote Account is specified in [SIMD 185 (Vote
+Account v4)](https://github.com/solana-foundation/solana-improvement-documents/pull/185)
+
+## Motivation
+
+Adding more staked validators to a blockchain does come with costs. At the
+very least, the votes and corresponding rewards need to be processed by
+every voting validator. Therefore, every additional staked validator will
+put a bit more pressure on the chain.
```
rogerANZA: will put a bit more pressure... --> increases the number of messages for every validator.

proposals/0357-alpenglow_validator_admission_ticket.md
rogerANZA: also below: just 2 SOL (not SOLs) IMHO, like 1.6 SOL above.

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,206 @@
...
+
+Right now every voting validator pays voting fee on any vote transaction
+included in a block. The voting fee adds up to ~2 SOLs per epoch if the
+validator votes most of the time. It is a burden, yet at the same time it's an
+economic barrier to having too many voting validators on the chain.
```
rogerANZA: on chain. (or just ... validators.)

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,206 @@
...
+
+Of course, VAT is only a temporary solution to maintain the current economic
```
rogerANZA: without of course, just "The VAT is..."

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,206 @@
...
+equilibrium. This proposal intentionally strives to keep voting validator
+protocol costs similar to pre-Alpenglow consensus.
+
+## New Terminology
+
+- **Validator Admission Ticket**: The 1.6 SOL charged once per epoch to every
```
rogerANZA: (VAT)

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,206 @@
...
+
+- Identity account for receiving block rewards
+
+- Vote authority account for signing vote transactions
+
+After Alpenglow, we will still have roughly the same accounts serving the same
+purpose:
+
+- Vote account: This account must contain the correct BLS pubkey corresponding
+to vote authority keypair; it continues to keep all the vote credits and all
```
rogerANZA: somewhere a "." would be great (sorry for the many nits)

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,206 @@
...
+it doesn’t contain vote information any more.
+
+- Identity account: This is the account for the 1.6 SOL VAT; it continues
+receiving block rewards and commissions.
+
+- Vote authority account: For signing BLS messages
+
+## Detailed Design
+
+### What is considered valid vote account
```
rogerANZA: a valid vote account

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,206 @@
...
+current vote authority ed25519 private key.
+
+In summary, a valid vote account in an Alpenglow epoch must contain:
+
+- a BLS pubkey
+
+- at least 1.6 SOL VAT fee plus the necessary storage rent amount for a new
+epoch in its corresponding identity account
+
+When the staked validators for a new epoch is calculated, the leader will
```
rogerANZA: ... are calculated

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,206 @@
...
+perform the following operations:
+
+- When there are more than 2,000 valid validators, sort all valid vote accounts
+by descending order of stake. We deterministically sort all validators by
```
rogerANZA: 2x same sentence

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,206 @@
...
+### How to implement the checks
+
+1. When a new bank crosses the epoch boundary (bank.epoch() >
+parent_bank.epoch()), we calculate the participating staked validators for the
+next epoch (bank.epoch() + 1)
+
+2. Perform stake activation and deactivation, so we will use the intended stake
+for the new epoch
+
+3. The calculation goes through all vote accounts and filter the following:
```
rogerANZA: (grammar)

proposals/0357-alpenglow_validator_admission_ticket.md
rogerANZA: if we do it in between epochs (as it is described above), then having it at the end of epoch e should be enough, right?

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,206 @@
...
+
+Validators not providing BLS Pubkey or desired fee will not be able to
+participate in an Alpenglow epoch regardless of their stake. Also, only the
+nodes selected by this process will receive votes or certificates from other
+selected validators in real time.
+
+## Security Considerations
+
+Validator operators need to ensure they have enough fund and correct BLS
+pubkey before end of epoch s to partipate in epoch s+2. This poses some new
```
rogerANZA: s -> e
s+2 -> e+1
(s is not so good because we usually use it for slot)
the math is slight off, see above.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Done.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Done.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Added. I think it doesn't quite work to combine the two into one paragraph, so I added a separate paragraph.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Done.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Changed.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Done.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Done.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Added. Thanks for all the feedback!

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Done.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Done.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Oops, removed.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Changed, does this look better?

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Hmm, at end of epoch e we calculate the validators for e+2 already. So to be in e+1 you need to have enough stake at end of epoch e-1. Does that make sense?

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Changed.

rogerANZA: (the e-1 -> e proposal might have been premature... maybe there is something that needs to stabilize first)

rogerANZA: Wen: Yes we really have to go back to e-1 in the both places. Makes just everything much easier, because then the chain can settle more easily.

wen-coding: Oops, did I somehow miss this one. Changed.

I believe we calculate two epochs ahead (not one epoch ahead) just because we are worried things will oscillate at the end of an epoch.

Say you calculate staked validators at end of e-1 for e, but different people might have different opinions which block is "end of e-1", this can lead to consensus divergence.

If we can prove this can't happen in Alpenglow, then we might change this. But I prefer to keep it as is for Alpenglow launch.

proposals/0357-alpenglow_validator_admission_ticket.md
rogerANZA: Yes, it's good now. While it's in principle possible to go from e-1 to e directly (Quentin said Sui does this), it is a major engineering effort. Just lowering epoch times to 1h (eventually) also solves many of the practical problems. This way we can be sure that we have a finalized block before we do the computation. But technically we could do that whole between epochs computation that you describe as soon as we have a finalized state in epoch e. Not sure whether we want to mention this in this document, but it might be good.

wen-coding: Totally agreed, I think after Alpenglow launches we can explore both directions. But I wouldn't mention those in this SIMD since we don't have a concrete proposal yet.

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,213 @@
...
+
+## Impact
+
+Validators not providing BLS public key or desired fee will not be able to
+participate in an Alpenglow epoch regardless of their stake. Also, only the
+nodes selected by this process will receive votes or certificates from other
+selected validators in real time.
+
+Validator operators need to ensure they have enough fund and correct BLS
+public key before end of epoch e-1 to participate in epoch e+1. This poses some
```
rogerANZA: Maybe here 1 sentence: "Why e-1? Because we want to be absolutely sure that we have a finalized state before we enter epoch e+1. Specifically, if the last slots of epoch e-1 are not finalized at the end of the epoch, we have enough slots in epoch e to finalize epoch e-1's state."

wen-coding: Sounds good, added, but reworded the first sentence.

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,219 @@
...
+
+- Vote authority account: For signing BLS messages
+
+## Detailed Design
+
+### What is considered a valid vote account
+
+The VAT discussion in SIMD 326 proposes:
+
+- The VAT is a temporary measure before we implement a better mechanism
```
t-nelson: this point doesn't really add anything to the simd. if anything it risks calling it into question. suggest removal

proposals/0357-alpenglow_validator_admission_ticket.md
t-nelson: another benefit of depositing VAT into the vote account is that it is more secure.  the vote account withdraw authority private key is not required to be resident on the host machine whereas the identity private key must be available to sign gossip messages. operators could be spared the calculus of deciding how much SOL to keep available for VAT vs risk of funds loss due to host intrusion

the only reasons rewards are paid to the gossip identity today are that
a) the private key must be hot
b) the using the same key for tx fee payer and vote authority is one fewer transaction signatures, reducing the base fee
c) the fee payer needs funded

iirc the vote account was the rewards recipient pre-mb before someone realized the above cost savings was available

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,219 @@
...
+
+In summary, a valid vote account in an Alpenglow epoch must contain:
+
+- a BLS public key
+
+- at least 1.6 SOL VAT fee plus the necessary storage rent amount for a new
+epoch in its corresponding identity account
+
+When the staked validators for a new epoch are calculated, the leader will
+perform the following operations:
```
t-nelson:
```suggestion
When the staked validators for a new epoch are calculated, all validators must
perform the following operations:
```

this statement partially duplicates and partially contradicts the last statement in this section. correct it here, delete it there

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,219 @@
...
+2001, then we remove all validators with stake S and less. If there are fewer
+than 2,000 valid validators, pick all of them.
+
+- Deduct 1.6 SOL VAT fee from each picked vote identity account once
+
+- Mark the fee burned and write the result into the bank
+
+All validators must perform the same operation, or the resulting bankhash will
+be different.
+
```
t-nelson:
```suggestion
```

duplicative of first statement in section

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,219 @@
...
+- a BLS public key
+
+- at least 1.6 SOL VAT fee plus the necessary storage rent amount for a new
+epoch in its corresponding identity account
+
+When the staked validators for a new epoch are calculated, the leader will
+perform the following operations:
+
+- When there are more than 2,000 valid validators, sort all valid vote accounts
```
t-nelson: sorting is fine for now with the 2k limit, but is not generally a scalable operation

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,219 @@
...
+be different.
+
+### How to implement the checks
+
+1. When a new bank crosses an epoch boundary (bank.epoch() >
+parent_bank.epoch()), calculate the participating staked validators for the
+next epoch (bank.epoch() + 1). This is the same as now.
+
+2. Perform stake activation and deactivation so that the intended stake values
+are used for the new epoch.
```
t-nelson: the timing of this operation is insufficiently specified. an implementer needs to know when it happens in relation to at least feature gate activation, rewards payouts and stake (de)activation. not sure if there are other dependencies that could be in conflict. a lot happens on the epoch boundary

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,219 @@
...
+
+3. The calculation goes through all vote accounts and filters those that meet
```
t-nelson:
```suggestion
3. The calculation iterates all vote accounts and filters those that meet
```

this list is effectively unbounded. the naive approach described here may not be practical

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,219 @@
...
+the following criteria:
+
+  - The corresponding identity account has a balance of at least 1.6 SOL
+
+  - The account has a valid BLS compressed public key (i.e., it can be
+    correctly decompressed)
```
t-nelson: this seems unnecessarily. we can do it once when the bls vote authority is set and fail the instruction if decompression fails. no "invalid" vote account should be allowed to exist

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,219 @@
...
+- at least 1.6 SOL VAT fee plus the necessary storage rent amount for a new
+epoch in its corresponding identity account
+
+When the staked validators for a new epoch are calculated, the leader will
+perform the following operations:
+
+- When there are more than 2,000 valid validators, sort all valid vote accounts
+by descending order of stake. If some validator with stake S is in position
+2001, then we remove all validators with stake S and less. If there are fewer
+than 2,000 valid validators, pick all of them.
```
t-nelson: this is duplicative of section [4] below. seems we don't need them both

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,219 @@
...
+happen here, but it doesn’t contain vote information any more.
+
+- Identity account: This is the account for the 1.6 SOL VAT; it continues
+receiving block rewards and commissions.
+
+- Vote authority account: For signing BLS messages
+
+## Detailed Design
+
+### What is considered a valid vote account
```
t-nelson: this section seems wholly duplicative of the next. i don't think we need them both

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,219 @@
...
+
+When the staked validators for a new epoch are calculated, the leader will
+perform the following operations:
+
+- When there are more than 2,000 valid validators, sort all valid vote accounts
+by descending order of stake. If some validator with stake S is in position
+2001, then we remove all validators with stake S and less. If there are fewer
+than 2,000 valid validators, pick all of them.
+
+- Deduct 1.6 SOL VAT fee from each picked vote identity account once
```
t-nelson:
```suggestion
- Deduct 1.6 SOL VAT fee from each picked gossip identity account once
```

there is no such thing as a "vote identity." there is only linkage of a vote account to a gossip identity

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,219 @@
...
+
+4. If the number of filtered accounts exceeds 2,000, sort them according to the
+following rules and select the top 2,000. Otherwise, return the entire list:
+
+  - Sort in descending order of stake (largest to smallest)
+
+  - If multiple validators have exactly the same amount of stake and including
+  all of them would exceed the 2,000 limit, then exclude all of them
+
+5. Subtract 1.6 SOL from the corresponding vote identity account for each
```
t-nelson:
```suggestion
5. Subtract 1.6 SOL from the corresponding gossip identity account for each
```

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,219 @@
...
+validator in the accepted list from the previous step
+
+6. Record the VAT fee subtraction in the bank, which reduces the bank’s
+capitalization. This occurs before any transactions are processed in the new
+epoch’s bank
```
t-nelson: the second part of this section relates to timing and should be earlier

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,219 @@
...
+
+7. The epoch stakes for epoch bank.epoch() + 1 are generated from the list
+produced in step 4. These epoch stakes are saved in bank snapshots, so if the
+network restarts from a snapshot taken from any bank within the same epoch as
+bank.epoch(), the 1.6 SOL fee will not be charged again
```
t-nelson: i don't understand how this is relevant. the vat deduction all happens in a single slot (baring forks), which is the same slot we distribute inflation reward commissions. how would we run this logic again?

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,219 @@
...
+## Operation Considerations
+
+- To be included in epoch e+1, validator operators must ensure the vote
+identity account has at least 1.6 SOL before epoch e-1 ends
+
+- Validator operators must ensure they have valid BLS public key specified in
+their vote account
+
+  - The ledger tools will be updated to verify validity of the BLS public key
+when performing updates
```
t-nelson: what does this mean? accounts aren't in the ledger. what updates?

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,219 @@
...
+
+  - We will not allow removing the BLS public key
+
+  - The written BLS public key must be generated by the vote authority ed25519
+keypair and change with the vote authority keypair, the keygen tools and
+ledger tools will be updated
```
t-nelson: these are relevant to whatever simd specifies the vote program instruction for setting the bls key, not this one

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,219 @@
...
+we also need to return the fee collected.
+
+## Impact
+
+Validators not providing BLS public key or desired fee will not be able to
+participate in an Alpenglow epoch regardless of their stake. Also, only the
+nodes selected by this process will receive votes or certificates from other
+selected validators in real time.
+
+Validator operators need to ensure they have enough fund and correct BLS
```
t-nelson:
```suggestion
Validator operators need to ensure they have enough funds and correct BLS
```

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,219 @@
...
+public key before end of epoch e-1 to participate in epoch e+1. This poses some
+new operation challenges.
```
t-nelson: the bls stuff can be tidied up after alpenglow activation with a new initialize instruction and deprecating the old one. then the we are operationally identical to today which is just "keep your account topped up"

t-nelson (commented):
i'd urge a reconsideration of moving vat to the vote account. it has many benefits

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Yup, I think we should have the new better approach long before we hit more validators.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Removed.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Removed.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Removed.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: I'm fine with either one, but chose identity account mostly because that's where the current vote fees come from, and where rewards are paid, so it's not disruptive to current practice.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Yup, do you think I should write that explicitly into the SIMD?

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: I guess I can move this to SIMD 387, removed.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Sorry I'm confused, what do you mean here?

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Also, are you implying we should send rewards to vote account instead in Alpenglow?

wen-coding: Wait, I think @tigarcia explained to me a lot of people don't actually have the private key to their vote accounts any more, because we didn't really have any operations requiring that.

Are we implying we need everyone to recreate vote accounts?

proposals/0357-alpenglow_validator_admission_ticket.md
t-nelson: > This occurs before any transactions are processed in the new
epoch’s bank

this is important timing information for the implementer. this operation cannot be done after or in parallel to, tx execution

proposals/0357-alpenglow_validator_admission_ticket.md
t-nelson: i think i'll have the same questions no matter which simd it's in :sweat_smile:

proposals/0357-alpenglow_validator_admission_ticket.md
t-nelson: no. they are only relevant to the vote program itself

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: https://github.com/solana-foundation/solana-improvement-documents/pull/387 See if wrote this better over there? I originally thought when doing vote authority change we can verify locally the BLS proof of possession makes sense. But Maybe we should not do that and rely on the transaction succeeding in vote program instead.

I planned to ask you to review that SIMD after it gets more feedback

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Got it, moving to bullet 2. Or should I specify it at top of this section?

Actually I think the bank's capitalization changes happen when the bank is frozen? (There might be other burned fees.) But the reduction of 1.6 SOL definitely happens before any user transactions. Updated.

t-nelson:
i think there was a bad rebase here? several other simds are showing modifications by this pr now

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: I'm afraid there are some vote accounts which are staked but the users never bothered adding the BLS public key, they don't work in Alpenglow and I need to remove them from staked validators.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: I assume you mean the first part of this section says the same thing as the second part, removed.

wen-coding:
> i think there was a bad rebase here? several other simds are showing modifications by this pr now

Oops, sorry about that, rebased.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Okay, scratch my false statement above, I think choosing vote account is fine. The only caveat is we need to move block rewards and prio fee destination from identity account to vote account, then people don't need to constantly top up their vote accounts for VAT fees.

t-nelson: > we need to move block rewards and prio fee destination from identity account to vote account

yes but keep that out of this simd. it's a separate optimization and doesn't need to muddy vat

wen-coding: Okay, I changed this SIMD to deduct VAT from vote account, we can put rewards into vote account in another SIMD.

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,196 @@
...
+## Different Vote Related Accounts
+
+Before Alpenglow, we have the following accounts related to voting (the
+identity account and vote authority account can be the same):
+
+- Vote account for saving all the vote states and receiving commission
+
+- Identity account for receiving block rewards and priority fees
+
+- Vote authority account for signing vote transactions
```
buffalojoec: I think it would be valuable here to mention who currently pays the voting fee, since the post-Alpenglow list below specifies who pays the VAT.

buffalojoec: Vote authority account or just keypair? Afaik this doesn't actually need to exist on-chain (no balance, or anything else). Same with post-Alpenglow.

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,196 @@
...
+voting validator 1.6 SOL per epoch, it replaces the current voting fee at ~2
+SOL per epoch, and reduces the likelihood there are too many voting validators
+immediately after Alpenglow launches.
+
+The general VAT concept has already been accepted with the governance vote on
+SIMD 326.
+
+Everything specified below are protocol-level changes, they need to be
+implemented by all Solana clients. The client specific data structure
+changes are omitted.
```
buffalojoec, line 32: Are you talking about runtime-level data structures or vote program account changes? I feel like on-chain Vote account layout should be specified in this proposal, even if we just say something like "Vote State v4 with an empty `votes` list".

buffalojoec: Later in the proposal, you mention that vote accounts "doesn’t contain vote information any more". That somewhat vague statement made me wonder if we should be a little more explicit about vote state. Lmk what you think!

proposals/0357-alpenglow_validator_admission_ticket.md
buffalojoec: > no "invalid" vote account should be allowed to exist

Yeah I'm also not really sure how we'd pull this off with the v3 -> v4 vote state migration. The program does the migration for you and sets a bunch of defaults for new fields, like the compressed BLS pubkey.

Making the migration manual and requiring a BLS pubkey & proof be provided is kind of the same thing as the flow proposed in [SIMD-0387](https://github.com/solana-foundation/solana-improvement-documents/pull/387).

Seems like this filter is necessary for now.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Changed

wen-coding: That's true, removed it since it's not too relevant in SIMD 357, it's more relevant in SIMD 387.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: I previously wrote something like that, people complained I should keep the scope of this SIMD smaller. So I think I'll just talk about deducting VAT behavior here. Changed "doesn't contain vote information any more" to "the `votes` list in vote state will be empty". Hopefully this is clearer.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Yeah, unfortunately for security reasons, no one other than the users can set the BLS pubkey here. This is not something we can do automatically for every user during v3->v4 migration. So I think our best path forward is:
1. complete SIMD 387 and add the tool so users can add BLS pubkey
2. tirelessly poke users so they add BLS pubkey
3. filter out any account without BLS pubkey when we really launch Alpenglow

proposals/0357-alpenglow_validator_admission_ticket.md
buffalojoec: Yeah okay that works for me, as long as it's the only notable vote state change, in which case it's not really a state change only a usage change.

proposals/0357-alpenglow_validator_admission_ticket.md
t-nelson: i'm not sure how we're missing each other on this point. surely either the bls key is an option or migrate sets it to a sentinel value. my point was that we only need to check the decompression when user sets the key. we can do a cheaper test here

wen-coding: Ah I see your point now, okay, I'm changing it to only check existence of a BLS pubkey.

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,191 @@
...
+purpose:
+
+- Vote account: This account must contain the correct BLS public key
+corresponding to vote authority keypair. It continues to keep all the vote
+credits and all validator identity/authority/commission information updates
+happen here, but the `votes` list in vote state will be empty. This will be
+the account where the 1.6 SOL VAT is deducted from.
+
+- Identity account: It continues receiving block rewards and priority fees.
+But we may change that in a separate SIMD.
```
t-nelson:
```suggestion
```

i think this distracts from the topic at hand

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,191 @@
...
+
+- When there are more than 2,000 valid validators, sort all valid vote accounts
+by descending order of stake. If some validator with stake S is in position
+2001, then we remove all validators with stake S and less. If there are fewer
+than 2,000 valid validators, pick all of them.
+
+- Deduct 1.6 SOL VAT fee from each picked vote account once
+
+- Mark the fee burned and write the result into the bank
+
```
t-nelson: these points seem unrelated to the heading?

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,191 @@
...
+  - The account has a BLS public key
+
+4. If the number of filtered accounts exceeds 2,000, sort them according to the
+following rules and select the top 2,000. Otherwise, return the entire list:
+
+  - Sort in descending order of stake (largest to smallest)
+
+  - If multiple validators have exactly the same amount of stake and including
+  all of them would exceed the 2,000 limit, then exclude all of them
```
t-nelson:
```suggestion
  - Do not tiebreak. If total valid vote accounts exceed the limit of 2000,
  omit all accounts whose stake is equal to that of the 2001st account
```

i think this is more clear?

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,191 @@
...
+
+5. Subtract 1.6 SOL from the vote account for each validator in the accepted
+list from the previous step
+
+6. Record the VAT fee subtraction in the bank, which reduces the bank’s
+capitalization. The recording of fee subtraction occurs when the bank is
+frozen even though fee reduction happens before any transaction in the
+bank is executed.
```
t-nelson: these steps aren't particularly precise. we should probably just say to move the lamports directly into the incinerator account?

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Operation Considerations
+
+- To be included in epoch e+1, validator operators must ensure the vote
+account has at least 1.6 SOL before epoch e-1 ends
+
+- Validator operators must ensure they have BLS public key specified in
+their vote account
+
```
t-nelson: this is apparent from (3) in the previous section?

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,191 @@
...
+- We will not allow removing the BLS public key
+
```
t-nelson: not relevant to this simd. hopefully it is already in the vote account one

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,191 @@
...
+Specifically, if the last slots of epoch e-1 are not finalized at the end
+of the epoch, we have enough slots in epoch e to finalize epoch e-1's state.
+
+## Security Considerations
+
+Keeping funds in vote accounts for VAT is safer than keeping funds in identity
+accounts. Because the keypair for identity account needs to be in a hot wallet
+to support real-time validator operations, while the vote authorized withdrawer
+keypair does not need to be in a hot wallet.
+
```
t-nelson, line 145:
```suggestion

In furtherance of improving operational security, a future simd may change
the deposit location of validator rewards to the vote account. This would
eliminate the necessity for any funds to be controlled by a hot keypair.
```

we can mention moving the reward deposit location here

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,191 @@
...
+Validator operators need to ensure they have enough funds and correct BLS
+public key before end of epoch e-1 to participate in epoch e+1. This poses some
+new operation challenges.
+
+We are requiring correct setup of vote accounts at end of epoch e-1 for
+inclusion in epoch e+1 because of the following reason: We want to be
+absolutely sure that we have a finalized state before we enter epoch e+1.
+Specifically, if the last slots of epoch e-1 are not finalized at the end
+of the epoch, we have enough slots in epoch e to finalize epoch e-1's state.
+
```
t-nelson: we're beating a dead horse with this "valid vote account" thing

the material impact is that the voting set is now strictly capped at 2000. we can also mention that operators must now keep their vote account topped up instead of their autorized-voter/gossip-identity

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,191 @@
...
+- at least 1.6 SOL VAT fee plus the necessary storage rent amount for a new
+epoch in the vote account
+
+When the staked validators for a new epoch are calculated, all validators must
+perform the following operations:
+
+- When there are more than 2,000 valid validators, sort all valid vote accounts
+by descending order of stake. If some validator with stake S is in position
+2001, then we remove all validators with stake S and less. If there are fewer
+than 2,000 valid validators, pick all of them.
```
alexpyattaev: It remains ambiguous what happens when there are exactly 2000.

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,191 @@
...
+
+- Deduct 1.6 SOL VAT fee from each picked vote account once
```
alexpyattaev: "Once" is not needed here

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,191 @@
...
+
+3. The calculation iterates all vote accounts and filters those that meet
+the following criteria:
+
+  - The account has a balance of at least 1.6 SOL
+
+  - The account has a BLS public key
+
+4. If the number of filtered accounts exceeds 2,000, sort them according to the
+following rules and select the top 2,000. Otherwise, return the entire list:
```
alexpyattaev, line 103: This is literally a dup of text above.

alexpyattaev (commented):
Suggesting some edits

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Sorry, which one are you talking about? "Mark the fee burned and write the result into the bank", or everything above?

t-nelson: apparently github doesn't show the comment context on the files view anymore...

Screenshot from 2025-11-26 13-13-43

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Hmm, I thought "When there are more than 2,000 valid validators" implies that when there are exactly 2,000 we skip this step, do you have suggestions how to make that even more clear?

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: I wanted to assure the users we will only deduct 1.6 SOL once, but hmm I guess it's optional.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Hmm, I'm confused, what you are highlighting is the exact VAT implementation every validator should perform, I thought SIMD is specifically used to communicate these things so we don't end up with a consensus breaking mismatch?

proposals/0357-alpenglow_validator_admission_ticket.md
alexpyattaev: Translating the current text to rust:
```rust
if (N > 2000){
     validators.sort();
     validators.truncate(2000);
     return validators;
}
else if N < 2000 {
    return validators;
}
else {
    panic!("This will happen when N == 2000");
}
```

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Oh, do you mean I should remove the text here and just leaves the "How to implement the checks"?

proposals/0357-alpenglow_validator_admission_ticket.md
alexpyattaev: Once would imply we never do it again, would it not?

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Removed.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: OK, guess we can remove this section.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Done.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Changed.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Removed this section.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Okay, removed the above section.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Err, first of all, we won't panic on exactly 2000. Also I believe the algorithm should be pseudo-code level instead of relying on Rust. And it's hard to describe exactly what we are doing using minimum rust code.

Would it help if I say "When the number of validators > 2000"?

alexpyattaev: I'd go with something like this:
```
When there are more than 2,000 valid vote accounts, sort all valid vote accounts
by descending order of stake, keep the top 2000 and remove the rest. Otherwise use all valid vote accounts.
```
Note the change from valid validators to valid vote accounts (since we define valid vote account but not valid validator).

wen-coding: Err, I removed this section already. Sorry. See if the "How to implement the checks" section is clear enough for you.

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Intended to say once per Epoch, but I guess it's obvious.

proposals/0357-alpenglow_validator_admission_ticket.md
alexpyattaev: Yeah looks great thank you!

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,152 @@
...
+## Summary
+
+This SIMD describes how the validator admission ticket (VAT) collection
+described in [SIMD 326](https://github.com/solana-foundation/solana-improvement-documents/pull/326)
+will be implemented. Specifically, how it affects validator operation
+procedures.
+
+The validator admission ticket is a mechanism translating the current cost of
+voting into a similar economic equilibrium for Alpenglow. By charging every
+voting validator 1.6 SOL per epoch, it replaces the current voting fee at ~2
```
t-nelson, line 23: nit: we repeat 1.6 SOL throughout the document. replace with simply VAT other than this first instance

proposals/0357-alpenglow_validator_admission_ticket.md
```diff
@@ -0,0 +1,152 @@
...
+
+  - Do not tiebreak. If total valid vote accounts exceed the limit of 2000,
+  omit all accounts whose stake is equal to that of the 2001st account
+
+5. Subtract 1.6 SOL from the vote account for each validator in the accepted
+list from the previous step
+
+6. Record the VAT fee subtraction in the bank, move the lamports directly into
+the incinerator account.
+
```
t-nelson, line 115: not a change for the simd, but wherever we're tracking the agave implementation, we need to emit an appropriate [`RewardInfo`](https://github.com/anza-xyz/solana-sdk/blob/d893d59066773b0bb7b3fde79ae761e8d6f0af9f/reward-info/src/lib.rs#L37) entry for each of these debits in the corresponding block. it is unacceptable to lose this accounting information

wen-coding: Hmm, I did a quick search for `RewardInfo` in Agave, I don't see a case where the reward is negative. Would it break someone's tool if we suddenly have a negative number in `RewardInfo`? Should we add it to some other entries?

proposals/0357-alpenglow_validator_admission_ticket.md
wen-coding: Sure, I kept the first one and the one in `New Terminology`, removed all other ones.

proposals/0357-alpenglow_validator_admission_ticket.md
t-nelson: the `RewardInfo::lamports` field was converted from `u64` to `i64` specifically to record rent debits (gh indexing of old labs repo is busted, so i can't find it). likely this has been removed since rent collection was disabled

wen-coding: Okay, in that case I'll just add it to this SIMD so we don't forget to do it. Changed.

t-nelson: implementation details do not belong in simds. this is not an agave todo list

wen-coding: Okay, removed. Just for my education, which team is supposed to do this?

lidatong (approved):

proposals/0357-alpenglow_validator_admission_ticket.md
t-nelson: the field is not under consensus. it's only metadata

wen-coding: Sounds good, any more comments on this SIMD?

PR #358, opened 2025-09-18
SIMD-0358: Minimal ELF Binary Encoding

deanmlittle:


proposals/0358-minimal-elf-binary-encoding.md
```diff
@@ -0,0 +1,177 @@
...
+### ELF Header Requirements
+
+| Field        | Value / Rule                                          |
+| ------------ | ----------------------------------------------------- |
+| `e_type`     | `ET_EXEC (0x0002)`                                    |
+| `e_machine`  | `BPF (0xF7)`                                          |
+| `e_entry`    | **start of `.text`**, directly preceded by `.rodata`  |
+| `e_phnum`    | `0`                                                   |
+| `e_shnum`    | `0`                                                   |
+| `e_shstrndx` | `0`                                                   |
```
ripatel-fd, line 72: - This breaks basic debug tooling like obj2yaml, readelf, and elftools. This is hardly an ELF file. Might as well use a custom container format then.
- So if we use ELF, then we should at least support the most basic features like program headers (3 program headers is really not a lot of overhead, the section table would stay optional)

proposals/0358-minimal-elf-binary-encoding.md
```diff
@@ -0,0 +1,177 @@
...
+
+> Note: In an sBPF v0 program, the equivalent artifact would be more than 
+> double this size due to additional ELF scaffolding.
+
+## Backwards Compatibility
+
+* Existing sBPF v0 binaries will remain valid moving forwards; this SIMD 
+introduces an **additional** accepted format.
+* In the runtime, minimally encoded binaries is a breaking change and must be 
+enabled by **feature gate** activation during rollout.
```
ripatel-fd, line 155: IMO, not worth a breaking change to save 3 program headers (96 bytes), compared to Alex sBPF v0 ELF proposal (which is still compatible with debug tooling like readelf).
Once people have built a workflow and security tooling around sBPF v0, it's just going to break immediately again once this proposal gets accepted.

proposals/0358-minimal-elf-binary-encoding.md
```diff
@@ -0,0 +1,177 @@
...
+Solana validators execute programs inside the rBPF/eBPF VM. Today’s artifacts 
+often include superfluous ELF structures (program headers, section headers) 
+that are primarily useful for debugging, but not needed by the runtime. 
+Removing them enables:
+
+* **Smaller binaries** (measurably reducing on-chain and network overhead).
+* **Deterministic layout** that simplifies the VM's memory scaffolding.
+* **Simpler, faster JIT** (no dynamic reloc ations, fewer failure modes).
+* **Safer memory access** (rigidly enforce `R^X` on `.text`, `R` on `.rodata`)
+* **Consistent data alignment** (all data segments are 8-byte aligned)
```
ripatel-fd, line 34: https://github.com/solana-foundation/solana-improvement-documents/pull/189

SIMD-0189 has all the same benefits. The parsing in SIMD-0189 is more straightforward: The segments are laid out in the ELF file itself, whereas this proposal derives them from e_entry and the file size.

proposals/0358-minimal-elf-binary-encoding.md
```diff
@@ -0,0 +1,177 @@
...
+
+The VM assumes contiguous placement in the above order.
+
+### ELF Header Requirements
+
+| Field        | Value / Rule                                          |
+| ------------ | ----------------------------------------------------- |
+| `e_type`     | `ET_EXEC (0x0002)`                                    |
+| `e_machine`  | `BPF (0xF7)`                                          |
+| `e_entry`    | **start of `.text`**, directly preceded by `.rodata`  |
```
Lichtso, line 69: Unfortunately the [ELF spec](https://refspecs.linuxfoundation.org/elf/gabi4+/ch4.eheader.html) encodes this as a virtual address, not a file offset:

> e_entry
>     This member gives the virtual address to which the system first transfers control, thus starting the process. If the file has no associated entry point, this member holds zero.

Thus standard tooling might require a program header with an identity mapping. Though I haven't tired it out.

proposals/0358-minimal-elf-binary-encoding.md
```diff
@@ -0,0 +1,177 @@
...
+| `e_phnum`    | `0`                                                   |
+| `e_shnum`    | `0`                                                   |
+| `e_shstrndx` | `0`                                                   |
+
+All remaining ELF header fields are ignored by the loader unless otherwise 
+specified by existing Solana runtime rules, leaving room for us to reintroduce
+program headers in the future should the need arise.
+
```
Lichtso, line 77: The alignment requirements should be mentioned here too, not just in the motivation. Also, not just the beginning, but also the length (of the bytecode) needs to be 8 byte aligned. Thus we also need to check that the file length is a multiple of 8 bytes as the bytecode is always at the end in this design.

proposals/0358-minimal-elf-binary-encoding.md
```diff
@@ -0,0 +1,177 @@
...
+header tables or violates the fixed order.
+
+### Runtime Behavior
+
+When executing a minimally encoded binary, the runtime must:
+
+1. **Disable relocation/JIT relocation** logic (there are no relocations in
+this format).
+2. **Resolve static call targets**
+3. **Adhere to expected memory layout** (`.rodata` is mapped read-only, `.text`
```
Lichtso, line 95: So far this only describes the file layout, we need to specify the virtual address space layout too. I suggest that we copy the one specified in [SIMD-0189](https://github.com/solana-foundation/solana-improvement-documents/pull/189) (virtual address layout only, not file layout) and have the read-only region before the bytecode region.

This not only makes it a bit more intuitive as that would be the same order as the file layout but also means that references to the read-only region can be expressed in 32 bits and no longer require 64 bit immediate loads. Doing so can save quite a bit in filesize (and CUs too if we count `lddw` as two instructions) simply because `lddw` won't be required to access the read-only data anymore.

proposals/0358-minimal-elf-binary-encoding.md
```diff
@@ -0,0 +1,177 @@
...
+introduces an **additional** accepted format.
+* In the runtime, minimally encoded binaries is a breaking change and must be 
+enabled by **feature gate** activation during rollout.
+* Tooling (SDKs/build scripts) should add a “minimal” output mode that emits 
+this layout while continuing to support current formats.
+
+## Alternatives Considered
+
+* **Keep current ELF with headers present but ignored.** Reduces diff risk but 
+fails to realize size and simplicity gains.
```
Lichtso, line 162: In [SIMD-0189](https://github.com/solana-foundation/solana-improvement-documents/pull/189) we also started with a hard-coded number but then switched to a minimum because multiple teams approached us and said that they would like to insert arbitrary sections (unrelated to debug info IIRC).

Supporting that would require skipping over these which only costs us an additional multiplication and addition on loading. Allowing for more does still allow them to not exist, thus does not cost a single byte if unused.

proposals/0358-minimal-elf-binary-encoding.md
```diff
@@ -0,0 +1,177 @@
...
+*No other sections or tables are present.* In particular:
+
+* **No Program Header Table** (`e_phnum = 0`)
+* **No Section Header Table** (`e_shnum = 0`, `e_shstrndx = 0`)
+
+The VM assumes contiguous placement in the above order.
+
+### ELF Header Requirements
+
+| Field        | Value / Rule                                          |
```
Lichtso, line 65: Is there a reason not to restrict the other header fields (like the ELF magic) too like [SIMD-0189](https://github.com/solana-foundation/solana-improvement-documents/pull/189/files) does? That can be done efficiently with 8 byte or even 16 byte comparisons (as the validator will be able to load the ELF host aligned from loader-v4 onward).

proposals/0358-minimal-elf-binary-encoding.md
```diff
@@ -0,0 +1,177 @@
...
+ecosystem and further complicates tooling compatability.
+
+## Impact
+
+Minimal encoding reduces program binary size and complexity, resulting in 
+cheaper program deployments, simpler tooling, fewer failure modes and more 
+efficient execution. These benefits carry on to building and maintaining
+custom linker tooling required to enable future upstream eBPF compatibility
+(in fact, in our above example, they are already compatible!)
+
```
Lichtso, line 173: Another thing that should be mentioned somewhere is that this design requires the linker to always place the entrypoint function first so that it is at the start of the bytecode.

proposals/0358-minimal-elf-binary-encoding.md
```diff
@@ -0,0 +1,177 @@
...
+## Motivation
+
+Solana validators execute programs inside the rBPF/eBPF VM. Today’s artifacts 
+often include superfluous ELF structures (program headers, section headers) 
+that are primarily useful for debugging, but not needed by the runtime. 
+Removing them enables:
+
+* **Smaller binaries** (measurably reducing on-chain and network overhead).
+* **Deterministic layout** that simplifies the VM's memory scaffolding.
+* **Simpler, faster JIT** (no dynamic reloc ations, fewer failure modes).
```
Lichtso, line 32: Dynamic relocation happens during ELF loading, before JIT compilation as interpretation also requires relocations to be applied.

proposals/0358-minimal-elf-binary-encoding.md
deanmlittle: This is not true for the local version of the binary that is yet to be linked to a minimal encoding, and due the predictable layout of the binary, it is also completely reversible for the deployed version. That information is simply not needed by the loader or the runtime at all. The current "3 headers" in V0 is also not a requirement. You can have one program header, or even zero program headers. It also allows you to place rodata in the text section.

proposals/0358-minimal-elf-binary-encoding.md
ripatel-fd: > In https://github.com/solana-foundation/solana-improvement-documents/pull/189 we also started with a hard-coded number but then switched to a minimum because multiple teams approached us 

@Lichtso Wait, but teams could still inject section headers that are just ignored. Things like Neodyme's security.txt would have a section but not a program header. Therefore, suggest reverting SIMD-0189 back to a fixed program header count

Lichtso: you mean only the sections should be variable in count, but the program headers fixed?

proposals/0358-minimal-elf-binary-encoding.md
deanmlittle: Yes, we should perform the existing checks on the ELF header. This is a *diff* in the checks that we perform compared to V0, with the exception of the machine ID which is mentioned for posterity as we used our own machine ID for a period of time.

proposals/0358-minimal-elf-binary-encoding.md
deanmlittle: In reality, every binary that isn't custom-linked with multiple text sections will simply already be pointing to the exact right file offset which is superfluously redefined in the program section headers.

tlambertz:
I do see value in reducing elf parsing complexity, getting rid of extra sections and reloc's, that definitely was a source of bugs before. [SIMD-189](https://github.com/solana-foundation/solana-improvement-documents/pull/189) tries to do something similar.  What is the advantage of this proposal over that one?

I have not verified, but suspect that the removal of program/section header tables would break most elf parsing tools. If that is the case, we should not be calling it an "elf" anymore, could get rid of even more header, and just roll a fully custom format.

But there definitely is some value in being able to just pull a binary from chain and parse it as-is, we have done this a number of times in the past, for example to look at program similarity (ie we know that one program has a bug, and want  to find forks/similar programs).

Last I checked there were about 13k ELFs deployed on mainnet, so a couple hundred bytes per elf doesn't amount to much, like 1MB per 100 bytes saved. Putting very clear restrictions on what the elf header is allowed to look like, as done in SIMD-189, seems the cleaner solution to me. I don't like fully ignoring the header, but it could be checked just once on deployment if it's sufficiently static. But all off-chain tooling could just use a default elf parser.

@ripatel-fd mentioned security.txt above. It adds [another section](https://github.com/neodyme-labs/solana-security-txt/blob/a68ced075fce8f10498cc1c7f0d1195e11750b3c/security-txt/src/lib.rs#L17), including the security-txt metadata, which wouldn't be compatible with any of the proposed changes. This isn't too big of a problem though, as most (all?) parsing tools do needle-in-a-haystack search anyways. Parsing elfs turns out to be quite annoying, especially in browser/javascript where lots of UI tools are running. (eg explorer.solana.com here: https://github.com/solana-foundation/explorer/blob/a704d9b8248de4d398f9faf421c69bfda496a51d/app/utils/security-txt.ts#L42)

Right now, the security.txt crate still creates a separate link_section, but we could easily include the same info as a static string anywhere in the binary instead. That is technically breaking, but I don't know of any consumers of this header. When rolling a custom binary format, we could also think about including info such as name/website/source-link/... as optional header fields.

proposals/0358-minimal-elf-binary-encoding.md
Lichtso: Don't think that is how it works. If you use a linker script and e.g. LLD it would place the entry point at the virtual address `0x100000000` and thus encode the file header that way too, right?

deanmlittle: As I said, I cannot speak for what contrived examples could be created by custom linking, but from what I have seen, by default, none of our tooling does this. There is one .text section entry, the entrypoint is always at the start of the .text section entry, this is what is pointed to by `e_entry` in the ELF Header, as well as the `entrypoint` entry in the section headers.

deanmlittle:
> I do see value in reducing elf parsing complexity, getting rid of extra sections and reloc's, that definitely was a source of bugs before. [SIMD-189](https://github.com/solana-foundation/solana-improvement-documents/pull/189) tries to do something similar. What is the advantage of this proposal over that one?

189 increases the cost to deploy programs which matters a lot for high performance, single task stuff like oracle updates. By contract, 358 lowers the cost of program deploys and removes additional offset tracking that needs to be handled by the linker.

> I have not verified, but suspect that the removal of program/section header tables would break most elf parsing tools. If that is the case, we should not be calling it an "elf" anymore, could get rid of even more header, and just roll a fully custom format.

Abandoning ELF is listed as one of the possible alternatives. Historically, it has not been a popular idea. Strictly speaking, the ELF format does not require program/section headers at all. This is strictly a requirement of the loader executing them, and ours, for whatever reason, already doesn't care about program headers unless your binary contains text-based relocations.

That said, even if this is the case, if you are the developer of said program, you have the unstripped version locally for debugging purposes, and if you are getting the program from the chain, due to the rigid layout of the binary proposed, it is entirely possible to just relink the binary with headers using the same tooling that minimally encodes it in the first place. This could even be built into `solana program dump` if it became a major devex hassle.

> Right now, the security.txt crate still creates a separate link_section, but we could easily include the same info as a static string anywhere in the binary instead. That is technically breaking, but I don't know of any consumers of this header. When rolling a custom binary format, we could also think about including info such as name/website/source-link/... as optional header fields.

We should not be putting arbitrary data like this in binaries. Loading this into memory every time we execute the program does nothing to benefit performance. You can create metadata specific programs, like a `SecurityTxtProgram` that enables the upgrade authority of a program to place their security.txt into a PDA derived from the program address that can be displayed by block explorers. The same thing is done with IDLs. It's a much better pattern to follow. 

> I do see value in reducing elf parsing complexity, getting rid of extra sections and reloc's, that definitely was a source of bugs before. [SIMD-189](https://github.com/solana-foundation/solana-improvement-documents/pull/189) tries to do something similar. What is the advantage of this proposal over that one?

Until 1 hour ago, it also proposed to add two additional unnecessary program headers which is even more bloat. I also don't think it makes sense to enforce that programs that use no `.rodata` have a `readable` program header. Right now in a v0 program, if I don't invoke a syscall, I can deploy a binary with zero program headers.

jacobcreech:
> @ripatel-fd mentioned security.txt above. It adds [another section](https://github.com/neodyme-labs/solana-security-txt/blob/a68ced075fce8f10498cc1c7f0d1195e11750b3c/security-txt/src/lib.rs#L17), including the security-txt metadata, which wouldn't be compatible with any of the proposed changes. This isn't too big of a problem though, as most (all?) parsing tools do needle-in-a-haystack search anyways. Parsing elfs turns out to be quite annoying, especially in browser/javascript where lots of UI tools are running. (eg explorer.solana.com here: https://github.com/solana-foundation/explorer/blob/a704d9b8248de4d398f9faf421c69bfda496a51d/app/utils/security-txt.ts#L42)

> Right now, the security.txt crate still creates a separate link_section, but we could easily include the same info as a static string anywhere in the binary instead. That is technically breaking, but I don't know of any consumers of this header. When rolling a custom binary format, we could also think about including info such as name/website/source-link/... as optional header fields.

We plan on recommending devs to not place their security.txt in the header anymore once we have [program metadata program tooling](https://github.com/solana-program/program-metadata) ready for wider adoption. This may be a different discussion, but I don't think security.txt should be a blocker for anything here.

proposals/0358-minimal-elf-binary-encoding.md
deanmlittle: 64-bit BPF program headers are actually 56 bytes in length, so 168 bytes in total, and if adding headers back is required for tooling compatibility, it's actually very easy to just relink with headers using the same toolchain that links it without them in the first place.

deanmlittle: Unless the idea was to use 32-bit program headers on a 64-bit binary? That sounds far more cursed than what I'm proposing.

tlambertz:
Thanks! I might have to build myself such a minimal elf and just go test. Do you have any such binary ready to go?

Regards security.txt, agreed! Thats a much better solution. Back then, we wanted to give an easy way for *all* programs to have such metadata, and there are some quite weird deploy mechanism, like wormhole, which is it's own upgrade authority. Tying a key to metadata updates in that case will require program code changes to support that.

deanmlittle:
> Thanks! I might have to build myself such a minimal elf and just go test. Do you have any such binary ready to go?
> 
> Regards security.txt, agreed! Thats a much better solution. Back then, we wanted to give an easy way for _all_ programs to have such metadata, and there are some quite weird deploy mechanism, like wormhole, which is it's own upgrade authority. Tying a key to metadata updates in that case will require program code changes to support that.

Blueshift is working on a minimal linker for upstream bpf support that will be compatible with current v0 binaries provided no unsupported insns are used. It would be trivial to also handle for this much simpler binary layout. I will link it as soon as we open source something. We can alos trivially reverse the binary in a way that would make it look similar to a SIMD-189 binary. I think there is a reasonable argument to be made here that we get everything SIMD-189 gets us for free, there are fewer failure modes, nothing will break, program deploys get cheaper and we remove a lot of question marks.

ripatel-fd:
> I think there is a reasonable argument to be made here that we get everything SIMD-189 gets us for free, there are fewer failure modes, nothing will break, program deploys get cheaper and we remove a lot of question marks.

@deanmlittle Can you list out the failure modes and question marks in SIMD-0189 that this spec doesn't have? As far as I understand, SIMD-0358 implies memory protection instructions, and SIMD-0189 explicitly lays them out via program headers. SIMD-0358 forces some fields to be zero that SIMD-0189 ignores. Is there any other difference?

proposals/0358-minimal-elf-binary-encoding.md
```diff
@@ -0,0 +1,177 @@
...
+complexity, simplifying future upstream BPF compatibility.
+
+## Motivation
+
+Solana validators execute programs inside the rBPF/eBPF VM. Today’s artifacts 
+often include superfluous ELF structures (program headers, section headers) 
+that are primarily useful for debugging, but not needed by the runtime. 
+Removing them enables:
+
+* **Smaller binaries** (measurably reducing on-chain and network overhead).
```
LucasSte, line 30: #189 also brings smaller binaries, because it removes the restriction of `sh_offset` to be equal to `sh_addr`. As a consequence, we don't need to pad file sections so that their virtual address matches their file offset.

proposals/0358-minimal-elf-binary-encoding.md
```diff
@@ -0,0 +1,177 @@
...
+* **No Program Header Table** (`e_phnum = 0`)
+* **No Section Header Table** (`e_shnum = 0`, `e_shstrndx = 0`)
+
+The VM assumes contiguous placement in the above order.
+
+### ELF Header Requirements
+
+| Field        | Value / Rule                                          |
+| ------------ | ----------------------------------------------------- |
+| `e_type`     | `ET_EXEC (0x0002)`                                    |
```
LucasSte, line 67: Changing to executable also means that we would need to adapt existing Rust code, since right now we write contracts as if they were a lib. Emitting an executable also implies either using a main function, or tweaking your code to use `#[no_main]`. Having said that, I believe we could maintain compatibility with existing programs, requiring only a recompilation for having the new headers if we stick to `ET_DYN`.

proposals/0358-minimal-elf-binary-encoding.md
LucasSte: @Lichtso is right. `e_entry` points to the virtual address of the entry point, not the file offset. If we want to place the text section at address `0x100000000`, and the entrypoint function is the first one, `e_entry` will contain `0x100000000`.

We do need to specify a virtual address for sections so that relocations can be resolved. For the text sections, we allow the `callx` instructions, which means at some point we would need to load the right address to the register to be called, so the relocation must resolve it correctly, being the function address stored in the rodata section or in the text section.

proposals/0358-minimal-elf-binary-encoding.md
LucasSte: That might be true for the unlinked binary, yes, but as a consequence of that, we cannot have tools that surface to users the code coverage, since we want to execute a program and map the executed instructions to the code lines.

There is a discussion happening about this here: https://github.com/anza-xyz/agave/discussions/7709

Lichtso:
> We do need to specify a virtual address for sections so that relocations can be resolved. For the text sections, we allow the callx instructions, which means at some point we would need to load the right address to the register to be called, so the relocation must resolve it correctly, being the function address stored in the rodata section or in the text section.

Section headers and relocations are only necessary during linking right? For the on-chain upload / deployment they don't matter and this SIMD is only about what gets uploaded, not the immediate products. Program headers might be necessary for all tooling I suspect. Otherwise you can't map `e_entry`. The program-runtime could ignore that, though at that point it becomes the question why have an ELF header at all if it is not readable by any tooling.

LucasSte:
> Section headers and relocations are only necessary during linking right? For the on-chain upload / deployment they don't matter and this SIMD is only about what gets uploaded, not the immediate products. 

Yes, we only need the virtual addresses during link time so that we can resolve relocations, and possibly map the `e_entry` field.

> Program headers might be necessary for all tooling I suspect. Otherwise you can't map `e_entry`. The program-runtime could ignore that, though at that point it becomes the question why have an ELF header at all if it is not readable by any tooling.

If we repurpose `e_entry`, won't we be leaving the ELF format behind?

cyanpencil:
> Abandoning ELF is listed as one of the possible alternatives. Historically, it has not been a popular idea.

I'm working on tooling to statically analyze programs and I had to write a custom ELF parser anyways because any normal tool (except `llvm-readelf` and so on from the solana toolchain) would very slightly be incorrect (due to small idiosyncrasies like the relocations). 
Writing a parser for a much simpler but custom format would be less work, at this point. If optimizing performance for oracle programs is a top priority, a simple and optimized custom format would be better both for CUs and for tool devs at this point. 

However, those are only my 2 cents from someone new in the field, it is just my opinion.

I also like what @deanmlittle proposed:
>  This could even be built into solana program dump if it became a major devex hassle.

You could absolutely make `solana program dump` give you a nice, Linux-friendly ELF that can be loaded in any tool, and make everyone's day instantly a bit better :)

LucasSte:
> I'm working on tooling to statically analyze programs and I had to write a custom ELF parser anyways because any normal tool (except `llvm-readelf` and so on from the solana toolchain) would very slightly be incorrect (due to small idiosyncrasies like the relocations). 

@cyanpencil  What problems did you face with the tooling?

> Writing a parser for a much simpler but custom format would be less work, at this point. If optimizing performance for oracle programs is a top priority, a simple and optimized custom format would be better both for CUs and for tool devs at this point.

Can we measure how much savings we gain by using this proposal instead of #189? Specially for an average program and for an oracle too.

Compared to v0, #189 decreases the size of p-token in about 2kb: from 90kb (92488 bytes) to 88kb (89912 bytes). Most of this space is consumed by instructions, not headers.

Supposing we gain 500 bytes (I did not measure, only speculating) from this proposal compared to #189. For p-token, it would represent a 0.5% reduction. I don't think such a small size saving outweighs losing compatibility with off-the-shelf tooling and with existing ELF linkers.

deanmlittle:
For high performance single purpose programs, like a doppler oracle using in Prop AMMs, the most invoked programs on the whole chain right now, the current binary size is 464 bytes. SIMD-189 would make it 384 bytes if you ditch the section headers. This SIMD would make it 272 bytes.

cyanpencil:
Let me clarify: the tooling issues I've mentioned in the previous post are relative to v0. While I did not test I'm pretty sure that #189 would fix the vast majority of them (if not all).

My use case (tool dev) is very niche and not very important; the crux of the decision is @deanmlittle's use case where a few bytes might matter a lot. 
However, I do realise that ditching most of the ELF format would require a bunch more work in the linker and in other parts.

deanmlittle:
> > I think there is a reasonable argument to be made here that we get everything SIMD-189 gets us for free, there are fewer failure modes, nothing will break, program deploys get cheaper and we remove a lot of question marks.
> 
> @deanmlittle Can you list out the failure modes and question marks in SIMD-0189 that this spec doesn't have? As far as I understand, SIMD-0358 implies memory protection instructions, and SIMD-0189 explicitly lays them out via program headers. SIMD-0358 forces some fields to be zero that SIMD-0189 ignores. Is there any other difference?

Until about 1-2 hours after I posted this SIMD, the two acftually diverged by quite a bit more than that. SIMD-0189 has since been modified to become much more similar to SIMD-0358.

It isn’t merely implied by SIMD-0358; it’s enforced by the format: the region between the ELF header and the entrypoint offset is defined as read-only, and the remainder of the binary is executable. By contrast, adding program headers introduces additional mutability.

Concretely, here are the deltas I see today:
• Memory protection model: 0358 encodes R/X via a fixed layout; 0189 expresses protections via program headers.
• Field constraints: 0358 requires certain fields to be zero; 0189 does not.
• Program headers: 0189 requires a superfluous readable program header even for binaries without .rodata; 0358 does not.
• Section headers: 0189 (as written) still allows optional section headers; 0358 explicitly disallows them.

From a failure-mode perspective, 0189’s reliance on mutable headers creates extra surface area (e.g., malformed/omitted/mis-aligned headers, RO/exec discrepancies at load time) that 0358 avoids by construction. If I’ve missed an 0189 constraint that eliminates these cases, happy to be corrected—pointers to the exact clauses would help.

Net: Everything 0189 aims to achieve (memory protection, loader guidance) is provided by 0358’s fixed layout with fewer moving parts and cheaper deploys, while avoiding the header-related edge cases above.

deanmlittle:
> Let me clarify: the tooling issues I've mentioned in the previous post are relative to v0. While I did not test I'm pretty sure that #189 would fix the vast majority of them (if not all).
> 
> My use case (tool dev) is very niche and not very important; the crux of the decision is @deanmlittle's use case where a few bytes might matter a lot. However, I do realise that ditching most of the ELF format would require a bunch more work in the linker and in other parts.

It actually results in *less* work in the linker as there are fewer moving parts.

Lichtso:
> Until about 1-2 hours after I posted this SIMD, the two actually diverged by quite a bit more than that. SIMD-0189 has since been modified to become much more similar to SIMD-0358.

I did apply the three or so changes to the ELF header fields as agreed upon in Chicago and removed the two program headers for stack and heap in SIMD-0189 on your suggestion / request. Likewise SIMD-0358 also had a few omissions which, when filled with detail end up very similar to the other SIMD. So, both have converged and that is good as that moves us closer to consensus.

The biggest difference between the two is that SIMD-0189 is more explicit in its encoding and SIMD-0358 is more implicit, which at the end of the day makes a constant difference of 56 * 2 = 112 bytes, no matter how big the program. It is also true that SIMD-0189 because of its explicit nature can have malformed inputs which need to be rejected by the loader and that an implicit formulation would be nice.

However, because there is a fundamental difference between the concept of file offsets and virtual addresses the way this SIMD currently stands would get us back into a situation where only our (Solana specific) own ends (linker and loader) would be able to understand it and all other external tooling (compiler, debugger, coverage, profiler, instrumentation, fuzzer, etc.) would not, which is precisely what we aimed to support in the first place.

In other words currently, this is not a minimal binary encoding (because you could further omit most of the ELF header fields, but we decided against that direction) and it is not an ELF binary encoding because you need at least one program header. Thus, as it currently stands, it has not reached its proclaimed goal of being a "Minimal ELF Binary Encoding" yet.

deanmlittle:
> In other words currently, this is not a minimal binary encoding (because you could further omit most of the ELF header fields, but we decided against that direction) and it is not an ELF binary encoding

Sorry, but this is completely incorrect. A program header is not at all a requirement of the ELF. LLD will also link a headerless ELF just fine so long as it doesn't contain any dynamic relocations. In fact, all doppler oracles with <7 copies deployed on mainnet today already contain zero program headers. Program headers are only a requirement if the loader imposes it as one. Our loader happens to already not require them. If we abandoned the ELF header entirely, we could indeed make something slightly smaller that is extremely easy to relink back into an ELF file, but then we actually would be doing what you're saying and abandoning the ELF. If there is some appetite for that, I'm open to ideas. Historically, it has not been a palatable idea. At the end of the day, we really only care about the size of the text section, rodata section if one exists, and version. Right now a doppler oracle costs about $1 to deploy. It could be <$0.50. If you want to onboard trillions of dollars of RWAs, we're going to need a lot more cheap, efficient oracle feeds.

Lichtso:
> Sorry, but this is completely incorrect. A program header is not at all a requirement of the ELF.

It depends, the [ELF spec](https://refspecs.linuxbase.org/elf/elf.pdf) says that program headers are not necessary for relocatable files, and only relevant for dynamic libraries and executable files. That itself is still being discussed (see https://github.com/solana-foundation/solana-improvement-documents/pull/189#discussion_r2417843779). My point still stands that the mapping between file offsets and virtual addresses has to be declared somehow. So yes, we can make custom Solana only tooling which implicitly loads certain program headers but no other existing third party tooling (think debuggers, fuzzers, etc.) will be doing that.

> deployed on mainnet today already contain zero program headers

I am not debating whether or not it is possible but if it breaks or is compatible with third party tooling which we agreed on in Chicago is a goal.

> Right now a doppler oracle costs about $1 to deploy. It could be <$0.50. If you want to onboard trillions of dollars of RWAs, we're going to need a lot more cheap, efficient oracle feeds.

Isn't the fiat price dependent on the market value of a highly volatile asset? Making it a moving goal post no matter what we do on the SIMDs. The other thing is we are talking about a fixed improvement here, not one proportional to the program size. Thus, it over-proportionally matters for minimal programs which only have a few instructions, correct?

deanmlittle:
Withdrawn in favor of latest updates to #189

PR #363, opened 2025-09-19
SIMD-0363: Simple Alpenglow Clock

rogerANZA:


ksn6:
Could you articulate how these clock conditions interact with fast leader handover / a changing parent + the fact that your approach still works under these conditions?

Separately - with this SIMD, it may be worth including a "parent block (slot, hash)" field in the block footer in addition to the existing clock. This would be the "final parent" of the block, taking fast leader handover into account. Two good things happen as a result of this:

(1) given the final parent, we can skip a block with a bad clock quickly on shred ingest rather than after replay, since the first shred of the last ~~FEC set~~ block marker** will contain both the clock + last parent.

(2) repair in a block with an UpdateParent marker gets slightly better because now there are two shreds with the "final parent" that race each other

EDIT: fixed an issue with comment (1) above.

proposals/0363-simple-alpenglow-clock.md
```diff
@@ -0,0 +1,119 @@
...
+- Clock Variable in Solana Sysvar Cluster Data (unix_timestamp) is computed
+  differently. The variable is still accessed by Clock::get() or by passing the
+  sysvar account to a program as an account parameter. 
+
+
+## Detailed Design
+
+In Alpenglow, the current block leader includes an updated integer clock value
+(Rust system time in nanoseconds) in its block *b* in slot *s* in the block
+footer. This value is bounded by the clock value in the parent block.
```
ksn6, line 40: At what point should we get the system clock time? When it is the time to create our leader slot? When we actually send out the block footer? Or just refer to what the current clock sysvar says?

rogerANZA: You insert your system time when you produce the slice with the clock info in the block marker.

rogerANZA:
> Could you articulate how these clock conditions interact with fast leader handover / a changing parent + the fact that your approach still works under these conditions?

This should be mentioned in the fast leader handover SIMD in my opinion. (1) If a parent is changed and we consider everything before the parent change irrelevant, then we must include the clock again. (2) If the metadata of the block before the parent change is still relevant even after the parent change, then we don't include it again. In any case, we must have exactly 1 valid clock entry per block. But mentioning in the fast handover SIMD is more natural because it's the same exception for all the cases.
 
> Separately - with this SIMD, it may be worth including a "parent block (slot, hash)" field in the block footer in addition to the existing clock. This would be the "final parent" of the block, taking fast leader handover into account. 

We could decide to have the clock always in the last slice (if we have other information that always goes in the last slice). If this is the way to go, then we can describe it here as well.

Thanks for these good questions.

proposals/0363-simple-alpenglow-clock.md
ksn6: I mean - what precisely do you mean by "system time when you produce the slice?"

E.g. is this:

- The beginning of block production for a slot?
- The end of block production for a slot, post block building?
- When we actually create the block footer, post block building, and pre-shredding for turbine dissemination?

The specific choice isn't really enforceable, although, for default implementation purposes / consistency, it's probably worth opining on the specifics in the SIMD.

ksn6:
> (1) If a parent is changed and we consider everything before the parent change irrelevant, then we must include the clock again

What do you mean by "we must include the clock again?"

The block footer, which is where the clock value will reside, will appear exactly once per block, after the final entry batch (see [SIMD 0307](https://github.com/solana-foundation/solana-improvement-documents/pull/307/files#diff-55c89cf8fd9d0d5503ad1cbebffda984367ef1f2a1e897a0b17f5b4475d3692eR121-R122)).

Could we mention SIMD 0307 + specify that it's specifically the `block_producer_time_nanos: u64` field in the footer that will store the clock you're referring to in this SIMD?

> We could decide to have the clock always in the last slice (if we have other information that always goes in the last slice). If this is the way to go, then we can describe it here as well.

The clock won't always be in the last *slice / FEC Set*, but rather in the last *block marker*, which will eventually span multiple FEC Sets.

This is why I'm suggesting that we place parent block information into the footer as part of this SIMD - if (1) the clock and (2) the final parent are both included within the same shred, we can run the clock check in this SIMD exactly once, directly in shred ingest, prior to replay.

proposals/0363-simple-alpenglow-clock.md
rogerANZA: All of these are possible (and enforceable). This depends on how we organize meta-data in a block (i.e., our discussion in Chicago). If we may change the parent, then the header (the first slice) is the wrong place, but the footer (the last slice) should be alright.

rogerANZA:
> This is why I'm suggesting that we place parent block information into the footer as part of this SIMD

Sure, footer is okay. In fact, line 40 already says footer.

ksn6:
> Sure, footer is okay. In fact, line 40 already says footer.

I think there's a bit of confusion here. Yes, we're in agreement that the clock should go into the footer; e.g., SIMD 0307 specifies this. This isn't what I'm referring to, though.

To clarify - at the moment, the *only* fields included in the block footer are:
  - `user_agent`: `Vec<u8>` denoting the validator user agent (e.g., Agave / Firedancer / etc., with mods)
  - `block_producer_time_nanos: u64`: the clock field that this SIMD aims to impose constraints on

I'm saying that, in this SIMD, we should consider proposing a new third field to the block footer of type `(Slot, Hash)` which denotes the (parent, block ID) of the block associated with this block footer. This third field will allow us to impose the clock check on shred ingest.

proposals/0363-simple-alpenglow-clock.md
ksn6: For now, it looks like nothing prevents a validator from reporting a timestamp that is off by +/- 50 ms, equating to an honest validator reporting a timestamp that is at the beginning of block production vs. maybe even somewhere in the middle of block production or even the end of block production.

To somewhat make the notion of a clock uniform across validator implementations, we should probably specify roughly at what point in a leader's production of a block the timestamp should be captured.

If this can somehow be enforced, that's even better imho.

rogerANZA:
> SIMD 0307 specifies this. [...] I'm saying that, in this SIMD, we should consider proposing a new third field to the block footer of type `(Slot, Hash)` which denotes the (parent, block ID) of the block associated with this block footer. This third field will allow us to impose the clock check on shred ingest.

I don't disagree. But this should rather be in SIMD 307.

proposals/0363-simple-alpenglow-clock.md
rogerANZA: > For now, it looks like nothing prevents a validator from reporting a timestamp that is off by +/- 50 ms...

Well, I would say it can even be up to 800 ms wrongly reported. Nothing we can do here. 

But we established that the clock should be in the block footer, and we established that it should be captured before putting it it there, so what's still unclear?

ksn6: > All of these are possible (and enforceable).

> Well, I would say it can even be up to 800 ms wrongly reported. Nothing we can do here.

This part is a bit unclear - these two statements seem a bit contradictory. I agree with the second statement you're making re. being off by up to 800 ms, though.

> But we established that the clock should be in the block footer, and we established that it should be captured before putting it it there, so what's still unclear?

The part that's unclear - at what exact part of the leader block production phase should we have leaders set the block timestamp?

Should the timestamp be associated with when the leader starts producing their block? When the leader conclusively knows what the block parent is? When the leader actually constructs the footer itself?

I'm aware that it's impossible to enforce the particular point in time, as you pointed out re. the 800 ms piece. This being said, it's worth having a sane default.

ksn6:
FYI - after a few conversations, looks like we'll be punting on placing a third field in the block footer denoting the parent. We plan on accomplishing this via other means (can elaborate if there's interest) in later work.

proposals/0363-simple-alpenglow-clock.md
rogerANZA: "We assume that a correct leader inserts its correct local time..."

Maybe this is not clear enough? When you insert the time value in the footer you insert your local time at that point. This way we have the least (additional) skew. Note that we will still have a systematic lag since the slice then has be encoded, sent through Rotor/Turbine, and decoded on the receiver side. We might consider to take that systematic lag into account.

rogerANZA:
> I'm saying that, in this SIMD, we should consider proposing a new third field to the block footer of type `(Slot, Hash)` which denotes the (parent, block ID) of the block associated with this block footer. This third field will allow us to impose the clock check on shred ingest.

Okay, now I understand your argument.

Why only include the parent and the slot of the parent, and not also the actual time of that parent slot? Then our check is even easier because all the data is already there...

Would you agree that this is a slippery slope? Including the same information a second time is problematic in my opinion. Now we additionally would have to check that the second inclusion of the information is equal to the first inclusion of the information. What if not? Then the block is still skipped?

proposals/0363-simple-alpenglow-clock.md
```diff
@@ -0,0 +1,119 @@
...
+of slot *p* be *c* (in nanoseconds). For the clock value in slot *s* to be
+correct, the clock value of block *b* must be strictly higher than *c*, and at
+most *c*+(*s*–*p*) x 2*T*, where 2*T* is two times the block time, currently *T*
+= 400 ms. We assume that a correct leader inserts its correct local time as long
+as it is within the allowed bounds. If the correct local time is out of bounds,
+the leader inserts the minimum or maximum allowed time.
+
+If the clock value for slot s is not within the bounds (*c*, *c*+(*s*–*p*) x
+2*T*], the proposed block *b* is invalid, and validators vote skip. Currently
+2*T* = 8e8.
```
rogerANZA, line 51: My original proposal was to add up to 1 second. The new proposal is to make it completely symmetric (+/- 400 ms). However, it's not really symmetric because we also have to consider the time it takes to encode/disseminate/decode the information, which adds something like 100ms to the process. Should we explicitely add these X = 100 ms and make it 2T+X?

rogerANZA: ... but in 3 out of 4 cases it's the same leader, in which case it's X = 0. So maybe keep it this slightly asymmetric way.

rogerANZA:
To include the parent (again) is reasonable for repair. But this is orthogonal to this clock SIMD, isn't it?

proposals/0363-simple-alpenglow-clock.md
```diff
@@ -0,0 +1,119 @@
...
+it is better if programs work with just reasonably accurate clocks.
+
+## Impact
+
+Developers and validators must not rely on a highly precise system clock. If an
+application relies on a highly accurate clock, it should consider alternative
+sources.
+
+## Security Considerations
+
```
OliverNChalk, line 185: Assuming my understanding is correct that:

- If no blocks are produced then the next leader can set the time to be the same as last (no change) or up to 2 times time since last block;
- OR; If no slots are produced (no skips) that the block will simply just halt with no chance for the next leader to place the correct time.

Then I think the following risk scenario is worth considering:

- Money market (MM), let say marginfi, uses pyth pull oracles where a user is able to update the oracle and then use the MM.
- Chain halts for 1 hour, now the clock is 1 hour behind, imagine the chain halt coincided with market volatility.
- Attacker uses 1 hour old pyth payload to open insolvent positions on MM.

This is an example where the clock is not off by a few seconds but can be off by hours. In this scenario most current MMs would be vulnerable as they use the following code:

```rust
        let price = price_feed_account
            .get_price_no_older_than_with_custom_verification_level(
                clock,
                max_age,
                feed_id,
                MIN_PYTH_PUSH_VERIFICATION_LEVEL,
            )
            .map_err(|e| {
                debug!("Pyth push oracle error: {:?}", e);
                let error: MarginfiError = e.into();
                error
            })?;
```
Which in turn relies on the following check:

```rust
        check!(
            price
                .publish_time
                .saturating_add(maximum_age.try_into().unwrap())
                >= clock.unix_timestamp,
            GetPriceError::PriceTooOld
        );
```

This check will be completely broken until the clock catches up, allowing stale prices to be pushed.

### Caveat

The same issue exists in the current clock which I believe will have the price for block N use the votes from N-1 which of course will be pre halt and thus stale. That said, it will correct within a few slots as opposed to this clock which will have a much longer vulnerability window.

rogerANZA: > Assuming my understanding is correct that:
> 
> * If no blocks are produced then the next leader can set the time to be the same as last (no change) or up to 2 times time since last block;

Basically, yes. However, the new value has to be strictly higher (just by 1 tick though).

> ... Chain halts for 1 hour ...

Wait-what? So you're saying that the whole chain is down for a full hour? None of expected 9,000 blocks in that hour appended? I would say that in this case we have much bigger problems, don't we?

What we could do of course is to narrow the time the leader can choose in such a case, in the most extreme case even narrow it down to basically 1 hour +/- 400 ms. This is an oversimplification because if the chain was really down for 1 hour, our emergency protocol would kick in, and slots would eventually get exponentially longer)

But, essentially, we could change the formula, and give the leader a narrower window of choice if the chain was down for a very long time.

Would that make it better?

OliverNChalk: I think the core protocol is sound and desirable in 99.99% of operation. In the 0.01% where we have a chain halt it would be ideal if:

- The protocol restarted with the correct time
- OR; There was a way for onchain applications to detect that a halt/emergency mode had occurred

The former of the two is desirable as it protects applications that rely on semi accurate time without those applications needing to change their logic.

> So you're saying that the whole chain is down for a full hour? None of expected 9,000 blocks in that hour appended? I would say that in this case we have much bigger problems, don't we?

It's entirely feasible that the Solana blockchain halts during a liquidation cascade that promptly reverts. Now if money markets come back online after the recovery they should have minimal bad debt. However, if the clock lags then these money markets will observe all of those prices as the clock catches up. In this case an attacker will be able to open positions at the maximally depressed prices resulting in much more bad debt and potentially blowing up all on chain lending protocols in the worst case.

The reason chain halts are bad are:

1. Continuous services stop (CLOBs, payments). **not scary**
2. Protocol assumptions breakdown (liquidations, oracle update timeliness). **very scary**

Unfortunately the only way to address 2.liquidations is to have 100% uptime. However, we can address oracle update timeliness by providing an accurate post restart clock.

> But, essentially, we could change the formula, and give the leader a narrower window of choice if the chain was down for a very long time.
>
> Would that make it better?

This sounds perfect, how easy is this to achieve?

rogerANZA: > This sounds perfect, how easy is this to achieve?

Not difficult. The question is how exactly we would be doing it. 

1. The simplest way is narrowing the window (as suggested in the previous answer). 

There might be more elaborate ways (also easy to implement): 

2. For instance, after a long outage without any blocks, have the next 3 or 5 (or your favorite odd number) leaders propose a time and we take the median as a new starting time. But there are downsides to this solution, in particular we would have 3 (or 5) blocks without a good time.

rogerANZA: RE solution 1: For a 1 hour outage, you say that choosing a new time in [epsilon,2 hours] is too much freedom. What makes most sense? [1 hour - delta, 1 hour + delta] for what delta? What's the maximum delta that still makes sense? (A larger delta is better because it increases the chances that the actual time is in the interval.)

OliverNChalk: If Alpenglow halts at slot 100 (lets say 0600) and then leaders restart at 0700 (which would have been slot 200), will the first block be 101 or 200?

Ideally we could have validators post restart only vote to confirm blocks that are within 5s of their current system time. This seems simple to implement provided there isn't some hardcoded notion of clock correctness (beyond monotonic requirement) that would contradict our desire to have NTP synced time post restarts?

rogerANZA: We're in trouble if validators only confirm blocks which are close enough to their system time. This is actually an attack vector for various blockchains. This is why our clock correctness can only depend on the status of the chain, i.e. we only have a "syntactic" (and not a "semantic") criterion.

OliverNChalk: > RE solution 1: For a 1 hour outage, you say that choosing a new time in [epsilon,2 hours] is too much freedom. What makes most sense? [1 hour - delta, 1 hour + delta] for what delta? What's t

For the money market use case it looks like being accurate within 5 minutes is okay and would have minimal impact even during extreme volatility.

For other use cases such as CLOB orders that expire after N seconds, then the delta can be much greater (i.e. if we're offline for 1 hour and most traders use a 10s expiry, then so long as we come back >10s past the previous block, their cancels will be processed).

For my understanding, the difference between this and NTP is that with this approach we can just use machine local timekeeping to determine elapsed time +/- a delta. With NTP it could contradict the solana clock and if the network had vastly different results it would prevent consensus from forming. Conversely provided operators have access to continuous local time they can more confidently guarantee a restart?

I'm curious, this would mean validators cannot use SystemTime when measuring time since last block - because system time is affected by NTP. How would validators track elapsed time, additionally, does this solution still work if validators migrate their hosts?

rogerANZA: The clock will be much more accurate than 5 minutes, except in a case when *the internet is broken*. And yes, our clock is also fed by NTP. The crucial part is that we don't base consensus on time. All the systems that do that are in hot water.

rogerANZA: So basically what I get is that if the system had an outage, we should make sure that the clock actually advanced (say at least half of the outage), and then we're good.

rogerANZA: Thanks for the feedback by the way, I was looking for an application expert.

OliverNChalk: > So basically what I get is that if the system had an outage, we should make sure that the clock actually advanced (say at least half of the outage), and then we're good.

Well there are two quality of services:

- Expiring stale quotes (orders, cached oracles, etc). **this is covered by 50%+ of outage time being included**
- Ensuring stale quotes cannot enter the system. **this requires us to guarantee the clock will never drift more than X minutes (probably 5 minutes is a reasonable target)**

So if downtime is an hour then we can tolerate ~92% - ~108% of real time for the first block and still meet the 5 minute SLA.

OliverNChalk: I have two questions about failure modes:

- If no blocks or skips are produced for 1 hour (9000 slots) is recovery automatic once enough nodes come online and if so do they just produce slot 9001?
- If no blocks are produced but skips are is recovery automatic and if so would the first slot be 9001?

If recovery is automatic in all scenarios then I would say it makes sense not to introduce NTP. However, if it's expected that recovery from a liveness failure is manual then it may not be terrible to introduce NTP as some operator oversight/involvement is expected

rogerANZA: This very much depends on the disaster. In a "normal" disaster scenario, we continue to produce slots (and just skip them). In a "all internet down" scenario however, we don't do that, and we indeed don't increase slot numbers. In that case, if the disaster lasts for X time, we need X time to recover the clock (basically by speeding up the clock by a factor of 2). If you think that this is unacceptable, then we need to fix it. There are probably several ways how to fix it. One I sketched a few messages ago ("2"). 

Basically we would need some sort of consensus for a new starting clock value. To be ultra-secure, we could have a program where all validators send their proposed time. The time can only be in the future of the current clock, and we need a certain % of the stake to submit their bid, and then we take the stake median of the bids as the new clock value. Something like that. Note however that it would need time to get to an agreement, and now we're talking about a situation where the whole internet was down for an extended time!! This is why I would prefer the alternative where the clock speeds up by a factor of 2 until we have the right time again.

OliverNChalk: Hmm, I guess the question might boil down to, is it worse to:

- Take a dependency on NTP (but only during a restart/major slot skip).
- OR; Accept it's possible for the clocking to be 30m+ out of sync while its catching up.

If we take a dependency on NTP conditional on some critical failure having occurred prior, what do you see as the increased risk to the protocol? I would assume we now have a risk that a % of our validators have poisoned NTP and thus we cannot restart the protocol until the NTP issues are manually corrected/overridden?

proposals/0326-alpenglow.md
damotatts-cyber: 0x44af94ecacc4aba4a7a74f051d96b6113bff5702

proposals/0363-simple-alpenglow-clock.md
rogerANZA: A dependency on NTP is a big NO from my side. It directly affects consensus, and I cannot prove correctness of consensus anymore. And the 30 minute skew you only get after a 30 minute outage (which we will never have, fingers crossed).

BUT: It's good that you raised the question, and that we are aware of it. I will (eventually) think more seriously about it. At this point, I would suggest that we eventually have a special program which can be used to forward the blockchain time. Anybody can suggest to vote to jump to a future time, and then all the stakers can suggest a time, and then we take the median of that. If enough stake participates (send their time vote to the program in the allowed time frame), we jump there. I don't think we need to immediately implement this, but it's good to have the thought ready if we need it.

OliverNChalk: Okay, that sounds fair. Only thing to flag is that it would be ideal if the clock time was fixed before any defi transactions were processed. For instance if it took 32 slots (random number i chose) to fix the clock then ideally for the first 32 slots no user txs are processed to prevent the looting of vulnerable defi protocols.

Though agreed, ideally we don't have a 30 minute outage ever again... or at least it doesn't coincide with large market volatility

qkniep: We could try to expose this information to the protocols and have the vulnerable protocols do this logic of dismissing all transactions for some time. But I would definitely not want to halt *all* user transactions just because *some* protocols have a bug.

OliverNChalk: If there's a way to expose that the clock is/may be out of sync to the application layer then I agree that would solve this. My fear was that doing so would be roughly equivalent to reaching consensus on the clock time itself but if that's not the case then perfect.

ksn6:
@rogerANZA here are implementation details we'll want to include in the SIMD:

**Implementation**

In proposing a block, a leader will include a special marker called a "block footer," which stores a UNIX timestamp (in nanoseconds). As of the writing of this SIMD, the `block_producer_time_nanos` field of `BlockFooterV1` stores the clock:

```
/// Version 1 block footer containing production metadata.
///
/// The user agent bytes are capped at 255 bytes during serialization to prevent
/// unbounded growth while maintaining reasonable metadata storage.
///
/// # Serialization Format
/// ```text
/// ┌─────────────────────────────────────────┐
/// │ Producer Time Nanos          (8 bytes)  │
/// ├─────────────────────────────────────────┤
/// │ User Agent Length            (1 byte)   │
/// ├─────────────────────────────────────────┤
/// │ User Agent Bytes          (0-255 bytes) │
/// └─────────────────────────────────────────┘
/// ```
#[derive(Clone, PartialEq, Eq, Debug)]
pub struct BlockFooterV1 {
    pub block_producer_time_nanos: u64,
    pub block_user_agent: Vec<u8>,
}
```

Upon receiving a block footer from a leader, non-leader validators will:

- Locally update the clock sysvar associated with the bank
- During replay, validate the bounds of this clock sysvar with respect to the parent bank's clock sysvar and proceed as outlined in "Detailed Design"

Separately, could you add my name to the authors list please (`ksn6`)? Now that we have the ability to disseminate block components within the Alpenglow prototype, I'll be working on the above items to implement the clock checks.

ksn6:
> We could try to expose this information to the protocols and have the vulnerable protocols do this logic of dismissing all transactions for some time. But I would definitely not want to halt all user transactions just because some protocols have a bug.

> If there's a way to expose that the clock is/may be out of sync to the application layer then I agree that would solve this. My fear was that doing so would be roughly equivalent to reaching consensus on the clock time itself but if that's not the case then perfect.

@OliverNChalk @qkniep - any further updates on this?

In the interest of getting this SIMD across the finish line, I'd strongly prefer coming up with a solution to this in a follow-up SIMD, while keeping this SIMD about the design for when the chain is online.

OliverNChalk:
Works for me - happy for @qkniep to propose exposing additional information to DeFi apps in a subsequent proposal

ksn6:
FYI, we're implementing the clock checks in this PR: https://github.com/anza-xyz/alpenglow/pull/597.

@rogerANZA additional implementation details to include in this SIMD:

The clock sysvar, as currently implemented, has second resolution. This SIMD proposes a clock with greater resolution (nanoseconds). Accordingly, we introduce an off-curve account where we store the clock value upon processing the block footer (at the end of a block).

In particular, we calculate the off-curve account as follows:

```rust
/// The off-curve account where we store the Alpenglow clock. The clock sysvar has seconds
/// resolution while the Alpenglow clock has nanosecond resolution.
static NANOSECOND_CLOCK_ACCOUNT: LazyLock<Pubkey> = LazyLock::new(|| {
    let (pubkey, _) =
        Pubkey::find_program_address(&[b"alpenclock"], &agave_feature_set::alpenglow::id());
    pubkey
});
```

where `Pubkey::find_program_address` is a part of `solana-address-1.0.0`. In practice, the `Pubkey` ends up having value `BKRDmw2hTDSxQK4mitpK7eCWkNUvCvnaWqm1NZmGDTUm`.

The clock sysvar is updated at the end of each block by simply dividing the nanosecond timestamp by `1_000_000_000` and rounding down (i.e., integer division). For now, we do not expose the nanosecond clock to SPL programs via a sysvar; we may choose to do so in the future.

AshwinSekar:
Might be worth mentioning somewhere that for program developers using the clock sysvar the semantics have slightly changed:
it's no longer an estimation of the start time of the current block but an estimation of the time at which the last slice of the parent block was produced.

I hope no one is using the clock at this resolution to be impacted by this, but would be good to call it out anyway.

rogerANZA:
> ```rust
> off-curve account
> ```



> FYI, we're implementing the clock checks in this PR: [anza-xyz/alpenglow#597](https://github.com/anza-xyz/alpenglow/pull/597).
> 
> @rogerANZA additional implementation details to include in this SIMD:
> 
> The clock sysvar, as currently implemented, has second resolution. This SIMD proposes a clock with greater resolution (nanoseconds). Accordingly, we introduce an off-curve account where we store the clock value upon processing the block footer (at the end of a block).
> 
> In particular, we calculate the off-curve account as follows:
> 
> ```rust
> /// The off-curve account where we store the Alpenglow clock. The clock sysvar has seconds
> /// resolution while the Alpenglow clock has nanosecond resolution.
> static NANOSECOND_CLOCK_ACCOUNT: LazyLock<Pubkey> = LazyLock::new(|| {
>     let (pubkey, _) =
>         Pubkey::find_program_address(&[b"alpenclock"], &agave_feature_set::alpenglow::id());
>     pubkey
> });
> ```
> 
> where `Pubkey::find_program_address` is a part of `solana-address-1.0.0`. In practice, the `Pubkey` ends up having value `BKRDmw2hTDSxQK4mitpK7eCWkNUvCvnaWqm1NZmGDTUm`.
> 
> The clock sysvar is updated at the end of each block by simply dividing the nanosecond timestamp by `1_000_000_000` and rounding down (i.e., integer division). For now, we do not expose the nanosecond clock to SPL programs via a sysvar; we may choose to do so in the future.

ok. 

(for the record: I was told to go for nanoseconds because it's "easier." just computing with seconds would had been possible too.)

proposals/0363-simple-alpenglow-clock.md
```diff
@@ -0,0 +1,126 @@
...
+type: Core
+status: Review
+created: 2025-09-19
+---
+
+## Summary
+
+Since the clock sysvar computation is incompatible with Alpenglow, we need a new
+design. In this document we suggest a simple replacement solution, which should
+be accurate enough for all use cases.
```
topointon-jump, line 17: Just throwing this out there - should we consider deprecating the `unix_timestamp` and `epoch_start_timestamp` clock sysvar fields entirely? We could mark them as deprecated in the next release, and set them to zero in a feature gate a little while after that.

If we are considering changing the semantics of the timestamps, we should also consider getting rid of them entirely.

Is there a good reason not to do this? Will this break a lot of use-cases today? Genuinely asking. If removing these fields would break a lot of important use-cases then we should probably keep them, if not then it's worth considering.

OliverNChalk: `unix_timestamp` is used very broadly within defi, its probably not feasible to ever remove that. would there be another trustless way to access the current time on chain? a notion of time is required for things like interest calculations.

rogerANZA: Unfortunately I don't know how many applications are using the clock (and how). I asked people about this before I wrote this SIMD, but nobody was able to tell me. The consensus was then to have some reasonably good solution no matter what. This is basically what this SIMD is about.

topointon-jump: @OliverNChalk if the timestamp fields are removed, then slot number could be used as a rough proxy for time. I'm not sure if this is sufficient.

OliverNChalk: This would require us to expose the configured slot time target and would likely drift fairly significantly over time? Not to mention this still breaks marginfi, kamino, and all other money markets, not to mention meme coin launchpads/AMMs that use time as a trigger for pool unlock.

In an ideal world we have decent accuracy in all cases. In a okay world we have this proposal which gives us decent accuracy in the average case. I don't think it's possible to remove the concept of time from DEFI as time is very fundamental to finance (i.e. funding/interest rates). If we remove time then now all of defi will have to ask Pyth for the current time which introduces another trust assumption making all of solana finance dependent on a 3rd party chain

t-nelson: the intended uses of today's block timestamp are these
1. enforcing paper contract dates on chain (eg. stake account lockups)
1. real-world (tax) accounting

all other uses are a misuse of this timestamp as it is not what the consumer believes it to be. that said, its misuse is prevalent, so we must retain its availability as its removal would be extremely disruptive to the ecosystem today

attempts to use `slot_height * target_block_time` will fall apart since they lack corrective inputs. each time we drive down the block time target would need treatment. we've already accumulated several hours worth of slot-time/wall-time skew due to outages

that the clock sysvar timestamp is unsuitable for alpenglow is fine, so long as the solution to alpenglow clock is capable of emitting a timestamp suitable for populating the clock sysvar timestamp

ksn6:
@rogerANZA any updates here?

In particular, please see:
- https://github.com/solana-foundation/solana-improvement-documents/pull/363#issuecomment-3464172711
- https://github.com/solana-foundation/solana-improvement-documents/pull/363#issuecomment-3524325922

rogerANZA:
> @rogerANZA any updates here?
> 
> In particular, please see:
> 
> * [SIMD-0363: Simple Alpenglow Clock #363 (comment)](https://github.com/solana-foundation/solana-improvement-documents/pull/363#issuecomment-3464172711)
> * [SIMD-0363: Simple Alpenglow Clock #363 (comment)](https://github.com/solana-foundation/solana-improvement-documents/pull/363#issuecomment-3524325922)

sure, feel free to integrate these two comments in the doc?

rogerANZA:
Would be good to get a review here, so that we can finally merge this back.

proposals/0363-simple-alpenglow-clock.md
```diff
@@ -0,0 +1,198 @@
...
+    pub block_producer_time_nanos: u64,
+    pub block_user_agent: Vec<u8>,
+}
+```
+
+Upon receiving a block footer from a leader, non-leader validators will:
+
+- Locally update the clock sysvar associated with the bank
+- During replay, validate the bounds of this clock sysvar with respect to the
+  parent bank's clock sysvar and proceed as outlined in "Detailed Design"
```
topointon-jump, line 120: Quick question about this - how does replay work? Do we update the clock sysvar from the block footer and _then_ start replaying? If so, doesn't that mean we can't start replaying until we've received the block footer?

AshwinSekar: no, it happens when we replay the block footer. So effectively we are setting the clock sysvar at the end of the block for use in the child block

rogerANZA: "during replay" should be more precise?

ksn6: Yeah, useful to clarify this.

@rogerANZA - I updated the doc. The relevant part:

Non-leader validators must apply the following block footer logic after transactions in a block have been replayed:

- Locally update the clock sysvar associated with the bank
- Validate the bounds of this clock sysvar with respect to the parent bank's clock sysvar as outlined in "Detailed Design"

For avoidance of doubt - while replaying transactions in a block with slot S, the clock sysvar has value specified in S.parent’s block footer.

rogerANZA:
> Hey @rogerANZA, this PR has been inactive for 60 days and is now marked as **stale**.
> 
> If you're still working on this, please comment below to let us know. Otherwise, this PR will be automatically closed in **7 days**.
> 
> To keep this PR active, simply leave a comment explaining the current status or any blockers you're facing.

This should be merged. Not sure who can do that.

bw-solana:
@AshwinSekar can you SME this one?

I think we've implemented and tested this one in the AG feature branch, so should have a pretty good idea if it's viable or not 😄

proposals/0363-simple-alpenglow-clock.md
```diff
@@ -0,0 +1,198 @@
...
+        &[b"alpenclock"],
+        &agave_feature_set::alpenglow::id(),
+    );
+    pubkey
+});
+```
+
+where `Pubkey::find_program_address` is a part of `solana-address-1.0.0`. In
+practice, the `Pubkey` ends up having value
+`BKRDmw2hTDSxQK4mitpK7eCWkNUvCvnaWqm1NZmGDTUm`.
```
AshwinSekar:
```suggestion
`HQcg2uM8uUqfRprvypQGcU7qucUZJY3odWjJVYce2a6C`.
```

AshwinSekar (commented):
looks good, matches the implementation we have.

Please update the account pubkey and remove the `.idea` directory

rogerANZA:
everything good?

AshwinSekar (approved):
approving as SME

proposals/0363-simple-alpenglow-clock.md
```diff
@@ -0,0 +1,198 @@
...
+Since the clock sysvar computation is incompatible with Alpenglow, we need a new
+design. In this document we suggest a simple replacement solution, which should
+be accurate enough for all use cases.
+
+## Motivation
+
+So far, individual votes also include a timestamp, and the clock sysvar was
+updated every slot by calculating the stake-weighted median of validator vote
+provided times. With Alpenglow, individual votes do not go on chain anymore.
+Consequently, we need a new clock design.
```
bw-solana, line 24: should probably mention something here about BLS aggregation and unique timestamps in the vote payload not playing nicely together

proposals/0363-simple-alpenglow-clock.md
```diff
@@ -0,0 +1,198 @@
...
+- Locally update the clock sysvar associated with the bank
+- During replay, validate the bounds of this clock sysvar with respect to the
+  parent bank's clock sysvar and proceed as outlined in "Detailed Design"
+
+## Alpenglow Clock Variable Storage
+
+The clock sysvar, as currently implemented, has second-resolution. This SIMD
+proposes a clock with greater resolution (nanoseconds). Accordingly, we
+introduce an off-curve account where we store the clock value upon processing
+the block footer (at the end of a block).
```
bw-solana, line 127: nit: make it explicit that this means the value is stored after all txs in the block have been executed

proposals/0363-simple-alpenglow-clock.md
```diff
@@ -0,0 +1,198 @@
...
+/// ├─────────────────────────────────────────┤
+/// │ User Agent Bytes          (0-255 bytes) │
+/// └─────────────────────────────────────────┘
+/// ```
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct BlockFooterV1 {
+    pub block_producer_time_nanos: u64,
+    pub block_user_agent: Vec<u8>,
+}
+```
```
bw-solana, line 114: this is what I'm currently seeing in master:
```rust
pub struct BlockFooterV1 {
    pub bank_hash: Hash,
    pub block_producer_time_nanos: u64,
    #[wincode(with = "WincodeVec<u8, FixIntLen<u8>>")]
    pub block_user_agent: Vec<u8>,
    pub final_cert: Option<FinalCertificate>,
    pub skip_reward_cert: Option<SkipRewardCertificate>,
    pub notar_reward_cert: Option<NotarRewardCertificate>,
}
```

AshwinSekar: I think you could probably just put in a `...` for the other fields and keep the focus on `block_producer_time_nanos`.
They're covered by other SIMDs:
`bank_hash` https://github.com/solana-foundation/solana-improvement-documents/pull/298/changes#diff-f022e6f12bcbae45d2d8512b4dc201f5d96f1de8a24318616b7c82393dce868eR161
`*_reward_cert` https://github.com/solana-foundation/solana-improvement-documents/pull/428/changes#diff-3369408e3148ac266f9f9bfb8733b12056781168a47208dc308ed2c586f6c188R94
`final_cert` https://github.com/solana-foundation/solana-improvement-documents/pull/402/changes#diff-715a0ef0aabe5a1547fd3ac944371b1ca20c2316bee6a0d0bdc980a72eb6a9dbR129

bw-solana: could probably leave it as
```rust
pub struct BlockFooterV1 {
    pub bank_hash: Hash,
    pub block_producer_time_nanos: u64,
    ...
```
knowing that this field is offset 32 bytes into the footer is the necessary piece

t-nelson: aren't these details for the footer simd itself? here we should only care about layout and interpretation of the relevant fields.

---

we mention the user agent in this section, but i'm assuming it's unrelatated?

---

fields should explicitly state signedness and endianness in addition to size. don't rely on language specific types to relay this information. ideally don't include them at all

proposals/0363-simple-alpenglow-clock.md
```diff
@@ -0,0 +1,198 @@
...
+/// The off-curve account where we store the Alpenglow clock. The clock sysvar
+/// has seconds resolution while the Alpenglow clock has nanosecond resolution.
+static NANOSECOND_CLOCK_ACCOUNT: LazyLock<Pubkey> = LazyLock::new(|| {
+    let (pubkey, _) = Pubkey::find_program_address(
+        &[b"alpenclock"],
+        &agave_feature_set::alpenglow::id(),
+    );
+    pubkey
+});
+```
```
t-nelson, line 141: This probably more implementation detail that belongs here, but probably a reasonable stand in since there's no PDA SIMD to point to

---

Aside: I happened across this in the code. We can skip the runtime generation by resolving the bump seed once, then using [Address::derive_program_const](https://docs.rs/solana-address/latest/solana_address/struct.Address.html#method.derive_address_const) to assign a constant. We should be explicit about the bump seed value anyway

proposals/0363-simple-alpenglow-clock.md
```diff
@@ -0,0 +1,198 @@
...
+
+where `Pubkey::find_program_address` is a part of `solana-address-1.0.0`. In
```
t-nelson, line 143: This crate stuff for sure does not belong in a SIMD. Remember we're supposed to be implementation agnostic here

proposals/0363-simple-alpenglow-clock.md
```diff
@@ -0,0 +1,198 @@
...
+practice, the `Pubkey` ends up having value
+`HQcg2uM8uUqfRprvypQGcU7qucUZJY3odWjJVYce2a6C`.
+
+The clock sysvar is updated at the end of each block by simply dividing the
+nanosecond timestamp by `1_000_000_000` and rounding down (i.e., integer
+division). For now, we do not expose the nanosecond clock to SPL programs via a
+sysvar; we may choose to do so in the future.
```
t-nelson, line 150: > we may choose to do so in the future

NAK. They already don't understand the second-resolution clock. God knows what religious nonsense would come from nanos

proposals/0363-simple-alpenglow-clock.md
```diff
@@ -0,0 +1,198 @@
...
+program accessing the clock is poorly designed and relies on a very accurate
+clock, such an extended bribe could be profitable. If at the end of the bribe
+period the clock was Delta (seconds) off, then it will take about Delta seconds
+for the clock to go back to normal.
+
+## Backwards Compatibility
+
+This new clock is not as accurate as the current clock. So those users that run
+programs that access the clock might need to adapt their code if they need a
+high accuracy. We got the impression that this is not an issue in practice.
```
t-nelson, line 197: @jacobcreech is it true?

PR #370, opened 2025-09-24
SIMD-0370: Remove Compute Unit Block Limit

topointon-jump:


proposals/0370-remove-compute-unit-block-limit.md
```diff
@@ -0,0 +1,115 @@
...
+        // Block verification (including unified scheduler) case;
+        // collect and check transaction costs
+        let tx_costs = get_transaction_costs(bank, &commit_results, batch.sanitized_transactions());
+        check_block_cost_limits(bank, &tx_costs).map(|_| tx_costs)
+```
+
+Each validator client will also need to have a way of aborting the execution of
+a block if the timeout has been exceeded. As this will be highly dependent on
+each client's implementation, and does not affect the consensus algorithm,
+the mechanism for this is out of scope for this SIMD.
```
jacobcreech, line 77: This goes into a timeout instead of the CU cap. Does this mean we're getting rid of CUs in totality with this SIMD?

cavemanloverboy:
incompatible with async exec

proposals/0370-remove-compute-unit-block-limit.md
topointon-jump: Getting rid of CUs entirely is more complicated, as it affects the block producers and the user-facing APIs. Removing the limit, which is redundant with the Alpenglow timeout, is a good first step in that direction.

proposals/0370-remove-compute-unit-block-limit.md
```diff
@@ -0,0 +1,115 @@
...
+if we were to continue manually increasing the block limit, which will result
+in a better experience for users and on-chain app devs as block space will
+become less scarce.
+
+This also makes it easier to reduce slot times - as we don't have to reason
+independently about the block compute unit limit and the slot time.
+
+## Security Considerations
+
+We will need to make sure that larger blocks can be properly disseminated
```
alessandrod, line 92: This should probably be expanded to explain that it's safe if a large block can't be propagated in time (which I'm assuming is just another case of slow block gets skipped)

alessandrod (commented):
>  Removing the limit today has tangible benefits for the ecosystem and end
  users, as described above. These benefits can be realized today, without
  waiting for the future architecture of the network to be fleshed out.

Strong +1 from me

bw-solana:
confirming that we would still keep some block limits as part of this proposal (e.g. maximum shred limit)

proposals/0370-remove-compute-unit-block-limit.md
```diff
@@ -0,0 +1,115 @@
...
+in the replay stage, effectively reverting (`apply_cost_tracker_during_replay`).
+
+## Motivation
+
+The current incentive structure for validator clients and program
+developers is broken. The capacity of the network is determined not
+by the capabilities of the hardware but by the arbitrary block compute
+unit limit. Removing the limit means that:
+
+- Block producers are incentivized to continually improve their
```
akhi3030, line 34: Just to cover all ground, if just the block producer improves their machine and the rest of the network don't, then they are increasing the chances of their blocks being skipped.  In other words, would there need to be some coordination between the nodes in the network so that they all upgrade their machines at the same time?

proposals/0370-remove-compute-unit-block-limit.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Summary
+
+This is a post-Alpenglow proposal.
+
+In Alpenglow, voter nodes broadcast a `SkipVote` if they do not manage to
+execute a block in time. This makes the block compute limit redundant:
+any blocks which take too long to execute are skipped.
+
+This SIMD therefore removes the block compute unit limit enforcement
```
akhi3030, line 24: As there was some discussion about it in Chicago, it might be worth clarifying whether or not we will keep the per transaction limit.  I believe the rough consensus was that we still keep the per transaction limit.

proposals/0370-remove-compute-unit-block-limit.md
deanmlittle: Producing a slightly larger block than the majority of the network *is* the coordination. https://en.wikipedia.org/wiki/Stackelberg_competition

akhi3030: thanks for the link!  Makes sense.

proposals/0370-remove-compute-unit-block-limit.md
deanmlittle: I see no reason for there to be a protocol limit around the computational complexity of a transaction that a validator wishes to accept. If we are already accepting the premise that validators can do the math on how big their blocks should be to balance earnings against network acceptance, the only real question to consider is how a transaction broadcaster would signal to validators how many nanoseconds of compute they are requesting.

deanmlittle:
+1

proposals/0370-remove-compute-unit-block-limit.md
akhi3030: I agree with that.  I just meant that in Chicago we covered that making this change should be separate from this proposal as it will have other far reaching consequences.

proposals/0370-remove-compute-unit-block-limit.md
```diff
@@ -0,0 +1,115 @@
...
+Removing the static block compute unit limit may impact the asynchronous execution
+and certain variants of multiple concurrent proposers designs.
+
+However, it makes sense to remove this limit now:
+
+- There are several variants of the multiple concurrent proposers design being
+  debated, some of which have asynchronous execution and some of which have
+  synchronous execution. In some of these designs the block limit is needed
+  and in some of them it is not. Therefore, removing the limit now does not
+  preclude us from realizing multiple concurrent proposers.
```
KirillLykov, line 106: I think this part is considered to be the main obstacle. Do we have an idea how a hypothetical MCP could look like in the absence of the block limit? To sort of proof existence of meaningful solution.

aeyakovenko:
> Removing the limit today has tangible benefits for the ecosystem

Not really. Blocks aren't full.  So if the block limit was removed there would be little or no tangible difference to end users in terms of price or latency or to staked node revenues.

alessandrod:
> > Removing the limit today has tangible benefits for the ecosystem
> 
> Not really. Blocks aren't full. So if the block limit was removed there would be little or no tangible difference to end users in terms of price or latency or to staked node revenues.

The phrasing isn't great because by today it means once alpenglow ships. If by that time blocks aren't full devs are going to have some questions for you.

aeyakovenko:
There is a bunch of hand wavy arguments in the proposal 

* since blocks aren't full right now there would be no impact to users. We haven't had any time where demand would spike median fees or average fees significantly. So it's not even clear that burst capacity would be meaningful 

* there is no incentive for the rest of the network to increase hardware to support some crazy leaders big block.  If anything they have an incentive to fail it and steal all the mev. 

Main issue I have with this is that it blocks MCP. The point of IBRL is so end to end user costs go down, including spreads, and chain revenues go up. All the perf improvements help, like increasing block limits, but it all follows Amdhals law.

eugene-chen:
Why this instead of committing to a schedule of CU increases, i.e. changing the default to be gradually increasing the CU limit?

There is a ton of impact to validator games that is difficult to model and not contemplated in the proposal.

alessandrod:
> There is a bunch of hand wavy arguments in the proposal
> 
>     * since blocks aren't full right now there would be no impact to users. We haven't had any time where demand would spike median fees or average fees significantly. So it's not even clear that burst capacity would be meaningful

Before the increases we had months where we were always at 48M CUs. During $TRUMP/$MELANIA the blocks were extra full and while perhaps median fees didn't spike significantly, latency did and people complained.

You're right that blocks aren't full right now, but we've had cases where burst capacity would have helped.

>     * there is no incentive for the rest of the network to increase hardware to support some crazy leaders big block.  If anything they have an incentive to fail it and steal all the mev.

There is also no incentive to pack crazy big blocks. And FWIW the cluster as it is today is already way overprovisioned for the kind of load we have (it's provisioned for how bad agave used to be memory/iops wise). 
 
> Main issue I have with this is that it blocks MCP. 

How far away is MCP realistically? We can put CUs back if need be? It's not like the current implementation is great anyway, the whole estimate/refund code in agave is anti IBRL and needs to be rewritten anyway.

alessandrod:
> Why this instead of committing to a schedule of CU increases, i.e. changing the default to be gradually increasing the CU limit?

CUs are wildly inaccurate, and their implementation at least in agave is even worse. Historically, they were used to bound replay time, which post AG will happen with a timer. 

> There is a ton of impact to validator games that is difficult to model and not contemplated in the proposal.

Can you say more? This is why we're here :)

proposals/0370-remove-compute-unit-block-limit.md
```diff
@@ -0,0 +1,115 @@
...
+  that they don't skip blocks produced by high-performing block producers.
+
+Together, these two incentives produce a flywheel effect where
+block producers continually improve their performance, which in turn
+increases the average capacity of the validator client set, which in
+turn makes it possible for block producers to safely push the limits,
+and so on. The net result is that the capacity of the network is
+governed by market forces - if demand is there, the capacity of the
+network will increase to meet it.
+
```
rogerANZA, line 52: I have always been a big fan of the idea of removing the compute limit completely, since Alpenglow has no issue with that. (BTW: Since this is an Alpenglow SIMD, I would like to see "Alpenglow" in the title.) Block proposers are incentivized to up their hardware game, which is great! 

However, I have questions: 

1) Clearly a super-advanced proposer cannot pack its block arbitrarily because it still needs to get the 2x60% or 1x80% approval rating. It might be a difficult game to try to figure out what speed is currently possible. Maybe not impossible, but certainly harder than just greedily packing to the max.

2) At some point some nodes may fall behind. Not only in replaying the execution by the leader, but even to catch up with the chain advancing. Now these nodes will drop out. New (beginner, probably also in hardware) nodes cannot even enter the system because they might not be able to catch up from a snapshot. Are we okay with that? Nodes which fall back may have to up their game, but they might also just leave.

3) If the speed advances during an epoch, we might fall below the 60/80 thresholds, at which point we only get skips, and we basically have to enter Alpenglow's disaster scenarios.

4) Being closer to a leader also helps because you can start executing earlier. So as a side effect, we might get centralization.

I think all these problems are solvable, so I've always been a big advocate of ditching the limit. But the questions need answers. For instance, it would help tremendously if we made sure that the leader's work (to put together the block) is bigger than the work of replaying/checking by the followers. E.g., the leader might give advice for parallel execution, or even stuff like rollups. That way the leader still has an incentive to go hard, but we have a sane system at the same time. I think for this proposal to pass, we need something like that.

akhi3030: Agree with the points above.  

> Being closer to a leader also helps because you can start executing earlier. So as a side effect, we might get centralization.

Another type of centralization that we might see is that if the bigger validators keep upgrading to more expensive hardware, the smaller ones who cannot afford to upgrade would be forced to drop out.  So as a result, we might end up with fewer big validators.

proposals/0370-remove-compute-unit-block-limit.md
bw-solana: My understanding is there are 3 schools of thought on this:
1. MCP w/ sync execution is possible and Maher is cooking up a design
2. MCP requires async to be viable
3. It doesn't matter because we'll just add block limits back if we decide to do MCP w/ async

Given this is a post-Alpenglow SIMD (i.e. we have the luxury of some time on our hands), I think it would be great to see what Maher comes back with to see if MCP w/ sync looks viable.

We may still decide to go forward w/ timeouts but at least we would be doing it with a common understanding of the downstream impact

eugene-chen:
> > There is a ton of impact to validator games that is difficult to model and not contemplated in the proposal.
> 
> Can you say more? This is why we're here :)

Agree this is why we are here :) Here are a couple examples.

- One degenerate equilibrium is the bottom 20% of validators continually get nuked, eventually leading to an effectively very small validator set.
- As validators end up "close to the edge" of profitability / losing independent stake / keeping up with the chain, there is more incentive to colocate with other stake to be able to replay and vote in time, leading to data center centralization.
- Under the no-CU-limit regime, leaders should be more incentivized to build and stream their blocks continuously, rather than waiting to blast their blocks out in the last 50ms of their slot. I believe this is a good thing but it's difficult to forecast the second-order effects.

I understand CUs are far from perfect, but at least they are predictable, and the limits themselves can also be increased in a predictable manner. That will create much less thrash in validator economics.

proposals/0370-remove-compute-unit-block-limit.md
igor56D: this is the first I'm hearing of Stackelberg so I may be missing something, but how does this apply to leader monopoly block production? Even in MCP, proposers aren't directly competing for the same block space afaik.

proposals/0370-remove-compute-unit-block-limit.md
```diff
@@ -0,0 +1,115 @@
...
+unit limit. Removing the limit means that:
+
+- Block producers are incentivized to continually improve their
+  performance, so that they can pack more transactions into a block
+  and earn more revenue. If a single client is more performant than
+  others, the other clients need to improve their performance to the
+  same level or risk earning less rewards for validators running their
+  software, which will in turn impact their market share.
+
+- Validator clients with below-average performance are incentivized to
```
igor56D, line 41: a below-average validator isn't incentivized to improve performance if the leader still requires their vote to get a proposed block accepted. In fact, I'd argue the *leader* may be incentivized to *slow down* block production in order to not get skipped.

The argument can go both ways, and which direction prevails probably depends on how rewards are distributed across the set of transactions under consideration more than it depends on the dynamics of this resource game between leaders and followers.

igor56D: I think a more formal case can be made for how this may reduce the gap in replay capacity between median validators and tail validators. There are benefits to that, but whether this causes the entire cluster to perform better is not a given.

igor56D (commented):
Thanks for putting this together!

Main concerns:

- uncertainty: how will a leader decide how large to make their block? It isn't sufficient to say this is an implementation detail + leaders are incentivized to do this accurately based on local heuristics. It's highly non trivial and can impact transaction throughput, liveness, etc in significant ways. Given that profitability of transactions is heavily skewed, the profit maximizing strategy may very well be to include only that small, profitable subset rather than risk getting skipped.
- geographic centralization/colocation: timeouts in alpenglow w/ async bound block dissemination time only. Without async, timeouts also bound replay time. This means replay and dissemination consume from the same time budget rather than being considered independently. Because of the uncertainty around packing blocks mentioned above, leaders are incentivized to colocate with other validators in order to reduce the amount of this time budget consumed by dissemination. Also, validators further away will always spend more of the budget on dissemination rather than profitable block packing.
- This proposal is only guaranteed to increase throughput if the existing limits are far below the capacity of the vast majority of the current validator set. This actually probably is true right now, but the case still needs to be made for why removing the explicit limits would be better than just increasing them, as we're currently doing. I'm not convinced that median validators have a strong incentive to get faster, blocks exceeding their replay capacity will always get skipped.

My guess is that this will likely preserve something very similar to the existing CU tracking system because the need to track resource usage in a way that's reasonably reproducible across most of the validators in the cluster doesn't go away. The difference (and potential benefit) is that the tracking doesn't need to be deterministic and there's no enshrined, explicit limits.

I'd expect these resource tracking systems to be even more complicated and arbitrary than what can be achieved with an enshrined resource model. An enshrined model provides a predictable reference point for client implementations, hardware configurations, etc to target/coordinate around. With no reference point, the cluster may drift towards more heterogeneity, with each validator having wildly different performance characteristics. It's not just a matter of fast vs slow, but fast vs slow at which tasks under which circumstances.

My opinion is that there's significant complexity lurking beneath the surface level simplicity of this proposal. More detail would probably help.

igor56D:
This was touched on by @rogerANZA, but how will this proposal affect/be affected by dynamically adjustable alpenglow timeouts? Timeouts are increased/decreased (based on factors like slots observed since last finalization) in order to adapt to changing network conditions.

If explicit limits are removed, the timeout will have the additional use of informing slot capacity during block production. There could be complicated and problematic interplay between these two uses:

- If more slots are being skipped, is it because the timeout is too low or because the leaders have packed too many transactions?
- Should followers respond by adjusting the timeout upwards or should leaders respond by packing fewer txs?
- If the network suffers a temporary disruption leading to increased timeouts, will leaders respond by packing larger blocks after the network recovers, using the new increased timeout?

proposals/0370-remove-compute-unit-block-limit.md
alessandrod: > I have always been a big fan of the idea of removing the compute limit completely, since Alpenglow has no issue with that. (BTW: Since this is an Alpenglow SIMD, I would like to see "Alpenglow" in the title.) Block proposers are incentivized to up their hardware game, which is great!
> 
> However, I have questions:
> 
>     1. Clearly a super-advanced proposer cannot pack its block arbitrarily because it still needs to get the 2x60% or 1x80% approval rating. It might be a difficult game to try to figure out what speed is currently possible. Maybe not impossible, but certainly harder than just greedily packing to the max.

A proposer knows the upper bound to execution time since it executes while packing so this part is easy. Then it needs to have an estimate of how long it takes to propagate the shreds, which doesn't seem like it's something that would be highly variable? This doesn't seem _too_ hard. 

>     2. At some point some nodes may fall behind. Not only in replaying the execution by the leader, but even to catch up with the chain advancing. Now these nodes will drop out. New (beginner, probably also in hardware) nodes cannot even enter the system because they might not be able to catch up from a snapshot. Are we okay with that? Nodes which fall back may have to up their game, but they might also just leave.

This was actually the case for a while when agave sucked, or in the transition between 1.18 and 2.0, and 2.0 and 2.1 since new releases were significantly faster.

Today restarting is much much faster than it used to be - 3 minutes vs 20m+ a year ago - and this is with some known replay inefficiencies that we're fixing right now. I'd say that replay can get ~50% faster than  today once those issues are fixed.

Ultimately the bottleneck in replay is shred propagation. Given ~cluster avg hw, replay is pretty much guaranteed to be significantly faster. 

>     3. If the speed advances during an epoch, we might fall below the 60/80 thresholds, at which point we only get skips, and we basically have to enter Alpenglow's disaster scenarios.

Shorten epochs too 😋 

>     4. Being closer to a leader also helps because you can start executing earlier. So as a side effect, we might get centralization.

Isn't this the case with alpenglow and CUs too, assuming we set CUs close to hw capacity?

> 
> I think all these problems are solvable, so I've always been a big advocate of ditching the limit. But the questions need answers. For instance, it would help tremendously if we made sure that the leader's work (to put together the block) is bigger than the work of replaying/checking by the followers.

For non malicious blocks, I think this is guaranteed to be the case. For malicious blocks, timely replay isn't guaranteed with CUs either.

alessandrod:
> > > There is a ton of impact to validator games that is difficult to model and not contemplated in the proposal.
> > 
> > 
> > Can you say more? This is why we're here :)
> 
> Agree this is why we are here :) Here are a couple examples.
> 
>     * One degenerate equilibrium is the bottom 20% of validators continually get nuked, eventually leading to an effectively very small validator set.

I'm not sure how this happens: in order for it to happen, demand would have to be up only, and software improvements would have to match it. Seems unlikely that demand would grow so fast, and software definitely doesn't improve as fast. Today an agave validator that is too fast and has too many cores performs _worse_ than a validator with fewer cores because of thread contention bugs. 

>     * As validators end up "close to the edge" of profitability / losing independent stake / keeping up with the chain, there is more incentive to colocate with other stake to be able to replay and vote in time, leading to data center centralization.

I don't think this would drive being closer to the edge of profitability. With AG it becomes cheaper to run validators, and current validators are already way over provisioned for the kind of load we get. I don't think we get to more than say 2x of current hw requirements in the foreseeable future.


The cluster already includes good hardware, it's the software that needs to accelerate. Competition between agave and firedancer has already undeniably improved the state of the network. I guess removing CUs provides even more incentive to compete (without playing silly timing games).


>     * Under the no-CU-limit regime, leaders should be more incentivized to build and stream their blocks continuously, rather than waiting to blast their blocks out in the last 50ms of their slot. I believe this is a good thing but it's difficult to forecast the second-order effects.

And this is why I personally care about this change 😋 

> I understand CUs are far from perfect, but at least they are predictable, 

[REDACTED]

> and the limits themselves can also be increased in a predictable manner. That will create much less thrash in validator economics.

My gut feeling says that timing games will become harder, latency will be drop much lower, but not much else will change. But I concede that fixed CUs makes reasoning about some things conceptually easier.

proposals/0370-remove-compute-unit-block-limit.md
yewman: > a below-average validator isn't incentivized to improve performance if the leader still requires their vote to get a proposed block accepted. In fact, I'd argue the _leader_ may be incentivized to _slow down_ block production in order to not get skipped.

I agree. I don't think it is ceratin how the incentives will manifest, especially in the short term.

bw-solana:
Block Limits:

- Overhead of tracking hinders perf
- Constant maintenance (e.g. re-tune based on new HW, large testing burden before increasing limits)
- Overly restrictive for typical case to protect from worst case downside (or exposed to tail case risk)

Timeouts:
- GTO market forces to maximize validator block rewards may not align with what broader Solana community desires:
  - skip rate likely to be higher
  - Incentive is to build blocks where significant portion of stake (~half?) votes skip, delaying finalization latency
- Unpredictable HW requirements (and economics) for validator operators
- Default leader packing heuristic still needs to be maintained and built (probably looks a lot like current block limits to start).  But this heuristic can be iterated on much faster than enshrined limits.
- Individual heuristic mods will be hard to reason about for cluster debug.
- Precludes finalizing blocks before execution (async)

proposals/0370-remove-compute-unit-block-limit.md
rogerANZA: @alessandrod What's the maximum imaginable ratio between the fastest block producer and the slowest block replayer?

rogerANZA:
That brings me back to my question in the other thread: If the fastest leader is not faster than the slowest replayer/follower, then I don't see a problem. The reason to have CU limits is because the fastest leader is actually faster, and then we get all sorts of problems. (Just think of an extended stretch of fast leaders: If they lose <40% of the validators, then it's the slow validators' problem; if they lose >40% of the slow validators, then it's the leaders' problem.)

rogerANZA:
> This was touched on by @rogerANZA, but how will this proposal affect/be affected by dynamically adjustable alpenglow timeouts? Timeouts are increased/decreased (based on factors like slots observed since last finalization) in order to adapt to changing network conditions.
> 
> If explicit limits are removed, the timeout will have the additional use of informing slot capacity during block production. There could be complicated and problematic interplay between these two uses:
> 
> * If more slots are being skipped, is it because the timeout is too low or because the leaders have packed too many transactions?
> * Should followers respond by adjusting the timeout upwards or should leaders respond by packing fewer txs?
> * If the network suffers a temporary disruption leading to increased timeouts, will leaders respond by packing larger blocks after the network recovers, using the new increased timeout?

What you are addressing are dynamic timeouts, which are part of the disaster handling, i.e., if nothing works anymore. Leaders do not have an incentive to go there because they lose all their blocks on the way to there. And as soon as we have a good block (a good leader) we are back to normal. So it takes a lot of time and resources to steer towards the disaster scenario (and the leaders who are doing it lose a lot ... the current white paper suggests 10 seconds), and then it takes a single good leader to get back to normal. I don't think we have to worry about this very much.

igor56D:
> If the fastest leader is not faster than the slowest replayer/follower, then I don't see a problem

You're right, this is straightforward if every leader can safely assume the cluster can always keep up with them. If we fail to hold up this guarantee perfectly then leaders will start self throttling: getting skipped is much much worse than under-packing a block. Also, faster/slower isn't absolute but workload dependent (due to caching, parallelism, etc), which increases the likelihood of violating the "fastest leader <= slowest replayer" condition unexpectedly imo.

I don't think we can ship this without a reliable system for self throttling. Without it, we'd need to somehow ensure we're not too much faster than the other validators in the cluster when we optimize the clients, release new versions, upgrade hardware, etc. CU limits happen to be an existing (imperfect) solution for self throttling, though making the limits dynamic may be worth looking into.

alessandrod:
> That brings me back to my question in the other thread: If the fastest leader is not faster than the slowest replayer/follower, then I don't see a problem. The reason to have CU limits is because the fastest leader is actually faster, and then we get all sorts of problems. (Just think of an extended stretch of fast leaders: If they lose <40% of the validators, then it's the slow validators' problem; if they lose >40% of the slow validators, then it's the leaders' problem.)

Assuming non-malicious blocks, I'd expect the ratio between fastest / slowest _with meaningful amounts of stake_ to be very low 1.x.  Or in other words, there are things that can be very slow in execution for everyone (including leaders), or typical execution is almost insignificant and all the work is waiting for shreds. 

Unfortunately our current metrics make it impossible to get an actual ratio, but we can fix that.

rogerANZA:
> the ratio between fastest / slowest

Yes, figuring out the distribution of this ratio would be super helpful in the discussion.

PR #376, opened 2025-10-07
SIMD-0371: Relaxing Transaction Signature Verification

Rexicon226:


t-nelson:
fwiw we only moved to verify_strict 'cause status cache used to be keyed directly off signature, so the malleability was unacceptable for tx replay protection. it's now keyed off message hash, so we're free to relax

Rexicon226:
Interesting, thanks for the insight.

We should think hard now if there are any usecases in the future which could require non malleable signatures. This proposal can easily be amended to achieve strongly binding signatures, while retaining all of the benefits mentioned. All that's needed is to reject small-order A points, which is extremely cheap to do in this context.

As mentioned in the written proposal, the main reason I chose this equation, one which does not achieve SBS, is because there is an existing, well proven, rust library for Agave to use. Both Sig and Firedancer use from-scratch ed25519 implementations, so there is no difference in which version to use for us. Agave could trivially implement the proposed algorithm with ed25519-dalek primatives, but I don't know if that's something they are interested in doing.

If we will never need SBS, might as well use the well-proven library and more relaxed algorithm. If we might, and would like to not break backwards compatibility with malleated signatures, then we could opt for the slightly stricter algorithm and perhaps losen it later.

t-nelson:
too be clear, status cache using the signature as a key was a bug and vulnerability. no similar use cases dependent upon unique signatures should be permitted so long as we're using ed25519 dsa.

---

one thing we _do_ need to consider is that the first signature on a transaction has become the canonical identifier. there are likely some social games that could be played with malleable signatures. i'm not sure that's something that we need to care too much about though

Rexicon226:
> too be clear, status cache using the signature as a key was a bug and vulnerability. no similar use cases dependent upon unique signatures should be permitted so long as we're using ed25519 dsa.

Agreed.

> one thing we do need to consider is that the first signature on a transaction has become the canonical identifier. there are likely some social games that could be played with malleable signatures. i'm not sure that's something that we need to care too much about though

That's an interesting one! I assume you're talking about things like wallets showing the transaction signature in their view. I feel like this would be more impactful if it were related to the addresses of the accounts themselves, rather than the transaction signatures, as that's where I've seen most "social" attacks target ([a talk I attended a couple months ago](https://arxiv.org/html/2501.16681v1)). I don't want to ramble about speculations though, and I am no expert on the subject, so I would love to hear other people's opinions and some examples. It's always seemed like an interesting attack vector to me.

proposals/0376-verify-strict.md
```diff
@@ -0,0 +1,205 @@
...
+
+- All signatures valid under `verify_strict` remain valid under ZIP-215
+verification.
+- A small class of signatures previously rejected may now be accepted.
+
+This upgrade should be possible to do with two seperate feature gates. The
+first one will be to use both verification equations, and once enough
+stake has migrated, we can then disable the `verify_strict` one through
+another feature gate. This is still up for discussion though, while it
+should definitely be doable, it will require a bit of careful thought on how.
```
samkim-crypto: What would the use of both verification equation precisely mean? Would this result in double the cost for verification before we disable the `verify_strict`?

proposals/0376-verify-strict.md
```diff
@@ -0,0 +1,205 @@
...
+Switch to using the verification equation described in
+[ZIP-215](https://zips.z.cash/zip-0215) for Ed25519 EdDSA signature
+verification.
+
+### Algorithm:
+
+Given a message $M$, public key $A$, signature ephemeral point $R$, and
+scalar $S$.
+
+1. Reject the signature if $`S \notin \{0, ..., L - 1\}`$.
```
samkim-crypto:
```suggestion
1. Reject the signature if $S \notin \{0, ..., L - 1\}$.
```

proposals/0376-verify-strict.md
```diff
@@ -0,0 +1,205 @@
...
+### Application:
+
+This proposal specifically targets usages of `verify_strict`, replacing
+them with the Algorithm described above. This includes replacing the equation
+used for verification of transaction signatures, gossip packet signatures, shred
+packet signatures, and the Ed25519 precompile program.
+
+Section 3.2 of [Taming the many EdDSAs](https://eprint.iacr.org/2020/1244.pdf)
+explains the relationship between batched and single cofactored verifications,
+proving them to be compatible. As a result, they can be used interchangeable,
```
samkim-crypto:
```suggestion
proving them to be compatible. As a result, they can be used interchangeably,
```

samkim-crypto (approved):
I am in favor of this as well. I think relaxing the SBS requirement while maintaining SUF is the right way forward.

proposals/0376-verify-strict.md
Rexicon226: This is probably the roughest part of the proposal, I wasn't really sure how a migration would look like at the time of writing. Now that I've had a bit more time to think about it, is there any issue with just placing it behind a single feature gate? This proposal affects behaviour around weird signatures, ones which honest validators would never be creating anyways.

proposals/0376-verify-strict.md
Rexicon226: Thanks! 83038f2d2f50c0253183afdd2ce00af15426abea

proposals/0376-verify-strict.md
samkim-crypto: Yeah I don't think there is any big issue placing it under a single feature gate. I guess the worry could be that some signatures that are accepted by `verify_strict`, gets rejected by a ZIP-215 implementation. We know that this doesn't happen theoretically, but maybe there are some subtleties in the individual implementations. Perhaps we can have the new verification relation run on a background thread to detect whether we see any such signatures. If we don't see any of these signatures for a certain period for each validator clients, then we can go ahead and activate the single feature gate to transition to ZIP-215.

I'd be interested to hear other people's thoughts. Would be good to have this decided and written in the SIMD.

0x0ece: Agree, one single gate is sufficient as all previous signatures are accepted by the new verify.

It should be very straightforward, since epoch X/block Y, it's ZIP-215 for everything sigverify.

Rexicon226: What would this "verification relation" be? It sounds somewhat similar to what I was discussing in the original text with running them both at the same time.

What do you propose happen if a non backwards compatible signature is found? It's all testnet so I guess it isn't too critical, but it would be nice to not bring down the entire cluster :).

I rewrote this section locally to describe a single feature gate, all that's left any details regarding this idea, if we were to do it.

samkim-crypto: I guess if a non backwards compatible signature is found, then this would mean a previous transaction that was once valid would be invalid under the new scheme and that could haunt us. Mathematically this will never happen, so I guess as long as we have good testing, then it should be fine. I think live double verification is not actually necessary. This would only be activated after thorough testing of all sig verify logic in each of the validator clients are consistent as well. Let's specify single feature gate in the simd and I think this should be fine to merge on my side.

Rexicon226: Understood.

Rexicon226: ce91b1e748cb907c61e3260e4437727ea3c1dc87

proposals/0376-verify-strict.md
```diff
@@ -136,11 +136,11 @@ which ZIP-215 achieves by rejecting $S$ scalars which do not fit into $l$.
...
 
-This upgrade should be possible to do with two seperate feature gates. The
-first one will be to use both verification equations, and once enough
-stake has migrated, we can then disable the `verify_strict` one through
-another feature gate. This is still up for discussion though, while it
-should definitely be doable, it will require a bit of careful thought on how.
+This upgrade will require one feature gate. Once this feature gate is active,
+ZIP-215 will be used instead of `verify_strict` for all EdDSA signature
+verifications. Even though the equation is theoretically backwards compatible,
+there could be unfoseen incompatibilities in the `ed25519-zebra` library,
```
samkim-crypto: unfoseen --> unforeseen

Also, the feature gate is not necessarily for the backwards compatibility of `ed25519-zebra`. We need some feature gate coordination here since want ZIP-315 to be active on all validator clients at the same time. If a ZIP-315 signature is invalid in agave v3, but valid in v4, that can create a fork. Similarly if a ZIP-315 compliant signature is valid on agave, but invalid in fd, then that can create a fork. We need some coordination using a feature gate.

Rexicon226: Ah, thanks for the clarification. Still new to this feature gates thing :P. 
I will just remove the extra details, I feel they're implied.

Rexicon226: Is this good? ba7adc6027f7e17393450dcb90eca78343968067

samkim-crypto: Looks good thanks!

samkim-crypto:
@Rexicon226 I think we should refine how we frame the security trade-offs. Currently, the proposal implies that we might care less about SBS, but technically zip-215 preserves both SUF and SBS for all properly generated keys. 

The nuance is that for weak keys (torsion points), zip-215 actually breaks both properties (whereas `dalek-strict` maintains them by banning those keys entirely).

I think we should explicitly mention that we are okay with this trade-off because:
1. No private keys: Weak keys are just torsion points and lack a proper private scalar. More importantly, because the verification equation loosens for these keys, they cannot secure funds (signature can be forged by anyone). Therefore, no honest user would ever use then, and they effectively serve only as spam/attacker addresses.
2. Quarantined malleability: It is true that signatures become malleable on weak keys (allowing non-unique tx ids), but this behavior is isolated entirely to these specific attacker-controlled accounts. Since honest wallets always generate prime-order keys, a regular user's TXID is always unique. 
  While dumb apps might get confused by the duplicate TXIDs, this is acceptable because the confusion is self-inflicted and contained. The attacker is effectively paying network fees to corrupt the indexing of their own spam history. Since the collision never involves an honest user's TXID, no user's data is ever at risk.

(This blog post has a great breakdown of the sig verification variants for reference: https://hdevalence.ca/blog/2020-10-04-its-25519am/)"

t-nelson (approved):
i'm fine with this as is

@deanmlittle any disrespect due here?

deanmlittle:
> i'm fine with this as is
> 
> 
> 
> @deanmlittle any disrespect due here?

I agree with point #1 raised by @samkim-crypto, wrt to #2, signature malleability is already ~infinite by simply using non-canonical randomness in the EdDSA signing process so I don't see this as introducing any new malleability vector into the protocol. Lgtm.

PR #377, opened 2025-10-09
SIMD-0377: eBPF ISA compatibility

LucasSte:


proposals/0377-ebpf-isa-compatibility.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Alternatives Considered
+
+We have considered diverging from the eBPF standard by introducing new opcodes 
+and creating specific instructions to the Solana environment. We discarded 
+such an approach to be compatible with the existing LLVM eBPF code generation.
+
+Another consideration was bringing new eBPFv4 instructions to the Solana 
+environment. They are a superset of eBPFv3 and does not conflict with it, 
+however they are not yet stabilized, so they can change in future LLVM releaes.
```
mschneider: What do you mean by "not yet stabilized"? Why do llvm releases even matter? Isn't the linux-kernel version the actual definition of the isa that should be relevant for the definition of the solana runtime?

LucasSte: We have relied on LLVM for code generation for some years now. The LLVM eBPF target (and the GCC target as well) brings in some instructions that are not yet fully adopted by the kernel, and not necessarily appear in the documentation.

The oldest example is `callx` (indirect calls). We have supported `callx` with the LLVM encoding since Solana started, but LLVM has changed its encoding for `callx` recently (see https://github.com/llvm/llvm-project/commit/c43ad6c0fddac0bbed5e881801dd2bc2f9eeba2d), rendering it incompatible with our VM.

What I wanted to point out was to avoid experimenting with v4 before it is officially supported.

proposals/0377-ebpf-isa-compatibility.md
```diff
@@ -0,0 +1,182 @@
...
+
+SUB32_IMM and SUB64_IMM must perform the operation `src = src - imm`.
+
+### Dynamic stack frames
+
+Aiming a closer compatibility to eBPF, the implementation of dynamic stack 
+frames is going to change.
+
+The R10 register must continue to be the frame pointer, i.e. pointing to the 
+highest address accessible in a function. As such, the stack will grow upwards.
```
ripatel-fd: Isn't it pointing one byte past the highest addressable byte?

ripatel-fd (approved):

proposals/0377-ebpf-isa-compatibility.md
```diff
@@ -0,0 +1,125 @@
...
+
+The virtual machine must initialize the frame pointer `R10` at 4096 bytes to 
+allow stack space for the entry function. After that, the it must increment 
+the frame pointer in 4096 at each internall call (`callx` or `call` with 
+source field set to one).
+
+Functions may optionally include an instruction `add64 r10, imm` on their
```
alessandrod: I'm struggling to understand what's the point of this. The original dynamic frames design was to avoid having to preallocate/memset 4k * MAX_DEPTH. Looks like this design still requires that and adds extra complexity?

proposals/0377-ebpf-isa-compatibility.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Motivation
+
+The eBPF target on the Rust compiler emits code by default for eBPFv1, whose 
+only incompatibility with the Solana virtual machine is the `callx` 
+instruction. Aiming to prioritize Solana programs and decrease their CU 
+consumption, we want to be compatible with at least the current eBPF version 
+(v3), which brings in new instructions. In order for that to be possible, we 
+must modify our virtual machine to support eBPF integrally.
+
```
deanmlittle, line 31:
```suggestion

## Dependencies

This proposal depends on the following previously accepted proposals:

- **[SIMD-0178]: SBPF Static Syscalls**
- **[SIMD-0189]: SBPF stricter ELF headers**

    This SIMD requires static syscalls which requires stricter ELF headers.

[SIMD-0178]: https://github.com/solana-foundation/solana-improvement-documents/pull/178
[SIMD-0189]: https://github.com/solana-foundation/solana-improvement-documents/pull/189

```

I think we should make it explicit that we need the above two SIMDs to activate first in order for this one to be activated.

proposals/0377-ebpf-isa-compatibility.md
```diff
@@ -0,0 +1,125 @@
...
+source field set to one).
+
+Functions may optionally include an instruction `add64 r10, imm` on their 
+prologue to increment the frame pointer relative to the already supplied 4096 
+bytes, providing `4096 + imm` bytes of frame space for them.
+
+When a function returns, the virtual machine must automatically restore the 
+frame pointer register with the value used in the caller, so programs do not 
+need to emit any instruction to adjust the pointer in the epilogue of each 
+function.
```
deanmlittle:
```suggestion
```

As this is unrelated to upstream compatibility and the implementation remains a point of contention, I think it would be best to split this into a second SIMD. As an opt-in feature, there's no reason we need to version for this, it's not critical to shipping SBPFv3 and would be pretty trivial to activate at a later date if we agree this is the way we want to implement dynamic stack frames in the future.

LucasSte: > As an opt-in feature, there's no reason we need to version for this, it's not critical to shipping SBPFv3 and would be pretty trivial to activate at a later date if we agree this is the way we want to implement dynamic stack frames in the future.

Both LimeChain and JetBrains are developing tooling that relies on dynamic stack frames being enabled, so it would be unwise to remove the feature, as it would break their work.

> the implementation remains a point of contention

What is your concern about this approach?

proposals/0377-ebpf-isa-compatibility.md
LucasSte: The idea behind this approach is to provide dynamic stack frames without breaking compatibility with upstream eBPF as it is right now. If eBPF LLVM ever supports dynamic stack frames, but in a different way, we can change our implementation to match what has been adopted there.

> The original dynamic frames design was to avoid having to preallocate/memset 4k * MAX_DEPTH.

The implementation we have right now with the original design (sbpfv1) still allocates the same amount of memory. Is your idea for users to request in the transaction how much stack space they want? If that is the case, this proposal still allows this, since `add64 r10, imm` can also decrease the frame size of a function, if the immediate value is negative.

LucasSte: I also wanted to remain conservative on the design to be honest. So far, the LLVM maintainers have had some consensus on the features, but not on the implementation. I think it is fair to assume that whatever option we adopt for dynamic stack frames right now may not be the same one adopted upstream in the future.

proposals/0377-ebpf-isa-compatibility.md
Lichtso: > Both LimeChain and JetBrains are developing tooling that relies on dynamic stack frames being enabled, so it would be unwise to remove the feature, as it would break their work.

We can increase the default stack frame size in debug mode to solve that.

deanmlittle: I'm not saying "remove this feature and throw it in the trash", it's clear you think it's a reasonable approach. I'm saying this has absolutely nothing to do with `eBPF ISA Compatibility`. If you split this out into a separate SIMD, we can avoid derailing shipping this SIMD on time and direct whatever criticisms we have of the dynamic stack frame design to the new SIMD. I don't think they should be bundled into one proposal like this, especially when it would do no harm to introduce this feature later if we decided we wanted it.

proposals/0377-ebpf-isa-compatibility.md
deanmlittle: Since this been split out from the SIMD, we can resolve this.

PR #384, opened 2025-10-22
SIMD-0384: Alpenglow Migration

carllin:


proposals/0371-alpenglow-migration.md
```diff
@@ -0,0 +1,203 @@
+---
+simd: '0371'
```
AshwinSekar:
```suggestion
simd: '0384'
```
they want us to match the PR #

proposals/0371-alpenglow-migration.md
```diff
@@ -0,0 +1,203 @@
...
+   as to avoid the beginning of an epoch. After slot `S`:
+   - Turn off packing user transactions in blocks. Any transactions not
+     belonging to the vote program will cause a block to be marked dead during
+     replay by all correct validators.
+   - Turn off optimistic confirmation/commitment reporting over RPC.
+   - Turn off rooting blocks in TowerBFT to prevent losing optimistically
+     confirmed blocks that could qualify as the Alpenglow genesis blocks.
+
+2. After the migration boundary slot `S`, wait for some block `B > S` to be
+   `strongly optimistically confirmed` with `>=82%` of the stake voting.
```
AshwinSekar: nit: maybe we drop this in the `New Terminology` section above

proposals/0371-alpenglow-migration.md
AshwinSekar: Also edit the title of this PR

proposals/0371-alpenglow-migration.md
```diff
@@ -0,0 +1,203 @@
...
+
+5. Anytime a correct validator receives a `genesis certificate` for a slot `B`
+   (either constructed themselves, received through replaying a block, or received
+   from all-to-all broadcast), they:
+   - Broadcast the certificate to all other validators via the Alpenglow
+     all-to-all mechanism, which guarantees delivery system via its retry
+     mechanism.
+   - We initialize the alpenglow `votor` module with `G` as genesis, and disable
+     TowerBFT for any slots past `G`
+   - In block production pack the `genesis certificate` in the headers of any
```
AshwinSekar: Could we specify the `BlockMarker` and the layout here? I don't think it needs to be in the header as it's only used for first alpenglow block(s), easier to make it a separate marker.

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,261 @@
...
+    Specifies `BlockMarker`, an means of disseminating metadata in a block
+
+
+## New Terminology
+
+Strong optimistic confirmation:
+
+- A block `B` is strong OC if there exists a block `T` such that `B` is the parent
+  of `T` and `slot(B) + 1 = slot(T)` and `T` contains vote transactions for `B` from
+  at least `82%` of stake.
```
wen-coding, line 62: nit: I assume the assumption here is we don't do this while crossing Epoch boundary to make the stake calculation complicated, but we don't have to specify that here.

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,261 @@
...
+
+## Detailed Design
+
+### Migration Handoff
+
+1. Pick a migration boundary slot `S` as follows. Let `X` be the rooted slot in
+   which the feature flag is activated. Let the migration slot S be `X + 5000`,
```
wen-coding: nit: this sounds like a suggestion, I believe we've made a decision on this value being 5000 already?

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,261 @@
...
+   as to avoid the beginning of an epoch. After slot `S`:
+   - Turn off packing user transactions in blocks. Any transactions not
```
wen-coding: nit: do you mean "user transactions" here or do you mean "non-vote transactions", I believe they are roughly equivalent, but non-vote might be more precise?

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,261 @@
...
+     belonging to the vote program will cause a block to be marked dead during
+     replay by all correct validators.
+   - Turn off optimistic confirmation/commitment reporting over RPC.
```
wen-coding: nit: "Turn off" is a bit vague, do you mean these RPC will refuse to answer or they will return an error code? Might be good to be specific so users can prepare.

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,261 @@
...
+   - Turn off rooting blocks in TowerBFT to prevent losing optimistically
+     confirmed blocks that could qualify as the Alpenglow genesis blocks.
```
wen-coding: nit: could we tell users what "Turn off" means here? What would they observe differently. Would that mean `TowerSync` containing more than 32 votes?

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,261 @@
...
+
+2. After the migration boundary slot `S`, wait for some block `B > S` to reach
+   strong optimistic confirmation.
```
wen-coding, line 73: nit: I'm guessing "wait for" here means every validator actively searching for? Otherwise it's not clear who sends the signal when validators wait for the block.

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,261 @@
...
+
+3. Find the latest ancestor block `G` of `B` from before the migration boundary
```
wen-coding: nit: I think this SIMD might be easier to read if we define `G` and `S` in New Terminologies section.

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,261 @@
...
+   slot `S`. Cast a BLS vote (the genesis vote) for `G` via all to all. Note
+   validators should have filled out their BLS keys prior to the feature flag
+   activation, and will sign this genesis vote with this BLS key.
```
wen-coding: I think we will have SIMD 357 activated at least an Epoch before the migration, so we can kick out all validators without BLS pubkeys from staked validators.

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,261 @@
...
+   Alpenglow. Validators will periodically refresh genesis votes until this
+   `genesis certificate` is observed. During this period they perform regular
+   TowerBFT consensus for all blocks.
+
+5. Anytime a correct validator receives a `genesis certificate` for a slot `G`
+   (either constructed themselves, received through replaying a block, or received
+   from all-to-all broadcast), they:
+   - Broadcast the certificate to all other validators via the Alpenglow
+     all-to-all mechanism, which guarantees delivery system via its retry
+     mechanism.
```
wen-coding: nit: "which guarantees delivery system via its retry mechanism", what does this mean?

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,261 @@
...
+   - We initialize the alpenglow `votor` module with `G` as genesis, and disable
+     TowerBFT for any slots past `G`
+   - In block production pack the `genesis certificate` as a `GenesisBlockMarker`
+     for any blocks that are *direct* children of `G`. This means anybody 
+     replaying any of the initial Alpenglow blocks must see the `genesis certificate`.
+   - Delete all blocks and shreds with slot greater than `slot(G)`.
```
wen-coding: nit: maybe mention Delete from where?

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,261 @@
...
+   - We exit vote only mode, enable Alpenglow rooting, and re-enable RPC
```
wen-coding: nit: would be nice to have "enter vote only mode" somewhere above

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,261 @@
...
+   *replaying* the header of a block, they store the certificate in a
+   `migration success` off-curve account 
+   `Pubkey::find_program_address(&["carlgration"], alpenglow::id())`. 
+   This means all snapshots descended from the block will contain this account 
+   and signal to validators that they should initiate Alpenglow after unpacking 
+   the snapshot.
+
+7. Alternatively, anytime a correct validator that has not yet detected a
+   `genesis certificate`, but receives an Alpenglow finalization certificate for
+   some block `X`, they should repair/replay all the ancestors of `X`
```
wen-coding: Hmm, if I'm an attacker, can I just spam everyone with a fake Alpenglow finalization certificate in the future (e.g. 2 epochs later) and make them stuck in the repair/replay loop? Might be good to specify we only accept blocks in the current Epoch so we can verify the BLS signatures?

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,261 @@
...
+
+8. Once an Alpenglow finalization certificate is received via all-to-all or via
+   replaying a block, validators can stop broadcasting the genesis certificate
```
wen-coding, line 122: oh, we do manually retry? Can we specify the time interval of certificate broadcast retry somewhere?

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,261 @@
...
+
+This `63%` correct validators is then sufficient to run Alpenglow and produce
+a finalized Alpenglow block, which will induce a repair/transition from any
+other correct/lagging validators.
+
+Thus from 1. we know some correct validator must eventually migrate because
+eventually they must receive a genesis certificate, then we know from 2 that
+eventually all correct validators must migrate.
+
+For paper, see here.
```
wen-coding: Are you supposed to give a link here?

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,261 @@
...
+block.
+
+This means that eventually 82% of validators will cast a genesis vote for the
+same genesis block. Because these genesis votes are reliably delivered via
+all-to-all, some correct validator will eventually get a genesis certificate.
+
+2. Next we show that once a correct validator migrates, then all correct
+   validators will eventually migrate.
+
+There's two ways for correct validators to migrate:
```
wen-coding: nit: There are ?

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,261 @@
...
+
+1. A genesis certificate
+2. An Alpenglow finalization certificate
+
+The first correct validator to migrate must have gotten a `82%` genesis
+certificate. We assume at most `19%` of the cluster is Byzantine, then at least
+this correct validator will continuosly broadcast the genesis certificate until
+they see an Alpenglow finalization certificate. 
+
+This means eventually `82% - 19% = 63%` correct nodes will eventually receive
```
wen-coding: nit: remove one eventually ?

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,261 @@
...
+   will still think it has reached `bank.max_tick_height()`. This last tick is
+   necessary to coordinate with banking stage and broadcast to properly end the
+   packing/dispersion of a block. Eliminating it is possible, but a load of
+   risky work.
+3. Change `blockstore_processor::verify_ticks()` to turn off tick verification.
+
+### Duplicate block handling
+
+1. Turn off tower duplicate block handling
+2. Turn off epoch slots
```
wen-coding, line 265: When do we turn this off? At the first Alpenglow block?

proposals/0384-alpenglow-migration.md
carllin: yeah we bring this up later where we say "Let the migration slot S be `X + 5000`,
   as to avoid the beginning of an epoch. After slot `S`:"

proposals/0384-alpenglow-migration.md
carllin: I believe right now in the code `MIGRATION_SLOT_OFFSET = 32`, we can change this though

proposals/0384-alpenglow-migration.md
carllin: yeah we mean simple vote transactions, will update

proposals/0384-alpenglow-migration.md
carllin: we mean the core code won't even notify the RPC services of any updates, will clarify

proposals/0384-alpenglow-migration.md
carllin: turn off means no roots will be set in the core code. TowerSync can never contain more than 32 votes because it's implemented as a fixed length queue, but the root distance will grow indefinitely until migration succeeds

proposals/0384-alpenglow-migration.md
carllin: the signal is the strong optimistic confirmation which is detected like any optimistically confirmed block

proposals/0384-alpenglow-migration.md
carllin: edit nvm, was looking at tests, it's already correctly set https://github.com/anza-xyz/alpenglow/blob/5e1f448123da6821ea062f85c15d549969185f10/votor-messages/src/migration.rs#L65-L70

proposals/0384-alpenglow-migration.md
carllin: updated!

proposals/0384-alpenglow-migration.md
carllin: updated!

proposals/0384-alpenglow-migration.md
carllin: it's retried via the same 10 second standstill timer in the certificate pool, will update

proposals/0384-alpenglow-migration.md
carllin: changed to remove vote only mode, we just start packing non-vote transactions again

proposals/0384-alpenglow-migration.md
carllin: removed link

proposals/0384-alpenglow-migration.md
carllin: yup, updated

proposals/0384-alpenglow-migration.md
carllin: it's retried via the same 10 second standstill timer in the certificate pool, will update

proposals/0384-alpenglow-migration.md
carllin: yeah you would only accept Alpenglow finalization certificates you can verify.

And verification implies knowing the validator set, so it must be at most one epoch ahead

carllin: updated to specify you must be able to verify the certificate

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,282 @@
...
+
+    Specifies `BlockMarker`, an means of disseminating metadata in a block
+
+
+## New Terminology
+
+Migration boundary slot:
+
+- Slot at which Alpenglow migration begins
+- After slot this boundary slot:
```
wen-coding: nit: this sentence sounds weird

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,282 @@
...
+4. If we observe `>=82%` genesis votes for the ancestor block `G`, this
+   consitutes the `genesis certificate`, and `G` is the genesis block for
+   Alpenglow. Validators will periodically refresh genesis votes every
+   `GENESIS_VOTE_REFRESH` = 400ms (i.e. once a slot) until this
+   `genesis certificate` is observed. During this period they perform regular
+   TowerBFT consensus for all blocks.
+
+5. Anytime a correct validator receives a `genesis certificate` for a slot `G`
+   (either constructed themselves, received through replaying a block, or received
+   from all-to-all broadcast), they:
```
wen-coding, line 90: You didn't say but I suppose the first thing they will do (if they received it rather than constructed it) would be verify the certificate is valid. (And do we need to list validity checks?)

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,282 @@
...
+#### Liveness argument
+
+1. First we show that eventually at least one correct validator should see an
+   Alpenglow genesis certificate.
+
+We assume the the cluster must eventually run under normal network conditions,
+so blocks past the migration boundary slot `S` should be `strongly optimistically
+confirmed`.
+
+Next, Until a migration certificate is observed, no correct validators will
```
wen-coding:
```suggestion
Next, until a migration certificate is observed, no correct validators will
```

proposals/0384-alpenglow-migration.md
carllin: updated!

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,283 @@
...
+   consitutes the `genesis certificate`, and `G` is the genesis block for
+   Alpenglow. Validators will periodically refresh genesis votes every
+   `GENESIS_VOTE_REFRESH` = 400ms (i.e. once a slot) until this
+   `genesis certificate` is observed. During this period they perform regular
+   TowerBFT consensus for all blocks.
+
+5. Anytime a correct validator receives a `genesis certificate` for a slot `G`
+   (either constructed themselves, received through replaying a block, or received
+   from all-to-all broadcast), they:
+   - Verify the certificate against the BLS keys fr the epoch.
```
wen-coding: nit: fr -> for?

wen-coding (approved):

lidatong (dismissed):

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,283 @@
...
+
+This proposal depends on the following accepted proposal:
+
+- **[SIMD-0326]: Alpenglow**
+
+    Requires Alpenglow to be implemented in order to migrate
+
+- **[SIMD-0307]: Add Block Footer**
+
+    Specifies `BlockMarker`, an means of disseminating metadata in a block
```
bw-solana: `s/an/a/`

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,283 @@
...
+
+
+## New Terminology
+
+Migration boundary slot:
+
+- Slot at which Alpenglow migration begins
+- After this boundary slot:
```
bw-solana: can we specify inclusive or exclusive?

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,283 @@
...
+- A block `B` is strong OC if there exists a block `T` such that `B` is the parent
+  of `T` and `slot(B) + 1 = slot(T)` and `T` contains vote transactions for `B` from
+  at least `82%` of stake.
+
+## Detailed Design
+
+### Migration Handoff
+
+1. Pick a "migration boundary slot" (defined above) `S` as follows. Let `X` be
+the rooted slot in
```
bw-solana: maybe specify first* rooted slot?

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,283 @@
...
+   which the feature flag is activated. Let the migration slot S be `X + 5000`,
+   as to avoid the beginning of an epoch.
+
+2. After the migration boundary slot `S`, wait for some block `B > S` to reach
+   strong optimistic confirmation.
+
+3. Find the latest ancestor block `G` of `B` from before the migration boundary
```
bw-solana: nit: `s/latest/most recent/`

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,283 @@
...
+   TowerBFT consensus for all blocks.
+
+5. Anytime a correct validator receives a `genesis certificate` for a slot `G`
+   (either constructed themselves, received through replaying a block, or received
+   from all-to-all broadcast), they:
+   - Verify the certificate against the BLS keys for the epoch.
+   - Broadcast the certificate to all other validators via the Alpenglow
+     all-to-all mechanism. Validators will continually retry broadcasting this
+     certificate every 10 seconds via the certificate pool standstill timer so
+     long as a finalized Alpenglow certificate for a higher slot isn't detected.
```
bw-solana, line 95: nit: `s/so long as/until/` ?

bw-solana: ~is there a chance some small percentage of stake doesn't see the genesis cert for whatever reason (network issue maybe) before the rest of the cluster sees a finalize cert for slot `<Alpenglow Gensis Slot> + 1` ?~

~This would mean the rest of the cluster stops broadcasting the genesis cert. In theory this node will see the new Alpenglow finalization cert, but will it be able to proceed without seeing the genesis cert? Is there some means of reconstructing this cert? I seem to recall leader must stick this cert in the first alpenglow block or some such~

Edit: This is answered later

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,283 @@
...
+   - We initialize the alpenglow `votor` module with `G` as genesis, and disable
+     TowerBFT for any slots past `G`
+   - In block production pack the `genesis certificate` as a `GenesisBlockMarker`
+     for any blocks that are *direct* children of `G`. This means anybody
+     replaying any of the initial Alpenglow blocks must see the
+     `genesis certificate`.
+   - Delete all blocks and shreds with slot greater than `slot(G)` from
+   blockstore, and reset all associated state (replay, AccountsDB) to block
+   `G`, as is currently done when we rollback duplicate blocks.
```
bw-solana, line 104: what if someone doesn't and now holds duplicate shred slot/index with differing data.. can they generate a duplicate proof?

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,283 @@
...
++---------------------------------------+
+| Genesis Slot                (8 bytes) |
++---------------------------------------+
+| Genesis Block ID           (32 bytes) |
++---------------------------------------+
+| BLS Signature             (192 bytes) |
++---------------------------------------+
+| Validator bitmap length     (8 bytes) |
++---------------------------------------+
+| Validator bitmap      (max 512 bytes) |
```
bw-solana, line 149: why max 512B?

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,283 @@
...
+be done in a few steps to mitigate the amount of code changes:
+
+1. Before the end of each Alpenglow block, set the bank tick height to
+   `bank.max_tick_height() - 1`
+2. Change tick producer to only make 1 ending tick per block, so that each bank
+   will still think it has reached `bank.max_tick_height()`. This last tick is
+   necessary to coordinate with banking stage and broadcast to properly end the
+   packing/dispersion of a block. Eliminating it is possible, but a load of
+   risky work.
+3. Change `blockstore_processor::verify_ticks()` to turn off tick verification.
```
bw-solana, line 258: most of this seems implementation specific. I think the important part is #2 (only put 1 tick at end of block). We should probably specify how many hashes for this tick

proposals/0384-alpenglow-migration.md
```diff
@@ -0,0 +1,283 @@
...
+## Alternatives Considered
+
+Alpenswitch where we pick fixed slot intervals `N` at which to attempt to
+optimistically migrate to Alpenglow. On failure fallback to TowerBFT we try
+again at the next slot interval. This is more painful to implement because
+of the transition back and forth between Alpenglow and TowerBFT
+
+## Impact
+
+Validators will run Alpenglow after the migration boundary
```
bw-solana: might be nice to spell out what is expected as far as liveness impact here. In the happy case, I believe we expect ~1 slot and a consensus round worth of lost liveness.

Of course, malicious or offline block producers could extend this, but that's no different than normal runtime.

proposals/0384-alpenglow-migration.md
carllin: clarified to be the feature activation slot

proposals/0384-alpenglow-migration.md
carllin: I think its because we have at most 2000 validators right? @AshwinSekar

proposals/0384-alpenglow-migration.md
carllin: clarified it's only one hash because its equivalent to setting poh to the low power mode tick producer

proposals/0384-alpenglow-migration.md
carllin: So if somebody sends a duplicate proof to a validator:

1. After the validator detected a certificate of successful alpenglow migration, then the validator should already turned off all TowerBFT duplicate handling so that proof should be ignored. (confirming with @AshwinSekar here there isn't a race between turning off TowerBFT duplicate block handling and alpenglow certificate detection). 
2. Before this validator has a certificate of successful alpenglow migration, but some correct validator in the cluster has put together a migration certificate. Then we should repair the block for the migration certificate.
3. Before this validator has a certificate of successful alpenglow migration, but no correct validator in the cluster has put together a migration certificate. Then we handle this like a regular TowerBFT duplicate block

@AshwinSekar I think only tricky case is 2?

AshwinSekar: yeah 2 is fine and is handled once we see the genesis certificate or a future finalization certificate.

From the slashing angle, yes unless we want to bump shred version here there will be a duplicate proof for all blocks in the migrationary window. Ideally this should just be one block. Luckily I dropped the ball on SIMD-0204 so there shouldn't be any impact.

proposals/0384-alpenglow-migration.md
AshwinSekar: This is the maximum certificate size currently supported, corresponds to 4096 validators.

wen-coding: We will limit the number of validators at 2000 per SIMD 357, so you can limit it at 2000/8=250 bytes. That said, this is an one-time certificate, so we probably don't care that much.

proposals/0384-alpenglow-migration.md
bw-solana: sounds good. I suppose once we implement SIMD-0204, we can think through complexity of excluding proofs in this migration range

AshwinSekar: implementation only has a 1 epoch lookback, so as long as we activate 204 at least a couple epochs post alpenglow we should be fine

lidatong (approved):

PR #296, opened 2025-06-05
SIMD: Increase Transaction Size Limit

jacobcreech:
Today's transaction size limit is capped at 1232 bytes for the payload, significantly limiting some of the usecases that developers on Solana can do.

This proposal raises that limit to 4kb, with potential for raising higher on a new transaction format.

PR #385, opened 2025-10-25
SIMD-0385: Transaction V1 Format

jacobcreech:
I have separated the Transaction V1 format from the larger transaction size SIMD #296. Both of these SIMDs would eventually have the same feature gate if they are accepted.

### DIFF
--- a/proposals/0385-transaction-v1.md
+++ b/proposals/0385-transaction-v1.md
@@ -53,9 +53,9 @@ sizes.
 ```
 VersionByte (u8)
 LegacyHeader (u8, u8, u8) 
-NumInstructions (u8)
 TransactionConfigMask (u32) -- Bitmask of which config requests are present.
 LifetimeSpecifier [u8; 32]
+NumInstructions (u8)
 NumAddresses (u8)
 Addresses [[u8; 32]] -- Length matches NumAddresses
 ConfigValues [[u8; 4]] -- Length equal to the popcount (number of set bits)


proposals/0385-transaction-v1.md
```diff
@@ -0,0 +1,262 @@
...
+unable to support developer features that are required for today's applications.
+
+This proposal presents a new transaction format, `v1`, that is designed to 
+get rid of the need for compute budget instructions, address lookup tables, 
+and more to speed up transaction ingestion and processing.
+
+## Motivation
+
+The current live transaction formats, v0 and legacy, have a number of 
+inefficiencies and limitations that adversely impace the performance of the
```
apfitzge: impace -> impact

apfitzge (dismissed):

proposals/0385-transaction-v1.md
```diff
@@ -0,0 +1,262 @@
...
+By introducing a new transaction format, `v1`, we can address these
+issues and improve the performance of transactions ingestion, processing, and 
+save developers space in their transaction by including compute budget 
+information in the transaction header itself.
+
+## New Terminology
+
+- 'v1 transaction' - A new transaction format that is designed to enable larger
+  transactions sizes while not having the address lookup table features 
+  introduced in v0 transactions.
```
tao-stones: Also mention `v1` does not need (or support) compute budget instructions?

proposals/0385-transaction-v1.md
```diff
@@ -0,0 +1,262 @@
...
+  of which the first is the fee payer
+- `num_readonly_signed_accounts` addresses for which the transaction contains
+  signatures and are loaded as readonly
+- `num_addresses-num_required_signatures-num_readonly_unsigned_accounts`
+  addresses for which the transaction does not contain signatures and are
+  loaded as writable
+- `num_readonly_unsigned_accounts` addresses for which the transaction does
+  not contain signatures and are loaded as readonly
+
+Any section with 0 addresses is skipped.
```
tao-stones: nit - the sentence is a bit ambiguous since both "section" and "skipped" aren’t clearly defined. It may be clearer to just remove it.

proposals/0385-transaction-v1.md
```diff
@@ -0,0 +1,262 @@
...
+The instructions will still consume compute-units if included and be processed 
+as a successful no-op instruction.
+
+For the cost-model, all TxV1 transactions are treated as if requests are 
+present.
+For all fields, if the bit(s) are not set, the minimum allowed value is used:
+
+- If bits [0, 1] are not set, the priority-fee is 0.
+- If bit 2 is not set, the requested compute-unit-limit is 0.
+- If bit 3 is not set the requested accounts data size limit is 0.
```
tao-stones: note - in v0/legacy, the default of this is "MAX_LOADED_ACCOUNTS_DATA_SIZE_BYTES". Setting it to `0` in `v1` may be intentional, just want to pointing it out.

proposals/0385-transaction-v1.md
```diff
@@ -0,0 +1,262 @@
...
+- [0, 1] - total lamports for transaction priority-fee. 8-byte LE u64.
+- [2] - compute-unit-limit. 4 byte LE u32.
+- [3] - requested loaded accounts data size limit. 4 byte LE u32.
+- [4] - requested heap size. 4 byte LE u32.
+
+For 2 bit fields, such as priority-fee, both bits MUST be set. If only one of 
+the bits is set, the transaction is invalid and cannot be included in blocks.
+
+For TxV1 transactions, any ComputeBudgetProgram instructions are ignored for 
+configuration, even if they are invalid.
```
tao-stones, line 233: Curious — why ignore the ComputeBudget instruction rather than reject the transaction outright?

proposals/0385-transaction-v1.md
```diff
@@ -0,0 +1,262 @@
...
+The transaction must not contain trailing data after the signatures field.
+There are no padding fields.
+
+This new `v1` transaction format notably does not include address lookup
+tables.
+
+### Transaction Constraints
+
+Any transaction violating these constraints will be considered invalid and will 
+not be included in the chain. Violations are considered sanitization failures:
```
tao-stones, line 201: nit - “invalid” and “sanitization failures” are mentioned in several places. It might help to consolidate all validation criteria into a single section for easier reference.

proposals/0385-transaction-v1.md
apfitzge: it's intentional but yes it is different so probably best to call that out

proposals/0385-transaction-v1.md
apfitzge: then i'd still have to scan over the ixs to validate 😄

tao-stones: makes sense. it could be a runtime error if txv1 includes ComputeBudget program. This can happen when  CB program is deprecated.

proposals/0385-transaction-v1.md
tao-stones: Compiled the sanitization rules below, do I miss anything?
- Version must be 129.
- NumInstructions <= 64.
- NumAddresses <= 64 and >= num_required_signatures + num_readonly_unsigned_accounts.
- Signature count matches num_required_signatures and is capped (<= 12).
- No duplicate addresses.
- Each instruction account index must be < num_addresses.
- No trailing data after signatures.
- Transaction total size cap (<= 4096).

robpal1990:
While the idea of increasing the transaction size to 4kB is great, in the absence of ALTs the limit of 64 accounts is way too low. Current limit w/o ALTs is 35, why bump the tx size 4x but the account number by less than 2x?

This is **especially** important for DEX aggregation. A single DEX swap on average touches 10+ accounts (this varies between DEX-es, some are at 8, some at 15+) and the current way of doing things is to shove all pool-specific accounts (token accounts, accounts holding DEX state, configs, token programs etc.) into one ALT and greatly reduce the amount of data and accounts used. 

If I understand correctly, current cap is at 256 accounts touched this way (https://docs.anza.xyz/proposals/versioned-transactions/?utm_source=chatgpt.com#limitations) and with other limitations -- 1.4M CU and 1232 bytes tx size -- this allows for ~10 swaps per transaction. Having a limit at 64 greatly reduces possible route space, realistically to 4-5 swaps per tx. This will result in worse prices for all people swapping on Solana.

I suggest bump the account limit to 128. This is still not great, but way less limiting than a hard cap at 64.

tao-stones:
> I suggest bump the account limit to 128.

There is [feature gate](https://github.com/solana-labs/solana/issues/27241) to bump limit to 128.

proposals/0385-transaction-v1.md
0xalpharush: I think technically the duplicate address [check](https://github.com/anza-xyz/solana-sdk/blob/9c5609a62b342bc993cc40eb069ea94439dc563c/transaction/src/sanitized.rs#L288) is [performed](https://github.com/search?q=repo%3Aanza-xyz%2Fagave%20validate_account_locks&type=code) following sanitization and it also checks the lock limit, but I don't see why this couldn't be done earlier.

Legacy also [requires](https://github.com/anza-xyz/solana-sdk/blob/9c5609a62b342bc993cc40eb069ea94439dc563c/message/src/legacy.rs#L115-L116) the fee-payer to not be a program index and the fee-payer [be](https://github.com/anza-xyz/solana-sdk/blob/9c5609a62b342bc993cc40eb069ea94439dc563c/message/src/legacy.rs#L106) RW.

Maybe beyond the scope of this proposal.... Would it be possible to [demote](https://github.com/anza-xyz/solana-sdk/blob/9c5609a62b342bc993cc40eb069ea94439dc563c/message/src/legacy.rs#L508-L510) additional executable-account write-locks by requiring the tx to statically provide indices of program keys only invoked in CPI and not top-level instructions? It would eliminate needless locking and JIT compilation.

cavemanloverboy:
i haven't even read this thing but can we reserve some header space for things in the future? success/failure fees, for example. we should reserve 12 or 16 bytes in the header

tao-stones:
> i haven't even read this thing but can we reserve some header space for things in the future? success/failure fees, for example. we should reserve 12 or 16 bytes in the header

Not opposed to it if expect to use it soon. If the need isn’t immediate, we could also keep things simpler for now and introduce it later as part of a “v2”.

cavemanloverboy:
> > i haven't even read this thing but can we reserve some header space for things in the future? success/failure fees, for example. we should reserve 12 or 16 bytes in the header
> 
> Not opposed to it if expect to use it soon. If the need isn’t immediate, we could also keep things simpler for now and introduce it later as part of a “v2”.

if u promise v2 can be done quickly, let's just wait

proposals/0385-transaction-v1.md
jstarry: We definitely need to revisit write lock demotion but I'm not sure this list would make a big difference. I think it's usually a safe assumption to load the program cache for all programs included in a transaction because they'll probably be invoked.

proposals/0385-transaction-v1.md
```diff
@@ -0,0 +1,262 @@
...
+
+#### TransactionConfigMask
+
+TransactionConfigMask is explained below in detail.
+
+#### LifetimeSpecifier
+
+LifetimeSpecifier is the same as the Recent Blockhash field in prior
+transaction formats. It has been renamed to clarify its use without changing
+its meaning.
```
2501babe, line 111: does this mean that it is in fact a recent blockhash, or that it can also be a durable nonce? we have been debating how to improve or replace nonce transactions, ideally a way to offer similar functionality but require they be normal blockhash transactions from the pov of the runtime

if no one objects, i would suggest officially not supporting nonce transactions for this transaction format. we can always change our mind and support it, but its very hard to change our mind and unsupport it

tao-stones: I support this SIMD to be explicit on not supporting Nonce transaction and compute budget instructions.

jacobcreech: For nonce transactions I am going to make a different SIMD addressing that. Nonce transactions are too controversial not to be in a separate SIMD, but also should not hold up transaction v1

apfitzge:
> i haven't even read this thing but can we reserve some header space for things in the future? success/failure fees, for example. we should reserve 12 or 16 bytes in the header

This is the point of the resources mask it's [done](https://github.com/solana-foundation/solana-improvement-documents/pull/385/files/f99f36dde65e69496c3b4b0778d3c5e66550fa31#diff-dbd54cc962c00d4b548ccd7f2250d7403ce75365c1f2bd154461582a142fb3a5R211). Please read, I'm very confident you know how.

proposals/0385-transaction-v1.md
```diff
@@ -0,0 +1,264 @@
...
+A new transaction format, `v1`, is proposed to enable larger transaction
+sizes.
+
+### Transaction V1 Specification
+
+```
+VersionByte (u8)
+LegacyHeader (u8, u8, u8) 
+NumInstructions (u8)
+TransactionConfigMask (u32) -- Bitmask of which config requests are present.
```
jstarry, line 56: I think ideally this is 4 byte aligned and so `NumInstructions` should move down, maybe above `NumAddresses`? cc @cpubot

proposals/0385-transaction-v1.md
```diff
@@ -0,0 +1,264 @@
...
+LifetimeSpecifier [u8; 32]
+NumInstructions (u8)
+NumAddresses (u8)
+Addresses [[u8; 32]] -- Length matches NumAddresses
+ConfigValues [[u8; 4]] -- Length equal to the popcount (number of set bits)
+  of TransactionConfigMask. See section TransactionConfigMask for details.
+InstructionHeaders [(u8, u8, u16)] -- Length of NumInstructions. Values are
+  (ProgramAccountIndex, NumInstructionAccounts, NumInstructionDataBytes)
+InstructionPayloads [InstructionPayload] -- Length = NumInstructions.
+  Each InstructionPayload is the concatenation of the following byte arrays:
```
cpubot, line 66: Do we anticipate validator or runtime code paths typically requiring to access `InstructionAccountIndexes` and `InstructionData` at the same time, or will they primarily consumed in separate phases (e.g., accounts-only pre-execution validation and lock planning vs. data access during execution)?

With v1 moving compute-budget configuration into the header and Alpenglow deprecating vote transactions, (removing one of the main pre-execution consumers of `InstructionData` as far as I can tell) would it make sense to split instruction payloads into separate contiguous regions for account indices and data? This could in theory improve cache locality and reduce unnecessary data fetches during ingestion without materially affecting execution

topointon-jump (approved):

proposals/0385-transaction-v1.md
tao-stones: `InstructionAccountIndexes` serves the same purpose as `CompiledInstruction.accounts`: it specifies which accounts are passed to the program for execution, together with `InstructionData`. Given that these two are always consumed together, it would be reasonable to place `InstructionAccountIndexes` and `InstructionData` in a contiguous memory region.

tao-stones (approved):

bw-solana (approved):

PR #387, opened 2025-10-27
SIMD-0387: BLS Pubkey Management in Vote Account.

wen-coding:
- Add BLS Pubkey Management in Vote Account for Alpenglow.

### DIFF
--- a/proposals/0387-bls-pubkey-management-in-vote-account.md
+++ b/proposals/0387-bls-pubkey-management-in-vote-account.md
@@ -137,7 +137,7 @@ case).
 
 See "Security Considerations" for why the fields are needed.
 
-During PoP calculation, the cli will generate the BLS keypair, then use the BLS
+During PoP calculation, the CLI will generate the BLS keypair, then use the BLS
 private key to sign this message to generate the signature, compress it, and
 save it in `authorized_voter_bls_proof_of_possession`.
 


proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,180 @@
...
+(as BLS operates over a different elliptic curve), so we can’t naively reuse
+the current ed25519 public keys in vote accounts either. We have to add a BLS
+public key into each vote account and make that a requirement a little before
+Alpenglow launches per [SIMD 357](https://github.com/solana-foundation/solana-improvement-documents/pull/357).
+
+## New Terminology
+
+- **BLS public key**: The public key used in BLS signatures. Its size is 48
+bytes in vote accounts, the field name is bls_pubkey_compressed (we will
+uncompress it later into 96 bytes, but that’s internal implementation details).
```
buffalojoec: I wouldn't mention the field name here, since it's an `Option` anyway. Just describe the actual compressed BLS public key data structure itself, agnostic of its `VoteStateV4` interface.

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,180 @@
...
+associated BLS keypair on demand.
+
+### User Operations Support
+
+Since a user’s BLS keypair always changes with the vote authority keypair, we
+list three different scenarios below.
+
+1. Creating a new vote account
+
+After VoteStateV4 (SIMD 185) is enabled everywhere, when users use
```
buffalojoec: nit: might as well link this SIMD, too.

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,180 @@
...
+create-vote-account to create a new vote account, the correct BLS public key
```
buffalojoec: It wasn't clear to me at first that these kebab-case references are commands from the Solana CLI. Typically we refrain from detailing implementation details, such as SDKs and CLIs, in SIMDs.

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,180 @@
...
+with the new BLS public key. Since we need to verify the caller really holds
+the BLS public key claimed, the caller needs to prove possession of the BLS
+keypair to the vote program by signing the BLS public key with the
+corresponding BLS keypair. Then the vote program can do a syscall to actually
+verify the validity of BLS public key. If the verification fails, the
+transaction will fail.
+
+Since BLS verification is expensive (in the order of ~1 millisecond), these
+operations changing BLS public key will need to have correct CU specified to
+succeed (actual numbers to be measured and published later).
```
buffalojoec: Are you saying we should change the Vote program to use different CU consumption amounts per-instruction? That should be specified in here, and we should probably include the new values themselves.

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,180 @@
...
+  InitializeAccountWithBLS(VoteInitWithBLS),
+  AuthorizeCheckedWithBLS(VoteAuthorizeWithBLS),
+  AuthorizeCheckedWithSeedAndBLS(VoteAuthorizeCheckedWithSeedAndBLS)
+}
+```
+
+- InitializeAccountWithBLS(VoteInitWithBLS)
+
+The main difference between this and InitializeAccount is that VoteInitWithBLS
+will contain a new field in addition to the ones in VoteInit:
```
buffalojoec: @jstarry I didn't see plans to change `VoteInit` in [SIMD-0123](https://github.com/solana-foundation/solana-improvement-documents/pull/123), [SIMD-0232](https://github.com/solana-foundation/solana-improvement-documents/pull/232), or [SIMD-0291](https://github.com/solana-foundation/solana-improvement-documents/pull/291), but do we want to consider a consolidated new `VoteInit` that will allow configuration of both collectors and commissions as basis points? Basically capture all of those + BLS stuff in `InitializeAccountV2`.

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+In addition to the original checks, it also checks that proof of possession is
+correctly signed.
+
+- AuthorizeCheckedWithBLS(VoteAuthorizeWithBLS)
+
+Changing the vote authority public key and BLS public key for existing vote
+account without using seed: (we can also use it to add BLS public key to
+existing vote account, just check vote authority public key didn’t change)
```
buffalojoec: If the BLS key is set, I can't use the old instruction to change my vote authority, right? Otherwise, the BLS key would be out-of-step. So, we need to deprecate and fully disable that instruction in order to add this one.

You could alternatively just add a new variant for the instruction input. Then you can use the existing instruction, and the SIMD can specify Vote program behavior that rejects updates to voter address without corresponding BLS proof of possession, if BLS key is set.

buffalojoec: I think the alternative above would also allow `AuthorizeCheckedWithSeed` to just have a behavioral change and also capture the new variant in `authorization_type`.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Changed, does this look better?

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Done, added all related SIMDs to Dependencies section.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Removed.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Added. I don't think we have the actual numbers now, need to test that.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: That would be awesome. Let me know what you prefer.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: I actually only care about touching vote authority pubkey, for other changes I'm fine to have old instructions go through, but it's going to be confusing to user, so maybe better to move everything to new instructions.

If capturing the new variant in authorization_type does the work, then I'm fine with it. All I care is:
- Must specify BLS pubkey when creating new account
- Whenever vote authority pubkey changes, BLS pubkey must change with it

proposals/0387-bls-pubkey-management-in-vote-account.md
buffalojoec: Is this section describing CLI or program behavior, though?

proposals/0387-bls-pubkey-management-in-vote-account.md
buffalojoec: Yeah, but I'd make sure you specify that it's in a compressed form.

proposals/0387-bls-pubkey-management-in-vote-account.md
buffalojoec: I don't see why we couldn't do a new variant in `VoteAuthorize`. I think the program would just have to have a feature gate that changes its behavior once the new variant is activated. It should disallow changes to the `Voter` variant if a BLS pubkey is set.

We would need this behavior with the existing instruction anyway, so IMO it's cleaner to just add a variant to the input rather than a new instruction.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Mostly trying to describe vote program behavior, you have a good point I should add CLI section as well

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: I'm totally fine with that. @jstarry what do you think?

proposals/0387-bls-pubkey-management-in-vote-account.md
buffalojoec: Actually I was thinking the program should be directly specified, not the CLI. If you want this behavior to be respected by the Vote program after this SIMD is implemented, it has to very clearly detail the required changes to the Vote program as a result. Right now it's kind of murky what this section is describing.

wen-coding: Got it, let me rework this section a bit and I'll ping you when it's done.

proposals/0387-bls-pubkey-management-in-vote-account.md
jstarry: Yes, I think it could be nice to have a v2 initialize instruction. I wasn't (and still not) too concerned about it because the v4 defaults in the original initialize are fine and you can change the new v4 fields by appending instructions in the same transaction for setting commission, collectors, and bls key. If we add v2 init, I'm fine with it getting added in this SIMD such that it

> allow[s] configuration of both collectors and commissions as basis points? Basically capture all of those + BLS stuff in InitializeAccountV2

as you've laid out.

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,191 @@
...
+
+### New instructions to vote programs
+
+In the following instructions, there is a proof_of_possession field associated
+with the new BLS public key. Since we need to verify the caller really holds
+the BLS public key claimed, the caller needs to prove possession of the BLS
+keypair to the vote program by signing the BLS public key with the
+corresponding BLS keypair. Then the vote program can do a syscall to actually
+verify the validity of BLS public key. If the verification fails, the
+transaction will fail.
```
jstarry: There's no precedent for native programs making syscalls that I know of. We should probably add a BLS program which can verify the BLS key proof of possession. Then, the vote program can invoke the BLS program to verify the proof.

proposals/0387-bls-pubkey-management-in-vote-account.md
jstarry: Chatted a bit with @buffalojoec about this. I don't mind either approach (new variant in `VoteAuthorize` or new instruction). But I think that if we go with a new instruction it would be pretty nice because we could remove the clock sysvar account input and make it specific to the vote authority and not need to worry about supporting "WithSeed" because that's really only useful for the withdraw authority anyways.

In either case, @buffalojoec is right that we can't allow voters to use the old instruction (or variant) to change the vote authority if a BLS key is already set. This behavior needs to be detailed in this SIMD.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: @samkim-crypto confirmed it's about 801us to 1.15ms in benchmark. I'll check with @tao-stones the actual CU numbers we would need.

samkim-crypto: Yes it is 801us on my local machine and 1.15ms on my remote dev machine. My remote dev machine matches the ed25519 sigverify cost pretty closely, so 1.15 is probably a more correct number.

But the precise CU would depend on how/where we execute the verification: one syscall, multiple syscalls inside the vote program or CPI into a BLS program.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Done.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: @jstarry @buffalojoec I tried to create https://github.com/anza-xyz/solana-sdk/pull/424, is that close to what you had in mind?

proposals/0387-bls-pubkey-management-in-vote-account.md
tao-stones: To be consistent, using [current conversion rate](https://github.com/anza-xyz/agave/blob/master/cost-model/src/block_cost_limits.rs#L8) of 30cu per 1 micro-sec, 1.15ms converts to 34,500 CU. It exceeds [3,000 MAX builtin allocation CUs](https://github.com/anza-xyz/agave/blob/master/program-runtime/src/execution_budget.rs#L47). (As a reference, currently all vote instructions have 2,100 CUs). So the v2 vote needs to be excluded from "builtin " list. Also, to echo @buffalojoec 's point, are all instructions do same BLS verifications therefore cost same CUs, or they can be different?

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Changed.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Added that to "Alternatives Considered".

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Thanks! The new instructions/new variants all do the same BLS verification and they should cost the same amount of CUs.

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,206 @@
...
+instructions will be disallowed to change vote authority, they will result
+in transaction errors. These include:
+
+```rust
+// Will be totally forbidden, use InitializeAccountV2
+InitializeAccount(VoteInit),
+// Forbidden when VoteAuthorize is VoteAuthorize::Voter
+AuthorizeChecked(VoteAuthorize),
+// Forbidden when authorization_type is VoteAuthorize::Voter
+AuthorizeCheckedWithSeed(VoteAuthorizeWithSeedArgs),
```
jstarry: We should have a period of time where both legacy and alpenglow instructions are active to give us time to migrate offchain tooling like the CLI commands for creating vote accounts. But as discussed, we should not allow using the `VoteAuthorize::Voter` variant if the BLS pubkey is already configured for a vote account.

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,206 @@
...
+Whenever a new BLS public key is being updated in the vote account, we need
+to perform BLS verification on its validity, see "Security Considerations"
+for details. We plan to implement this by calling BLS library from the vote
+program, see "Alternatives Considered" for comparison with other solutions.
+
+Since BLS verification is expensive (in the order of ~1 millisecond), these
+operations changing BLS public key will need to have correct CU specified to
+succeed (actual numbers to be measured and added to this PR).
+
+#### Disallow change of vote authority by old instructions
```
jstarry, line 94: `VoteInstruction::Authorize` and `AuthorizeWithSeed` also exist. We need to address changes for those instructions as well

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Changes to vote program
+
+Whenever a new BLS public key is being updated in the vote account, we need
+to perform BLS verification on its validity, see "Security Considerations"
+for details. We plan to implement this by calling BLS library from the vote
+program, see "Alternatives Considered" for comparison with other solutions.
+
+Since BLS verification is expensive (in the order of ~1 millisecond), these
+operations changing BLS public key will need to have correct CU specified to
```
jstarry: Can you clarify that CU's must be consumed right before doing BLS signature verification? And cc @tao-stones for discussing the CU number given time for BLS verification.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Changed, how does this look?

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Added.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Yup, already asked Tao and he estimate 1.15ms running time translates to 34,500 CU. Updated that in the PR.

tao-stones: 34,500 CU for BLS verification is based on current 30cu per 1us - questioned [here](https://github.com/solana-foundation/solana-improvement-documents/pull/387#discussion_r2492080890) if vote program be excluded from "builtin". @jstarry if this is the case?

proposals/0387-bls-pubkey-management-in-vote-account.md
jstarry: Yeah we should exclude these from the builtin list.. We could exclude the vote program completely since AFAIK simple vote transactions don't go through that code path and other vote related transactions are relatively rare and then 3k -> 200k default cost per vote ix is not very impactful.

wen-coding: Sounds good, do I need to write that into this SIMD?

tao-stones: Yeah, I think so — excluding the Vote program from the built-in list would affect consensus. It’d be great if you could include what @jstarry mentioned in the section where the Vote program is defined.

wen-coding: Added, see if what I added make sense?

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,221 @@
...
+
+- VoteStateV4 is specified in [SIMD 185](https://github.com/solana-foundation/solana-improvement-documents/pull/185)
+and it adds an optional BLS public key field
+
+- Requiring BLS public key for Alpenglow is specified in [SIMD 357](https://github.com/solana-foundation/solana-improvement-documents/pull/357)
+
+## New Terminology
+
+- **BLS public key**: The public key used in BLS signatures. It will be
+compressed and stored as a 48 bytes vector in vote account.
```
buffalojoec:
```suggestion
compressed and stored as a 48-byte array in vote account.
```

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Detailed Design
+
+BLS keypairs can be generated randomly like ed25519 keypairs. But to save the
+users some trouble on keypair management, we chose to derive their BLS keypair
+used in Alpenglow votes based on their ed25519 vote keypair. In other words,
+with an existing ed25519 vote keypair, the users can safely regenerate the
+associated BLS keypair on demand.
+
+### User Operations Support
```
buffalojoec: In my opinion, you can probably replace this whole section with a single paragraph and just put it directly on line 61 under the `## Detailed Design` prelude. Maybe something like this:

"When users create vote accounts, they must register their BLS public key by storing it in the newly created vote account. When they modify their vote authority, they must re-register the new corresponding BLS key."

The wording "automatically updated" is confusing, and pertains to client-side (CLI) handling, which shouldn't be included in a SIMD.

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,221 @@
...
+### Changes to vote program
+
+Whenever a new BLS public key is being updated in the vote account, we need
+to perform BLS verification on its validity, see "Security Considerations"
+for details. We plan to implement this by calling BLS library from the vote
+program, see "Alternatives Considered" for comparison with other solutions.
+
+Since BLS verification is expensive (around 1.15ms), these operations changing
+BLS public key will need to have consume 34,500 CU right before BLS signature
+verification happens.
```
buffalojoec: This is a little bit unclear. Maybe we can say something like this?

```suggestion
Since BLS verification is expensive (around 1.15ms), each verification will cost
34,500 CUs. Any Vote program instruction that performs a BLS verification will
therefore add 34,500 CUs per verification on top of its baseline cost. As a result,
Vote program instructions - which currently all cost 2,100 CUs - will have
differentiated CU costs depending on whether they include BLS verification. The
updated CU values are detailed in later sections.

Note the 34,500 CUs for BLS verification will be consumed immediately before
the verification is performed.
```

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,221 @@
...
+
+Currently the vote program is still using the builtin CU, and each instruction
+consumes 3,000 CU. We propose to completely remove the vote program from using
+the built in CU. Because simple vote transactions do not use the code path
+where the CU limit matters, and there are very few other transactions executed
+in the vote program. We can make those transactions consume correct CU based
+on the amount of CPU they would consume.
```
buffalojoec: Let me know what you think of this - I was also a bit confused by this paragraph:

```suggestion
Currently the vote program is allocated a budget of 3,000 CUs in the validator's
builtin program cost modeling mechanism. Simple vote transactions (containing a
`Vote` instruction) already bypass this mechanism, and other Vote program
instructions that may use BLS verification are fairly infrequent. As a result, the Vote
program will be removed from builtin program cost modeling.
```

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,221 @@
...
+
+We need to have the proof of possession in the instruction inputs so we can
+guard against BLS rogue-key attack. If anyone is allowed to randomly choose a
+public key, then an attacker can select a particular public key which interacts
+with other participants' keys so a forged aggregate signature verifies even
+though not all honest parties actually signed.
+
+## Alternatives Considered
+
+### Moving BLS verification to a syscall and/or a different program
```
buffalojoec, line 243: Meh, I think I would just simply place a paragraph about BLS verification being a separate program and then just mention below it that we could explore extracting it into a syscall or on-chain program again later if the Vote program is migrated to an on-chain BPF program post-Alpenglow. Can probably keep it short.

proposals/0387-bls-pubkey-management-in-vote-account.md
buffalojoec: Yep, I think that's exactly what I had in mind!

proposals/0387-bls-pubkey-management-in-vote-account.md
buffalojoec: While I can appreciate the ability to remove the sysvar accounts, my opinion is that it feels a bit like scope creep and I think the new `VoteAuthorize` variant is a cleaner solution right now. `Vote` and `VoteSwitch` will still need Clock anyway, and we can always just remove the need for such accounts, in a completely backwards-compatible way, either separate from this SIMD or whenever the program is migrated on-chain post-Alpenglow.

proposals/0387-bls-pubkey-management-in-vote-account.md
buffalojoec: Is this comment still relevant if we don't do new instructions and instead do the new `VoteAuthorize` variant? I think no - since it appears outdated - but wanted to double-check as I review!

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,221 @@
...
+in the vote program. We can make those transactions consume correct CU based
+on the amount of CPU they would consume.
+
+#### Disallow change of vote authority by old instructions
+
+After the feature gate associated with this SIMD is activated, the previous
+instructions will be disallowed to change vote authority after off-chain tools
+are upgraded, they will result in transaction errors. These include:
+
+```rust
```
buffalojoec, line 123: Up to you nit: this could probably just be a bullet list in markdown and not a code snippet, but it's pedantic, so your call!

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,221 @@
...
+// Will be forbidden after off-chain tools are upgraded, use InitializeAccountV2
```
buffalojoec: I wonder if we can just save the deprecation of this instruction for another SIMD? As per [SIMD-0357](https://github.com/solana-foundation/solana-improvement-documents/pull/357), vote accounts with an invalid BLS pubkey won't be included in Votor, so we already have an incentive for people to use any of these new methods to set the BLS key.

In my opinion, we can just add a nice, simple SIMD to deprecate/remove this instruction later.

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,221 @@
...
+
+pub enum VoteAuthorize {
+    Voter,
+    Withdrawer,
+    VoterWithBLS(VoterWithBLSArgs),
+}
+```
+
+We only allow the new variant in AuthorizeCheck instruction. Calling the new
+variant in any other instruction will fail. Upon receiving the AuthorizeCheck
```
buffalojoec: What's the reason for this constraint? `AuthorizeChecked` requires the signature of the new authority, but as long as the proper BLS pubkey and proof are provided, can't both `Authorize` and `AuthorizeChecked` do the verification? Afaik, we don't need the new authority's ed25519 signature - only its BLS proof - to verify the provided BLS key, right?

If so, we might as well just allow this new variant across all instructions.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Done.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Done.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Does the current change look correct then?

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: I think forbid `VoteAuthorize::Voter` when the BLS pubkey is already configured is probably enough, this guarantees once you have a BLS pubkey it is always valid. For those people rejecting to write BLS pubkey to their vote account, allowing them to do `VoteAuthorize::Voter` is not end of the world, we will kick them out eventually per SIMD 357.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Okay, this is from the wrong impression that we plan to not use Authorize any more, removed.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Removed.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Done.

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,184 @@
...
+```rust
+InitializeAccountV2(VoteInitV2),
+```
+
+```rust
+pub const BLS_PUBLIC_KEY_COMPRESSED_SIZE: usize = 48;
+pub const BLS_SIGNATURE_COMPRESSED_SIZE: usize = 96;
+
+pub struct VoteInitV2 {
+  pub note_pubkey: Pubkey,
```
samkim-crypto: This probably means `node_pubkey`?

wen-coding: Oops, fixed.

bw-solana:
@qkniep any concerns?

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,222 @@
...
+(for example, 60% of the validators voted to skip a slot). The ed25519
+signatures we currently use are not the best fit for this purpose, so instead
+we will be using the Boneh–Lynn–Shacham (BLS) aggregate signature scheme to
+sign Alpenglow votes.
+
+However, the BLS public key is entirely different from an ed25519 public key
+(as BLS operates over a different elliptic curve), so we can’t naively reuse
+the current ed25519 public keys in vote accounts either. We have to add a BLS
+public key into each vote account and make that a requirement a little before
+Alpenglow launches.
```
bw-solana: wordsmith: "Each validator must add a BLS public into their vote account before the network enables Alpenglow in order to vote"

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,222 @@
...
+- Alpenglow is specified in [SIMD 326](https://github.com/solana-foundation/solana-improvement-documents/pull/326)
+
+- VoteStateV4 is specified in [SIMD 185](https://github.com/solana-foundation/solana-improvement-documents/pull/185)
+and it adds an optional BLS public key field
+
+- Requiring BLS public key for Alpenglow is specified in [SIMD 357](https://github.com/solana-foundation/solana-improvement-documents/pull/357)
+
+## New Terminology
+
+- **BLS public key**: The public key used in BLS signatures. It will be
```
bw-solana: is this really new? I believe we reference this in 357 all the time

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,222 @@
...
+compressed and stored as a 48-byte array in vote account.
+
+## Detailed Design
+
+BLS keypairs can be generated randomly like ed25519 keypairs. But to save the
+users some trouble on keypair management, we chose to derive their BLS keypair
+used in Alpenglow votes based on their ed25519 vote keypair. In other words,
+with an existing ed25519 vote keypair, the users can safely regenerate the
+associated BLS keypair on demand.
```
bw-solana: we should be clear as to whether it MUST be this way or not. This reads as if it's just 1 (perceived easier) option. Is it okay if I generate a BLS pubkey that is completely unrelated to my ed25519 key, stick that in my vote account and use it to sign?

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,222 @@
...
+```
+
+Upon receiving the transaction, if the parameter is of the new variant, the
+vote program will perform a BLS verification on submitted BLS public key and
+associated proof of possession. The transaction will fail if the verification
+failed. Otherwise the vote authority change will be recorded in vote account.
+
+## Impact
+
+When starting a validator, the users are supposed to provide all ed25519
```
bw-solana: `s/the users/ operators/`

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,222 @@
...
+keypairs like before. The BLS keypair will automatically be derived from the
+vote authority keypair (if that’s missing, then the identity keypair is used
+like now). The operations needed to switch the keypair and the operations
+needed to switch to a standby node are the same as today.
```
bw-solana, line 215: this section is a little unclear to me. There are a few distinct time windows, and we should probably be clear about which one(s) we refer to here:

1. Before feature gate activation (no impact)
2. After feature gate activation (we can verify PoP and update vote account key)
3. After key has already been verified and vote account is up to date w/ BLS pubkey

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,222 @@
...
+status: Review
+created: 2025-10-27
+feature: (fill in with feature key and github tracking issues once accepted)
+---
+
+## Summary
+
+This proposal specifies in detail how a BLS public key can be generated by
+users via updated existing tools and how they can put the generated BLS public
+keys into their vote accounts for voting in Alpenglow.
```
bw-solana, line 19: We should mention a critical piece of the proposal here, which is how the network verifies PoP

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Changed.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Removed.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Done.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Proof of Possession Calculation actually is the verification method as well, reworked that section.

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Good point, I clarified it a big, how does this look?

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: It's actually possible, and we have no way to verify the two are related. So users can do:
1. Generate a random BLS keypair and put that in vote account
2. Change code to use the given BLS keypair instead of using the vote authority keypair

The protocol works as long as the users use the same BLS keypair in step 1 and 2.

The long term plan after Alpenglow migration is that we should get rid of vote authority keypair and have BLS keypair only, then users supply BLS keypair to validator execution.

Clarified it a bit.

bw-solana: might be nice to specify what the "we chose to derive" pertains to. I believe what it means is that we (Azna) chose to make the Agave client default implementation infer the BLS keypair from the ed25519 keypair and use this for signing votes.

I have mixed thoughts as to whether this client implementation detail belongs in a SIMD.

lidatong (dismissed):

proposals/0387-bls-pubkey-management-in-vote-account.md
wen-coding: Yeah, I originally didn't want to put it in the SIMD, but then the users may have the question: "Do I need to generate a BLS keypair and put it in my wallet?" So this is user visible.

We mainly chose current route so during the migration epoch, we don't need to tell users: "Please explicitly load both keypairs in your wallet, you will need the BLS keypair suddenly in the middle".

After Alpenglow launches it matters less, we can relax this association and remove vote authority ed25519 keypair completely.

Changed.

bw-solana: yeah, makes sense. I'm assuming FD will do this the same way. Might be good to spell out this is the default client behavior (as opposed to something in protocol)

wen-coding: Sure, changed it to be:
```
The association of BLS keypair with vote authority ed25519 keypair is the
default client behavior to simplify Alpenglow launch. After Alpenglow launches
we may get rid of ed25519 vote keypair and allow users to randomly generate BLS
keypairs.
```

How does this look?

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,251 @@
...
+validator votes to succinctly prove certain state transitions can safely happen
+(for example, 60% of the validators voted to skip a slot). The ed25519
+signatures we currently use are not the best fit for this purpose, so instead
+we will be using the Boneh–Lynn–Shacham (BLS) aggregate signature scheme to
+sign Alpenglow votes.
+
+However, the BLS public key is entirely different from an ed25519 public key
+(as BLS operates over a different elliptic curve), so we can’t naively reuse
+the current ed25519 public keys in vote accounts either. Each validator must
+add a BLS public into their vote account before the network enables Alpenglow
```
qkniep:
```suggestion
add a BLS public key into their vote account before the network enables Alpenglow
```

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,251 @@
...
+
+- `chain_id` is the genesis hash of the chain.
+
+- `authorized_voter_pubkey` is the authorized_voter Ed25519 public key.
+
+- `bls_pubkey_bytes` is the compressed new BLS public key (48 bytes).
+
+See "Security Considerations" for why the fields are needed.
+
+During PoP calculation, the cli will generate the BLS keypair, then use the BLS
```
qkniep:
```suggestion
During PoP calculation, the CLI will generate the BLS keypair, then use the BLS
```

proposals/0387-bls-pubkey-management-in-vote-account.md
```diff
@@ -0,0 +1,251 @@
...
+
+We need to put `authorized_voter_pubkey` in Proof of Possession calculation
+because a replay attack exists:
+
+- User A wants to update vote authority, calculates PoP signature
+
+- The attacker sees this PoP signature in the transaction and sends in another
+transaction grabbing the BLS public key before user A
+
+- Now user A cannot use the BLS public key generated for his own vote account
```
qkniep, line 236: Do we say that we check whether the public key is already in use? I don't see it.

wen-coding: We don't enforce it anywhere, but we can emit a warning in CLI if the public key is already in use. The reasons being:
1. it's hard to perform cross account check in vote program transactions
2. it should be cryptographically hard to generate the same private key as others
3. the PoP proof above guarantees you can't easily claim a BLS keypair you don't own by replay attack
So without checking we should still be okay.

bw-solana (approved):

wen-coding:
> ✅ All approvals received! @wen-coding, you can now merge this by commenting `/merge`.
> 
> ✅ **Status:** Ready to merge

/merge

PR #388, opened 2025-10-28
SIMD-0388: BLS12-381 Elliptic Curve Syscalls

samkim-crypto:
Add BLS12-381 elliptic curve syscalls

### DIFF
--- a/proposals/0388-bls12-381-syscalls.md
+++ b/proposals/0388-bls12-381-syscalls.md
@@ -105,20 +105,23 @@ being the byte-ordering of the base field (Fq) elements themselves:
    - `_BE` variants expect this 48-byte array in big-endian.
    - `_LE` variants expect this 48-byte array in little-endian.
 
-2. Fq^2 Element Ordering: An Fq2 element `(c0 + c1 * u)` is encoded as 96 bytes.
-   The standard's structural ordering is preserved: the 48-byte encoding of the
-   `c1` component (imaginary) comes first, followed by the 48-byte encoding of
-   the `c0` component (real). Each 48-byte component respects the `_LE` or `_BE`
-   flag.
+2. Fq^2 Element Ordering: An Fq^2 element `(c0 + c1 * u)` is encoded as 96 bytes.
+   - `_BE` variants: The Zcash standard's structural ordering is preserved. The
+     48-byte encoding of the `c1` component (imaginary) comes first, followed by
+     the 48-byte encoding of the `c0` component (real).
+   - `_LE` variants: The structural ordering follows the canonical memory layout
+     of Rust structs and Little-Endian polynomial representation. The 48-byte
+     encoding of the `c0` component (real) comes first, followed by the 48-byte
+     encoding of the `c1` component (imaginary).
 
 3. Compressed Point Control Bits: For compressed representations (used by
    `sol_curve_decompress`), the Zcash standard uses the 3 most-significant-bits
    of the entire byte string (e.g., the first bits of the 48-byte G1 array) for
-   flags (compression, infinity, and sign). This convention is retained for both
-   BE and LE variants to ensure a consistent encoding structure. The syscall
-   will always read these flags from the same most-significant bits of the byte
-   array, while the remaining bits will be interpreted as the field element
-   according to the specified endianness.
+   flags (compression, infinity, and sign).
+   - `_BE` variants: The flags are located in the 3 most significant-bits of the
+     first byte (offset 0) of the 48-byte array.
+   - `_LE` variants: The flags are located in the 3 most-significant-bits of the
+     last byte (offset 47) of the 48-byte array.
 
 ### New Curve ID Constants
 


proposals/0388-bls12-381-syscalls.md
```diff
@@ -0,0 +1,181 @@
...
+foundation for many efficient and modern zero-knowledge proof systems, such as
+those based on Groth16. While the BN254 curve enables basic cryptography and
+zero-knowledge applications, it does not provide a 128-bit security level, which
+is insufficient for high-security protocols.
+
+The BLS12-381 curve is a modern, widely adopted standard for a pairing-friendly
+curve that achieves a 128-bit security level. It is adopted in other major
+ecosystems, such as Ethereum, which already supports it via a precompile.
+
+Adding support for the BLS12-381 curve is important for enabling Alpenglow consensus
```
wen-coding, line 39: nit: we could link to SIMD 326 here.

proposals/0388-bls12-381-syscalls.md
```diff
@@ -0,0 +1,181 @@
...
+as well. Alpenglow intends to process standard vote transactions by verifying
```
wen-coding: nit: process consensus votes, we don't have concept of vote transactions any more.

proposals/0388-bls12-381-syscalls.md
```diff
@@ -0,0 +1,181 @@
...
+BLS signatures in a dedicated BLS sigverify stage within the validator logic, so
```
wen-coding: hmm, I think whether it's dedicated BLS sigverify stage is internal implementation, just say BLS verification on votes doesn't need the syscall, we are calling the library directly.

proposals/0388-bls12-381-syscalls.md
```diff
@@ -0,0 +1,181 @@
...
+a BLS12-381 syscall is not necessary for the vote transactions themselves.
+
+However, when a validator registers their BLS public key, it must do so using a
+regular on-chain transaction that will pass through the normal transaction
+pipeline. This transaction must contain a BLS Proof of Possession (PoP) to
```
wen-coding: nit: Do you think we should maybe list "BLS Proof of Possession" in New Terminology section? This is new to the chain.

proposals/0388-bls12-381-syscalls.md
```diff
@@ -0,0 +1,181 @@
...
+validate ownship of the BLS public key. This PoP is a cryptographic proof that
+is crucial to prevent a "rogue-key-attack", where a malicious validator could,
+roughly speaking, try to register another validator's public key as their own.
```
wen-coding: nit: we can remove "roughly speaking" maybe?

proposals/0388-bls12-381-syscalls.md
```diff
@@ -0,0 +1,181 @@
...
+
+1. The addition, subtraction (negation) and scalar multiplication in G1
+2. The pairing operation
+3. The decompression operations in G1 and G2
+
+These operations are sufficient to enable standard pairing-based cryptography
+applications, for example, Groth16 proof verifier on BLS12-381 and BLS signature
+or Proof Of Possession (PoP) verifier. Additional operations are needed to
+support more advanced ZK or cryptography applications, but these are outside the
+scope of this SIMD for now.
```
wen-coding: nit: remove "for now".

proposals/0388-bls12-381-syscalls.md
```diff
@@ -0,0 +1,181 @@
...
+elliptic curve group operations. This function takes in a `curve_id`, `group_op`,
+and two scalar/points that are encoded in little-endian. It interprets the
+curve points according to the curve id, and applies the group operations
+specified by the `group_op`. Currently, the syscall supports the curve25519
+edwards and ristretto representations.
+
+This function can be extended to support the addition, subtraction, and scalar
+multiplication in BLS12-381 G1. We propose adding new `curve_id` constants for
+BLS12-381. The BLS12-381 inputs should be interpreted as affine points unlike
+how Curve25519 points are currently interpreted by the syscall.
```
wen-coding: nit: they use inputs differently? Maybe we just list which is used by BLS12-381 and which is used by Curve25519?

proposals/0388-bls12-381-syscalls.md
```diff
@@ -0,0 +1,181 @@
...
+define_syscall!(fn sol_curve_pairing_map(
+    curve_id: u64,
+    point: *const u8,
+    result: *mut u8
+) -> u64);
+```
+
+This function will take in a curve id and a point that is represented in its
+compressed representation (encoded in little-endian). It will decompress the
+compressed point into an affine representation and write the result.
```
wen-coding: Hmm, do you mean, if I have a compressed pubkey and compressed signature, I need to do three syscalls to verify proof of possession? Would that be more expensive than one syscall?

proposals/0388-bls12-381-syscalls.md
```diff
@@ -0,0 +1,181 @@
...
+encoded in little-endian. It should then compute the pairing product
+`e(P1, Q1) * ... * e(Pn, Qn)`.
+
+### Decompression Operations in G1 and G2
+
+For the decompression operations in G1 and G2, we propose adding a dedicated
+syscall for general decompression `sol_curve_decompress`.
+
+```rust
+define_syscall!(fn sol_curve_pairing_map(
```
0x0ece: nit: sol_curve_decompress

proposals/0388-bls12-381-syscalls.md
samkim-crypto: Right, the PoP will actually use 4 syscalls:
1. sha256 to hash the public key
2. decompress public key
3. decompress signature (PoP)
4. pairing

There is some additional overhead for invoking a syscall itself. However, the cryptographic operations themselves are relatively much more costly, so the actual overhead of invoking multiple syscall as opposed to executing each of these in one larger syscall should be almost negligible I believe. Unlike votes, PoP is done once for each validator during BLS public key registration, so I think the cost should be reasonable.

proposals/0388-bls12-381-syscalls.md
```diff
@@ -0,0 +1,212 @@
...
+This function is not actually instantiated at the moment. We propose updating
+this function's signature to take in an array of points in G1 and an array of
+pointst in G2 to support batch pairings.
+
+```rust
+define_syscall!(fn sol_curve_pairing_map(
+    curve_id: u64,
+    g1_points: *const u8,
+    g2_points: *const u8,
+    result: *mut u8
```
ananas-block, line 254: Is the intention for the pairing result to be a `bool` (bn254 syscall) or the actual result?

Returning the actual result would reduce the number of pairs necessary to verify a Groth16 proof by one from 4 to 3 (12k CU for the current bn254 syscall).

proposals/0388-bls12-381-syscalls.md
```diff
@@ -0,0 +1,212 @@
...
+## Summary
+
+This proposal introduces a new family of syscalls to provide native support for
+cryptographic operations on the BLS12-381 elliptic curve. These syscalls will
+expose:
+
+1. The addition, subtraction (negation) and scalar multiplication in G1
+2. The pairing operation
+3. The decompression operations in G1 and G2
+
```
deanmlittle, line 23: We should also add G2 addition and multiplication. It unlocks Halo2 proofs, hash to G2 onchain, and verification of external Alpenglow quorums so we can actually have real, L1-verified L2s on mainnet.

proposals/0388-bls12-381-syscalls.md
```diff
@@ -0,0 +1,212 @@
...
+// Reserve indices 2 and 3 in case we want to support affine representations of
+// curve25519 points in the future
+// pub const CURVE25519_EDWARDS_AFFINE_LE: u64 = 2;
+// pub const CURVE25519_EDWARDS_AFFINE_BE: u64 = 2 | 0x80;
+// pub const CURVE25519_RISTRETTO_AFFINE_LE: u64 = 3;
+// pub const CURVE25519_RISTRETTO_AFFINE_BE: u64 = 3 | 0x80;
+
+// New Curve ID
+pub const BLS12_381_LE: u64 = 4;
+pub const BLS12_381_BE: u64 = 4 | 0x80;
```
deanmlittle, line 146: Is there any demand for BE? 🤔

proposals/0388-bls12-381-syscalls.md
```diff
@@ -0,0 +1,212 @@
...
+([SIMD 326](https://github.com/solana-foundation/solana-improvement-documents/pull/326))
+as well. Alpenglow consensus votes themselves do not require a BLS12-381
+syscall as they are processed differently than regular transactions.
+
+However, when a validator registers their BLS public key, it must do
+so using a regular on-chain transaction that will pass through the normal
+transaction pipeline. This transaction must contain a BLS Proof of Possession
+(PoP) to validate ownership of the BLS public key. This PoP is a cryptographic
+proof that is crucial to prevent a "rogue-key-attack", where a malicious
+validator could try to register a a key that is an algebraic combination of
```
Rexicon226:
```suggestion
validator could try to register a key that is an algebraic combination of
```

proposals/0388-bls12-381-syscalls.md
```diff
@@ -0,0 +1,212 @@
...
+    result: *mut u8
+) -> u64);
+```
+
+This function is not actually instantiated at the moment. We propose updating
+this function's signature to take in an array of points in G1 and an array of
+pointst in G2 to support batch pairings.
+
+```rust
+define_syscall!(fn sol_curve_pairing_map(
```
Rexicon226, line 249: With this syscall prototype, how will you know how many pairs are provided in the arrays? `sol_alt_bn128_group_op` takes in a length parameter.

proposals/0388-bls12-381-syscalls.md
```diff
@@ -0,0 +1,212 @@
...
+    curve_id: u64,
+    point: *const u8,
+    result: *mut u8
+) -> u64);
+```
+
+This function is not actually instantiated at the moment. We propose updating
+this function's signature to take in an array of points in G1 and an array of
+pointst in G2 to support batch pairings.
```
Rexicon226:
```suggestion
points in G2 to support batch pairings.
```

proposals/0388-bls12-381-syscalls.md
```diff
@@ -0,0 +1,212 @@
...
+) -> u64);
+```
+
+The function would interpret `g1_points` as an array of BLS12-381 points in G1
+`[P1, ..., Pn]` and `g2_points` as an array of BLS12-381 points in G2
+`[Q1, ..., Qn]`. Both inputs are interpreted as affine representations
+encoded in either little-endian or big-endian depending on the curve id. It
+should then compute the pairing product `e(P1, Q1) * ... * e(Pn, Qn)`.
+
+### Decompression Operations in G1 and G2
```
Rexicon226, line 283: Would it be possible to clarify the decompression validation behaviour here? Or at least link to where it might be explained. 
What qualifies a valid encoding?
Are we following [the IETF draft?](https://www.ietf.org/archive/id/draft-irtf-cfrg-bls-signature-05.html)?

proposals/0388-bls12-381-syscalls.md
```diff
@@ -0,0 +1,212 @@
...
+
+// Reserve indices 2 and 3 in case we want to support affine representations of
+// curve25519 points in the future
+// pub const CURVE25519_EDWARDS_AFFINE_LE: u64 = 2;
+// pub const CURVE25519_EDWARDS_AFFINE_BE: u64 = 2 | 0x80;
+// pub const CURVE25519_RISTRETTO_AFFINE_LE: u64 = 3;
+// pub const CURVE25519_RISTRETTO_AFFINE_BE: u64 = 3 | 0x80;
+
+// New Curve ID
+pub const BLS12_381_LE: u64 = 4;
```
Rexicon226, line 145: Apologies, this question may have already been answered, but what is the difference between `BLS12_381_LE` and `BLS12_381_G1_LE`?

proposals/0388-bls12-381-syscalls.md
```diff
@@ -0,0 +1,212 @@
...
+
+This proposal introduces a new family of syscalls to provide native support for
+cryptographic operations on the BLS12-381 elliptic curve. These syscalls will
+expose:
+
+1. The addition, subtraction (negation) and scalar multiplication in G1
+2. The pairing operation
+3. The decompression operations in G1 and G2
+
+Group operations on G2 and Poseidon hash support is outside the scope of this
```
Rexicon226: By "Poseidon hash support", you mean adding a syscall which is poseidon hash over bls12-381 instead of bn254, right? Just want to make sure I understand. Not a problem with the SIMD.

proposals/0388-bls12-381-syscalls.md
samkim-crypto: Yes, I am in support to add G2 group ops as well eventually, but I wanted to limit the scope of the current proposal since I don't want this to be a blocker for alpenglow (we need the syscalls to verify PoP of BLS public keys).

We might need more development time to add support for G2 (esp. with fd) and maybe we can run into issues like G2 ops being too expensive and what not, which could delay feature gate activation, which can affect alpenglow. I don't think it hurts to separate out G2 ops in a separate SIMD and therefore a separate feature gate.

proposals/0388-bls12-381-syscalls.md
samkim-crypto: At least the [rfc](https://www.ietf.org/archive/id/draft-irtf-cfrg-bls-signature-05.html#name-bls12-381) for bls signatures cite the [zcash](https://github.com/zkcrypto/pairing/blob/34aa52b0f7bef705917252ea63e5a13fa01af551/src/bls12_381/README.md#serialization) serialization, which encodes things in big-endian. The `blst` library also serializes using big-endian, so the current bls sig implementation uses BE as well.

deanmlittle: wasn't aware of that. strange choice, but fair enough!

proposals/0388-bls12-381-syscalls.md
samkim-crypto: Yeah I was thinking of actually returning the result in the target group, but I realized that I didn't make that clear. I added a clarifying statement.

proposals/0388-bls12-381-syscalls.md
samkim-crypto: Oh yes this is a nice catch! We do need to add specify the lengths for each of the inputs. I updated to add the lengths.

proposals/0388-bls12-381-syscalls.md
samkim-crypto: This is a good point that I don't fully specify it in the doc. This should follow the Zcash specification. I updated the doc with the specification.

proposals/0388-bls12-381-syscalls.md
samkim-crypto: For something like the pairing function, we just need to specify the curve and hence `BLS12_381_{LE,BE}`. But for functions like group op and decompress, we need to specify either G1 or G2 and hence we have `BLS12_381_G1_{LE,BE}` and `BLS12_381_G2_{LE,BE}`. I discuss it below where the constants are specified.

proposals/0388-bls12-381-syscalls.md
samkim-crypto: Right. I mean extending the poseidon syscall to work over bls12-381 in addition to bn254 (maybe by extending [this](https://github.com/anza-xyz/agave/blob/master/poseidon/src/lib.rs#L118)).

proposals/0388-bls12-381-syscalls.md
Rexicon226: What point is there in having them as separate parameters, how could they ever be different lengths?

I do like the SoA approach, rather than AoS approach taken by the alt-bn128 syscall, but I don't see a reason to have two length parameters when the implementation would just end up comparing them, and erroring anyways.

proposals/0388-bls12-381-syscalls.md
Rexicon226: This is much better, thank you.

proposals/0388-bls12-381-syscalls.md
Rexicon226: Nice, makes sense.

proposals/0388-bls12-381-syscalls.md
Rexicon226: Hm, makes sense. I could see ways to simplify it to just share them, but there is no need for that I imagine, it's all so trivial.

proposals/0388-bls12-381-syscalls.md
Rexicon226: I have wondered about this. As I understand it, BLS signature verification and such only cares about 

```math
e(p_1, q_1) \cdot e(p_2, q_2)^{-1} = 1
```

So moving the identity check out into the ebpf makes that usecase slower. I agree there should be a way to get the result of the miller eval out, esp for circuit constraints and stuff, but there are usecases for the identity case.

Perhaps we could specify which happens based on the curve id? I wouldn't want it as another argument, since that just increases register pressure for no real reason.

proposals/0388-bls12-381-syscalls.md
samkim-crypto: Yeah I think the alt-bn128 syscall pushes a lot of expensive work onto the BPF program. In Groth16 verifier, for example, the G1 points (from the proof) and G2 poinits (from the verifying keys) generally live in different memory and are almost never stored together in the interleaved format.

In terms of one length parameter vs two, I think having a byte length parameter for each array parameter to a function is just a direct and more canonical C-style API. An alternative would be to pass a `num_pairs` parameter and I think this is fine as well, but the runtime would then need to calculate `expected_g1_len = num_pairs * 96` and `expected_g2_len = num_pairs * 192` and use this number to parse the arrays. I think just having the byte lengths directly follow the arrays feels more natural to me, but I am open to taking in `num_pairs` if others feel strongly about this.

proposals/0388-bls12-381-syscalls.md
samkim-crypto: This is a reasonable point. But for the identity check operation, what we need is essentially just a `memcmp`, right? After the result is returned, we just need to compare the result to a hard-coded identity element in the target group, which I can't imagine it being too expensive. For applications like groth16, pinnocchio, groth-maller, etc. we would be saving an entire pairing which is in the realm of tens of thousands of CUs (12k for alt_bn128).

Rexicon226: True, it would just be a 576 byte memcmp (12 * 48), which is 100 CUs (`max(100, 576 / 250)`). So I guess it doesn't matter too much. The reason I brought this up is because of how trivial it is to support.

proposals/0388-bls12-381-syscalls.md
Rexicon226: I personally like the `num_pairs` approach myself. If you pass in byte lengths, the runtime still needs to check
```
g1_bytes % (2 * 48) = 0
g2_bytes % (4 * 48) = 0
if (g1_bytes / (2 * 48) != g2_bytes / (4 * 48)) return error
```
which leads to yet another thing the caller can get wrong.
I feel that in most callsites of this syscall, the program will already have the number of pairs. 
Perhaps someone else could chime in and clarify which is more ergonomic for the users?

As for it being more C-style, eh, maybe. APIs like `memcmp` and `memmove` already take in one length parameter which is for two parameters. I think this is comparable.

proposals/0388-bls12-381-syscalls.md
mertimus: I don't quite get the risk here, what do we mean by too expensive and what not? blocks are relatively empty and it's a safe change that can enable a ton of ecosystem teams on top to build new things that they need to compromise on today

samkim-crypto: I will actually go ahead and add G2 group ops in this SIMD since there is a lot of demand for this. 

I was initially hoping to add G2 group ops in a separate SIMD/feature gate to not block alpenglow. Adding G2 would obviously require more development time to test, assign the appropriate CUs, make sure that all the edge cases are handled correctly among validator clients, etc. and I wanted to minimize the chance of alpenglow launch being blocked by the activation of BLS12-381 syscall.

But the most recent alpenglow proposal was [updated](https://github.com/solana-foundation/solana-improvement-documents/pull/387/commits/a8e44f94212516ae791818591bc87cedf53d67eb) to not rely on the bls12-381 syscall, but to perform the BLS PoP verification inside the vote program at launch. So I think we can now have a more complete BLS12-381 syscall proposal.

proposals/0388-bls12-381-syscalls.md
samkim-crypto: I was looking at the design again and yes, I agree that the `num_pairs` design could be slightly cleaner. I updated the function signature accordingly.

deanmlittle:
@0x0ece any objections to approving this?

0x0ece (approved):

proposals/0388-bls12-381-syscalls.md
```diff
@@ -0,0 +1,341 @@
...
+1. Field Check: The coordinate(s) are valid field elements (i.e., less than the
+   modulus `p`).
+2. On-Curve Check: The point satisfies the curve equation.
+3. Subgroup Check: The point is in the correct prime-order subgroup `r`.
+
+This syscall is specifically for uncompressed affine points. To validate a
+compressed point, programs can use the `sol_curve_decompress` syscall, which
+performs identical validation checks as part of its decompression process.
+
+While it is possible to validate an uncompressed affine poinnt `(x, y)` by using
```
Rexicon226:
```suggestion
While it is possible to validate an uncompressed affine point `(x, y)` by using
```

Rexicon226 (approved):

jacobcreech (approved):

PR #389, opened 2025-10-28
SIMD-0389: Reduce Account Creation Cost and Introduce Supervisory Controller

igor56D:


proposals/0389-dynamic-min-deposit.md
```diff
@@ -0,0 +1,199 @@
...
+Depends on SIMD-0392 or another protocol change to allow for non-disruptive
+minimum deposit increases.
+
+## Motivation
+
+Introduce a real-time price signal for state demand and apply back pressure to
+prevent runaway state growth. The proposal gives the protocol more tuning knobs
+(controller parameters such as deadband H, bounds, step sizes, clamps, target growth)
+than a single fixed constant, enabling better behavior targeting. With an
+average on-chain state growth currently around ~200 MB per epoch, targeting 1
```
brooksprumo: >With an average on-chain state growth currently around ~200 MB per epoch,

Where is the 200 MB number from?

I plotted all my account state updates from Discord, and see higher numbers. Not a ton higher, at least recently. Around the beginning of the year, growth was much higher.

Here's the change in on-chain account data size each week, averaged per day. This is just the `data` field of each account summed up. Thus does not include any other fields e.g. owner, lamports, etc.
Image

Here's a more comprehensive look. It's the change of account storage file size each week, again averaged per day. This is the actual size Agave uses to store accounts. This has some storage overhead though. And it doesn't include any per account costs elsewhere in the system (e.g. in the accounts index).
Image

proposals/0389-dynamic-min-deposit.md
```diff
@@ -0,0 +1,199 @@
...
+
+The controller updates `min_deposit` every slot based on measured state growth
+but is intended to be inactive during normal conditions (holding at the 0.1x
+floor) and only engage when sustained growth exceeds a safety threshold.
+
+**State tracking:**
+
+- The integral accumulator `I` MUST be tracked in fork-aware bank state.
+- Each slot measures realized state growth `G_slot` (bytes of new account data
+  allocated minus freed/deallocated data).
```
brooksprumo: What is "account data" here, specifically? Is it just the `data` field of Account? Or does it include any account metadata?

Note that this tracking must be deterministic cluster-wide, which doesn't exist today. We have approximations for on-chain account data size, and number of accounts, but work would need to be done here first.

proposals/0389-dynamic-min-deposit.md
```diff
@@ -0,0 +1,199 @@
...
+  discontinuities and gaming incentives.
+- Dynamic rent w/ eviction: more complete but controversial and
+  complicated. This proposal will allow a reduction in state allocation cost
+  sooner.
+
+## Impact
+
+- Dapp developers: Creation and reallocation costs vary over time; programs can
+  query the Rent sysvar for current rates. Significant minimum-deposit
+  reduction is expected.
```
brooksprumo, line 219: I've heard that variable costs can be a pain. I don't have useful numbers for how much is too much though. This impacts regular users too, not just dapp developers.

This in turn impacts RPC calls and the CLI, which return the rent amount. 

Not saying these are deal breakers, just noting them.

proposals/0389-dynamic-min-deposit.md
igor56D: I came up with the 200MB number by just eyeballing more recent metrics before you posted this data. I'll update it to be more accurate and specific about what the number is actually measuring.

proposals/0389-dynamic-min-deposit.md
igor56D: Good point, I haven't updated this since putting together the [PoC implementation](https://github.com/anza-xyz/agave/pull/9051). AFAIK we do deterministically track *new* state growth in the cost tracker to enforce limits and in the bank's `accounts_data_size_delta_on_chain` field, but neither turned out to be sufficient for this use-case so I added a new bank field to track what the controller is actually interested in (net state growth).

I'll update the proposal to be more specific:
1. the state contribution of each account is `acc.data.size() + solana_rent::ACCOUNT_STORAGE_OVERHEAD`, ie the same way we do it for min_balance checks.
2. an account with post-exec balance of 0 and non-zero pre-exec balance is considered deleted, so we subtract its contribution from the net state growth.

proposals/0389-dynamic-min-deposit.md
igor56D: The intention is for rent to stay constant at 0.1x. It's only dynamic if the safety threshold is exceeded, which shouldn't happen under normal circumstances. I don't think this is avoidable regardless of the how we choose to reduce rent because there's always the possibility that rent will need to be increased if growth becomes excessive.

The main difference with this specific proposal is that it proactively increases rent if our estimate for the safe threshold is exceeded while manual intervention will probably only increase rent after real issues are observed. This can be both better and worse depending on how you look at it. Worse if the safety threshold turns out to be too conservative and nothing bad happens if its exceeded, because rent gets increased without a good reason. Better if real issues are prevented rather than responded to after the fact.

Overall, I'm partial to the proactive approach because most people believe that (1) demand for state won't skyrocket after a rent decrease and (2) the amount of state growth the client can handle significantly exceeds demand, so rent will most likely stay constant at 0.1x because the controller is never engaged. If we want to avoid the bad outcome we can always increase the safety threshold, which essentially functions like a resource constraint similar to the CU limit.

One other potential issue is that the controller may not respond effectively to an issue (e.g. a bunch of tiny accounts are created that blow up the index but don't exceed the growth threshold), but this just brings us back to manual intervention, which is what we would've done anyway without the controller. If the index growth rate is particularly concerning, we can increase `solana_rent::ACCOUNT_STORAGE_OVERHEAD` so the minimum per-account contribution is higher.

Sorry for the essay haha. I'll add some of this to the proposal.

proposals/0389-dynamic-min-deposit.md
brooksprumo: Gotcha, makes sense (w.r.t. data size + rent storage overhead).

I was the one that added the Bank::accounts_data_size fields. The hard part is on startup to get the deterministic number of accounts and accounts_data_size. The delta per bank during replay is (should be?) easier, and more or less done.

brooksprumo: And this part is probably the main thing for me that gives me concern on shipping fast. Is a controller nice? Sure. Is it necessary? I don't know, I'm leaning more towards no than yes currently, given a 90% decrease. Maybe it becomes more useful with larger decreases.

I am confident that we could do `new rent = old rent / 10` and have that ready in v4.0. The signals I'm getting say that shipping fast is a priority. How fast is fast enough? I don't know. Can we get an impl of this SIMD into v4.0? I'm not sure! Luckily you've already been working on it, so you may be in a better position to comment.

igor56D: > I am confident that we could do new rent = old rent / 10 and have that ready in v4.0

The rollout for that proposal will most likely be incremental, not a direct 10x reduction, and also include a waiting period between reduction stages to observe effects. From my PoV, most of the core part of the implementation for this SIMD is already done and I don't see why it can't be shipped quickly if everyone is aligned.

> The hard part is on startup to get the deterministic number of accounts and accounts_data_size

I was assuming we can persist the fields we want by adding them to `BankFieldsToSerialize` or `ExtraFieldsToSerialize`. I'm not very familiar with the what happens during startup to initialize the current bank but it doesn't sound too difficult to add an extra field. Am I missing something?

brooksprumo: > The rollout for that proposal will most likely be incremental, not a direct 10x reduction

Why? 


> most of the core part of the implementation for this SIMD is already done and I don't see why it can't be shipped quickly if everyone is aligned.

Working on getting the accounts data size to be deterministic needs to be done first. IMO it's an unknown, thus I don't feel confident this can skip quickly.

Luckily the accounts data size part can be worked on immediately/doesn't need to wait for SIMD approval.


> I was assuming we can persist the fields we want by adding them to `BankFieldsToSerialize` or `ExtraFieldsToSerialize`. I'm not very familiar with the what happens during startup to initialize the current bank but it doesn't sound too difficult to add an extra field. Am I missing something?

The field is already in the snapshot, that's not the hard part. The hard part is keeping it deterministic across the cluster. The last time I checked (which was like 1-2 years ago), it still wasn't.

igor56D: Makes sense. I'll look into this

igor56D: Pushed an update (e2ef30d49ae573d795bfe4daa4b95648bd044355). Determinism is achieved by backing up the necessary bank fields into dedicated sysvars, which are subject to the snapshot integrity check through the accounts lthash.

igor56D: two deltas are tracked by the bank:
1. `accounts_data_size_delta`: only tracks the change in the actual data portion of the overall account state. Calculated as the sum of the resize_delta (already tracked deterministically by the runtime) across all transactions in a slot. Zero balance accounts have their data size subtracted off separately post-execution because the resize_delta doesn't account for them.
2. `accounts_num_delta`: tracks change in the number of accounts. Determinism is pretty easy to achieve here with pre and post exec balance checks.

proposals/0389-dynamic-min-deposit.md
```diff
@@ -0,0 +1,249 @@
...
+authors:
+  - Igor Durovic (anza)
+category: Standard
+type: Core
+status: Idea
+created: 2025-10-27
+feature: (fill in with feature key and github tracking issues once accepted)
+supersedes:
+superseded-by:
+extends:
```
brooksprumo:
```suggestion
```

proposals/0389-dynamic-min-deposit.md
```diff
@@ -0,0 +1,249 @@
...
+---
+
+## Summary
+
+Reduce the minimum deposit rate by 10x (set baseline
+`min_deposit = 0.1 * min_balance_legacy`) and make this the new floor.
+
+Introduce a supervisory integral controller targeting 2 GB/epoch state growth
```
brooksprumo: Why 2 GB now?

proposals/0389-dynamic-min-deposit.md
```diff
@@ -0,0 +1,249 @@
...
+feature: (fill in with feature key and github tracking issues once accepted)
+supersedes:
+superseded-by:
+extends:
+---
+
+## Summary
+
+Reduce the minimum deposit rate by 10x (set baseline
+`min_deposit = 0.1 * min_balance_legacy`) and make this the new floor.
```
brooksprumo: Since we're now likely going to have a constant reduction before the controller, we may want to specify either an exact number for the new min, or specify the exact value for the `min_balance_legacy`. As I believe we are not intending to do a constant reduction, then an additional 10x reduction. This applies to the whole PR.

proposals/0389-dynamic-min-deposit.md
```diff
@@ -0,0 +1,249 @@
...
+  - `num_accounts_delta` (count): the net change in the number of accounts. Zero
+    balance accounts are considered deleted.
+
+**`G_slot` calculation:**
+
+`G_slot` is the sum of the account data size growth plus the storage overhead
+associated with changes in the number of accounts:
+
+```
+G_slot = data_size_delta + ACCOUNT_STORAGE_OVERHEAD * num_accounts_delta
```
brooksprumo, line 85:
```suggestion
G_slot = (data_size_delta + ACCOUNT_STORAGE_OVERHEAD) * num_accounts_delta
```

proposals/0389-dynamic-min-deposit.md
```diff
@@ -0,0 +1,249 @@
...
+```
+
+where `ACCOUNT_STORAGE_OVERHEAD` is an existing in-consensus parameter used by
+rent (128 bytes), representing the per-account storage overhead.
+
+**Update rule (executed once per slot):**
+
+Let target growth per slot be `G_target = 2 GiB/epoch / slots_per_epoch`.
```
brooksprumo: Note that the two mentions prior both say 2 GB, not 2 GiB. We should be consistent. 

Looks like below we spell out 2 GB twice, so likely this is the one to be updated.


```suggestion
Let target growth per slot be `G_target = 2 GB/epoch / slots_per_epoch`.
```

proposals/0389-dynamic-min-deposit.md
```diff
@@ -0,0 +1,249 @@
...
+sysvar and to preserve compatibility with existing on-chain programs.
+
+The Rent sysvar account is `SysvarRent111111111111111111111111111111111`.
+
+### Persistence and snapshot integrity
+
+- The two per-slot deltas (`data_size_delta`, `num_accounts_delta`) and the
+  integral accumulator `I` MUST be backed up into dedicated sysvars so they are
+  covered by snapshot integrity checks via the accounts lthash. This avoids the
+  need to include these values in the bank hash.
```
brooksprumo: just a nit: Yes, we want these values in an on-chain account. Based on other discussions, it may or may not be a sysvar.

Also, the wording here is a little confusing to me. Because if we put these values into accounts then they are indeed included/part of the bank hash, which is the desire.

(Also the snapshot part can be removed IMO. We're not changing the snapshot, so we'll keep the existing behavior of account verification. (But even that is an agave implementation detail, not a protocol requirement.))

proposals/0389-dynamic-min-deposit.md
```diff
@@ -0,0 +1,249 @@
...
+- On startup, validators MUST load these values exclusively from the sysvars,
+  rather than any non-account serialized fields in the snapshot, ensuring that
+  integrity-checked values are used.
```
brooksprumo: The new fields aren't going into the snapshot (as there is no SIMD/proposal to do so). Thus may not need to mention otherwise, as to me it implied this could be possible.

proposals/0389-dynamic-min-deposit.md
```diff
@@ -0,0 +1,249 @@
...
+## New Terminology
+
+- `min_deposit`: The protocol-defined minimum lamports per byte required to
+  allocate account data. Baseline is 0.1x the legacy minimum-balance constant
+  (the floor). A supervisory integral controller MAY adjust it above the floor
+  only when safety thresholds are exceeded.
+
+## Detailed Design
+
+### High level
```
brooksprumo, line 48: Wdyt about putting all the constants declared here into accounts instead.

Specifically:
* the min deposit value
* the max deposit value (aka min_deposit_legacy)
* the deposit step values (aka the +10% and -5%)
* the I_min and I_max
* the growth target per epoch

Thus we'll be able to more easily change them later, and all validator clients will implicitly be in sync (no need for coordinated rollout of clients just to change a constant).

proposals/0389-dynamic-min-deposit.md
brooksprumo: Oh, sorry. I see that this PR defines data_size_delta as only the account data len. Explicitly without the storage overhead. Please disregard.

proposals/0389-dynamic-min-deposit.md
igor56D: Now that we have more clarity on the constant reduction plan, I can integrate that into this SIMD. The most straightforward way is to use the existing `lamports_per_byte` constant that's subject to reduction as the floor of the controller. That way this proposal doesn't clash with the others.

proposals/0389-dynamic-min-deposit.md
igor56D: I'll clean it up. Too many small adjustments has left a few inconsistencies like this.

proposals/0389-dynamic-min-deposit.md
igor56D: > Yes, we want these values in an on-chain account. Based on other discussions, it may or may not be a sysvar.

Using sysvars just seems like the most straightforward out-of-the-box solution for storing these values in "special" accounts, but I'm open to doing something else as well. I'll leave them as sysvars for now because I don't know what to replace that with but will leave a note that this may be subject to change. The important detail is that they're stored in state covered by the lthash that's only modifiable directly by the bank.

> Also, the wording here is a little confusing to me. Because if we put these values into accounts then they are indeed included/part of the bank hash, which is the desire.

The wording is ambiguous. I meant that we won't need to add the new bank fields to the top level bank hash calculation as first-class values because the lthash accounts for them already. Will clarify.

> (Also the snapshot part can be removed IMO. We're not changing the snapshot, so we'll keep the existing behavior of account verification. (But even that is an agave implementation detail, not a protocol requirement.))

agree, will remove

proposals/0389-dynamic-min-deposit.md
igor56D: I like it! Is there a special account type we should store these as? I'm only aware of sysvars.

brooksprumo: I don't know of anything off the top of my head, but I imagine we can probably do something similar to sysvars without the annoying bits. E.g. sysvars have special treatment for their balance. IMO that's unnecessary, as we can just set the balance correctly when creating the accounts. There may be other special things sysvars can do that we probably don't want.

We can likely define a struct/type for each of these accounts. Then on feature activation:
* create new accounts
* serialize the structs into them
* store 'em

There won't be a keypair for the accounts, so no one will be able to change their values. I think we can just do a find-replace on the word "sysvar" for something like "on-chain constant" or w/e.

proposals/0389-dynamic-min-deposit.md
igor56D: I updated the SIMD to use a different rationale for selecting the target. Let me know what you think.

proposals/0389-dynamic-min-deposit.md
igor56D: updated the target again and provided rationale in the SIMD

proposals/0389-dynamic-min-deposit.md
igor56D: updated to use constants. I decided against what I mentioned above, can't think of a clean way to repurpose the `lamports_per_byte` value as the floor price. It may be easiest for this SIMD to fully supersede 437.

proposals/0389-dynamic-min-deposit.md
igor56D: I like this idea but it's probably out of scope for this SIMD. I'd fully support a separate SIMD that moves protocol constants into on-chain accounts though.

igor56D:
This is currently soft-blocked by other rent related work but should be unblocked in about a month or so.

brooksprumo (commented):
Leaving a "review" so it doesn't show up in my "reviews requested" list.

I'd like to see how the existing rent reduction rollout goes before going further down this path.

PR #390, opened 2025-10-30
SIMD-0390: Arbitrary-Byte Seeds for Seeded Addresses

kevinheavey:
IBRL by removing useless work

### DIFF
--- a/proposals/0390-arbitrary-byte-seeds.md
+++ b/proposals/0390-arbitrary-byte-seeds.md
@@ -96,12 +96,12 @@ Derivation logic and collision resistance are unchanged, and the seed remains
 limited to 32 bytes. Removing UTF-8 validation does not introduce new attack
 vectors because the runtime never required the seed to be human-readable.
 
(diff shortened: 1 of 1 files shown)


Benhawkins18:
This was discussed on twitter, there is a likely better solution here suggested by @t-nelson  "SystemInstruction::InvokeSignedWithSeed", could you please consider making a new SIMD that uses this idea to address your motivation outlined here?

Benhawkins18:
In the future please consider, especially since this is an idea, making this a discussion and not a SIMD.

kevinheavey:
Discussion is here https://github.com/solana-foundation/solana-improvement-documents/discussions/378, I waited three weeks before opening a SIMD

kevinheavey:
@t-nelson's solution seems fine too but I don't have bandwidth to think about it and I won't be in the least upset if someone else takes it on

deanmlittle:
This is an obviously good idea.

PR #391, opened 2025-10-30
SIMD-0391: Stake Program Float to Fixed-Point

grod220:
Proposal to migrate floats in stake-interface to fixed point

### DIFF
--- a/proposals/0391-replace-stake-program-floating-point.md
+++ b/proposals/0391-replace-stake-program-floating-point.md
@@ -37,17 +37,17 @@ sync.
 
 - **Basis points (bps)**: An integer representation of a percentage where
   `bps = percent × 100`.
-  - 1 bps = 0.01%
-  - 1% = 100 bps
+    - 1 bps = 0.01%
+    - 1% = 100 bps
 
 - Formula variables
-  - **account_portion**: The amount of stake (in lamports) for a single
-    account that is eligible to warm up or cool down in a given epoch.
-  - **cluster_portion**: The total amount of stake (in lamports) across the
-    cluster that is in the same warmup/cooldown phase as `account_portion`
-    for the previous epoch.
-  - **cluster_effective**: The total effective stake in the cluster (in
-    lamports) for the previous epoch.
+    - **account_portion**: The amount of stake (in lamports) for a single
+      account that is eligible to warm up or cool down in a given epoch.
+    - **cluster_portion**: The total amount of stake (in lamports) across the
+      cluster that is in the same warmup/cooldown phase as `account_portion`
+      for the previous epoch.
+    - **cluster_effective**: The total effective stake in the cluster (in
+      lamports) for the previous epoch.
 
 ## Detailed Design
 
@@ -194,22 +194,59 @@ usage, which just puts the technical debt off to handle later.
 ## Impact
 
 - **Stake Interface**:
-  - Export new integer-based stake activation and deactivation logic for rust
-    consumers
-  - Deprecate the floating-point rate field while preserving binary layout
-    compatibility
+    - Export new integer-based stake activation and deactivation logic for rust
+      consumers
+    - Deprecate the floating-point rate field while preserving binary layout
+      compatibility
 
 - **Stake Program**: Feature gate v2 interface helpers in:
-  - Stake Merging
-  - Stake Splitting
-  - Stake Redelegation
+    - **Stake Merging**: Stake calculations are used to determine if the
+      account is in a transient state, ensuring that merges are rejected if the
+      account is not effectively fully active or inactive.
+    - **Stake Splitting**: Stake calculations are used determine if the source
+      stake is currently active (effective stake > 0). This status is required
+      to correctly enforce rent-exempt reserve prefunding requirements for the
+      destination account.
+    - **Stake Redelegation**: The account's cooldown status is determined with
+      stake calculations and confirms that effective stake is exactly zero
+      before allowing redelegation.
+    - **Stake Withdrawal**: When withdrawing from a deactivated account, stake
+      calculations are used to determine the remaining effective stake.
 
-- **Validator Clients (Agave & Firedancer)**: Feature gate fixed-point math
-  in:
-  - Runtime & stake logic for stake activation and deactivation calculations
-  - Stake cache & history for effective stake derivation and history
-    aggregation
-  - Inflation rewards for calculation of stake-weighted rewards
+- **Validator Clients (Agave & Firedancer)**: Clients must feature gate the
+  transition from floating-point to fixed-point arithmetic in all
+  consensus-critical operations involving effective, activating, or
+  deactivating stake. The following operations require updates:
+    - **Stake Activation and Deactivation**: When querying a stake delegation's
+      status for a given epoch, the validator _computes how much of the
+      delegation's stake has completed warmup or cooldown_. This requires
+      walking through epochs from the delegation's activation or deactivation
+      point, computing the allowed stake change at each epoch boundary to
+      determine the portion that transitioned. The result categorizes the
+      delegation's lamports into effective, activating, and deactivating
+      buckets.
+    - **Epoch Boundary Stake History**: At each epoch boundary, the validator
+      iterates over all stake delegations and _computes their activation status_
+      as of the concluding epoch. These per-delegation values are summed to
+      produce the cluster-wide totals (effective/activating/deactivating) that
+      form the new stake history entry. This entry is then used as input for
+      subsequent epoch calculations.
+    - **Stake Cache Updates**: The validator maintains a cache mapping vote
+      accounts to their delegated stake. When a stake account is
+      created/modified/closed, the cache entry for the associated vote account
+      must be updated. This requires _computing the delegation's effective stake_
+      contribution before and after the change to correctly adjust the cached
+      totals.
+    - **Vote Account Stake Totals**: At epoch boundaries, the validator
+      refreshes the stake distribution across vote accounts for the upcoming
+      epoch. For each vote account, it _sums the effective stake_ of all
+      delegations pointing to that account. These totals determine leader
+      schedule weights and fork choice voting power.
+    - **Inflation Rewards**: Reward calculation iterates over each epoch in a
+      vote account's credit history. For each epoch, the validator _computes the
+      delegation's effective stake_ at that epoch, multiplies by the earned vote
+      credits to produce points and accumulates these across epochs. The final
+      reward is proportional to the delegation's share of total cluster points.
 
 ## Security Considerations
 
--- a/proposals/0391-replace-stake-program-floating-point.md
+++ b/proposals/0391-replace-stake-program-floating-point.md
@@ -13,39 +13,40 @@ feature: (to be assigned upon acceptance)
 
 ## Summary
 
-Replace of all floating-point (`f64`) arithmetic within the Solana
-Stake Program's warmup and cooldown logic with a deterministic
-fixed-point implementation using integer arithmetic. This change is a
-prerequisite to the Stake Program's migration to a `no_std`
-Pinocchio-based implementation and ensures compatibility with upstream
-eBPF toolchains, which do not support floating-point operations.
+Replace all floating-point (`f64`) arithmetic within the Solana Stake Program's
+warmup and cooldown logic with a fixed-point implementation using integer
+arithmetic. The new logic expresses the warmup/cooldown rate in basis points
+(bps) and performs all proportional stake calculations using `u128`
+intermediates.
+
+This change is a prerequisite to the Stake Program's migration to a `no_std`
+Pinocchio-based implementation and ensures compatibility with upstream eBPF
+toolchains, which do not support floating-point operations.
 
 ## Motivation
 
 The Stake Program's use of `f64` presents two blockers to the upcoming roadmap:
 
 1. **Upstream eBPF incompatibility:** Standard eBPF strictly forbids
-   floating-point operations. While the solana fork (SBF) currently
-   supports `f64` via a deterministic (and inefficient) `soft-float`
-   compiler built-in, aligning with upstream standards requires
-   removing all `f64` usage from the program.
+   floating-point operations. While the solana fork (SBF) currently supports `f64`
+   via a deterministic (and inefficient) `soft-float` compiler built-in, aligning
+   with upstream standards requires removing all `f64` usage from the program.
 
 2. **Pinocchio migration inconsistency:** There is appetite for
-   converting the Stake Program to a highly efficient, `no_std`
-   Pinocchio implementation (reducing CU usage by +90%). These efforts
-   are undermined by the immense cost of soft-float operations.
-   [Benchmarking shows](https://solana.com/docs/programs/limitations#limited-float-support:~:text=Recent%20results%20show,Divide%20%20%20%20%20%209%20%20%20219)
-   a 22x performance penalty for a single multiplication of an `f32`
-   versus a `u64`. Using an `f64` with an operation
-   like division is even more complex. Further, doing the float
-   migration independently allows p-stake to enforce semantic
-   equivalence for its migration.
+   converting the Stake Program to a highly efficient, `no_std` Pinocchio
+   implementation (reducing CU usage by +90%). These efforts are undermined by the
+   immense cost of soft-float operations. [Benchmarking
+   shows](https://solana.com/docs/programs/limitations#limited-float-support:~:text=Recent%20results%20show,Divide%20%20%20%20%20%209%20%20%20219)
+   a 22x performance penalty for a single multiplication of an `f32` versus a
+   `u64`. Using an `f64` with an operation like division is even more complex.
+   Further, doing the float migration independently allows p-stake to enforce
+   semantic equivalence for its migration.
 
 ## Requirements
 
-The new implementation must be a replacement that precisely models the
-intent of the original logic. Any resulting differences in output
-should be minor and a direct result of improved numerical precision.
+The new implementation must be a replacement that precisely models the intent of
+the original logic. Any resulting differences in output should be minor and a
+direct result of improved numerical precision.
 
 ## New Terminology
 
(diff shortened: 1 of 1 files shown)


proposals/0391-replace-stake-program-floating-point.md
```diff
@@ -0,0 +1,138 @@
...
+    **(numerator: 9, denominator: 100)**.
+
+- The current flow
+  `(account_portion / cluster_portion) * (cluster_effective * rate)` is
+  reordered to
+  `(account_portion * cluster_effective * rate_numerator) / (cluster_portion * rate_denominator)`.
+  Instead of performing divisions early in the process (which truncates
+  intermediate results), all multiplications are performed first.
+
+- All intermediate multiplications are performed using `u128` to
```
deanmlittle: Similar to f64, u128 is also not supported by upstream BPF at this time. We can roll our own u128 math, replace it with a fast jit intrinsic mapping to a syscall, or several other fairly trivial workarounds, but if your calculation can reasonably fit into a u64, consider simply not making the numerator stupidly large.

proposals/0391-replace-stake-program-floating-point.md
```diff
@@ -0,0 +1,138 @@
...
+}
+```
+
+This preserves the exact memory size and layout of existing accounts.
+It is a legacy field anyway, with the actual rate being determined
+dynamically in functions.
+
+## Alternatives Considered
+
+Decimal scaling factor. Uses a uniform scaling factor represent rates
```
deanmlittle: this is a fine idea so long as precision is sufficient imo.

proposals/0391-replace-stake-program-floating-point.md
grod220: Given the intermediate products could risk an overflow, think the temp upscaling is necessary. I was under the assumption that u128 support was quite likely to be upstreamed (maybe @Lichtso has some thoughts though). Otherwise, looks like the [uint crate](https://crates.io/crates/uint) could be useful here.

proposals/0391-replace-stake-program-floating-point.md
```diff
@@ -0,0 +1,138 @@
...
+  doing so mostly through its use of
+  `stake.delegation.stake_activating_and_deactivating()`. Also, the new
+  `Delegation` struct definition (with its private field), will likely
+  impact how its being instantiated in a few areas.
+
+- **Agave:** Update the workspace dependency on
+  `solana-stake-interface`. Runtime stake processing already funnels
+  through Delegation::stake_activating_and_deactivating, so a dependency
+  bump automatically picks up the fixed-point math without touching
+  Agave code.
```
joncinque: The new calculation must only happen with the enabling of a feature gate, so we'll need to maintain some old behavior during the transition period, but we can discuss that in the implementation PR

proposals/0391-replace-stake-program-floating-point.md
```diff
@@ -0,0 +1,138 @@
...
+
+- **Firedancer:** Will also need to update their stake interface
+  dependency in lockstep with Agave.
```
joncinque: Firedancer uses their own implementation of everything, so they'll need to implement the logic outlined in the previous section

proposals/0391-replace-stake-program-floating-point.md
```diff
@@ -0,0 +1,138 @@
...
+
+1. **Unit tests:** Baseline of correctness by testing specific, known
+   scenarios and edge cases.
+2. **Differential Fuzzing (`proptest`):**
+   - An oracle function preserving the original `f64` logic will be
+     maintained for testing purposes only.
+   - The test will run the new integer implementation against the oracle
+     with millions of random inputs.
+   - Assert that the results are within a relative tolerance to account
+     for the increased precision.
```
joncinque: This is a bit vague. What is this tolerance supposed to be?

The exact difference will depend on the size of the stake and the amount on mainnet-beta, or even better, give examples of how different the calculation can be for small and large stake accounts.

But since we're essentially widening out the calculation to 128 bits instead of 64, the new calculation should only diverge past ~52 bits?

Either way, since the calculation is more accurate, this should actually give people *higher* amounts of activation / deactivation, correct?

proposals/0391-replace-stake-program-floating-point.md
Lichtso: > u128 support was quite likely to be upstreamed

There are multiple aspects to this: Being able to pass around and return `u128` etc., which could be upstreamed. And then there are u64 x u64 = u128 multiplications which are unlikely to be upstreamed.

So, in this specific case I would agree with @deanmlittle on trying to avoid the extra wide multiplications here.

grod220: Edit: noted solution within the SIMD

proposals/0391-replace-stake-program-floating-point.md
grod220: Tricky, but proposing we add deprecated `_legacy()` methods for this usecase. This means a version bump gets the new functionality automatically.

proposals/0391-replace-stake-program-floating-point.md
grod220: Adding benchmarks for the diffs.  The deviation can be higher or lower, but the diff is quite low in the vast majority of cases.

Below 2^53 (inputs and intermediate mul/div values), there is a tight agreement between floats and ints. Larger than that, `f64` starts skipping integers and rounding errors can accumulate. But still, the new impl is more precise and the diffs aren't that big.

proposals/0391-replace-stake-program-floating-point.md
```diff
@@ -0,0 +1,286 @@
...
+category: Standard
+type: Core
+status: Idea
+created: 2025-10-23
+feature: (to be assigned upon acceptance)
+---
+
+## Summary
+
+Replace all floating-point (`f64`) arithmetic within the Solana Stake Program's
```
t-nelson: rather than `f64` initially cite ieee754 double, then floating-point/double throughout

proposals/0391-replace-stake-program-floating-point.md
```diff
@@ -0,0 +1,286 @@
...
+warmup and cooldown logic with a fixed-point implementation using integer
+arithmetic. The new logic expresses the warmup/cooldown rate in basis points
+(bps) and performs all proportional stake calculations using `u128`
```
t-nelson: prefer "unsigned 128-bit integer" to `u128`. mention endianness whenever relevant (on the wire/disk)

proposals/0391-replace-stake-program-floating-point.md
```diff
@@ -0,0 +1,286 @@
...
+2. **Pinocchio migration inconsistency:** There is appetite for
+   converting the Stake Program to a highly efficient, `no_std` Pinocchio
+   implementation (reducing CU usage by +90%). These efforts are undermined by the
+   immense cost of soft-float operations. [Benchmarking
+   shows](https://solana.com/docs/programs/limitations#limited-float-support:~:text=Recent%20results%20show,Divide%20%20%20%20%20%209%20%20%20219)
+   a 22x performance penalty for a single multiplication of an `f32` versus a
+   `u64`. Using an `f64` with an operation like division is even more complex.
+   Further, doing the float migration independently allows p-stake to enforce
+   semantic equivalence for its migration.
+
```
t-nelson: i think we can focus on upstream ebpf compatibility angle here rather than rust `no_std` and pinnochio implementation details

t-nelson (commented):
firstly some basic stuff. i thought this was all documented in simd0001, but apparently that is not the case, so i'm going to elaborate here as a reference for foundation

simd's are intended to document protocol changes, not implementation changes. as such they should be written in a programming language and implementation agnostic way. imagine that you are sitting down to do a clean room implementation with only this document and knowledge of your intended target language. prefer prose. avoid code references where possible. where necessary, use pseudo code rather than a specific language

---

aside from the general corrections above and inline comments, a few things crossed my mind while reading this.
1. do we want to depend on native u128 support landing in upstream ebpf or should we have a contingency for wide/arbitrary integer arithmetic? if the latter, that would make sense as its own simd
1. i believe there are other instances of floating-point arithmetic in protocol, under consensus. would a generic "floating-point to fixed-point" migration simd be valuable to avoid needing to restate some of the details? we can just use that as a reference and stick to the specifics for each instance

grod220:
Thanks for the review @t-nelson. Overhauled this SIMD to align with the the spirit of your feedback on being implementation agnostic.

> do we want to depend on native u128 support landing in upstream ebpf or should we have a contingency for wide/arbitrary integer arithmetic?

This is currently in progress, but ~this SIMD should be on hold until the outcome of that work is merged or otherwise~ EDIT: The math requiring a 128-bit widening is necessary regardless of the environment. Support for it is already in progress upstream, but even if not, it should be emulated using 64-bit limbs.

> would a generic "floating-point to fixed-point" migration simd be valuable to avoid needing to restate some of the details

I'm not sure if that would reduce redundancy as the motivation is specific to an on-chain program wanting to be aligned with the upstream compiler reqs. Not sure if agave's motivations would be the same.

proposals/0391-replace-stake-program-floating-point.md
```diff
@@ -0,0 +1,225 @@
...
+  - Stake Merging
+  - Stake Splitting
+  - Stake Redelegation
+
+- **Validator Clients (Agave & Firedancer)**: Feature gate fixed-point math
+  in:
+  - Runtime & stake logic for stake activation and deactivation calculations
+  - Stake cache & history for effective stake derivation and history
+    aggregation
+  - Inflation rewards for calculation of stake-weighted rewards
```
topointon-jump: I think this is too vague - all of these changes are consensus-critical and are not just implementation details. I think the right balance here between specificity and client-agnosticism is to describe in detail what needs to be changed, for each bullet point here, but without client-specific references such as function names and struct definitions.

grod220: Helpful feedback. Went ahead and added that additional detail for this section. Also, for your reference, the feature gating work draft PR in agave is here: https://github.com/anza-xyz/agave/pull/9637.

proposals/0391-replace-stake-program-floating-point.md
```diff
@@ -0,0 +1,262 @@
...
+If the intermediate multiplication overflows the maximum representable
+unsigned 128-bit value, the numerator saturates to the maximum 128-bit value
+before division. The result is then clamped to `account_portion`. This ensures
+that overflow cannot amplify a stake change beyond the account's own portion
+(fail-safe rather than fail-open).
+
+### Minimum progress clamp
+
+Currently, when `account_portion > 0`, there is a granted minimum change of 1
+lamport per epoch so that small delegations do not get stuck in activating/
```
t-nelson, line 133: is this true? i thought sub-lamport rewards were deferred (`Delegation::credits_observed` not updated) until sufficient epochs passed that a full lamport would be paid out*. that is, if a an account were to be due 0.5 lamport per epoch, it would receive one lamport every other epoch.

\* however i am old and may be forgetting that this has changed at some point.

proposals/0391-replace-stake-program-floating-point.md
```diff
@@ -0,0 +1,262 @@
...
+
+In existing stake account data, there is an 8-byte field that historically
+stored warmup/cooldown rate value as a double-precision float. It is legacy
+and currently unused by any part of the program. To preserve backwards
+compatibility with existing stake account state, this SIMD does not change
+stake account layout or size. Instead, it reclassifies that field as 8 bytes
+of reserved data.
+
+The implementations should continue not using this field when computing warmup/
+cooldown values and setting it to zero when creating new stake accounts.
```
t-nelson: i don't think this belongs in this simd. the field is deprecated and, as the last sentence states, we continue not using it with the proposed changes

t-nelson (commented):
we need to be clear of signededness of arithmetic throughout

proposals/0391-replace-stake-program-floating-point.md
grod220: Removed

proposals/0391-replace-stake-program-floating-point.md
grod220: These are two different mechanisms:

- Warmup/cooldown (related to this SIMD) uses a .max(1) clamp to prevent tiny delegations from never activating/deactivating: [state.rs reference](https://github.com/solana-program/stake/blob/0e030308c9091a706070f2eef9fe062b42169d1f/interface/src/state.rs#L669-L670).
- Rewards (not covered by this SIMD) uses the deferral behavior you are referencing [inflation_rewards/mod.rs](https://github.com/anza-xyz/agave/blob/d34495da06e37442bc0b8aaa7515a92db1217116/runtime/src/inflation_rewards/mod.rs#L195-L200) so sub-lamport rewards won't update `credits_observed`. 

I'll add a specific note to this SIMD about this to clarify the distinction.

grod220 (commented):
Also added clarification on the signedness of the integers throughout.

proposals/0391-replace-stake-program-floating-point.md
```diff
@@ -0,0 +1,258 @@
...
+```
+
+Note: The division uses unsigned integer division so it truncates (rounds down).
+
+#### Widening arithmetic to 128-bit integers
+
+All inputs are unsigned 64-bit lamport quantities. Because of the extra
+multiplication, all values used in the formula should be widened to unsigned
+128-bit integers (or an exact emulation). The final value should then be cast
+back down to an unsigned 64-bit integer.
```
t-nelson:
```suggestion
back down to an unsigned 64-bit integer, truncating if necessary.
```

or should we fault?

proposals/0391-replace-stake-program-floating-point.md
```diff
@@ -0,0 +1,258 @@
...
+    (cluster_portion * BASIS_POINTS_PER_UNIT)
+```
+
+Note: The division uses unsigned integer division so it truncates (rounds down).
+
+#### Widening arithmetic to 128-bit integers
+
+All inputs are unsigned 64-bit lamport quantities. Because of the extra
+multiplication, all values used in the formula should be widened to unsigned
+128-bit integers (or an exact emulation). The final value should then be cast
```
t-nelson:
```suggestion
multiplication, all values used in the formula MUST be widened to unsigned
128-bit integers (or an exact emulation). The final value MUST then be cast
```

we're under consensus. we can't be offering suggestions. see [rfc2119](https://datatracker.ietf.org/doc/html/rfc2119). make any similar changes throughout

proposals/0391-replace-stake-program-floating-point.md
```diff
@@ -0,0 +1,258 @@
...
+back down to an unsigned 64-bit integer.
+
+Implementations that do not offer native unsigned 128-bit arithmetic must
+emulate it (for example via fixed-width limb arithmetic).
+
+#### Saturation and fail-safe behavior
+
+If the intermediate multiplication overflows the maximum representable
+unsigned 128-bit value, the numerator saturates to the maximum 128-bit value
```
t-nelson: we should specify saturating arithmetic throughout. justification can be provided in this section

proposals/0391-replace-stake-program-floating-point.md
```diff
@@ -0,0 +1,258 @@
...
+    - Export new integer-based stake activation and deactivation logic for rust
+      consumers
+    - Deprecate the floating-point rate field while preserving binary layout
+      compatibility
+
+- **Stake Program**: Feature gate v2 interface helpers in:
+    - **Stake Merging**: Stake calculations are used to determine if the
+      account is in a transient state, ensuring that merges are rejected if the
+      account is not effectively fully active or inactive.
+    - **Stake Splitting**: Stake calculations are used determine if the source
```
t-nelson:
```suggestion
    - **Stake Splitting**: Stake calculations are used to determine if the source
```

t-nelson (commented):
getting close now by my eye. probably worth grabbing fd/blueshift reviewers

proposals/0391-replace-stake-program-floating-point.md
grod220: Imperatives corrected 👍

proposals/0391-replace-stake-program-floating-point.md
grod220: Given the clamp to `account_portion` beforehand, the 128-to-64-bit cast will never truncate. Also, the saturations prevent the fault behavior. All of this should be clarified, will update this section.

grod220 (commented):
Will chase down additional reviewers 👍

deanmlittle (dismissed):
thanks for addressing all of the feedback here. lgtm.

t-nelson (dismissed):
lgtm now thanks!

cc/ @bw-solana for viz

proposals/0391-replace-stake-program-floating-point.md
```diff
@@ -0,0 +1,271 @@
...
+## Detailed Design
+
+### Rate representation (basis points)
+
+The current network warmup/cooldown rate is 9%. This means that, in any given
+epoch, at most 9% of the previous epoch's effective stake can be activated or
+deactivated.
+
+Currently, this figure is represented in floating-point: `0.09`. The new
+representation is an integer of basis points: `900`.
```
topointon-jump, line 73: Does this SIMD include any change to the value of `warmup_cooldown_rate` stored on-chain in the stake accounts? I don't think it does but could we clarify this here 🙏

proposals/0391-replace-stake-program-floating-point.md
```diff
@@ -0,0 +1,271 @@
...
+
+    # Narrow back to unsigned 64-bit integer
+    allowed_change = allowed_change_128 as unsigned 64-bit integer
+
+    # Minimum progress clamp
+    if allowed_change == 0:
+        return 1
+
+    return allowed_change
+```
```
topointon-jump: Can we use sat_mul_u128/sat_div_u128 functions in these pseudocode blocks? To make it clear that we are using saturating operations.

Would also be great to clarify the signedness and size of each of the variables, maybe declaring them with "types" such as `u64` `u128` etc.

proposals/0391-replace-stake-program-floating-point.md
```diff
@@ -0,0 +1,271 @@
...
+      delegations pointing to that account. These totals determine leader
+      schedule weights and fork choice voting power.
+    - **Inflation Rewards**: Reward calculation iterates over each epoch in a
+      vote account's credit history. For each epoch, the validator _computes the
+      delegation's effective stake_ at that epoch, multiplies by the earned vote
+      credits to produce points and accumulates these across epochs. The final
+      reward is proportional to the delegation's share of total cluster points.
+        - Note: Only the effective stake computation (warmup/cooldown) is
+          affected by this SIMD. The downstream reward-to-lamport conversion
+          and sub-lamport deferral logic remain unchanged.
```
topointon-jump, line 270: I had a look, and from what I can tell the point of divergence for all of these comes down to `stake_and_activating`.

Do you mind:
a) specifying whether this is true or not, and if not which other calculations need to change? I don't mean passing down in the call stack, but places where the actual calculation needs to be updated. I think it's just `stake_and_activating` but would be good to confirm.
b) outlining pseudocode for the fixed-point version of `stake_and_activating` (`stake_and_activating_v2` in Agave) and how this differs from the current version of `stake_and_activating`

proposals/0391-replace-stake-program-floating-point.md
grod220: Originally I had a section about this, but was requested to remove it due to irrelevance as that field is deprecated and continues to be: https://github.com/solana-foundation/solana-improvement-documents/pull/391#discussion_r2637980719

proposals/0391-replace-stake-program-floating-point.md
grod220: There was specific feedback to move away from u64/u128 usage [here](https://github.com/solana-foundation/solana-improvement-documents/pull/391#pullrequestreview-3532085803) and [here](https://github.com/solana-foundation/solana-improvement-documents/pull/391#discussion_r2582775262). Trying my best to figure out how to avoid rust-specific terms yet remain understandable. But I agree, this pseudocode could be better. Think adding better pseudocode conventions can help. Will update.

proposals/0391-replace-stake-program-floating-point.md
grod220: > the point of divergence for all of these comes down to stake_and_activating

It comes down to updating+feature gating two paths. We versioned the functions [in the PR](https://github.com/solana-program/stake/pull/152) w/ `v2`:
- Activation path:`stake_and_activating_v2`
- Deactivation path: `stake_activating_and_deactivating_v2`

Down the call stack, the v2 versions eventually call `rate_limited_stake_change`, where the fixed point math updates are.

> outlining pseudocode for the fixed-point version of stake_and_activating (stake_and_activating_v2 in Agave) 

The SIMD already provides the pseudocode for `rate_limited_stake_change`, which is the actual mathematical formula that needs to change. That's the spec-level detail.

`stake_and_activating` / `stake_and_activating_v2` are stake program/agave-specific functions that are responsible for quite a bit more (looping through epochs, checking history, etc). It's just their use specifically of calculating allowed stake change (internal or helper) that is relevant to the SIMD and needs feature gating.

proposals/0391-replace-stake-program-floating-point.md
topointon-jump: imo we should be able to implement this function just from looking at the pseudocode. some form of types would make that a lot easier

grod220: Let me know what you think about the most recent update. The `### Pseudocode conventions` section establishes types / functions used later on to hopefully make things clearer.

topointon-jump (commented):
Looks good to me - could we just wait for @ibhatt-jumptrading to have a look as well from the FD side 🙏

topointon-jump (approved):
Looks good to me and @ibhatt-jumptrading!

t-nelson (approved):

PR #392, opened 2025-10-30
SIMD-0392: Relaxation of post-execution min_balance check

igor56D:


proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,178 @@
...
+deprecated.
+
+### Proposed behavior
+
+For all write-locked accounts, post-execution account balance checks MUST
+verify:
+
+```
+min_balance = calculate_min_balance(acc.post_exec_size)
+if acc.pre_exec_balance > 0 and acc.post_exec_size == acc.pre_exec_size:
```
roryharr: Does this mean it could cost lamports to shrink an account? 
```
acc.post_exec_size == acc.pre_exec_size
```

What about
```
acc.post_exec_size <= acc.pre_exec_size
```
instead?

igor56D: Yes, min_balance might increase after shrinking an account if rent has increased enough. Unideal but it's closer to the current behavior and therefore less likely to break existing programs IMO. For example, a program might have a `realloc(acc, new_size)` function that uses the new size and the current rent to determine the min_balance requirement that doesn't distinguish between upwards and downwards reallocs. This example wouldn't break but the protocol behavior diverges from the program behavior, which is slightly concerning to me.

It's also just simpler to model "since the last alloc" than "since the last upwards realloc". That said, I'm open to changing my mind.

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,178 @@
...
+    min_balance = min(min_balance, acc.pre_exec_balance)
+
+assert(acc.post_exec_balance == 0 or acc.post_exec_balance >= min_balance)
```
t-nelson:
```suggestion
rent_exempt_min_balance = calculate_min_balance(acc.post_exec_size)
if acc.pre_exec_balance > 0 and acc.post_exec_size == acc.pre_exec_size:
    min_allowed_balance = min(rent_exempt_min_balance, acc.pre_exec_balance)

assert(acc.post_exec_balance == 0 or acc.post_exec_balance >= min_allowed_balance)
```

nit: avoid aliasing symbol names for clarity

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,178 @@
...
+```
+
+The check ensures that an account's balance is always at least the minimum
+rent_per_byte price since the last allocation multiplied by the account
+size:
```
t-nelson:
```suggestion
The check ensures that an account's balance is always rent-exempt as of
the last (re)allocation its data:
```

i'm not sure reiterating the rent-exempt min balance calculation here is helpful

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,178 @@
...
+- The first subsequent transaction write-locking the same account but not
+  performing any reallocations can either have (1) a balance higher than the
+  original min_balance or (2) a lower balance that's bounded below by a reduce
+  rent price.
+- With this, we can inductively prove that the post-execution balance of
+  every account is bounded below by the lowest rent since the last allocation.
+
+### Implementation details
+
+- The pre-execution balance MUST be captured before any instruction execution
```
t-nelson: earlier. we need to capture before fee collection

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,178 @@
...
+- No change to existing behavior
+
+**No size change:**
+
+- `post_exec_size == pre_exec_size`
+- The `min()` clause applies: `min_balance = min(calculate_min_balance(size),
+  pre_exec_balance)`
+- **This is the key behavioral change**: allows accounts to retain their
+  original rent price even if current rent increases
+
```
t-nelson, line 149: it may be worth discussing lamports balance changes too. especially the case where a credit increases a formerly rent-exempt account's balance to a value that is still less than the current rent-exempt reserve balance

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,178 @@
...
+- Enables safe rent increases without disrupting existing accounts: only newly
+  allocated accounts are subject to new rent prices.
+- Maintains the invariant that accounts always satisfy the minimum rent price
+  since the time of their most recent allocation.
+- No mechanism for forcing accounts to "upgrade" to current rent prices without
+  reallocation; this is intentional to preserve non-disruptiveness.
+- It's essentially free to adopt a new rent rate that's lower than the rent rate
+  an account was subject to at allocation time. This isn't problematic in itself
+  but it may cause users to be less sensitive to rent increases.
+
```
t-nelson, line 204: there's also potential for a "land grab" in advance of a rent increase and creation of secondary state markets which would be worth discussing. eg. should we consider enforcing current rent exemption on accounts whose owning program has changed in addition to on reallocation?

t-nelson (commented):
a few nits, but lgtm assuming we choose to go the grandfather route. i'm not totally sold that the disruption will be so big a deal

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,178 @@
...
+The check ensures that an account's balance is always at least the minimum
+rent_per_byte price since the last allocation multiplied by the account
+size:
+
+- As the base case, it's clear this holds
+  for new account creations or re-allocations: the post-execution rent price
+  and account size is used to determine the minimum balance.
+- The first subsequent transaction write-locking the same account but not
+  performing any reallocations can either have (1) a balance higher than the
+  original min_balance or (2) a lower balance that's bounded below by a reduce
```
brooksprumo: nit:
```suggestion
  original min_balance or (2) a lower balance that's bounded below by a reduced
```

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,178 @@
...
+  begins, based on the account's loaded state.
+- The pre and post-execution sizes are compared to determine if reallocation
+  occurred.
+- The `calculate_min_balance()` function uses the current rent_per_byte value,
+  which may vary based on active feature gates and rent policies.
+- Newly created accounts have `pre_exec_balance = 0` and are not subject to the
+  `min()` clause.
+- As before, 0 post-balance is allowed and equivelent to closing an account.
+
+### Edge cases
```
brooksprumo, line 122: If we raise rent, how does/should delegated stake amount change?

IIUC[^1], we compute the delegated stake amount based on the account's balance minus the rent exempt minimum. Thus, if we raise the rent, will that change delegated stake amounts? We probably want the delegated stake amount to remain constant; do we need to do any special handling here?

(Related, is if the delegation amount were to drop the minimum (which is currently only 1 lamport).)

[^1]: Note that I haven't looked at the stake code in a while, so I reserve the right to be wrong!

t-nelson: i think it Just Works :tm:. the rent-exempt reserve lamports are stored explicitly and there are no data resize operations without deactivation+close. so the reserve balance snapshot at invocation of `StakeInstruction::Initialize` should remain in line with the rules of this proposed change

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,178 @@
...
+## Summary
+
+To allow for non-disruptive rent increases, relax post-execution account
+minimum balance checks. The new lower-bound on post-exec balance is
+`min(acc.pre_exec_balance, cur_rent_per_byte * acc.size())`.
+
+This maintains the invariant that every account has a balance at or above the
+minimum balance requirement calculated at any point since the most recent
+allocation occurred. When enabled, only newly allocated accounts will be
+subject to rent increases.
```
2501babe, line 22: changing rent has a number of stake-related implications we will have to carefully consider how to handle

`Meta` contains the field `rent_exempt_reserve`, which we set during stake account initialization. we use `rent_exempt_reserve` and `Rent::minimum_balance()` interchangably because stake accounts have a fixed size of 200 and we assume neither value ever changes. this affects us on rent increases _or_ decreases. we could plausibly rewrite them for active/activating stakes when processing rewards and have the stake program update them when calling `DelegateStake`. alternatively we could introduce a `StakeStakeV3` that removes it

this is the only issue i can think of with a rent decrease. `Delegation` would be fine because the newly non-rent lamports would just be free lamports as if they were transferred in

increasing rent, however, has additional implications. we would have to be sure all code (agave, stake program, stake pools, third party projects like sanctum and marinade) can handle situations where an active stake account is brought to an effective delegation of 0 or an effective delegation below the minimum delegation (presently 1 lamport but expected to increase to 1 sol). this could be particularly problematic for downstream projects if the stake program needed to block operations like `Split`/`Merge` on stake accounts that used to be good

also we would need to rewrite `Delegation.stake` values for all active/activating stake accounts, again plausibly during the rewards period. no code is designed to expect that `lamports - rent - delegation` may ever be less than 0. unlike `rent_exempt_reserve` we could never remove the delegation value from `StakeState` because we need to track how much value was in the account at the time of delegation

however...

2501babe: ...im using a new comment because i realized while writing all that maybe we can just treat active/activating stake accounts as _still good_ instead of trying to true them up. keep `rent_exempt_reserve` the old number and pretend that is the rent. let the "new rent" lamports continue to act as delegated stake. rewrite `rent_exempt_reserve` in `DelegateStake` for already-initialized accounts and block delegation if it doesnt satisfy the new rules

there are still some weird edge cases (not being able to `Withdraw` "free" lamports [according to `rent_exempt_reserve`], possible issues with `Split`) but i think most of these go away if we increase the minimum delegation to 1 sol, since this would give us more than enough delegation to cover up for the runtime-enforced higher rent assuming we arent planning to 50x it :sweat_smile:

proposals/0392-relax-minimum-balance-check.md
2501babe: oh i could have just read your comment before writing an essay :laughing: 

yea i think this will require a little stake/rewards work but should be fine, the main things are:
1. never use `Rent` sysvar for live accounts (we already do this in stake but would need to check the stake pools)
2. rewrite `rent_exempt_reserve` when delegating instead of just when initializing
3. pay rewards using `rent_exempt_reserve` and `Delegation` as if new rent is still stake
4. `Merge` is fine just keep the `rent_exempt_reserve` of the destination
5. `Withdraw`, probably let people withdraw free lamports in live accounts down to `rent_exempt_reserve`
6. some possible edge cases in `Split`, `MoveStake`, `MoveLamports` for accounts in the neighborhood of 228288 lamports, we can handle them or we can raise minimum delegation to 1 sol and obviate them

t-nelson: > oh i could have just read your comment before writing an essay 😆

:joy: 

> yea i think this will require a little stake/rewards work but should be fine, the main things are:
> 
>     1. never use `Rent` sysvar for live accounts (we already do this in stake but would need to check the stake pools)
> 
>     ...
> 
>     3. pay rewards using `rent_exempt_reserve` and `Delegation` as if new rent is still stake
> 
>     4. `Merge` is fine just keep the `rent_exempt_reserve` of the destination
> 

yeah i think so long as we only sample `Rent` sysvar on `Initialize` wrt to stake accounts, and adhere to that sampled value for the

>     2. rewrite `rent_exempt_reserve` when delegating instead of just when initializing
> 

not opposed, but i don't think it would be strictly necessary to adhere to this proposal?

>     5. `Withdraw`, probably let people withdraw free lamports in live accounts down to `rent_exempt_reserve`
> 

that's how it already works, right? no change necessary

>     6. some possible edge cases in `Split`, `MoveStake`, `MoveLamports` for accounts in the neighborhood of 228288 lamports, we can handle them or we can raise minimum delegation to 1 sol and obviate them

i think `Split` is fine now that destination is required to be pre-funded for rent. it'll just sample `Rent` sysvar on `Initialize` as usual. i'll need a refresher on the logic for `MoveStake` and `MoveLamports`, but so long as they respect `Meta::rent_exempt_reserve`, should be no change required

2501babe: i think the important point for the proposal is 3, rewards should be paid out according to delegation even if theyre required for "new" rent. the rest are more stake program implementation details to mitigate downsides, eg 2 would ensure you eventually give the new rent up to rent if youre actively managing the account

under 6 `Split` et al would fail for very small stake accounts, eg you split a small amount to an account that met the _old_ rent exemption, the proposed change would block the transaction at the runtime level. but we can handle it at the program level

i dont think the proposal has any blockers from stake accounts as long was we honor the delegation regardless of actual rent

igor56D: > 5. Withdraw, probably let people withdraw free lamports in live accounts down to rent_exempt_reserve

The balance after execution cannot be lower than both the pre-exec balance and the *current* rent price. IOW the max partial withdrawal amount is `max(0, stake_acc.lamports() - current_rent)`, so withdrawing `stake_acc.lamports() - rent_exempt_reserve` will fail if `rent_exempt_reserve < current_rent`.

igor56D: I didn't say this explicitly in the proposal, but the goal can be narrowed somewhat to: allow non-disruptive rent increases back to the legacy rent-exempt value.

Given that, couldn't this all be solved by using the legacy rent exempt value for `rent_exempt_reserve`?  This will work with all existing stake accounts and ensures rent_exempt_reserve is always greater than the current rent.

2501babe: right, i mean whether to allow withdrawls from an active stake account that would shift more of the delegation to covering new rent obligations while continuing to serve as delegation

but yea this is an implementation detail of stake program, i just mean to say stake program will be fine as long as agave does 3 instead of modifying `delegation.stake` and/or `meta.rent_exempt_reserve`

proposals/0392-relax-minimum-balance-check.md
igor56D: I'm somewhat weary of directly encoding more edge-cases into these checks unless we're confident that it nearly completely prevents secondary markets from developing. Otherwise, it becomes a game of whack-a-mole that makes the protocol spec more convoluted. Additional checks are also more likely to break existing program logic.

Rent increases *should* be very infrequent, so the chances this will be a problem should be very low. Having a good plan for a prompt rent increase is more important imo.

What do you think?

proposals/0392-relax-minimum-balance-check.md
igor56D: sounds good. Does this require any modifications to this SIMD? Is another SIMD needed? Not sure what the protocol is for changes to core programs.

proposals/0392-relax-minimum-balance-check.md
t-nelson: i think the owning program check would severely impair secondary market supply. like today i could just deploy a generic program to create accounts of arbitrary size and supports a transfer mechanism with a result similar to `SystemInstruction::CreateAccount`. with the owning program check you'd have to actually pre-allocate accounts with the existing program which may or may not be more costly depending on the implementation

proposals/0392-relax-minimum-balance-check.md
igor56D: gave this some more thought and decided to use your suggestion. It's strictly more permissive so it shouldn't cause any more failures + it's guaranteed to not allow new state with below market rent.

proposals/0392-relax-minimum-balance-check.md
t-nelson: probably worth reviewing all of the stuff that touches stake accounts to ensure it respects `Meta::rent_exempt_reserve`. so long as that's the case no further simd required besides maybe an advisory that `Meta::rent_exempt_reserve` supersedes `Rent` sysvar so long as it is set (everwhere except `StakeInstruction::Initialize`)

proposals/0392-relax-minimum-balance-check.md
igor56D: What you're suggesting doesn't seem quite right. Updated to:
```
rent_exempt_min_balance = calculate_min_balance(acc.post_exec_size)
if acc.pre_exec_balance > 0 and acc.post_exec_size <= acc.pre_exec_size:
    min_allowed_balance = min(rent_exempt_min_balance, acc.pre_exec_balance)
else:
    min_allowed_balance = rent_exempt_min_balance

assert(acc.post_exec_balance == 0 or acc.post_exec_balance >= min_allowed_balance)
```

lmk if that's more clear.

proposals/0392-relax-minimum-balance-check.md
igor56D: 3019e59fb3328b387b8402a97743cc6f346f6f45

proposals/0392-relax-minimum-balance-check.md
igor56D: 3019e59fb3328b387b8402a97743cc6f346f6f45

proposals/0392-relax-minimum-balance-check.md
igor56D: 3019e59fb3328b387b8402a97743cc6f346f6f45

proposals/0392-relax-minimum-balance-check.md
igor56D: 3019e59fb3328b387b8402a97743cc6f346f6f45

proposals/0392-relax-minimum-balance-check.md
igor56D: 3019e59fb3328b387b8402a97743cc6f346f6f45

Added:

> Note that if the balance increases but is still below the current rent price,
  the new balance becomes the effective minimum balance for the given account.
  This means that the balance can no longer be reduced back to the original
  rent price.

proposals/0392-relax-minimum-balance-check.md
igor56D: Agreed, this fragments the unified global secondary market into many per-owner markets, which I expect would result in lower aggregate volume. My concern is that this will break existing program logic where ownership transfer doesn't include a balance check/top-off. In any case, it's another thing app devs need to be aware of when writing programs.

There's a tradeoff and I'm not personally knowledgeable enough on the program landscape to say confidently in what direction the tradeoff leans.

fwiw, an integral controller largely mitigates this: the land grab would just increase the accumulator faster, causing the rent increase to occur sooner. In other words, we don't have to worry about state growth exploding too quickly before we can respond. A secondary market *may* even be valuable here because it incentivizes people to sell lower value account space, but that's very speculative. I know you're skeptical of the controller approach but I figured it was worth mentioning since it's relevant to this topic.

proposals/0392-relax-minimum-balance-check.md
igor56D: 3019e59fb3328b387b8402a97743cc6f346f6f45

proposals/0392-relax-minimum-balance-check.md
t-nelson: > What you're suggesting doesn't seem quite right

at least the motivation was proved correct :wink:

proposals/0392-relax-minimum-balance-check.md
t-nelson: man dapp devs barely support closing accounts, let alone transferring them to different programs :joy:

proposals/0392-relax-minimum-balance-check.md
igor56D: @t-nelson regarding:

> I didn't say this explicitly in the proposal, but the goal can be narrowed somewhat to: allow non-disruptive rent increases back to the legacy rent-exempt value.

> Given that, couldn't this all be solved by using the legacy rent exempt value for rent_exempt_reserve? This will work with all existing stake accounts and ensures rent_exempt_reserve is always greater than the current rent.

Adding an explicit `MAX_RENT_EXEMPT_MINIMUM_PER_BYTE` and setting it to the current rent value would simplify things. What do you think? Do we need to reserve the option to increase the rent beyond the current value?

2501babe: >I didn't say this explicitly in the proposal, but the goal can be narrowed somewhat to: allow non-disruptive rent increases back to the legacy rent-exempt value.

i missed the second comment when writing my first, adding here what i said in dms: if we declare current rent to be a ceiling on future rent increases, stake program can just use that maximum for `rent_exempt_reserve` and this simd doesnt need to have any special language about stake accounts

incidentally wsol accounts also carry a rent-exemption in account data. we can also set it to the max but maybe we can just deprecate and ignore it in the program and use `Rent` itself. this change:

>Note that if the balance increases but is still below the current rent price, the new balance becomes the effective minimum balance for the given account.

obviates the only annoying edge case, ie adding sol to a wsol account that is below rent-exemption both before and after the transaction. im pretty sure the only reason accounts carry this value is because the instructions couldnt get `Rent` via syscall

t-nelson: > im pretty sure the only reason accounts carry this value is because the instructions couldnt get Rent via syscall

that would be a good reason, however i believe the actual reason is "because stake accounts do it" :melting_face: 

> Adding an explicit MAX_RENT_EXEMPT_MINIMUM_PER_BYTE and setting it to the current rent value would simplify things. What do you think? Do we need to reserve the option to increase the rent beyond the current value?

sorry adding where? generally speaking these constants are not constant (they are a bug) and any additional changes we make are taking this proposal away from the simplest form afaict

igor56D: confusing wording on my part. We don't need to add it to the protocol in this proposal, but change the goal of the proposal from:

- allow non-disruptive rent increases (to any rent value)

to:

- allow non-disruptive rent increases to the legacy rent-exempt value

So, the rent-exempt value at any time in the future is guaranteed to be less than or equal to `MAX_RENT_EXEMPT_MINIMUM_PER_BYTE`. That way we can alway assume the transferable balance of an account is `>= balance - max_rent`.

brooksprumo:
Hi, checking in on the status of the SIMD. Is it ready for another round of reviews? (If yes, I suggest requesting re-reviews from folks.) Has there been any testing on a local cluster to mimic a rent increase, esp w.r.t. stake accounts?

igor56D:
> Hi, checking in on the status of the SIMD. Is it ready for another round of reviews? (If yes, I suggest requesting re-reviews from folks.) Has there been any testing on a local cluster to mimic a rent increase, esp w.r.t. stake accounts?

1. I'm in the process of wrapping up the implementation for this and 0389. I'll push it up later today as a draft PR. Also working on setting up tests for a local cluster, which should be ready shortly as well.
2. I think the main open question right now is how (if at all) any core/system programs need to be updated as a result of this change. The stake program may require some tweaks to avoid using the rent sysvar for determining delegation amount. I believe @2501babe is looking into potential changes in the SPL token programs.

roryharr (approved):

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,183 @@
...
+## Detailed Design
+
+### Current behavior
+
+For all write-locked accounts, post-execution account balance checks
+currently verify:
+
+```
+min_balance = calculate_min_balance(acc.data_size_post)
+assert(acc.post_exec_balance == 0 or acc.post_exec_balance >= post_min_balance)
```
2501babe:
```suggestion
min_allowed_balance = calculate_min_balance(acc.post_exec_size)
assert(acc.post_exec_balance == 0 or acc.post_exec_balance >= min_allowed_balance)
```

use the same pseudovariable names as below

2501babe:
spl-token, spl-token-2022, and p-token have no issues, we would do similar to stake program and hardcode the maximum rent. im not aware of any programs other than these and stake that are affected. we can start the change process whenever this is accepted

we should also put that maximum in the simd, maybe under "new terminology" or "proposed changes"

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,183 @@
...
+
+- The pre-execution balance MUST be captured before any state is modified
+  (e.g. before fee collection, instruction execution, etc).
+- The pre and post-execution sizes are compared to determine if reallocation
+  occurred.
+- The `calculate_min_balance()` function uses the current rent_per_byte value,
+  which may vary based on active feature gates and rent policies.
+- Newly created accounts have `pre_exec_balance = 0` and are not subject to the
+  `min()` clause.
+- As before, 0 post-balance is allowed and equivelent to closing an account.
```
2501babe:
```suggestion
- As before, 0 post-balance is allowed and equivalent to closing an account.
```

proposals/0392-relax-minimum-balance-check.md
igor56D: 95ae824e092c36e0fa2dfd178f40cb0852677b8e

proposals/0392-relax-minimum-balance-check.md
igor56D: added rent maximum caveat: 95ae824e092c36e0fa2dfd178f40cb0852677b8e

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,208 @@
...
+authors:
+  - Igor Durovic (anza)
+category: Standard
+type: Core
+status: Idea
+created: 2025-10-30
+feature: (fill in with feature key and github tracking issues once accepted)
+supersedes:
+superseded-by:
+extends:
```
brooksprumo:
```suggestion
```

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,208 @@
...
+---
+
+## Summary
+
+To allow for non-disruptive rent increases, relax post-execution account
+minimum balance checks. The new lower-bound on post-exec balance is
+`min(acc.pre_exec_balance, cur_rent_per_byte * acc.size())`.
```
brooksprumo: Can we reuse the fn defined down in New Terminology here too?


```suggestion
`min(acc.pre_exec_balance, calculate_min_balance(acc.size()))`.
```

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,208 @@
...
+## New Terminology
+
+- `calculate_min_balance(acc_size) = acc_size * current_rent_per_byte`.
+  Note, `acc_size` includes both the account data size and the 128 overhead
+  bytes.
+- `legacy_rent_per_byte`: the fixed rent-exempt per-byte rate used prior to this
+  proposal.
+
+  ```
+  bytes = ACCOUNT_STORAGE_OVERHEAD + acc.data.len;
```
brooksprumo: nit:
```suggestion
  acc_size = ACCOUNT_STORAGE_OVERHEAD + acc.data.len;
```

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,208 @@
...
+   // 3,480.9112548828 on mainnet
```
brooksprumo: 1. leading space
2. mainnet-beta
3. units
4. isn't the value 6,960?


```suggestion
  // 6,960 lamports on mainnet-beta
```

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,208 @@
...
+
+### Implementation details
+
+- The pre-execution balance MUST be captured before any state is modified
+  (e.g. before fee collection, instruction execution, etc).
+- The pre and post-execution sizes are compared to determine if reallocation
+  occurred.
+- The pre-execution owner MUST be captured. If the account owner changes during
+  execution, the `min(pre_exec_balance, …)` clause MUST NOT apply; enforce the
+  current rent-exempt minimum for the post-exec size.
```
brooksprumo, line 115: Why must owner changes now use the rent_exempt_min_balance instead of the min()?

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,208 @@
...
+
+assert(acc.post_exec_balance == 0 or acc.post_exec_balance >= min_allowed_balance)
+```
+
+The check ensures that an account's balance is always rent-exempt as of
+the last (re)allocation of its data:
+
+- As the base case, it's clear this holds
+  for new account creations or re-allocations: the post-execution rent price
+  and account size is used to determine the minimum balance.
```
brooksprumo: The proposed behavior above for re-allocations that strictly shrink the account has different behavior that described here.

I'm not sure why we should special case the shrinking, but regardless, this text and the proposed behavior above should match IMO.

proposals/0392-relax-minimum-balance-check.md
igor56D: You're right, good catch.

proposals/0392-relax-minimum-balance-check.md
igor56D: It's intended to prevent a secondary market from developing: [context](https://github.com/solana-foundation/solana-improvement-documents/pull/392#discussion_r2487933240).

proposals/0392-relax-minimum-balance-check.md
igor56D: Added the special case shrinking after reviews and forgot to update this part ([context](https://github.com/solana-foundation/solana-improvement-documents/pull/392#discussion_r2479798341)). The basic idea is that we want to prevent the case where shrinking increases the overall rent for an account. Will update.

proposals/0392-relax-minimum-balance-check.md
igor56D: 830d5b412fca7e2be7ba1a9d90bd99159e67225b

proposals/0392-relax-minimum-balance-check.md
igor56D: 830d5b412fca7e2be7ba1a9d90bd99159e67225b

proposals/0392-relax-minimum-balance-check.md
igor56D: 830d5b412fca7e2be7ba1a9d90bd99159e67225b

proposals/0392-relax-minimum-balance-check.md
igor56D: fixed: 830d5b412fca7e2be7ba1a9d90bd99159e67225b

proposals/0392-relax-minimum-balance-check.md
igor56D: added context to the proposal: 830d5b412fca7e2be7ba1a9d90bd99159e67225b

proposals/0392-relax-minimum-balance-check.md
igor56D: updated: 830d5b412fca7e2be7ba1a9d90bd99159e67225b

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,211 @@
...
+
+- `calculate_min_balance(acc_size) = acc_size * current_rent_per_byte`.
+  Note, `acc_size` includes both the account data size and the 128 overhead
+  bytes.
+- `legacy_rent_per_byte`: the fixed rent-exempt per-byte rate used prior to this
+  proposal.
+
+  ```
+  acc_size = ACCOUNT_STORAGE_OVERHEAD + acc.data.len;
+  // 6,960 lamports on mainnet
```
brooksprumo:
```suggestion
  // 6,960 lamports on mainnet-beta
```

proposals/0392-relax-minimum-balance-check.md
brooksprumo: got it, thanks!

proposals/0392-relax-minimum-balance-check.md
igor56D: woops...

igor56D: d39f83f3ade3e9027b653bc753da692d1be4631e

t-nelson:
> ... assuming we choose to go the grandfather route.

when/where was this decided?

igor56D:
> > ... assuming we choose to go the grandfather route.
> 
> when/where was this decided?

When this proposal is merged, I guess. If there are alternatives you still think may be better then we can discuss them here.

t-nelson:
it's in the discussion https://github.com/solana-foundation/solana-improvement-documents/discussions/399#discussioncomment-14926957

igor56D:
> it's in the discussion [#399 (comment)](https://github.com/solana-foundation/solana-improvement-documents/discussions/399#discussioncomment-14926957)

We can continue the discussion there if you prefer. I believe I was the last person to comment on the topic ([ref](https://github.com/solana-foundation/solana-improvement-documents/discussions/399#discussioncomment-14928828)).

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,211 @@
...
+```
+rent_exempt_min_balance = calculate_min_balance(acc.post_exec_size)
+if acc.pre_exec_balance > 0
+   and acc.post_exec_size <= acc.pre_exec_size
+   and acc.post_exec_owner == acc.pre_exec_owner:
+    min_allowed_balance = min(rent_exempt_min_balance, acc.pre_exec_balance)
+else:
+    min_allowed_balance = rent_exempt_min_balance
+
+assert(acc.post_exec_balance == 0 or acc.post_exec_balance >= min_allowed_balance)
```
topointon-jump, line 87: Instead of `assert` could we specify:
- Where this check will happen (I'm assuming transaction finalization)
- What error code will be thrown if the check fails

igor56D: There's an existing post-execution minimum balance check. This proposal modifies that check so no new error codes are necessary.

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,211 @@
...
+  bytes.
+- `legacy_rent_per_byte`: the fixed rent-exempt per-byte rate used prior to this
+  proposal.
+
+  ```
+  acc_size = ACCOUNT_STORAGE_OVERHEAD + acc.data.len;
+  // 6,960 lamports on mainnet-beta
+  legacy_rent_per_byte = LAMPORTS_PER_BYTE_YEAR * EXEMPTION_THRESHOLD
+  legacy_min_balance = legacy_rent_per_byte * acc_size
+  ```
```
topointon-jump, line 52: Will this be part of the `calculate_min_balance` calculation? I don't see this `legacy_min_balance` used anywhere else in this proposal - am I missing something?

igor56D: you're not missing anything, I just included this to show how the min balance is currently calculated with references to the relevant constants in case people weren't familiar. Not really relevant to the rest of the content.

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,213 @@
...
+
+## Impact
+
+- Dapp developers: Enables non-disruptive rent increases. Existing accounts are
+  grandfathered at their creation-time rent price until they increase in size. New
+  accounts and accounts that have increased in size adopt the current rent price.
+  Minimal behavior change for typical applications; most applications do not
+  reduce account balances below their minimum during normal operation.
+- Validators: Minimal performance impact; requires storing pre-execution balance
+  (one u64) and size per write-locked account during transaction execution.
```
topointon-jump, line 183: Minor comment, but we'll also have to look up/store the `pre_exec_owner`

topointon-jump (approved):

bw-solana:
Waiting on 2 items before clicking the button for Anza:
1. Sign-off by someone from Anza runtime team (@2501babe or @buffalojoec )
2. Confirmation from @t-nelson that we're okay w/ the "grandfather accounts in" approach

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,213 @@
...
+approach achieves a similar goal with only runtime logic changes by using the
+pre-execution balance as a proxy for the minimum rent price since last
+allocation. The benefit of a dedicated metadata field is that the invariant
+can be made stricter: every account's balance is bounded below by the rent
+price at the most recent allocation rather than the minimum rent price *since*
+the last allocation.
+
+## Impact
+
+- Dapp developers: Enables non-disruptive rent increases. Existing accounts are
```
t-nelson, line 177: since rent will become dynamic, dapp developer must also ensure that they are querying the current rent price from the cluster rather than using constants directly. the latter is a skill issue, but surely there are people doing so

igor56D: this shouldn't break existing programs because the dynamic value should never exceed their hardcoded legacy value (though this requirement can change in the future). If new programs hardcode sub-max rent instead of reading from the rent sysvar, I'd call that a disqualifying skill issue.

proposals/0392-relax-minimum-balance-check.md
```diff
@@ -0,0 +1,213 @@
...
+allocation occurred. When enabled, only newly allocated accounts will be
+subject to rent increases.
+
+Sidenote: this proposal doesn't include any mechanism for increasing rent,
+but when such a mechanism is added in the future it should cap the
+effective rent-per-byte at the legacy rent-exempt per-byte rate. This
+limitation isn't strictly necessary but it avoids needing to significantly
+modify any core system programs (e.g. stake program). If increasing the
+per-byte rate beyond this cap is desirable then the relevant core programs
+need to be updated to be compatible.
```
buffalojoec, line 30: This works for me. I think we should have some more discussion about how all on-chain programs - not just the core programs - should properly handle rent rules according to a new controller model in the relevant SIMD (ie. [SIMD-0389](https://github.com/solana-foundation/solana-improvement-documents/pull/389)).

Relaxation of the checks according to this proposal seems like a good, agnostic first step though, IMO.

PR #397, opened 2025-11-01
SIMD-0298, update: Block Footer V2 -> V1

ksn6:
SIMD 0298 originally proposed upgrading the Block Footer to v2 from v1, given the addition of the block bank hash.

While adding a field to the footer would typically warrant a version increment, for simplicity, we propose sticking with v1. As of November 2024, zero clients disseminate block footers or block markers, making this an appropriate time to extend the version 1 format before widespread adoption.

### DIFF
--- a/proposals/0298-bank-hash-in-block-footer.md
+++ b/proposals/0298-bank-hash-in-block-footer.md
@@ -152,20 +152,25 @@ will be extended to include a new field. The updated footer structure will be:
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 | footer_version=1            (16 bits) |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+| bank_hash                  (32 bytes) |  ←NEW
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 | block_producer_time_nanos   (64 bits) |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 | block_user_agent_len         (8 bits) |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 | block_user_agent        (0-255 bytes) |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-| bank_hash                  (32 bytes) |  ←NEW
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 ```
 
-**Note on Versioning**: While adding a field to the footer would typically warrant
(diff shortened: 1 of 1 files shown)


proposals/0298-bank-hash-in-block-footer.md
```diff
@@ -140,28 +140,38 @@ The block footer structure outlined in SIMD 0307
...
+| variant=0                    (8 bits) |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-| length                      (16 bits) | 
+| length                      (16 bits) |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-| footer_version=2          (16 bits)   |  ←UPDATED 1->2
+| footer_version=1            (16 bits) |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-| block_producer_time_nanos   (64 bits) |  
+| bank_hash                  (32 bytes) |  ←NEW
```
akhi3030, line 155: nit: can we not mix units?  I know the original doc was already mixing it.  With mixed units, it is possible to accidentally to confuse their sizes.

proposals/0298-bank-hash-in-block-footer.md
```diff
@@ -140,28 +140,38 @@ The block footer structure outlined in SIMD 0307
...
-| bank_hash                  (32 bytes) |  ←NEW
+| block_user_agent        (0-255 bytes) |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 ```
 
+**Note on Versioning and Field Ordering**: While adding a field to the footer
+would typically warrant a version increment, we maintain `footer_version=1` for
+simplicity. Rather than appending `bank_hash` to the end of the footer, which
+would be the typical approach, we insert `bank_hash` towards the beginning to
+give it a fixed offset within the footer.
```
akhi3030, line 169:
```suggestion
would be the typical approach, we insert `bank_hash` towards the beginning so that all the fields have a fixed offset.
```

small nit.

ksn6: sure, works

proposals/0298-bank-hash-in-block-footer.md
ksn6: agreed that sticking with bytes everywhere is better (e.g., see [here](https://github.com/solana-foundation/solana-improvement-documents/blob/d9bafb05a5456281c35c62914fb967b0eff34277/proposals/simd-0337-parent-ready-update-marker.md))

this being said, I'd strongly prefer keeping changes in this SIMD localized + keep the stylistic changes to a minimum.

akhi3030: agreed with keeping the changes to a minimum.

MaxResnick:
lgtm

MaxResnick (approved):

jherrera-jump (approved):

PR #400, opened 2025-11-04
SIMD-0400: Epoch Stake Root (ESR) Sysvar

Atamanov:
Introduce a new sysvar, Epoch Stake Root (ESR), that calculates a Merkle root commitment to the finalized validator set and effective stake distribution for the next epoch (N+1) and makes it available on-chain. The ESR is written exactly once per epoch at the epoch boundary when the runtime finalizes the upcoming epoch’s effective stakes. The sysvar also includes a minimum inclusion threshold and dual stake totals to support percentage calculations.

### DIFF
--- a/proposals/0360-epoch-state-root.md
+++ b/proposals/0360-epoch-state-root.md
@@ -1,0 +1,621 @@
+---
+simd: '0360'
+title: Epoch Stake Root (ESR) Sysvar
+authors:
+  - Alexander Atamanov (alexander@twinelabs.xyz)
+category: Standard
+type: Core
+status: Idea
+created: 2025-01-14
+feature: (fill in with feature key and github tracking issues once accepted)
+---
+
+## Summary
+
+Introduce a new sysvar, Epoch Stake Root (ESR), that calculates a Merkle
+root commitment to the finalized validator set and effective stake
+distribution for the next epoch (N+1) and makes it available on-chain. The ESR
+is written exactly once per epoch at the epoch boundary when the runtime
+finalizes the upcoming epoch's effective stakes. The sysvar also includes a
+minimum inclusion threshold and dual stake totals to support percentage
+calculations.
+
+---
+
+## Motivation
+
+Light clients and external verifiers need a compact, trust-minimized way to
+follow validator-set transitions across epoch boundaries. Today,
+reconstructing the validator set for epoch N+1 requires scanning thousands of
+stake/vote accounts during epoch N and re-implementing stake warmup/cooldown
+and activation/deactivation rate limits. This is data- and compute-intensive
+and undermines lightweight clients.
+
+ESR collapses this into a single, consensus-committed 32-byte root with
+minimal metadata, enabling efficient verification on constrained devices and
+cross-chain verifiers without duplicating runtime logic.
+
+---
+
+## New Terminology
+
+- **Effective stake:** Activated stake per vote account after applying
+  warmup/cooldown and global activation/deactivation rate limits at the epoch
+  boundary.
+- **Validator identifier:** Vote account public key (`vote_pubkey`).
+- **Minimum inclusion threshold:** Validators with effective stake below this
+  threshold are excluded from the ESR tree. Their stake still contributes to
+  the global epoch total for percentage-of-all calculations.
+
+---
+
+## Detailed Design
+
+### Sysvar
+
+- **Name:** `EpochStakeRoot` (ESR)
+- **Address:** `ESR111111111111111111111111111111111111111` (fixed sysvar
+  address, base58).
+- **Write semantics:** Runtime-owned; updated only at epoch transition
+  `N → N+1`.
+- **Access:** Read-only to programs; cached similarly to other sysvars.
+
+Data layout (little-endian unless stated otherwise):
+
+```text
+struct EpochStakeRootV1 {
+    u8     version;              // = 2 (ESRv1.1)
+    u64    epoch;                // epoch the root applies to (N+1)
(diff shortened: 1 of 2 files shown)


proposals/0360-epoch-state-root.md
```diff
@@ -0,0 +1,621 @@
+---
+simd: '0360'
```
deanmlittle:
```suggestion
simd: '0400'
```
Could you please amend this? The SIMD number goes off the PR number, in this case, #400

proposals/0360-epoch-state-root.md
```diff
@@ -0,0 +1,621 @@
...
+  the global epoch total for percentage-of-all calculations.
+
+---
+
+## Detailed Design
+
+### Sysvar
+
+- **Name:** `EpochStakeRoot` (ESR)
+- **Address:** `ESR111111111111111111111111111111111111111` (fixed sysvar
```
deanmlittle:
```suggestion
- **Address:** `EpochStakeRoot1111111111111111111111111111` (fixed sysvar
```
Consider following the existing convention of using whole words for Sysvar addresses

proposals/0360-epoch-state-root.md
```diff
@@ -0,0 +1,621 @@
...
+    u64    leaf_count;           // number of INCLUDED validators
+                                 // (stake >= min_stake_lamports)
+    u128   total_active_stake;   // sum of ALL effective stakes across
+                                 // ALL validators
+    u128   tree_lamports;        // sum of INCLUDED stakes
+                                 // (>= min_stake_lamports)
+    u128   min_stake_lamports;   // threshold; below this stake excluded
+                                 // from ESR tree
+    u8     hash_id;              // 1 = SHA-256, 2 = Poseidon
+                                 // (BN254 params)
```
deanmlittle: Sysvars are non-interactive in the sense that you cannot pass in a variable to them to modify the outputted data. Consider either:
- making multiple sysvar accounts (I don't think this would be a popular choice)
- just using SHA256 (proving times are probably not really a major concern on such a small tree, and we will likely move beyond Poseidon and BN254 in the future)
- including multiple hashes, placing them at the end of the account so it is easily extendable in the future, and not leading the the `u8` version byte at the start so as to maintain alignment

deanmlittle:
Awesome. I proposed something similar to a few people recently. This would certainly improve the SIMD voting process, amongst a few other things.

proposals/0360-epoch-state-root.md
```diff
@@ -0,0 +1,621 @@
...
+  address, base58).
+- **Write semantics:** Runtime-owned; updated only at epoch transition
+  `N → N+1`.
+- **Access:** Read-only to programs; cached similarly to other sysvars.
+
+Data layout (little-endian unless stated otherwise):
+
+```text
+struct EpochStakeRootV1 {
+    u8     version;              // = 2 (ESRv1.1)
```
deanmlittle: Starting a struct with a u8 misaligns the rest of the data. If version is not super important, consider omitting it.

Atamanov:
Fixed @deanmlittle

proposals/0400-epoch-state-root.md
```diff
@@ -0,0 +1,460 @@
...
+- **Write semantics:** Runtime-owned; updated only at epoch transition
+  `N → N+1`.
+- **Access:** Read-only to programs; cached similarly to other sysvars.
+
+Data layout (little-endian unless stated otherwise):
+
+```text
+struct EpochStakeRootV1 {
+    u64    epoch;                // epoch the root applies to (N+1)
+    [u8;32] root;                // SHA-256 Merkle root over INCLUDED leaves
```
deanmlittle: Consider putting root last so if we did add expand this to other hashing algorithms in the future, they'd be grouped together.

Atamanov: done

t-nelson (commented):
we're proposing that the validator do non-protocol work, under consensus, at the most compute-contended time, to support support light-clients, which have been fundamentally broken by switching the state commitment to lattice hash?

Atamanov:
Thanks for the review. A few points:

**1. ESR is protocol work**

Verifying consensus requires knowing the validator set. Validators use `epoch_stakes` to verify vote signatures and weight fork choice. ESR commits to this same data, enabling external verification of consensus itself.

The runtime is the only place this can be done correctly. External processes would need to re-implement stake warmup/cooldown, activation rate limits, and match exactly what the leader schedule uses. Any divergence = security vulnerability.

**2. The data already exists—ESR just commits to it**

`EpochStakes` is already computed at epoch boundaries for leader schedule. ESR hashes this already-in-memory `(vote_pubkey, stake)` map: ~3-9ms of SHA-256 on ~1-3k entries, once every ~2 days.

**3. LT-Hash compatibility**

ESR and lattice hash are orthogonal:
- LT-Hash: account state commitment
- ESR: validator set commitment (needed to verify vote signatures regardless of state commitment scheme)

We have a fully working light-client implementation on mainnet using an Agave fork that computes validator set proofs off-chain. ESR provides the minimal on-chain commitment to the validation set which could be used as the root of trust.

Atamanov:
> Hey @Atamanov, this PR has been inactive for 60 days and is now marked as **stale**.
> 
> If you're still working on this, please comment below to let us know. Otherwise, this PR will be automatically closed in **7 days**.
> 
> To keep this PR active, simply leave a comment explaining the current status or any blockers you're facing.

It's still super important

jacobcreech:
@Atamanov could we move this to a discussion? I'm going to close for now.

PR #401, opened 2025-11-06
SIMD-0401: Stake program Pinocchio migration (`p-stake`)

grod220:
Proposal to replace the current version of the Stake Program (`Stake11111111111111111111111111111111111111`) with a CU-optimized Pinocchio version: `p-stake`.

### DIFF
--- a/proposals/0393-stake-program-pinocchio-migration.md
+++ b/proposals/0393-stake-program-pinocchio-migration.md
@@ -1,0 +1,162 @@
+---
+simd: "0393"
+title: "Stake program Pinocchio migration (p-stake)"
+authors:
+  - Gabe (Anza)
+  - Pete (Anza)
+  - Febo (Anza)
+  - Jon (Anza)
+category: Standard
+type: Core
+status: Idea
+created: 2025-11-06
+feature: (to be assigned upon acceptance)
+---
+
+## Summary
+
+This proposal outlines a plan to replace the current Solana Stake Program
+(`Stake11111111111111111111111111111111111111`) with a highly efficient,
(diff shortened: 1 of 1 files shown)


jstarry:
I don't feel convinced by the motivation section. Are stake instructions really using up so much block space that we need to optimize them? This feels like a much smaller win compared to optimizing the token program.

grod220:
> This feels like a much smaller win compared to optimizing the token program

True, but perhaps the SIMD should have put the second point in the motivation section first. There is a no-std wave coming where there will be increasing requirements to adopt it else be marooned w/ old tooling. Alignment with the upstream eBPF toolchain is a primary goal for folks working on the compiler/VM. Aligning with this earlier will allow us to not build new features that eventually have to be re-written in no-std rust. A quite nice side effect of this work is the CU savings.

jstarry:
Ok, I'm definitely out of the loop on no-std plans but that feels like something we can do later in favor of doing other things that are more impactful to developers now.

deanmlittle:
lgtm

grod220:
Bump. Related preliminary work still in progress: https://github.com/solana-foundation/solana-improvement-documents/pull/391

grod220:
Current blockers: https://github.com/solana-program/stake/pull/152#issuecomment-4037210985

grod220:
Status: currently, deprioritized to get [p-ATA](https://github.com/solana-program/associated-token-account/issues/196) out first

grod220:
Status: currently, deprioritized to get [p-ATA](https://github.com/solana-program/associated-token-account/issues/196) out first

PR #402, opened 2025-11-12
SIMD-0402: Finalization Cert in Block Footer

wen-coding:


starrcara86 (commented):
good to me

proposals/0402-final-certs-in-block-footer.md
```diff
@@ -0,0 +1,206 @@
...
++---------------------------------------+
+| bank_hash                  (32 bytes) |
++---------------------------------------+
+| block_producer_time_nanos   (8 bytes) |
++---------------------------------------+
+| block_user_agent_len         (1 byte) |
++---------------------------------------+
+| block_user_agent        (0-255 bytes) |
++---------------------------------------+
+| final_cert_len              (2 bytes) | ← NEW
```
ksn6: Could we have this be 1 byte, where it's `0x00` if there's not `final_cert` and `0x01` if there is a `final_cert`?

proposals/0402-final-certs-in-block-footer.md
```diff
@@ -0,0 +1,206 @@
...
++---------------------------------------+
+```
+
+If the `final_cert_len` is 0, then there is no `final_cert` following it,
+otherwise it specifies the length of `final_cert`.
+
+#### FinalCertificate Serialization
+
+If `final_cert` is present, it is serialized as a whole using
+`wincode::serialize`.
```
ksn6: we should explicitly write out the serialization format in tabular form as done above.

in particular:

- `notar_aggregate: Option<VotesAggregate>` should use a single `u8` to denote whether a `VotesAggregate` is present (i.e., `0x00` if not present and `0x01` if present)

- by default `wincode` will use `8` bytes to serialize the length of `bitmap`; we could probably get away with a fewer number of bytes, right? e.g., would 2 bytes work?

proposals/0402-final-certs-in-block-footer.md
wen-coding: hmm, 1 byte only encodes up to 255, we will have up to ~1300 bytes for 2000 validators, we can do length * 8, then we can encode the length in 1 byte, what will wincode prefer here?

ksn6: to clarify, I'm saying that `final_cert_len` shouldn't actually be a len; rather, it should purely denote a bool that states `true` or `false` as to whether there exists a `final_cert`

proposals/0402-final-certs-in-block-footer.md
wen-coding: Done.

proposals/0402-final-certs-in-block-footer.md
wen-coding: Changed.

proposals/0402-final-certs-in-block-footer.md
```diff
@@ -0,0 +1,234 @@
...
+category: Standard
+type: Core
+status: Review
+created: 2025-11-11
+feature: (fill in with feature key and github tracking issues once accepted)
+---
+
+## Summary
+
+This SIMD introduces addition of an Alpenglow finalization certificate to the
```
rogerANZA: something is wrong here (grammatically), I don't quite get the sentence.

rogerANZA: maybe we want to be more specific? we want to say that we add a finalization certificate such that somebody who just observes blocks understands that they are finalized without having to understand the all-to-all communication between the validators.

proposals/0402-final-certs-in-block-footer.md
```diff
@@ -0,0 +1,234 @@
...
+Before Alpenglow, validator votes were expressed as on-chain transactions that
+updated vote accounts, each functioning as a state machine tracking the voting
+activity of a staked validator. Alpenglow (SIMD-0326) moves away from this
+model by having validators send vote messages directly to one another,
+improving the speed and efficiency of consensus.
+
+This shift removes the on-chain visibility previously provided by vote
+transactions and vote account state, which zero-staked validators currently
+rely on to infer validator delinquency. It also affects any party that depends
+on the ability to observe votes on-chain. To address this, Alpenglow SIMD
```
rogerANZA: (remove "SIMD", it sounds wrong)

wen-coding: Done.

proposals/0402-final-certs-in-block-footer.md
wen-coding: Thanks for the suggestion. Changed.

You can also make comments in the Google doc if that's easier.

proposals/0402-final-certs-in-block-footer.md
```diff
@@ -0,0 +1,242 @@
...
++---------------------------------------+
+| bank_hash                  (32 bytes) |
++---------------------------------------+
+| block_producer_time_nanos   (8 bytes) |
++---------------------------------------+
+| block_user_agent_len         (1 byte) |
++---------------------------------------+
+| block_user_agent        (0-255 bytes) |
++---------------------------------------+
+| final_cert_present           (1 byte) | ← NEW
```
akhi3030, line 129: It might make sense to use this field to also indicate if the notar and skip reward certs are present.  So we are not using additional bytes for them.  Maybe we want to roll the reward certs into this SIMD?

proposals/0402-final-certs-in-block-footer.md
```diff
@@ -0,0 +1,242 @@
...
++---------------------------------------+
+| Version = 1                 (1 byte)  |
++---------------------------------------+
+| bank_hash                  (32 bytes) |
++---------------------------------------+
+| block_producer_time_nanos   (8 bytes) |
++---------------------------------------+
+| block_user_agent_len         (1 byte) |
++---------------------------------------+
+| block_user_agent        (0-255 bytes) |
```
akhi3030, line 127: in previous SIMDs, the feedback was to keep this field in the end so that all fields have a fixed offset.  This SIMD is introducing a new variable field.  So at the least, we may want to swap things around so that all the fixed length fields come first and then all the variable length ones.

proposals/0402-final-certs-in-block-footer.md
wen-coding: It is chosen this way because this is the natural translation of
```
struct {
final_cert: Option<FinalCertificate>,
}
```
from Rust in bincode/wincode serialization.

proposals/0402-final-certs-in-block-footer.md
wen-coding: See above, if we swap things around then natural bincode/wincode serialization doesn't work any more, we need to write customized serialization.

proposals/0402-final-certs-in-block-footer.md
ksn6: I like @akhi3030 's suggestion, though more strongly prefer avoiding custom serde.

I'm checking with the `wincode` authors to see whether this is something we can support. For now, though, would rather use different bytes to denote whether different certs are present.

proposals/0402-final-certs-in-block-footer.md
```diff
@@ -0,0 +1,242 @@
...
++---------------------------------------+
+| final_aggregate_bitmap_len  (2 bytes) |
++---------------------------------------+
+| final_aggregate_bitmap     (variable) |
++---------------------------------------+
+| notar_aggregate_present      (1 byte) |
++---------------------------------------+
+| notar_aggregate_signature  (variable) |
++---------------------------------------+
+| notar_aggregate_bitmap_len (variable) |
```
ksn6, line 158: `notar_aggregate_bitmap_len` should be 2 bytes

wen-coding: Wrote variable basically because notar_aggregate is optional here.

starrcara86 (approved):

PR #406, opened 2025-11-19
SIMD-406: Restrict the number of accounts in an instruction

LucasSte:


ptaffet-jump:
Can you clarify what happens if there are exactly 256 accounts?
> both user deployed programs and CPIs only allow 255 accounts

vs.

> if it references more than 256 accounts, and must throw 
`InstructionError::MaxAccountsExceeded

proposals/0406-maximum-instruction-accounts.md
```diff
@@ -0,0 +1,60 @@
...
+or precompiles.
+
+## Motivation
+
+Although there can only be a maximum of 35 account keys in the legacy message 
+format and 256 in the v0 format, instructions can have up to 65535 (`u16::MAX`) 
+accounts, each being a `u8` index referencing a transaction account.
+
+Allowing instructions to reference more accounts than those contained in the 
+transaction is pointless, because it entails some accounts will be aliased,
```
t-nelson: > pointless

is it? instruction account references map to roles as defined by that instruction. multiple roles may reasonably be filled by the same account

restricting instruction account references to a unique set would force role mapping on chain, either statically in the program or dynamically via instruction data. the latter would cost addition transaction byte bloat, both would incur additional cu cost and risk of error

this seems like a naive motivation

proposals/0406-maximum-instruction-accounts.md
```diff
@@ -0,0 +1,60 @@
...
+created: 2025-11-19
+feature: (fill in with feature tracking issues once accepted)
+---
+
+## Summary
+
+This SIMD imposes a hard restriction on the maximum number of accounts an 
+instruction may refer to. There is already a limit of 255 accounts during 
+serialization in ABI v0 and v1, but it does not apply to builtin invocations 
+or precompiles.
```
t-nelson:
```suggestion
This SIMD imposes a hard restriction on the maximum number of account
references that an instruction may declare. There is already a limit of
255 account rederences imposed during serialization in ABI v0 and v1,
but it does not current apply to builtin invocations or precompiles.
```

proposals/0406-maximum-instruction-accounts.md
LucasSte: My wording there might have been imprecise. What I wanted to say is that allowing instructions to reference more accounts than the number of pubkeys the message format allows (256) is not reasonable.

The argument for that is: for what the protocol offers today, that is not necessary. On one side, there is a hard limit of 255 for CPI and programs. On the other, builtins and precompiles don't use that many.

We will continue to allow account aliasing and instructions referencing more accounts than the number of pubkeys in a transaction, provided that you are not overusing the message format.

LucasSte:
> Can you clarify what happens if there are exactly 256 accounts?
> 
> > both user deployed programs and CPIs only allow 255 accounts
> 
> vs.
> 
> > if it references more than 256 accounts, and must throw
> > `InstructionError::MaxAccountsExceeded

Good point. We can keep erroring out at 255 for programs and CPI, or we can increase the serialization limit to 256. Any thoughts @Lichtso?

Lichtso:
Limit should be `u8::MAX` meaning 255, not 256 in all paths / cases.

proposals/0406-maximum-instruction-accounts.md
```diff
@@ -0,0 +1,60 @@
...
+
+## New Terminology
+
+None.
+
+## Detailed Design
+
+For every instruction that goes through program runtime (CPI, user deployed 
+program invocation, builtin invocation or precompiles), program runtime must 
+check if it references more than 256 accounts, and must throw
```
jstarry: This should be 255 instead of 256 to match other instruction account limits used in CPI and BPF account serialization. It also allows us to reserve `u8::MAX` (255) to indicate that a given account index is a duplicate or not used in an instruction.

jstarry: Ah just saw @Lichtso's [comment](https://github.com/solana-foundation/solana-improvement-documents/pull/406#issuecomment-3554670934) saying the same thing.

proposals/0406-maximum-instruction-accounts.md
LucasSte: https://github.com/solana-foundation/solana-improvement-documents/pull/406/commits/96dec4ec37e28d81e6e717b3516b61bca8f59593

proposals/0406-maximum-instruction-accounts.md
LucasSte: I rewrote the motivation in https://github.com/solana-foundation/solana-improvement-documents/pull/406/commits/9ce30cbcf0c5059695dfd69d8aa33850b434ee5f

LucasSte:
I addressed all the feedback in the previous three commits, including the reduction of the limit to 255.

Lichtso (approved):

topointon-jump (dismissed):

proposals/0406-maximum-instruction-accounts.md
```diff
@@ -0,0 +1,61 @@
...
+## New Terminology
+
+None.
+
+## Detailed Design
+
+For every instruction that goes through program runtime (CPI, user deployed 
+program invocation, builtin invocation or precompiles), program runtime must 
+check if it references more than 255 accounts, and must throw 
+`InstructionError::MaxAccountsExceeded` when that is the case.
```
buffalojoec: Where should this check occur exactly? When should it throw? Ideally as early as possible.

jstarry: Yeah agreed. I think that in addition to the program runtime restriction, we should also add a sanitization check for the number of accounts passed to top-level instructions.

LucasSte: > Where should this check occur exactly? When should it throw?

The exact position I was thinking is in `prepare_next_instruction` and `prepare_next_top_level_instruction`. I did not detail the function names since this is Agave specific. I'll write down "right before account deduplication" as an alternative.

> we should also add a sanitization check for the number of accounts passed to top-level instructions.

Yes, I also had that idea, but thought it would difficult to pass down a boolean through the sanitization trait we have. Since everyone also believes in that being useful, I am going to update the SIMD with that requirement.

LucasSte: > we should also add a sanitization check for the number of accounts passed to top-level instructions.

On a second thought, transaction sanitization is used in far too many places throughout the mono repo. It is going to be time consuming to get everything right there. Do you think the additional check is still worth it?

LucasSte: I mentioned where the check happens in https://github.com/solana-foundation/solana-improvement-documents/pull/406/commits/959ecc3e02746e45fd253c163cbe9c6497f8bdad.

I didn't yet write about sanitization, because I'm waiting for your opinion on that.

Lichtso: The check also being present in sanitization would be redundant and not be protocol visible anyway, right? It does not belong in the SIMD if it does not change the outcome of any transaction, it is an implementation detail.

LucasSte: @jstarry

proposals/0406-maximum-instruction-accounts.md
```diff
@@ -39,10 +39,12 @@ None.
...
-For every instruction that goes through program runtime (CPI, user deployed 
-program invocation, builtin invocation or precompiles), program runtime must 
-check if it references more than 255 accounts, and must throw 
-`InstructionError::MaxAccountsExceeded` when that is the case.
+A check must be inclided for every instruction that goes through program 
+runtime (CPI, user deployed program invocation, builtin invocation or 
+precompiles) to ensure it does not reference more than 255 accounts, and 
+throw `InstructionError::MaxAccountsExceeded` when that is the case. Such 
+verification must happend right before account deduplication for instruction 
+execution.
```
buffalojoec: nit: you've got a few typos in here.

```suggestion
A check must be included for every instruction that goes through program 
runtime (CPI, user deployed program invocation, builtin invocation or 
precompiles) to ensure it does not reference more than 255 accounts, and 
throw `InstructionError::MaxAccountsExceeded` when that is the case. Such 
verification must happen right before account deduplication for instruction 
execution.
```

proposals/0406-maximum-instruction-accounts.md
buffalojoec: > did not detail the function names since this is Agave specific. I'll write down "right before account deduplication" as an alternative.

I think this is as close as we can get without a spec. 😅

Maybe FD can help us clear up the exact location/timing of this check - @mjain-jump @topointon-jump ? The most important piece is specifying which errors this should eclipse and which ones it should not.

jstarry: We need to specify in which cases the protocol will reject the transaction for violating a protocol constraint (eg sanitization error) and in which cases the transaction can be committed to a block with a failed result. I think in both cases we typically try to define the error handling order (as @buffalojoec is suggesting) to ensure diff fuzzing doesn't have discrepancies

mjain-jump: agree with @jstarry . i wouldve suggested putting this check inside the transaction loading step to make it a fees-only transaction in this case but CPI complicates things. "right before account deduplication" is probably as good as we can get and is pretty easily replicable in our codebase, given the instr processing is quite underspecified.

mjain-jump: i think a future SIMD to formally define what checks happen / what order they're done in would be great, similar to https://github.com/solana-foundation/solana-improvement-documents/pull/311 so the location of these new checks would be more well-defined and easily implemented by any other validator client

topointon-jump (approved):

proposals/0406-maximum-instruction-accounts.md
buffalojoec: I actually feel like this should be a runtime error and therefore it should charge the sender a fee. It's in a similar vein to [SIMD-0191](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0191-enable-transaction-loading-failure-fees.md).

Just like what happens when someone overloads the validator and tries to load more data than the limit, they should pay a fee and get an error for trying to create too many instruction account references.

Outside the SIMD, sanitization would be a backstop, so if you exceed the max number of account refs you probably side-stepped that workflow and it may cost you a fee if you're not careful.

jstarry: I disagree with @buffalojoec. I think that checking the number of instruction accounts is pretty lightweight and so we should specify in this SIMD that exceeding 255 accounts in a top-level instruction will result in the transaction being rejected by the protocol (and not be valid and therefore not fee-paying). If a block includes such a transaction, the block is invalid. This is consistent with other tx sanitization constraints in the protocol and matches the proposed sanitization limit for number of accounts per ix in the v1 tx format: https://github.com/solana-foundation/solana-improvement-documents/pull/385/files#diff-dbd54cc962c00d4b548ccd7f2250d7403ce75365c1f2bd154461582a142fb3a5R209

```md
### Transaction Constraints

Any transaction violating these constraints will be considered invalid and will 
not be included in the chain. Violations are considered sanitization failures:

| value | max specified by this SIMD | prior max | max implied by format |
| --- | --- | --- | --- |
| accounts/instruction | 255 | 255 | 255 |
```

Lichtso: @jstarry What you suggest would be implemented as an additional check [here](https://github.com/anza-xyz/solana-sdk/blob/e9de9d8845a45cb4276634dcc7013ed7fa86d168/message/src/legacy.rs#L118), correct?

jstarry: Effectively, yes that's where it should go. But ideally the SIMD is not written specific to the Agave implementation. It can just say that this check will be added to the existing transaction sanitization checks.

topointon-jump: Agree with @jstarry

LucasSte: Just so I can compile all feedback and amend the SIMD, are we going to place the check in both runtime and sanitization?

Lichtso: If we throw in sanitization then the runtime check becomes unreachable. So, for the SIMD it is either or. Specifying both is inconsistent / self conflicting.

LucasSte: Regarding the implementation in Agave, the check in the program-runtime is easier to do, since we already have the feature gating infrastructure in place. Implementing it in sanitization involves more code changes throughout the monorepo.

I see people weighing on both approaches, but I still did not understood if one is preferable over the other. Does anyone have any strong reason for the check not to be in runtime?

I'm happy to change if that is the case, but for simplicity my preference would be runtime.

jstarry: Looks like we have most people agreeing on adding a protocol error for transactions which use too many accounts in an instruction. If you need some additional convincing, v1 Transactions will have a sanitization limit of 255 accounts per instructions, so I think it makes sense to limit ix accounts in legacy and v0 transactions in the same way. I think the Agave implementation should be fine, we can discuss later if you want.

LucasSte: Thanks for the feedback. I updated the SIMD to use only a check during sanitization. The latest version is in https://github.com/solana-foundation/solana-improvement-documents/pull/406/commits/5bc845a222a5f8258ba8a2d0909875f0469485d3.

proposals/0406-maximum-instruction-accounts.md
```diff
@@ -39,12 +39,11 @@ None.
...
-A check must be included for every instruction that goes through program 
-runtime (CPI, user deployed program invocation, builtin invocation or 
-precompiles) to ensure it does not reference more than 255 accounts, and 
-throw `InstructionError::MaxAccountsExceeded` when that is the case. Such 
-verification must happen right before account deduplication for instruction 
-execution.
+A new verification step must be included in message sanitization for message 
+formats legacy and v0. The verification step must examine every instruction in 
+a transaction, check if the former references more than 255 accounts, and 
+throw `SanitizationError::ValueOutOfBounds` otherwise. Transactions violating
```
jstarry, line 45: We don't need to specify the error here, it's an implementation detail. It's enough to say that the transaction and any block including this transaction will be rejected.

Lichtso: While the specific error code is not part of consensus it is still good to specify it for fuzzing between validator implementations. At least that is how we handled it in other SIMDs.

topointon-jump:
@LucasSte shall we merge this?

topointon-jump:
@bw-solana mind taking a look?

proposals/0406-maximum-instruction-accounts.md
```diff
@@ -0,0 +1,62 @@
...
+the validator to deduplicate them, even though they are unused.
+
+## New Terminology
+
+None.
+
+## Detailed Design
+
+A new verification step must be included in message sanitization for message 
+formats legacy and v0. The verification step must examine every instruction in
```
bw-solana, line 43: how about the new Transaction v1 format?

proposals/0406-maximum-instruction-accounts.md
```diff
@@ -0,0 +1,62 @@
...
+
+None.
+
+## Detailed Design
+
+A new verification step must be included in message sanitization for message 
+formats legacy and v0. The verification step must examine every instruction in 
+a transaction, check if the former references more than 255 accounts, and 
+throw `SanitizationError::ValueOutOfBounds` otherwise. Transactions violating 
+such a restriction must be considered invalid and not included in a block.
```
bw-solana: Should we mention that blocks should be declared dead if they include such transactions? Or is the block still good but the transaction does not execute?

Is this fine for bankless leader because we can perform static analysis and determine validity?

proposals/0406-maximum-instruction-accounts.md
topointon-jump: The new transaction v1 format also specifies this limit, so this applies to all transaction formats. Definitely worth specifying 👍

proposals/0406-maximum-instruction-accounts.md
topointon-jump: > Should we mention that blocks should be declared dead if they include such transactions?

Yes we should mention this.

> Is this fine for bankless leader because we can perform static analysis and determine validity?

Yes this is correct.

proposals/0406-maximum-instruction-accounts.md
LucasSte: The limit for v1 transactions won't be behind the feature gate implied by this SIMD. Is it worth specifying if that is the case?

topointon-jump: Maybe just mention it for clarity?

LucasSte: I mentioned v1 transactions and dead blocks (from the following thread) in https://github.com/solana-foundation/solana-improvement-documents/pull/406/commits/f600fbc2b0cb7f970f9db94d1a64c99338079f46

bw-solana (approved):

PR #417, opened 2025-11-27
Assign Ed25519 precompile to native loader

deanmlittle:


t-nelson:
i don't think this is necessary. it's only testnet. i'd far prefer to just rewrite the account on the next testnet restart than carry a testnet-only feature gate around. the inconsistencies in cluster feature sets have already been enough trouble historically

proposals/0417-assign-ed25519-precompile-to-native-loader.md
```diff
@@ -0,0 +1,87 @@
...
+
+## New Terminology
+
+N/A
+
+## Detailed Design
+
+Introduce a feature gated change to the bank that, upon activation, updates 
+the owner of the Ed25519 program to the Native Loader program if it is not 
+already owned by it, also explicitly flagging its account as executable,
```
jstarry, line 37: Note that the testnet ed25519 prcompile is already executable on testnet so flagging the account as executable isn't necessary

proposals/0417-assign-ed25519-precompile-to-native-loader.md
```diff
@@ -0,0 +1,87 @@
...
+## New Terminology
+
+N/A
+
+## Detailed Design
+
+Introduce a feature gated change to the bank that, upon activation, updates 
+the owner of the Ed25519 program to the Native Loader program if it is not 
+already owned by it, also explicitly flagging its account as executable, 
+leaving all other remaining fields unchanged.
```
jstarry: For consistency's sake I think we should also initialize the account's data to match mainnet-beta and devnet. Note that changing or not changing this has no other benefit or side effect that I'm aware of, I just care about consistency. Those clusters have the following data:

```
Length: 15 (0xf) bytes
0000:   65 64 32 35  35 31 39 5f  70 72 6f 67  72 61 6d      ed25519_program
```

proposals/0417-assign-ed25519-precompile-to-native-loader.md
deanmlittle: The reason I chose to do this is because although in the RPC the account is returned as executable, it belongs to the system program, which makes it unclear as to why or how it is executable, so rather than risk it being a display bug, I thought explicitly setting it would remove all doubt.

jstarry: Ok, it's not a display bug. The account is marked as executable on all clusters already.

```
solana account Ed25519SigVerify111111111111111111111111111 -ut
```

```
Public Key: Ed25519SigVerify111111111111111111111111111
Balance: 0.000000001 SOL
Owner: 11111111111111111111111111111111
Executable: true
Rent Epoch: 18446744073709551615
```

proposals/0417-assign-ed25519-precompile-to-native-loader.md
deanmlittle: fixed here 50e3a4715b0c980657a18f4d7b1c7fd65525becb

proposals/0417-assign-ed25519-precompile-to-native-loader.md
deanmlittle: noted, will still set it anyway as there's no risk in being explicit.

deanmlittle:
> i don't think this is necessary. it's only testnet. i'd far prefer to just rewrite the account on the next testnet restart than carry a testnet-only feature gate around. the inconsistencies in cluster feature sets have already been enough trouble historically

Up to you ser. Just doing what I can to try make things less broken. Seems everyone I've asked is cool with it now. If this makes it into 3.1.0, I think it's worth it. If it doesn't, I think your approach is better. Having to wait for a whole release cycle to fix testnet is an unacceptable delay imo.

jstarry (approved):

jstarry:
As written, the SIMD feature gate can be activated on all clusters. It's also quite simple so I would be in favor of backporting the implementation to v3.1. @t-nelson thoughts?

proposals/0417-assign-ed25519-precompile-to-native-loader.md
```diff
@@ -0,0 +1,89 @@
...
+type: Core
+status: Review
+created: 2027-11-27
+feature: (fill in with feature tracking issues once accepted)
+---
+
+## Summary
+
+This proposal restores nominal functionality of the Ed25519 precompile to 
+testnet witohut any changes to mainnet and devnet.
```
buffalojoec:
```suggestion
testnet without any changes to mainnet and devnet.
```

proposals/0417-assign-ed25519-precompile-to-native-loader.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Detailed Design
+
+Introduce a feature gated change to the bank that, upon activation, updates 
+the owner of the Ed25519 program to the Native Loader program if it is not 
+already owned by it, also explicitly flagging its account as executable, 
+and setting the data to `ed25519_program` to match all other clusters. All
+other fields remain unchanged.
+
+```rust
```
buffalojoec, line 41: I don't think the code snippet is valuable here. The spec should be agnostic anyway.

buffalojoec (commented):
Lgtm, I think the all-cluster activation method makes sense, too.

t-nelson:
> As written, the SIMD feature gate can be activated on all clusters. It's also quite simple so I would be in favor of backporting the implementation to v3.1. @t-nelson thoughts?

that's all good and well, but why? why waste all this time on testnet? [we can just ram it in](https://discord.com/channels/428295358100013066/838890116386521088/1445137874256134176). who cares?

jstarry:
> that's all good and well, but why? why waste all this time on testnet? [we can just ram it in](https://discord.com/channels/428295358100013066/838890116386521088/1445137874256134176). who cares?

I'd rather not use "hacks" to fix any cluster, including testnet. Using ledger tool to patch a cluster snapshot is pretty heavy handed for something trivial to fix via feature gate. And the fix via feature gate is simple enough to not waste much time.

t-nelson:
i would rather not no-op hard fork mainnet beta under any circumstances

https://github.com/anza-xyz/agave/pull/9352

jstarry:
> i would rather not no-op hard fork mainnet beta under any circumstances

I don't really mind doing that.

But your snapshot change looks fine to me too. It's not too different to how we deactivate features on testnet. Let's do that instead of a feature gate.

t-nelson:
yeah it's the same class of activity as "deactivate feature gate" or "forced destake" imo. if this was any other cluster i'd be more hestitant, but testnet is here for validator client dev convenience and a feature gate is not convenient

deanmlittle:
resolved by https://github.com/anza-xyz/agave/pull/9352

PR #418, opened 2025-11-27
SIMD-0418:  Enable Loader v2 to v3 Program Migrations

febo:


proposals/0418-migration-loader-v2-programs-to-loader-v3.md
```diff
@@ -0,0 +1,88 @@
...
+## Summary
+
+This proposal introduces a mechanism to migrate Loader v2 programs to Loader
+v3.
+
+## Motivation
+
+Loader v2 programs are immutable by nature, since the loader does not support
+program upgrades. There are essential Loader v2 programs (e.g., SPL Token) that
+would benefit from updates, but currently lack a sound mechanism to be upgraded.
```
buffalojoec: I think we might want to be careful how we word the motivation here. We aren't trying to sneakily make Loader v2 programs _upgradeable_, but rather we're trying to introduce a feature-gated, consensus-oriented mechanism by which we can introduce upgrades to very specific programs as agreed upon by the community.

proposals/0418-migration-loader-v2-programs-to-loader-v3.md
```diff
@@ -0,0 +1,88 @@
+---
+simd: '0418'
+title: Migration of Loader v2 Programs to Loader v3
```
buffalojoec: nit: I think something like this might be a tad more clear:

```suggestion
title: Enable Migration of Loader v2 Programs to Loader v3
```

...or maybe even something like:

```suggestion
title: Enable Loader v2 to v3 Program Migrations
```

What do you think?

buffalojoec (commented):
Looks pretty good to me!

proposals/0418-migration-loader-v2-programs-to-loader-v3.md
febo: Good point – revised the text in https://github.com/solana-foundation/solana-improvement-documents/pull/418/commits/6bb864299f562944dfedc80aee65e32da5ee30be.

proposals/0418-migration-loader-v2-programs-to-loader-v3.md
febo: Second one looks good - updated!

proposals/0418-enable-loader-v2-to-v3-program-migrations.md
```diff
@@ -0,0 +1,89 @@
...
+
+Each individual program migration should be controlled by its own feature gate,
+which triggers the migration process when activated.
+
+## Impact
+
+With this change, validator clients would no longer be required to implement
+changes to essential programs. Instead these programs could be updated just
+once. This reduces some engineering overhead on validator teams.
+
```
buffalojoec, line 88: I think this section isn't describing the proposed change.

buffalojoec (dismissed):
Lgtm

proposals/0418-enable-loader-v2-to-v3-program-migrations.md
febo: Indeed, updated in [9864471](https://github.com/solana-foundation/solana-improvement-documents/pull/418/commits/98644714231a0bca7c26a29cffad17f307b69f0e).

buffalojoec (dismissed):

proposals/0418-enable-loader-v2-to-v3-program-migrations.md
```diff
@@ -0,0 +1,89 @@
...
+Previously, Loader v2 programs could be “upgraded” by directly modifying their
+account data. This approach is no longer recommended due to the introduction of
+the global program cache, since it can cause inconsistencies if account data is
+modified in-place.
+
+## New Terminology
+
+N/A
+
+## Detailed Design
```
mjain-jump, line 41: Will this be an instruction that the program authority invokes, or handled by the validator itself? Who will pay for the rent for the additional space used for the program / programdata account metadata?

febo: It will be handled by the validator. Each migration will have its individual feature gate that invoke the procedure to migrate the account. 

Since it is necessary to create a buffer account with the program data for Loader V3, the rent from the buffer account is used to fund the migration and any excess is burned.

mjain-jump: Not sure if I'm missing anything, but...

Before: v2 program account containing n bytes
After: v3 program account containing 45 bytes, v3 programdata account containing 36+n bytes

Where is the buffer account coming from?

buffalojoec: You deploy the buffer ahead of time and then take some of its lamports to fund the new program & programdata excess rent. The rest are burned.

buffalojoec: Maybe we need to specify that.

febo: Updated on https://github.com/solana-foundation/solana-improvement-documents/pull/418/commits/831375fa35c13bca3b95f9d0c2bfda43dec55b2b

mjain-jump: Got it thanks for clarifying!

proposals/0418-enable-loader-v2-to-v3-program-migrations.md
```diff
@@ -0,0 +1,93 @@
...
+
+Migrating a Loader v2 program to Loader v3 involves creating a buffer
+account owned by the Loader v3 (BPF Upgradable Loader) that contains
+the ELF bytes of the program's implementation to be migrated. This
+buffer is deployed before the initiating the migration, and a portion
+of its lamports is later used to fund the new program and program data
+accounts. Any remaining lamports in the buffer are burned as part of
+the migration process.
```
mjain-jump, line 55: I'm still a little bit confused about this. If I have a program that I want to migrate, will there be a loader v3 instruction that lets me create a buffer account with the loader v2 program's ELF bytes? Can anyone create the buffer account to migrate any loader v2 program? Would be nice to have the full process detailed from start to finish here.

febo: Migration is only performed by the runtime and activated via a feature gate. There is no public API to upgrade Loader v2 programs. I clarified a couple of sections of the text to make this more explicit.

buffalojoec: I guess we are resting on a few of the procedures laid down in https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0088-enable-core-bpf-programs.md. I agree with Febo we can specify more in this proposal as well.

TL;DR:
1. Write new program to Loader V3 buffer using `LoaderV3::Write`
2. Add feature gate to migration path in Bank
3. Activate feature
4. Profit

buffalojoec (dismissed):

proposals/0418-enable-loader-v2-to-v3-program-migrations.md
```diff
@@ -0,0 +1,96 @@
...
+modified in-place.
+
+## New Terminology
+
+N/A
+
+## Detailed Design
+
+The migration from Loader v2 program to Loader v3 is only performed by the
+runtime and activated via a feature gate.
```
mjain-jump: So I'm guessing that at the epoch boundary when the feature is activated, you're going to migrate every program (maybe in batches similar to how PER is distributed)? How are you planning on collecting the full list of programs to be migrated? Or would it be a lazy approach (upgrade-on-invoke)?

proposals/0418-enable-loader-v2-to-v3-program-migrations.md
```diff
@@ -0,0 +1,96 @@
...
+
+Migrating a Loader v2 program to Loader v3 involves creating a buffer
+account owned by the Loader v3 (BPF Upgradable Loader) that contains
+the ELF bytes of the program's implementation to be migrated. This
+buffer is deployed before the initiating the migration, and a portion
```
mjain-jump:
```suggestion
buffer is deployed before initiating the migration, and a portion
```

proposals/0418-enable-loader-v2-to-v3-program-migrations.md
febo: It will be done case-by-case. The idea is that a feature gate only enables the migration of a single program.

buffalojoec (dismissed):

proposals/0418-enable-loader-v2-to-v3-program-migrations.md
topointon-jump: I'm a little confused as to what this SIMD is describing - are there any code changes that are feature-gated by the feature gate associated with this SIMD? Or is it just describing a process that will be followed by future feature gates, one for each program? Could we clarify this?

topointon-jump: Each of these future individual program upgrade feature gates will also need a SIMD - but they can reference this one.

mjain-jump:
> It will be done case-by-case. The idea is that a feature gate only enables the migration of a single program.

The title seems a bit misleading, you mentioned this is a migration for a single program but the title implies that its a general migration procedure for any program?

febo:
> > It will be done case-by-case. The idea is that a feature gate only enables the migration of a single program.
> 
> The title seems a bit misleading, you mentioned this is a migration for a single program but the title implies that its a general migration procedure for any program?

It is a generic procedure and can be used to migrate any Loader v2 program. But the program being migrated will need its own feature gate to trigger the migration procedure.

proposals/0418-enable-loader-v2-to-v3-program-migrations.md
febo: It is just describing the procedure to follow, there is no feature gate for it since it is only executed when a migration is triggered. Each future individual program upgrade can reference the procedure described in this SIMD.

topointon-jump: Could you make this clear in the SIMD text? Other than that looks good!

febo: Added the following to the first paragraph of the design description:

_This SIMD describes the procedure to follow to execute the migration. Each individual program upgrade will reference the procedure described in this SIMD when proposing a migration and specify its corresponding feature gate._

mjain-jump (approved):

topointon-jump (dismissed):

buffalojoec (commented):
The mechanism looks sound, and is specified properly. However, I believe the alternatives section could elucidate why we can't simply swap two Loader V2 program accounts as we did before.

IMO, retaining the program's owner seems preferred. Thinking about SPL-Token, it's been under Loader V2 for years. I'd hate to break offchain tooling if we don't have to.

> This approach is no longer recommended due to the introduction
of the global program cache, since it can cause inconsistencies if account data is
modified in-place.

This sounds like @Lichtso - maybe you can detail this risk here?

Lichtso:
> This sounds like @Lichtso - maybe you can detail this risk here?

The global program cache does not support upgrades of loader-v2 programs. That is because in loader-v2 there is no upgrade path and thus also no last upgrade slot metainfo. Without this metainfo the global program cache can't distinguish the old and new version across forks. When migrating to loader-v3 this would not be a problem because the owner is a distinguishing feature and loader-v3 does have the last upgrade slot metainfo for further future changes.

buffalojoec:
> > This sounds like @Lichtso - maybe you can detail this risk here?
> 
> The global program cache does not support upgrades of loader-v2 programs. That is because in loader-v2 there is no upgrade path and thus also no last upgrade slot metainfo. Without this metainfo the global program cache can't distinguish the old and new version across forks. When migrating to loader-v3 this would not be a problem because the owner is a distinguishing feature and loader-v3 does have the last upgrade slot metainfo for further future changes.

@febo can we add this to the proposal?

topointon-jump (approved):

proposals/0418-enable-loader-v2-to-v3-program-migrations.md
```diff
@@ -0,0 +1,104 @@
...
+
+* **Close the buffer account**:
+  Once both program and program data accounts have been created and populated,
+  the buffer account is closed and any remaining lamports are burned.
+
+In the slot immediately following the feature activation, the program will not be
+invocable. This status will last one slot, then the program will be fully operational.
+The program ID does not change by this migration process.
+
+Each individual program migration should be controlled by its own feature gate,
```
bw-solana, line 80: this would imply each step of the process described above happens at a different epoch boundary.

but technically we can enable multiple feature gates at once.. are clients expected to make this work?

febo: > this would imply each step of the process described above happens at a different epoch boundary.

Not necessarily, they could all happen in the same epoch boundary.

> but technically we can enable multiple feature gates at once.. are clients expected to make this work?

Each program migration is independent of each other, in the sense that each will call up the migration procedure described in the SIMD.

bw-solana: ah, bad parsing on my end.

All of the steps Create/Rewrite/Close happen at epoch boundary under one feature gate. But each relocation will use a separate feature gate to drive these steps. I get it now 🫡

proposals/0418-enable-loader-v2-to-v3-program-migrations.md
```diff
@@ -0,0 +1,104 @@
...
+which triggers the migration process when activated.
+
```
bw-solana, line 82: do we need to mention any details around which hashes these account data changes should impact?

buffalojoec (approved):

proposals/0418-enable-loader-v2-to-v3-program-migrations.md
febo: The account changes will be captured as any other account update and reflected on bank's accounts lattice hash and bank hash.

bw-solana (approved):

PR #430, opened 2025-12-18
SIMD-0430: Loader V3: Relax Program Buffer Constraints

deanmlittle:


buffalojoec:
Lgtm

buffalojoec (approved):

proposals/0430-relax-program-buffer-constraints.md
```diff
@@ -0,0 +1,76 @@
...
+## New Terminology
+
+No new terminology is introduced by this proposal.
+
+## Detailed Design
+
+The following features will be activated behind a feature gate:
+
+1. Remove owner check on program buffer accounts during program deployment, 
+requiring buffer to be owned by `BPFLoaderUpgradeab1e11111111111111111111111`.
```
Lichtso: This should probably specify that we would still expect the layout (specifically the ELF starting offset in the account) to be the same, even if the supplied buffer was not created by loader-v3.

Lichtso:
Overall looking good. I think the wording would benefit from distinguishing whether the SIMD is talking about initial deployments (`Deploy`), redeployments (`Upgrade`) or both in the various places these are mentioned.

proposals/0430-relax-program-buffer-constraints.md
```diff
@@ -0,0 +1,76 @@
...
+1. Remove authority check enforcing buffer shares an upgrade authority with
+program being updated.
```
Lichtso: This should specify that it only removes the `InstructionError::IncorrectAuthority` check of the authority matching the program, not the `InstructionError::MissingRequiredSignature` check of the authority being a signer.

Both happen right next to one another:
https://github.com/anza-xyz/agave/blob/ec8ba132a605eee519effa7c7ea2bdc1bc55b151/programs/bpf_loader/src/lib.rs#L534-L541

buffalojoec:
```suggestion
## Detailed Design

After the feature is activated, the program will no longer enforce the
following checks:
* `DeployWithMaxDataLen`: The signing authority no longer must match the
  authority stored on the buffer account.
* `Upgrade`: The signing authority no longer must match the authority
  stored on the buffer account.

Note that the authority account will remain in the same position for both
instructions and must still sign the transaction as before.
```

proposals/0430-relax-program-buffer-constraints.md
```diff
@@ -0,0 +1,76 @@
...
+2. If a buffer is owned by a third-party program, that program may retain 
+write access to the buffer account irrespective of its upgrade authority. 
+This creates a potential supply chain attack vector if the security 
+assumptions of the owner program are not carefully evaluated.
+
+## Backwards Compatibility
+
+This feature relaxes existing Loader V3 constraints and is therefore not 
+backwards compatible for consensus, necessitating a feature gate. For CLI and 
+tooling, the change is fully backwards compatible.
```
buffalojoec, line 157: It would be good to add a small line/paragraph mentioning that the signer for `DeployWithMaxDataLen` doesn't have to match the buffer now, but it's going to still be a required signer as a sanity check to ensure the deployer has control over the to-be upgrade authority.

buffalojoec (changes requested):
@Lichtso is correct - we can definitely make this proposal more clear.

I wrote up a draft implementation here: https://github.com/anza-xyz/agave/pull/9854

You can see the feature changes only the `IncorrectAuthority` error case, not the missing signature.
https://github.com/buffalojoec/solana/blob/bd7db804ac3cbd72276d4edac45c2b4ab62ffda3/programs/bpf_loader/src/lib.rs#L534-L541
https://github.com/buffalojoec/solana/blob/bd7db804ac3cbd72276d4edac45c2b4ab62ffda3/programs/bpf_loader/src/lib.rs#L709-L716

I left some comments about two things. For starters, I think it would be really good to mention that the signature is stil required, obviously since `Upgrade` needs to match it to the program's upgrade authority but also for `DeployWithMaxDataLen` as a sanity check for the to-be upgrade authority.

The second thing is that we can't actually remove the ownership check, since both instructions will modify the buffer account, and the program can't do that if the buffer is not owned by Loader V3. We can either keep the ownership check (as per my comment) or change the behavior to not modify buffer accounts, so they can be reused.
https://github.com/anza-xyz/agave/blob/7b6683976906e298e869c680a58f14035c114b06/programs/bpf_loader/src/lib.rs#L578-L584
https://github.com/anza-xyz/agave/blob/7b6683976906e298e869c680a58f14035c114b06/programs/bpf_loader/src/lib.rs#L823-L834

buffalojoec:
@deanmlittle one suggestion for the ownership check is to only **partially relax** it.

We could add an argument to both deployment instructions that tells Loader V3 what to do with the buffer account.

```
enum LoaderV3Instruction {
    DeployWithMaxDataLen { reuse_buffer: bool },
    Upgrade { reuse_buffer: bool },
}
```

* `reuse_buffer=true`:
  * No signature required.
  * No ownership check required.
  * Buffer account is not modified.
* `reuse_buffer=false`:
  * Signature of the buffer authority is required.
  * Buffer must by owned by Loader V3.
  * Buffer account is drained according to existing behavior.

---

Personally I like more imperative arguments, but we'd have to treat the absence of that byte as `false`, which is a bit uglier, but doable. Obviously the control flow would be reversed.

```
enum LoaderV3Instruction {
    DeployWithMaxDataLen { close_buffer: bool },
    Upgrade { close_buffer: bool },
}
```

deanmlittle:
> Personally I like more imperative arguments, but we'd have to treat the absence of that byte as `false`, which is a bit uglier, but doable. Obviously the control flow would be reversed.

I also prefer the idea of `close_buffer` as the additional argument. I think it is actually fine to change the default behavior to not close the buffer by default (in the program). The worse case scenario here would be that program deploys temporarily take up more rent for users who don't upgrade their CLI which is in no way a dealbreaker imo. As for the default behavior in the CLI, I am ambivalent about that. @jacobcreech do you have any thoughts on this?

buffalojoec:
> The worse case scenario here would be that program deploys temporarily take up more rent for users who don't upgrade their CLI which is in no way a dealbreaker imo.

The CLI can be updated ASAP to try to mitigate this. We can add the extra byte to pass `close_buffer=true` by default before the feature even goes live.

```rust
for instruction in [
    UpgradeableLoaderInstruction::DeployWithMaxDataLen { max_data_len: 1000 },
    UpgradeableLoaderInstruction::Upgrade,
] {
    let mut data = bincode::serialize(&instruction).unwrap();
    let data_len_original = data.len();
    
    // Add trailing bytes.
    data.extend_from_slice(&[0xDE, 0xAD, 0xBE, 0xEF]);
    let data_len_with_trailing = data.len();

    // Deserialize.
    let result: UpgradeableLoaderInstruction =
        limited_deserialize(&data, solana_packet::PACKET_DATA_SIZE as u64)
        .unwrap();

    assert_eq!(
        result,
        instruction,
        "Deserialized instruction should match original"
    );

    assert_eq!(
        data_len_with_trailing,
        data_len_original + 4,
    );
}
```

proposals/0430-relax-program-buffer-constraints.md
deanmlittle: resolved by https://github.com/solana-foundation/solana-improvement-documents/pull/430/commits/fcbf33470ae5bd45cdda0c991dcf07a44c7cb551

buffalojoec (dismissed):

mjain-jump (approved):
lgtm!

Lichtso:
> I also prefer the idea of close_buffer as the additional argument. I think it is actually fine to change the default behavior to not close the buffer by default (in the program). The worse case scenario here would be that program deploys temporarily take up more rent for users who don't upgrade their CLI which is in no way a dealbreaker imo. As for the default behavior in the CLI, I am ambivalent about that.

I agree with @Dean that closing the buffer does not need to be part of Upgrade, the accidental omission is easily corrected. Furthermore, this begs the question: Why even add the option to close the buffer as a parameter to the instruction interface? We can add an explicit `Close` of the buffer in the same TX in the CLI.

Since we split the authorities of program and buffer in this SIMD anyway, this also opens up work flows in which the buffer is recycled in a different TX by a different key pair altogether. In that sense separating `Upgrade` and `Close` buffer is a natural consequence.

topointon-jump (dismissed):

buffalojoec:
> I agree with @dean that closing the buffer does not need to be part of Upgrade, the accidental omission is easily corrected. Furthermore, this begs the question: Why even add the option to close the buffer as a parameter to the instruction interface? We can add an explicit `Close` of the buffer in the same TX in the CLI.
> 
> Since we split the authorities of program and buffer in this SIMD anyway, this also opens up work flows in which the buffer is recycled in a different TX by a different key pair altogether. In that sense separating `Upgrade` and `Close` buffer is a natural consequence.

I'm hesitant to add more changes to this proposal. The goal for this and the other Loader V3 SIMDs was to keep the change spread to a minimum and minimize disruptions to developers.

You're correct that we could include a `Close` instruction in the same transaction and that would preserve existing CLI workflows. However, removing automatic closure in `Upgrade` would disrupt anyone who does not use the CLI - such as CPI-based workflows.

I would personally rather stick to the original goal and focus on the wins we want to see, rather than add breaking changes to obtain modularity. Save that for Loader V4 IMO.

deanmlittle:
> > I also prefer the idea of close_buffer as the additional argument. I think it is actually fine to change the default behavior to not close the buffer by default (in the program). The worse case scenario here would be that program deploys temporarily take up more rent for users who don't upgrade their CLI which is in no way a dealbreaker imo. As for the default behavior in the CLI, I am ambivalent about that.
> 
> 
> 
> I agree with @Dean that closing the buffer does not need to be part of Upgrade, the accidental omission is easily corrected. Furthermore, this begs the question: Why even add the option to close the buffer as a parameter to the instruction interface? We can add an explicit `Close` of the buffer in the same TX in the CLI.
> 
> 
> 
> Since we split the authorities of program and buffer in this SIMD anyway, this also opens up work flows in which the buffer is recycled in a different TX by a different key pair altogether. In that sense separating `Upgrade` and `Close` buffer is a natural consequence.

If we were designing from first principles with a blank slate, I would totally agree with you! The reality today is that separating them entirely in the current loader has a few undesirable outcomes:

1. It would cease to be backwards -compatible with expected behavior in existing CPI/CLI workflows.

2. Splitting out the Close instruction means that, although overall cost is the same, upfront capital required for a deployment is actually doubled as you cannot directly reuse the lamports from closing the buffer to deploy the program. 

While these may not lead to catastrophic failure, they do represent major pain points for developers today and thus lead to undesirable outcomes.

For these two reasons, in the context trying to achieve an easy win of added flexibility without breaking existing functionality, I believe the current design is more suitable.

TMbpinHHdcnsS4q5LyhDH5iYEZSdX5bPFM:
اريد سحب الرصيد على رقم الهاتف

buffalojoec:
@Lichtso was there anything else you were looking for in this proposal?

As currently written, `close_buffer=true` preserves the exact same behavior as the current implementation. The new functionality is completely opt-in and *relaxes* checks (non-additive).

Lichtso (approved):

proposals/0430-relax-program-buffer-constraints.md
```diff
@@ -0,0 +1,133 @@
...
+advanced and flexible deployment workflows.
+
+## New Terminology
+
+No new terminology is introduced by this proposal.
+
+## Detailed Design
+
+The `DeployWithMaxDataLen` and `Upgrade` instructions will be updated to include
+an optional boolean input. If not provided, the default will be `true`.
```
bw-solana: nit: optional `close_buffer` boolean input

proposals/0430-relax-program-buffer-constraints.md
```diff
@@ -0,0 +1,133 @@
...
+## Alternatives Considered
+
+- Introduce a new loader that enables these behaviors by default
+- Only relax the authority constraint (similar outcome, but requires CPI)
+- Retain existing constraints and make no changes
+
+## Impact
+
+This change enables more flexible program deployment workflows, unlocking new 
+use cases for developers.
```
bw-solana: I think it would be good to illustrate some examples of impactful use cases that are enabled

proposals/0430-relax-program-buffer-constraints.md
bw-solana: > If not provided, the default will be `true`.

Does this mean that some older CLI that doesn't understand the `close_buffer` param exists will still work and the ix will deserialize properly (just won't be able to take advantage of the `close_buffer=false` case?)

proposals/0430-relax-program-buffer-constraints.md
```diff
@@ -0,0 +1,133 @@
...
+```
+
+The accounts required by the instructions are unchanged, but the signer
+requirements differ based on the value of the `close_buffer` option.
+
+For a value of `true`, existing behavior is preserved. The buffer account will
+be closed (lamports transferred to a designated recipient and account data
+zeroed).
+
+For a value of `false`, the buffer account is not modified, enabling reuse for
```
bw-solana, line 63: how do the rent lamports and spill account work in this case? Presumably the buffer account no longer funds the program data account for `Upgrade`

proposals/0430-relax-program-buffer-constraints.md
buffalojoec: Yep that's correct, existing tooling will continue to default to the existing behavior.

proposals/0430-relax-program-buffer-constraints.md
buffalojoec: Yes also correct. I can update the proposal to be more explicit about this. If you use someone else's buffer, you pay all the rent for the programdata account. In the end the net rent cost is the same, though, since you didn't pay for the buffer.

buffalojoec:
@deanmlittle I've created a PR on your branch for @bw-solana 's feedback: https://github.com/blueshift-gg/solana-improvement-documents/pull/4

joncinque:
Sorry I'm coming late to the game on this. The changes make sense to me, but why don't we just create two new instructions rather than hacking the existing ones?

Could be `DeployWithMaxLenAndKeepBuffer` / `UpgradeAndKeepBuffer` or `DeployWithMaxLenWithoutClose` / `UpgradeWithoutClose` or `DeployWithMaxLenAndMaintainBuffer` / `UpgradeAndMaintainBuffer` or something else. We could also add `Unchecked` somewhere in the name.

Then we don't have to do goofy things on the instruction serde that will certainly confuse people.

deanmlittle:
> Sorry I'm coming late to the game on this. The changes make sense to me, but why don't we just create two new instructions rather than hacking the existing ones?
> 
> Could be `DeployWithMaxLenAndKeepBuffer` / `UpgradeAndKeepBuffer` or `DeployWithMaxLenWithoutClose` / `UpgradeWithoutClose` or `DeployWithMaxLenAndMaintainBuffer` / `UpgradeAndMaintainBuffer` or something else. We could also add `Unchecked` somewhere in the name.

DeployWithMaxLen was a really stupid design choice to begin with. Having two additional variants of it is, imo, even worse than a few self-explanatory optional parameters on the existing instruction that will only ever be consumed by hardcore developers with a specific use case in mind. I think 3 instructions actually causes unnecessary code duplication and begs the question of "which one should I use?" if someone were to stumble accross them somehow.
 
> Then we don't have to do goofy things on the instruction serde that will certainly confuse people.

Unfortunately, the existing enum parameter design is already *extremely* goofy. Nobody but very experienced devs interact with the loader program at a library level anyway. Users primarily only consume it via Squads or the Solana CLI. As existing behavior remains unchanged, nobody is about to get confused by anything unless it's our CLI itself (tbh, many such cases).

bw-solana (approved):

joncinque:
> Having two additional variants of it is, imo, even worse than a few self-explanatory optional parameters on the existing instruction that will only ever be consumed by hardcore developers with a specific use case in mind.

It would be one additional variant of `DeployWithMaxLen` and one additional variant of `Upgrade`. The user-facing functions could even hide it by creating the different instruction variants depending on a bool flag, ie:

```
fn deploy_with_max_len(..., close_buffer: bool) -> Instruction {
    if close_buffer {
        Instruction::NewVariant
    } else {
        Instruction::OldVariant
    }
}
```

My point is just about the serialization format for these, which is unnecessarily complicated. Why do something complicated / non-standard when we can do something simple / standard?

> Unfortunately, the existing enum parameter design is already extremely goofy.

Maybe I'm missing something, but nothing particularly strange jumps out in https://github.com/anza-xyz/solana-sdk/blob/d8e311b30785b398f689528443ffd50db2a262f0/loader-v3-interface/src/instruction.rs#L18-L201

mjain-jump (approved):

topointon-jump (approved):

deanmlittle:
> My point is just about the serialization format for these, which is unnecessarily complicated. Why do something complicated / non-standard when we can do something simple / standard?

Fair enough. Personally, I would argue the entire "Enum args" thing we do here is completely "non-standard" in 2026. Every onchain program today splits the concept of a discriminator from the arguments of its instruction. What is or isn't "standard" is a little arbitrary here. So long as we have concept ACK, that's what matters for the purpose of this SIMD, and the concept hasn't really evolved since we last discussed it in BP Abu Dhabi. We're pretty set on not adding additional instructions, but if you have some opinions about the final implementation of the changes to the instruction, such as serialization or whatever, perhaps we can take this discussion to the PR itself? I believe @buffalojoec already has a PoC in progress.

buffalojoec:
> I believe @buffalojoec already has a PoC in progress.

https://github.com/anza-xyz/solana-sdk/pull/515

joncinque:
> We're pretty set on not adding additional instructions, but if you have some opinions about the final implementation of the changes to the instruction, such as serialization or whatever, perhaps we can take this discussion to the PR itself?

Yep, that's fine by me

buffalojoec:
> We're pretty set on not adding additional instructions, but if you have some opinions about the final implementation of the changes to the instruction, such as serialization or whatever, perhaps we can take this discussion to the PR itself?

Unfortunately, if the implementation is going to be blocked on whether we use new variants or input parameters, we'll have to iron that out here so that it's reflected in the spec.

We've historically modeled program instructions after what bincode supports. However, I've hacked up a simple version of the `loader-v3-interface` that doesn't use bincode or bytemuck and instead just does everything by hand. I'm sure @deanmlittle could be nerd-sniped into iterating on it even further.

https://github.com/buffalojoec/solana-sdk/commits/loader-v3-interface-zc/

Aside from squeezing some perf gains by not allocating a vector for ELFs and `Write` chunks, the rub is that optional parameters look less ugly when we don't use bincode. The second commit is me adding SIMD-0430 support, which is much slimmer than https://github.com/anza-xyz/solana-sdk/pull/515

I'm personally more in favor of getting away from the bincode era:
* The first four bytes - the discriminator - tell the dispatch which instruction this payload is for.
* The rest of the bytes are arguments; some mandatory and some optional.

You could make the argument that variances in account role requirements might warrant a separate variant, but we already use optional accounts, and programs in the wild are decoupling accounts and roles from their primary instruction dispatch discriminators in increasingly creative ways. There's no reason we can't do the same.

joncinque:
To simplify my point: why do something complicated when we can do something simple?

If we have to hand-write parsers, we've already lost, since libraries and frameworks should be able to generate these wherever possible.

Of course, in some cases we need more complicated approaches, but that's not the case here.

deanmlittle:
>Unfortunately, if the implementation is going to be blocked on whether we use new variants or input parameters, we'll have to iron that out here so that it's reflected in the spec.

I agree that would be the case, but is that what's happening here? We had this out in Abu Dhabi where we stressed the importance of trying to maintain backwards compatibility whilst unlocking more flexibility in the existing instructions. This SIMD has been live for 4 months and we have had it out already, but still collectively decided on extending the input parameters. Is there a good reason why we're reviewing this again now?

>We've historically modeled program instructions after what bincode supports. 

The great thing about this decision is we now have infinite room for optimization! 

>However, I've hacked up a simple version of the loader-v3-interface that doesn't use bincode or bytemuck and instead just does everything by hand. I'm sure @deanmlittle could be nerd-sniped into iterating on it even further.

Thanks for your hard work. Would be happy to see if we can do better. Give us a few days.

>I'm personally more in favor of getting away from the bincode era:

Kill it with fire.

>* The first four bytes - the discriminator - tell the dispatch which instruction this payload is for.
>* The rest of the bytes are arguments; some mandatory and some optional.

This is how all programs should be written in 2026. It seems this is the major point of contention. However, if it actually just boils down to a refactor of bad legacy code which changes the deserialization method used to consume the exact same data, I'm not sure that has anything to do with the SIMD discussion?

deanmlittle:
> If we have to hand-write parsers, we've already lost, since libraries and frameworks should be able to generate these wherever possible.

Simplicity.jpg
Screenshot 2026-04-14 at 6 14 22 PM

buffalojoec:
> To simplify my point: why do something complicated when we can do something simple?

I'm not really convinced the design is complicated, though.

> If we have to hand-write parsers, we've already lost, since libraries and frameworks should be able to generate these wherever possible.

It's not really that we *have* to, I'm just demonstrating that the only reason my SDK PR looks outrageous is because we've been married to bincode. There are probably tools that handle this just fine, without hand-rolling.

joncinque:
> There are probably tools that handle this just fine, without hand-rolling.

Every serialization library handles my suggestion with zero work. It also allows for zero-copy on instruction data with no other weirdness.

It's a nit, so I don't want to blow this out of proportion, but also, no one has concretely responded to my comments. If the real response is "I don't care", that's a bit unsatisfying, but also a valid response to a nit.

buffalojoec:
> It's a nit, so I don't want to blow this out of proportion, but also, no one has concretely responded to my comments. If the real response is "I don't care", that's a bit unsatisfying, but also a valid response to a nit.

Concretely I think the proposal as written and approved is a completely valid way to implement the toggle behavior and I prefer it over awkwardly adding new variants which dispatch to the same instruction but toggle one subroutine. If it's really just a nit, and there isn't a critical reason we can't continue with this implementation, I am in favor of proceeding. 

Surprisingly, Loader V3 actually doesn't have such duplicated variants as of yet. Even `ExtendProgramChecked` would have been an outright replacement for `ExtendProgram`.

Nonetheless, I've converted the interface to use `wincode`, and the support for such an argument in wincode is trivial: https://github.com/anza-xyz/solana-sdk/pull/689

PR #432, opened 2025-12-18
SIMD-0432: Loader V3: Reclaim Closed Program

buffalojoec:


deanmlittle:
lgtm!

deanmlittle (dismissed):

proposals/0432-loader-v3-reclaim-closed-program.md
```diff
@@ -0,0 +1,112 @@
...
+of funds and persistent state bloat, while preserving security guarantees for
+users who explicitly wish to permanently disable a program ID.
+
+## Security Considerations
+
+When closing a program, the program account must not be reassigned to the
+system program. Doing so would allow redeployment within the same slot,
+potentially corrupting the program cache. When tombstoning, assigning the
+account to itself ensures the address is permanently unusable, as self-owned
+accounts cannot be modified by any program.
```
Lichtso: > the program account must not be reassigned to the system program

is in conflict with

> In both workflows ... and reassigned to System.

Instead I would suggest to update the "Security Considerations" to simply list the two constraints and why they are needed:

1. There must be at most one redeployment of a program in the same Slot, so that the index key in the global program cache is unique.
2. The loader of an invoked program may not change in the same TX, so that the ABI and its alignment requirements do not change during CPI.

Both of these can be guaranteed by either placing a tombstone (thus finalizing an account) or emptying it to be rent collected at the end of the TX. Never assigning to the System program explicitly during the TX. All of this is already in the "Detailed Design" section, but you still **need to add a check** for the last deployment slot such that a program can not be closed in the same slot it was redeployed in. Otherwise somebody can repeat the following in a loop inside the same slot at the same program address: init, deploy, close, reclaim ... repeat.

proposals/0432-loader-v3-reclaim-closed-program.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Backwards Compatibility
+
+This change modifies the semantics of an existing Loader v3 instruction and
+therefore requires a feature gate for consensus safety.
+
+From a tooling perspective, the change is backwards compatible, though tooling
+updates are required to access the new explicit tombstoning behavior.
```
Lichtso, line 261: What about the programs which have been closed while funded and are currently stuck in limbo? How about allowing them to be re-closed properly so to speak (given the program account key still exists)?

proposals/0432-loader-v3-reclaim-closed-program.md
```diff
@@ -0,0 +1,112 @@
...
+tombstoning for users who require it.
+
+## New Terminology
+
+No new terminology is introduced by this proposal.
+
+## Detailed Design
+
+The `Close` instruction will be updated to include an optional boolean input. If
+not provided, the default will be `false`.
```
mjain-jump: wouldnt you want it to be true to maintain legacy behavior?

Lichtso: I think the main point of this SIMD is not to maintain legacy behavior, but to change behavior unless explicitly opted out of.

buffalojoec: > I think the main point of this SIMD is not to maintain legacy behavior, but to change behavior unless explicitly opted out of.

Yeah that's correct. It's intentionally a breaking change. You can preserve the legacy behavior if you want to by explicitly specifying.

The logic here is that the current behavior has shown to be a bit of a "gotcha", so we want to eliminate that footgun.

proposals/0432-loader-v3-reclaim-closed-program.md
buffalojoec: > > the program account must not be reassigned to the system program
> 
> is in conflict with
> 
> > In both workflows ... and reassigned to System.

The second snippet is referring to the programdata account, not the program account. However I agree the Security Considerations section could use clarification.

> Instead I would suggest to update the "Security Considerations" to simply list the two constraints and why they are needed:
> 
> 1. There must be at most one redeployment of a program in the same Slot, so that the index key in the global program cache is unique.
> 2. The loader of an invoked program may not change in the same TX, so that the ABI and its alignment requirements do not change during CPI.
>
> Both of these can be guaranteed by either placing a tombstone (thus finalizing an account) or emptying it to be rent collected at the end of the TX. Never assigning to the System program explicitly during the TX.

These are already existing constraints that are not altered by this proposal. You just want this section to state why this change is compliant, right? Mainly mention why we do the garbage collection rather than direct assign.

> All of this is already in the "Detailed Design" section, but you still **need to add a check** for the last deployment slot such that a program can not be closed in the same slot it was redeployed in. Otherwise somebody can repeat the following in a loop inside the same slot at the same program address: init, deploy, close, reclaim ... repeat.

Yeah that's true, if it's not tombstoned you could deploy, close, then redeploy over and over. We can add this check, so that this operation can't loop in the same slot and trip up the cache.

proposals/0432-loader-v3-reclaim-closed-program.md
buffalojoec: Nice idea! I was going to reply with a suggestion, but I think I managed to work it into a pretty clean commit. Let me know what you think!

Lichtso (approved):

mjain-jump (approved):

proposals/0432-loader-v3-reclaim-closed-program.md
```diff
@@ -0,0 +1,255 @@
...
+account, preventing reuse of the program ID. This behavior has led to several
+issues:
+
+- Loss of funds: Users frequently confuse close program with close buffer in the
+  Solana CLI, unintentionally irreversibly disabling programs.
+- Permanent account bloat: Tombstoned program accounts cannot be reclaimed and
+  accumulate indefinitely in the accounts database.
+- RPC performance degradation: getProgramAccounts against the loader v3 program
+  must return all program accounts, including closed ones, increasing response
+  size and latency.
```
topointon-jump, line 33: Do we know how many tombstoned program accounts there are? Is this actually a performance issue? I'm not against this proposal but I think in general we should be clear about the motivations for proposals, and keep the motivations list small - only including the most impactful ones.

topointon-jump (dismissed):

topointon-jump (commented):
Is there a risk that a user will accidentally get their program address taken over by someone else? Say if the account is reclaimable, and they don't realise it, or an adversary claims it during the window where the original user is trying to reclaim it... are we concerned about this scenario?

buffalojoec:
> Is there a risk that a user will accidentally get their program address taken over by someone else? Say if the account is reclaimable, and they don't realise it, or an adversary claims it during the window where the original user is trying to reclaim it... are we concerned about this scenario?

The only way someone could reclaim another person's tombstoned program is if they themselves held the program keypair. I did consider this scenario - since devs might have previously assumed this keypair was forever useless and thrown it out, or disregarded mistakes like accidentally committing it to GitHub - and I was apprehensive at first. However, we figured this would be an exceptionally rare case.

What do you think?

proposals/0432-loader-v3-reclaim-closed-program.md
buffalojoec: I just ran a scrape and found 49,210 Loader V3 program accounts on mb, 33,804 of which are tombstones.
If you want the script, lmk.

Lichtso (approved):

proposals/0432-loader-v3-reclaim-closed-program.md
```diff
@@ -0,0 +1,255 @@
...
+model allows safe address reuse by default while preserving explicit
+tombstoning for users who require it.
+
+## New Terminology
+
+No new terminology is introduced by this proposal.
+
+## Detailed Design
+
+The `Close` instruction will be updated to include an optional boolean input. If
```
bw-solana: nit: maybe mention again this is the Loader-v3 Close instruction here

proposals/0432-loader-v3-reclaim-closed-program.md
```diff
@@ -0,0 +1,255 @@
...
+metadata. The program account will then be assigned to itself, creating a
+permanent tombstone for the program.
+
+```
+                         Close { tombstone }
+                                |
+                    +-----------+-----------+
+                    |                       |
+             tombstone=false          tombstone=true
+                    |                       |
```
bw-solana: should we add "program was deployed in the current slot" branches to this picture?

bw-solana (dismissed):
LGTM.

Left a couple minor nits

buffalojoec:
Ahh gud op sec

proposals/0432-loader-v3-reclaim-closed-program.md
buffalojoec: https://github.com/solana-foundation/solana-improvement-documents/pull/432/changes/0877b2f45b4150f3837fe53c1f77682b0024f346

proposals/0432-loader-v3-reclaim-closed-program.md
buffalojoec: https://github.com/solana-foundation/solana-improvement-documents/pull/432/changes/0877b2f45b4150f3837fe53c1f77682b0024f346

bw-solana (approved):

proposals/0432-loader-v3-reclaim-closed-program.md
```diff
@@ -0,0 +1,261 @@
...
+when explicitly requested.
+
+## Motivation
+
+Today, closing an upgradeable program permanently tombstones its program
+account, preventing reuse of the program ID. This behavior has led to several
+issues:
+
+- Loss of funds: Users frequently confuse close program with close buffer in the
+  Solana CLI, unintentionally irreversibly disabling programs.
```
t-nelson, line 28: whether or not the proposal addresses this is up to the deployer

proposals/0432-loader-v3-reclaim-closed-program.md
```diff
@@ -0,0 +1,261 @@
...
+- Permanent account bloat: Tombstoned program accounts cannot be reclaimed and
+  accumulate indefinitely in the accounts database.
```
t-nelson, line 30: surely the accounts orphaned by closing the program massively outweigh the program itself?

proposals/0432-loader-v3-reclaim-closed-program.md
```diff
@@ -0,0 +1,261 @@
...
+  Loader v3 until garbage-collected at transaction end, preventing same-TX
+  redeployment. Additionally, closing without tombstoning is rejected if the
+  program was deployed in the current slot, preventing multiple-TX loops.
+
+- **Programdata account**: Fully deallocated and reassigned to System. This is
+  safe because programdata is not used for cache indexing or invocation.
+
+- **Tombstone**: When tombstoning, the program account is assigned to itself,
+  permanently locking the address. Self-owned accounts cannot be modified.
+
```
t-nelson, line 254: we aren't going to discuss how this leaves any accounts owned by an untombstoned, closed program open to be drained later by the deployer or anyone who manages to acquire the program id keypair?

we're talking about deployers too careless to make sure that they're performing the operation that they believe they are. why do we think they'll be careful with keys?

deanmlittle: imo, if you do the cost-benefit analysis on this, tombstoning programs has resulted in millions of dollars of funds being permanently locked and was a terrible choice. it really failed to protect anyone from anything and caused a whole lot of harm. my $0.02: "oh no, several cases of defi protocols and token sales with millions of dollars locked forever by our CLI having the worst UX imaginable can suddenly be recovered, a bunch of people can get their money back, a bunch of accounts might get cleaned up and people who throw their keypairs out into the wild learn why that's a bad idea. how will i ever live with the weight of this decision? 🤣 "

t-nelson: why has the cli not been made more intuitive in the meantime?

Lichtso: https://github.com/solana-labs/solana/pull/27108

I agree with @deanmlittle that the lost and revealed keys argument is negligible. Especially since the current status is that they don't have access to these funds, thus nothing changes in these cases.

proposals/0432-loader-v3-reclaim-closed-program.md
buffalojoec: Sure, but the motivation stands. Now this action is reversible.

mjain-jump (approved):

topointon-jump (approved):

bart1e:
I know that this PR has already been closed, but I believe that this SIMD may break the `Close` instruction when invoked via CPI (DAO or multisig cases). The problem is that executable accounts are not synchronized on CPI return and I'm providing more details below (the analysis below is relevant to the Agave validator).

The 13th step ("Clear the program account's data and resize to zero.") unconditionally modifies the program account by setting its data size to 0, which means that for each `Close` invocation, program account is modified. The issue is that during CPI calls, executable accounts are [skipped entirely in `translate_accounts_common`](https://github.com/anza-xyz/agave/blob/091099f16a/program-runtime/src/cpi.rs#L1061) - they are not added to the translated accounts list. This means that on CPI return, [`update_caller_account`](https://github.com/anza-xyz/agave/blob/091099f16a/program-runtime/src/cpi.rs#L904-L917) is never called for them, so the caller's serialized view of the program account retains the pre-`Close` state (in particular, non-zero data).

When the instruction completes, [deserialization](https://github.com/anza-xyz/agave/blob/091099f16a/program-runtime/src/serialization.rs#L661-L662) sees the stale `post_len` from the serialized buffer, but the actual account length is 0 (cleared by `Close`). The length mismatch is interpreted as a resize attempt, which will fail with `AccountDataSizeChanged` because the calling program does not own the account.

As a consequence, any program instruction that calls `Close` via CPI will fail after the activation of this SIMD in its current form.

PR #433, opened 2025-12-18
SIMD-0433: Loader V3: Set Program Data to ELF Length

buffalojoec:


proposals/0433-loader-v3-set-program-data-to-elf-length.md
```diff
@@ -0,0 +1,69 @@
...
+
+No new terminology is introduced by this proposal.
+
+## Detailed Design
+
+The `Upgrade` instruction will be updated to automatically resize the program
+data account to match the length of the ELF in the buffer being deployed.
+
+If the new ELF is larger than the current program data account, the upgrade will
+fail. The account must first be extended to the required size via the
```
deanmlittle:
```suggestion
fail. The account must first be extended to at least the required size via the
```

deanmlittle (dismissed):

proposals/0433-loader-v3-set-program-data-to-elf-length.md
```diff
@@ -0,0 +1,69 @@
...
+No new terminology is introduced by this proposal.
+
+## Detailed Design
+
+The `Upgrade` instruction will be updated to automatically resize the program
+data account to match the length of the ELF in the buffer being deployed.
+
+If the new ELF is larger than the current program data account, the upgrade will
+fail. The account must first be extended to at least the required size via the
+`ExtendProgram` instruction.
```
Lichtso: How does this interact with https://github.com/solana-foundation/solana-improvement-documents/pull/164? Or rather, why require an `ExtendProgram` instruction at all and not have that be automatic in `Upgrade` instead of failing if the account is too small? Obviously the required funds would still have to be deposited before hand.

Lichtso: IIRC Breakpoint we discussed the alternative workflow in which programs which want to self upgrade to a larger ELF would reassign their authority to a TX fee payer temporarily, then upgrade at top-level (to avoid the 10 KiB CPI account growth limit of ABI v1) and reassign back to the original authority, all with the atomicity of a single TX.

buffalojoec: This proposal is just simply the lightest lift to solve the issue of excess space and lamports. It's trivial: existing workflows already resize the account to be large enough for the ELF, but there's no way to reclaim excess if you overshoot it.

If we want to go all the way we can. My main concern would be what happens to any existing workflows that use `ExtendProgram`. Would we continue to support the instruction, perhaps post-#431, so as to not break anyone's flows? Or, since #431 is already breaking, do we just close #431 and make all resizing part of the upgrade?

Lichtso: > we just close https://github.com/solana-foundation/solana-improvement-documents/pull/431 and make all resizing part of the upgrade

That is what I was saying: "have that be automatic in Upgrade".

mjain-jump (approved):
lgtm!

proposals/0433-loader-v3-set-program-data-to-elf-length.md
buffalojoec: In my opinion, we should keep this SIMD focused on addressing the core issue, which is the inability to reclaim lamports for an oversized programdata account. I don't think we should bloat the change with more complexity at this time. If we want to add another step to enable more automation within the upgrade workflow and move away from `Extend*`, it's pretty trivial to do so in the future. However, nobody has asked for this afaik.

> IIRC Breakpoint we discussed the alternative workflow in which programs which want to self upgrade to a larger ELF would reassign their authority to a TX fee payer temporarily, then upgrade at top-level (to avoid the 10 KiB CPI account growth limit of ABI v1) and reassign back to the original authority, all with the atomicity of a single TX.

This sounds a little more hacky than it's worth just to avoid having to best-guess a size increase before running an upgrade. If we bake in resizing, you'll still have to best-guess the new size anyway, in order to top up rent. Seems like a small marginal benefit overall IMO.

Lichtso: > This sounds a little more hacky than it's worth just to avoid having to best-guess a size increase before running an upgrade.

This was not about guessing the size at all. It was about not being able to upgrade to larger programs in a single instruction, no matter if it is done in Extend or Upgrade, both are limited to +10 KiB in CPI of ABI v1.

buffalojoec: > > This sounds a little more hacky than it's worth just to avoid having to best-guess a size increase before running an upgrade.
> 
> This was not about guessing the size at all. It was about not being able to upgrade to larger programs in a single instruction, no matter if it is done in Extend or Upgrade, both are limited to +10 KiB in CPI of ABI v1.

I think you misunderstood my comment. I'm not implying you were describing the act of guessing the size. I'm saying the workflow you outlined would be required if we decided to handle both resizing events automatically via `Upgrade`. The goal of doing such a thing would be to avoid the extra step whereby developers must know their new program size ahead of time. My point is that - even if we get rid of `ExtendProgram` completely - devs will always have to know the new size of their program in order to top up rent, so the point is moot IMO.

Lichtso: Yes, we are on the same page and I understood your comment as such: "even if we get rid of ExtendProgram completely - devs will always have to know the new size of their program in order to top up rent, so the point is moot IMO" agreed, and I did not bring it up as a reason.

The reasons (for handling the size changes in `Upgrade`) are:

- To avoid having to deal with changes to `ExtendProgram` (such as SIMD-0431) which has been controversial in the past (see SIMD-0164).
- To have both size increases and decreases handled in a unified matter, instead of having one be implicit and the other explicit.
- To have only one instruction change the size, so that the CPI account size increase limit of ABI v1 only needs to be dealt with once per TX.

buffalojoec: I'm willing to adjust the proposal to support bidirectional resizing, but I don't think we should eliminate `ExtendProgram`/`ExtendProgramChecked` in this proposal.

We should seek more opinions from stakeholders on the solution you described earlier in this thread - temporarily changing the authority to avoid the 10KiB CPI limit in ABI v1 - and route the feedback into SIMD-0431 or SIMD-0164.

proposals/0433-loader-v3-set-program-data-to-elf-length.md
```diff
@@ -0,0 +1,124 @@
...
+lamports after the auto-resizing from `Upgrade`. This was rejected in favor of
+automatically refunding surplus lamports to the spill account during the
+upgrade itself.
+
+## Impact
+
+This proposal results in a lower program footprint in Accounts DB, incentivizes
+developers to upgrade to newer, more performant libraries and SDKs, and enables
+the recovery of surplus lamports, including those accidentally sent to the
+program data address.
```
Lichtso: Not debiting the Buffer account means the TX needs twice the funds to run, even if it only uses half in the end.

proposals/0433-loader-v3-set-program-data-to-elf-length.md
```diff
@@ -0,0 +1,124 @@
...
+### Buffer Account Lamports
+
+Regardless of SIMD-0430, after the current proposal, buffer accounts will no
+longer be debited or credited lamports during `Upgrade`. This is true for all
+ELF sizing cases - shrinking, growing, or remaining exactly the same.
+
+As mentioned in the previous section, any additional rent exemption required
+for growing an ELF region must be credited to the program data account before
+`Upgrade` is invoked.
+
```
Lichtso, line 76: This means the buffer is not emptied and thus also does not need to be writable.

varo0000000 (approved):
gh pr checkout 433

proposals/0433-loader-v3-set-program-data-to-elf-length.md
buffalojoec: Updated.

Lichtso (approved):

proposals/0433-loader-v3-set-program-data-to-elf-length.md
```diff
@@ -0,0 +1,128 @@
...
+    `Upgrade` instructions, which determines whether the buffer account is
+    closed after the operation
+
+[SIMD-0430]: https://github.com/solana-foundation/solana-improvement-documents/pull/430
+
+## Detailed Design
+
+The `Upgrade` instruction will be updated to automatically resize the program
+data account to match the length of the ELF in the buffer being deployed. This
+applies in both directions: the account may grow or shrink as needed.
```
mjain-jump: you say the programdata account will now be the same length as the ELF, but wouldn't it also need to include some metadata? (i.e. the bpf loader program state)

deanmlittle: Correct. It would still contain:
- discriminator
- slot
- option
- upgrade authority
- program ELF

buffalojoec: Yeah, sorry, maybe I can make that statement clearer in the doc. I was implying metadata + ELF.

buffalojoec: https://github.com/solana-foundation/solana-improvement-documents/pull/433/changes/43c07c48d9e0e2972d80f02fbcc3ffd6f1ca2e42

mjain-jump: thanks!

mjain-jump (approved):
lgtm!

topointon-jump (approved):

De-ASI-INTERFACE (approved):

De-ASI-INTERFACE (approved):

De-ASI-INTERFACE (approved):
Approve

buffalojoec:
@Lichtso @topointon-jump I wanted to decouple this from SIMDs 0430 and 0432, since there's really no reason to make them hard dependencies, and this proposal is trivial.

Also @bw-solana can we get your stamp from Anza side?

bw-solana:
Concept ACK.

Behavior is reasonable and decoupling shrink/grow id good. This obviously changes the semantics of intentionally over-allocated ProgramData accounts w/ `DeployWithMaxDataLen` or `ExtendProgram`

Can we call that out explicitly in Backwards Compatibility / Impact? Not fully operationally backwards-compatible for workflows that intentionally over-reserve ProgramData capacity

bw-solana (approved):

bw-solana (approved):

mjain-jump (approved):

topointon-jump (approved):

buffalojoec:
@Lichtso let me know if you spot anything with the latest revisions that you want to discuss. If not, I'll merge in a few days.

proposals/0433-loader-v3-set-program-data-to-elf-length.md
```diff
@@ -60,28 +48,30 @@ upgrade authority), so the resulting account size is
...
+the reduced rent requirement for the program data account will be refunded to
+the spill account.
 
-### Growing (New ELF is Larger)
+**Growing:** If the new ELF is larger than the current program data account's
+ELF region, the account will be extended to accommodate the new ELF. New
+lamports for the increased rent requirement are expected to be either credited
+to the program data account or available in the buffer account prior to the
+execution of the `Upgrade` instruction. If the available lamports are
+insufficient to satisfy the new rent requirement, the upgrade will fail.
```
Lichtso: One of the later commits took out the error code:

> fail with `InsufficientFunds`

proposals/0433-loader-v3-set-program-data-to-elf-length.md
```diff
@@ -60,28 +48,30 @@ upgrade authority), so the resulting account size is
...
+**Shrinking:** If the new ELF is smaller than the current program data account's
+ELF region, the account will be retracted to the new size. Surplus lamports from
+the reduced rent requirement for the program data account will be refunded to
+the spill account.
 
-### Growing (New ELF is Larger)
+**Growing:** If the new ELF is larger than the current program data account's
+ELF region, the account will be extended to accommodate the new ELF. New
+lamports for the increased rent requirement are expected to be either credited
+to the program data account or available in the buffer account prior to the
```
Lichtso, line 59: I know you want to decouple from SIMD-0430, but we still need to define what to do if the buffer is passed as read-only.

buffalojoec: I think it's tricky to add such a comment, since the program current _requires_ the buffer to be writable. We aren't changing that requirement here, and I believe any SIMDs that do should specify what the implications are for a read-only buffer.

In other words, I'd rather amend SIMD-0430. What do you think?

proposals/0433-loader-v3-set-program-data-to-elf-length.md
buffalojoec: I took it out on purpose, but I'm fine with including it, since we've done this for other program-related SIMDs like SIMD-0123.

buffalojoec:
@Lichtso as briefly mentioned in https://github.com/solana-foundation/solana-improvement-documents/discussions/588, the `Upgrade` instruction is allowed via CPI, unlike other programdata-modifying instructions like `ExtendProgram` and `Write`.

Could you elaborate why we felt `Upgrade` was safe to whitelist for CPI, and does that logic still hold after this SIMD is implemented?

The key aspect is that currently, `Upgrade` does not resize the programdata at all. It asserts the account is large enough and then simply copies from the buffer account directly. After this change, it will also _resize_ the programdata. I'd like to know if there is some reason why this would be dangerous in CPI.

Lichtso:
> if there is some reason why this would be dangerous in CPI.

That upgrade now also resizes is benign IMO, even in the CPI case.

> Could you elaborate why we felt Upgrade was safe to whitelist for CPI

https://github.com/solana-labs/solana/pull/14449

> and does that logic still hold after this SIMD is implemented?

Think so. This SIMD would not change anything about that.

buffalojoec:
> Think so. This SIMD would not change anything about that.

Considering the logic in that PR was "we need Upgrade for something", I would agree with you.

PR #436, opened 2025-12-22
SIMD-0436: Reduce lamports_per_byte from 6960 to 3480

igor56D:


waldruupi:
Nice🔥

proposals/0436-reduce-rent-exempt-minimum-by-2x.md
```diff
@@ -0,0 +1,82 @@
...
+cost more expensive on mainnet beta than on competitors, with no
+compelling justification provided by real resource costs.
+
+## New Terminology
+
+N/A
+
+## Detailed Design
+
+On feature activation, set the effective `lamports_per_byte_year`
```
deanmlittle:
```suggestion
On feature activation, set the effective `lamports_per_byte`
```

Consider using Agave 4.0 terminology. `lamports_per_byte_year` has been flagged deprecated as of 3.1 and `lamports_per_byte` will be used exclusively as of 4.0 which is the earliest this feature could activate.

proposals/0436-reduce-rent-exempt-minimum-by-2x.md
```diff
@@ -0,0 +1,82 @@
...
+```
+ACCOUNT_STORAGE_OVERHEAD = 128
+lamports_per_byte_year = 3480 // reduced from 6960 by this proposal
+exemption_threshold = 1.0
+effective_size = ACCOUNT_STORAGE_OVERHEAD + data_size_bytes
+min_balance = effective_size 
+                            * lamports_per_byte_year
+                            * exemption_threshold
+                       = effective_size * 3480
+```
```
deanmlittle:
```suggestion
```

As this feature won't ship until at least 4.0, one full release cycle after SIMD-0194 is activated in 3.1, I don't think we need all of this additional information about the SIMD. If you do think it's relevant, perhaps consider adding an `extends` field to the frontmatter as is customary in such cases?

proposals/0436-reduce-rent-exempt-minimum-by-2x.md
```diff
@@ -0,0 +1,82 @@
...
+
+## New Terminology
+
+N/A
+
+## Detailed Design
+
+On feature activation, set the effective `lamports_per_byte_year`
+to `current_value / 2` in the runtime, Rent sysvar, and RPC
+helpers (e.g., `getMinimumBalanceForRentExemption`).
```
deanmlittle:
```suggestion
to `current_value / 2` in the bank, followed by updating 
`DEFAULT_LAMPORTS_PER_BYTE` in all relevant SDKs post-activation.
```

proposals/0436-reduce-rent-exempt-minimum-by-2x.md
igor56D: Good call. I was on the fence whether to include this at all but thought the context would be useful for reviewers. A dropdown/hidden section would be appropriate.

proposals/0436-reduce-rent-exempt-minimum-by-2x.md
igor56D: Is there a reason you left the Rent sysvar out? The rent params in the bank are tracked independently from the sysvar, and I believe the latter isn't automatically updated if the former changes.

proposals/0436-reduce-rent-exempt-minimum-by-2x.md
igor56D: I'll just remove the Note, simplify `lamports_per_byte_year` to `lamports_per_byte` and remove references to the `exemption_threshold`. That should trim this down by removing the less relevant information.

proposals/0436-reduce-rent-exempt-minimum-by-2x.md
igor56D: replaced all uses of `lamports_per_byte_year` to `lamports_per_byte`: e99fd07a2f52bf6f92de967a58a8e5b30a759dd7

proposals/0436-reduce-rent-exempt-minimum-by-2x.md
igor56D: e99fd07a2f52bf6f92de967a58a8e5b30a759dd7

proposals/0436-reduce-rent-exempt-minimum-by-2x.md
```diff
@@ -0,0 +1,75 @@
...
+years ago that has since increased significantly in real terms
+due to increases in the SOL price. This has made state allocation
+cost more expensive on mainnet beta than on competitors, with no
+compelling justification provided by real resource costs.
+
+## New Terminology
+
+N/A
+
+## Detailed Design
```
brooksprumo, line 34: We also need a feature gate to raise the lamports_per_byte back to 6960 that ships alongside this one. Should we also include that here in this SIMD?

igor56D: I created a separate SIMD ([0438](https://github.com/solana-foundation/solana-improvement-documents/pull/438)) for the rent increase safeguard.

proposals/0436-reduce-rent-exempt-minimum-by-2x.md
deanmlittle: 🚀

proposals/0436-reduce-rent-exempt-minimum-by-2x.md
deanmlittle: By my understanding, the values in the Rent sysvar are in fact populated from the Bank! In the case of Agave, we first modify the `RentCollector` in our instance of the `Bank` then call this function to update the Sysvar account:

https://github.com/anza-xyz/agave/blob/436aeb987526cac73b6dfe41f35fb71c5b15f22f/runtime/src/bank.rs#L2297-L2304

This should mean the only thing we need to change from a consensus perspective at the slot boundary of activation is the values in those two places in our `Bank`.

Once this is live on all clusters, we would still want to follow up by updating the defaults in all of our SDKs to match the new mainnet values, as much of our testing relies upon `Rent::default()` under the hood, so it would be great for these defaults to match the mainnet values moving forwards for testing purposes. :)

igor56D: Right, the bank is treated as the source-of-truth but there's no automatic per-slot call to `update_rent` so we need to explicitly make the call when this feature is activated. I think we're basically saying the same thing though. I added the change you requested + also mentioned the sysvar to avoid any potential ambiguity. Good point on the SDKs and thanks for the feedback!

mmarzullotv:
hey guys, is this live now?

alnoki:
> hey guys, is this live now?

@mmarzullotv https://github.com/anza-xyz/agave/wiki/Feature-Gate-Tracker-Schedule

PR #437, opened 2025-12-22
SIMD-0437: Incremental Reduction of lamports_per_byte to 696

igor56D:


brooksprumo (approved):
This schedule works for me.

brooksprumo:
Here's a copy of the table I used to compute the numbers/steps:

constants |   |   | step reduction | cummulative reduction | amount to reduce by (relative to starting amount) | resulting lamports-per-byte | effective reduction
-- | -- | -- | -- | -- | -- | -- | --
starting rent | 6,960 |   | 10% | 10% | 626 | 6,334 | 9%
ending rent | 696 |   | 20% | 30% | 1,879 | 5,081 | 27%
difference | 6,264 |   | 40% | 70% | 4,385 | 2,575 | 63%
  |   |   | 20% | 90% | 5,638 | 1,322 | 81%
  |   |   | 10% | 100% | 6,264 | 696 | 90%

brooksprumo (approved):
:shipit:

brooksprumo:
We'll also want to update SIMD-436 to include `superseded-by`. My thought would be to include that here in this PR as well.

proposals/0437-incremental-rent-reduction.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Detailed Design
+
+This proposal defines five incremental reductions to the `lamports_per_byte` constant.
+Each reduction is controlled by its own feature gate. The values are absolute,
+starting from the current value of 6960.
+
+| Step | Target `lamports_per_byte` | Feature Gate |
+|------|----------------------------|--------------|
+| 1    | 6333                       | TBD          |
```
waldruupi: Can we communicate the % reduction in each stable in this table?

proposals/0437-incremental-rent-reduction.md
```diff
@@ -0,0 +1,96 @@
...
+| 2    | 5080                       | TBD          |
```
waldruupi: I know it's TBD, but any rough estimate on when the first would start? 
And duration between the steps?

proposals/0437-incremental-rent-reduction.md
```diff
@@ -0,0 +1,96 @@
...
+| 3    | 2575                       | TBD          |
+| 4    | 1322                       | TBD          |
+| 5    | 696                        | TBD          |
+
+On activation of each feature, the effective `lamports_per_byte` is updated
```
t-nelson: can we also add the revert feature gates here and discuss an approximate observation period for each step before advancing?

t-nelson:
@bw-solana for the poh constant -> variable transition, did we add a noop feature gate that confirmed the plumbing? i don't recall and that may not quite be reasonable now that we have multiple clients in the wild

bw-solana:
> @bw-solana for the poh constant -> variable transition, did we add a noop feature gate that confirmed the plumbing? i don't recall and that may not quite be reasonable now that we have multiple clients in the wild

yes, that was the first feature gate we activated. Basically sanity checking the parent --> child values overwriting transition

proposals/0437-incremental-rent-reduction.md
igor56D: I added criteria for each activation: 2c652fa241bcfa6a72d9079afe3c7d033e019029

I'm assuming this will be somewhat controversial. I separated SIMD-0436 (2x reduction) in case reaching consensus on the criteria delays smaller reductions.

proposals/0437-incremental-rent-reduction.md
igor56D: Added criteria per-reduction: 2c652fa241bcfa6a72d9079afe3c7d033e019029

proposals/0437-incremental-rent-reduction.md
igor56D: 2c652fa241bcfa6a72d9079afe3c7d033e019029

proposals/0437-incremental-rent-reduction.md
```diff
@@ -0,0 +1,147 @@
...
+#### Step 2 Criteria
+
+Depends on (1).
+
+Average net increase per-day following activation of (1):
+
+- In account data size: does not exceed 250MB (measured over at least 3
+  weeks).
+- In number of accounts: does not exceed 1.5M (measured over at least 3
+  weeks).
```
brooksprumo: We're close-ish to these numbers today, but in spurts. Sometimes we exceed 'em. Looking back six months from today we'd be fine. IOW, we don't exceed these numbers for three weeks straight. Who knows what the cluster usage will look like by the time we activate these features. We may not want to hard code values here in the SIMD, but instead describe how we'll determine these numbers instead. Wdyt?

igor56D: I based this off the historical data you collected and concluded that it'd be unlikely for these limits to be exceeded if we averaged over 3 weeks, since that should tolerate some spikes. The 3 weeks can be extended out in case it's an abnormally high activity period. It shouldn't take too long to activate (1) so I wouldn't expect conditions to change drastically. If they do then we can always make a new PR to this repo and update the numbers, right? These conditions aren't enforced by the code so no new releases would be required.

brooksprumo: > It shouldn't take too long to activate (1)

Probably six months from today is my estimate.

Edit: Maybe a little pessimistic, but absolute earliest would be agave v4.0.

brooksprumo (approved):

proposals/0437-incremental-rent-reduction.md
```diff
@@ -40,23 +40,75 @@ N/A
...
+
+If after the first 3 weeks these conditions aren't met then the timeline
+is extended until the averages since the activation of (1) satisfy the
+requirements.
+
+#### Step 3 Criteria
+
+Activation of SIMD-0389 (supervisory controller).
```
bw-solana: I was surprised to see the supervisory controller is a requirement for enabling step 3 (~63% reduction from current).

I lean towards removing this as a requirement in the SIMD. We can always decide to add some guardrails like this in based on what we see in step 1/2

brooksprumo: I agree.

Let's keep this SIMD only about static/fixed reductions.

brooksprumo:
~~@jacobcreech Can you dismiss my approval, please?~~

edit: nevermind!

proposals/0437-incremental-rent-reduction.md
igor56D: I included that to force some discussion on the reduction criteria, knowing it would be controversial 😆

My position is that the risk scales super-linearly to the percent reduction (e.g. risk of 6960 -> 3480 is lower than 3480 -> 1740). 6960 has been the constant for a very long time so we have decent visibility into the proximal [sample space](https://en.wikipedia.org/wiki/Sample_space). This visibility declines the farther we get from that value so criteria should reflect that on principle.

I (somewhat arbitrarily) decided on 50% as the threshold where increased risk justifies a more effective safeguard (controller), but that can be adjusted.

brooksprumo: We'd only activate the next step once proven safe. If at some point the next step is not safe, we wouldn't activate.

Controller can happen concurrently and doesn't need to be a prereq for any step here. Safety *is* a prereq here, of course.

If we want further rent reductions and no further static reductions are safe, then maybe the controller provides that safety. IMO it is orthogonal.

igor56D: The problem is we can't prove the next step is safe with 100% confidence, so we need to decide on criteria for each step that would reduce risk sufficiently. Risk reduction includes both observation/forecasting (longer periods of observation before riskier steps) and safeguards (latent rent increase feature gate, controller, etc). I disagree that the controller is completely orthogonal, it's in the same class as the rent increase safeguard.

That said, I don't think the criteria need to be perfectly explicit and precise in this SIMD. Maybe we can settle on "proceed to the next step if a risk analysis based on existing safeguards and empirical data indicates it is safe"?

brooksprumo: > That said, I don't think the criteria need to be perfectly explicit and precise in this SIMD. Maybe we can settle on "proceed to the next step if a risk analysis based on existing safeguards and empirical data indicates it is safe"?

Yes and yes 👍

proposals/0437-incremental-rent-reduction.md
igor56D: replaced hardcoded values with general principles for advancement through the activation schedule: 6616745e724232c02611efb82a8f66542be7ad08

proposals/0437-incremental-rent-reduction.md
igor56D: 6616745e724232c02611efb82a8f66542be7ad08

proposals/0437-incremental-rent-reduction.md
igor56D: per the discussion in the other [thread](https://github.com/solana-foundation/solana-improvement-documents/pull/437#discussion_r2696265151), I replaced the explicit reduction criteria with general principles: 6616745e724232c02611efb82a8f66542be7ad08

brooksprumo (approved):
:shipit:

bw-solana (approved):
lgtm

proposals/0437-incremental-rent-reduction.md
```diff
@@ -0,0 +1,116 @@
...
+## New Terminology
+
+N/A
+
+## Detailed Design
+
+This proposal defines five incremental reductions to the
+`lamports_per_byte` constant. Each reduction is controlled by its own
+feature gate. The values are absolute, starting from the current value
+of 6960.
```
topointon-jump, line 46: Could we also enumerate the feature gate names here? Makes it easier to track 🙌

topointon-jump (approved):

2501babe:
i just became aware of this simd and the corresponding feature gates... as we discussed in https://github.com/solana-foundation/solana-improvement-documents/pull/392 lowering rent *requires* us to change how `Rent` is used in the stake program (and possibly in the token programs) and cannot be done without this work

PR #443, opened 2026-01-06
feat: lock old PRs

jacobcreech:
PRs older than 60 days will be locked to avoid spam abuse

### DIFF
--- a/.github/workflows/lock-merged-prs.yml
+++ b/.github/workflows/lock-merged-prs.yml
@@ -1,0 +1,23 @@
+name: Lock Merged PRs
+
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


Benhawkins18:
Will this lock the PRs with long discussions that are still ongoing?

jacobcreech:
> Will this lock the PRs with long discussions that are still ongoing?

[According to the documentation](https://github.com/dessant/lock-threads?tab=readme-ov-file#lock-threads), it'll only lock closed/merged PRs that haven't had any activity in the past 60 days.

Benhawkins18:
I think this makes sense I prefer 60 days over 90. @t-nelson you ok with 60?

PR #449, opened 2026-01-24
SIMD-0449: Direct Account Pointers in Program Input

febo:
Serialize pointers for instruction accounts to allow direct account access
without requiring parsing of the account section.

### DIFF
--- a/proposals/0449-direct-account-pointers-in-program-input.md
+++ b/proposals/0449-direct-account-pointers-in-program-input.md
@@ -40,7 +40,8 @@ independent of the number of accounts.
 
 When the feature is activated, the VM serializes the slice of account
 pointers after the program ID in the program input parameters for any
-ABIv1 program. All other parameters remain unchanged:
+ABIv1 program (i.e., those that do not use Loader v1). All other parameters
+remain unchanged:
 
 ```
 - 8 bytes: number of accounts (little-endian)
--- a/proposals/0449-direct-account-pointers-in-program-input.md
+++ b/proposals/0449-direct-account-pointers-in-program-input.md
@@ -1,0 +1,107 @@
+---
+simd: "0449"
+title: Direct Account Pointers in Program Input
+authors:
+  - febo (Anza)
+category: Standard
+type: Core
+status: Review
+created: 2026-01-24
+feature: (fill in with feature tracking issues once accepted)
+---
+
+## Summary
+
+Serialize pointers for instruction accounts to allow direct account access
+without requiring parsing of the account section.
+
+## Motivation
+
+Currently, sBPF programs must parse the accounts section of the serialized
+input region to extract account boundary information in order to provide the
+program with a slice of accounts. This parsing accounts for most of the
+entrypoint's cost.
+
+Since account boundaries are already known to the VM when preparing the
+program input, they can instead be serialized directly into the input. This
+allows the entrypoint to derive the account slice without iterating over
+or parsing the accounts section.
+
+As a result, the entrypoint consumes a constant number of compute units,
+independent of the number of accounts.
+
+## New Terminology
+
+- **accounts slice pointer**: A slice of 64-bit (8-byte) pointers, with
+  one pointer per instruction account, stored at the end of the program
+  input parameters.
+
+## Detailed Design
+
+When the feature is activated, the VM serializes the slice of account
+pointers after the program ID in the program input parameters. All other
+parameters remain unchanged:
+
+```
+- 8 bytes: number of accounts (little-endian)
+- <variable>: accounts section
+- 8 bytes: length of instruction data (little-endian)
+- <variable>: instruction data bytes
+- 32 bytes: program ID
+- [u64; <number of accounts>]: slice of account pointers
+```
+
+Each account pointer is the address of the first byte of the account
+information. This is the same address that entrypoints previously had
+to compute when parsing the accounts section, but it is now provided
+directly as an input parameter. Duplicated accounts will have the
+same pointer value, i.e., they will point to the same account.
+
+The slice of account pointers is made available to all programs, under
+all loaders, regardless of whether it is read. Since this data is
+appended after the standard input parameters, existing programs are
+unaffected.
+
+Given that, after [SIMD-0321](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0321-vm-r2-instruction-data-pointer.md)
+is activated, programs will receive a
+direct pointer to the instruction data, programs can derive the location
+of the slice of account pointers without parsing the accounts section:
+
+- read the number of accounts from `r1`.
+- read the length of the instruction data from `r2 - 8`.
+- pass the following three references to the program entrypoint:
+  1. `program_id`: a `&[u8]` slice created from
+    `r2 + <length of the instruction data>` with length equal
+    to `32 bytes`.
+  2. `accounts`: a `&[AccountView]` slice created from
+    `r2 + <length of the instruction data> + 32` with length equal
+    to `<number of accounts>`.
+  3. `instruction_data`: a `&[u8]` slice created from `r2` with
+    length equal to `<length of instruction data>`.
+
+## Alternatives Considered
+
+While ABIv2 proposes a new serialization layout that provides a similar
+slice of accounts and likewise avoids parsing the accounts section,
+the approach proposed in this SIMD is comparatively simpler and fully
+backward compatible.
+
+## Impact
+
+On-chain programs are positively impacted by this change. The program
+entrypoint complexity is significantly reduced, along with the number of
+compute units consumed.
+
+The implementation is relatively simple, as it relies on information that
+is already available during program input serialization.
+
+## Security Considerations
+
+Since programs currently do not read data from the input parameters
+beyond the program ID, this change does not introduce any security
+concerns.
+
+## Backwards Compatibility
+
+This feature is fully backwards compatible since no program entrypoint
+reads data from the input parameters after the program ID.
--- a/proposals/0449-direct-account-pointers-in-program-input.md
+++ b/proposals/0449-direct-account-pointers-in-program-input.md
@@ -48,6 +48,7 @@ parameters remain unchanged:
 - 8 bytes: length of instruction data (little-endian)
 - <variable>: instruction data bytes
 - 32 bytes: program ID
+- 0-7 bytes: padding bytes to align offset to 8-bytes
 - [u64; <number of accounts>]: slice of account pointers
 ```
 
@@ -74,8 +75,8 @@ of the slice of account pointers without parsing the accounts section:
     `r2 + <length of the instruction data>` with length equal
     to `32 bytes`.
   2. `accounts`: a `&[AccountView]` slice created from
-    `r2 + <length of the instruction data> + 32` with length equal
-    to `<number of accounts>`.
+    `r2 + <length of the instruction data> + 32 + padding` with length
+    equal to `<number of accounts>`.
   3. `instruction_data`: a `&[u8]` slice created from `r2` with
     length equal to `<length of instruction data>`.
 
--- a/proposals/0449-direct-account-pointers-in-program-input.md
+++ b/proposals/0449-direct-account-pointers-in-program-input.md
@@ -92,6 +92,32 @@ On-chain programs are positively impacted by this change. The program
 entrypoint complexity is significantly reduced, along with the number of
 compute units consumed.
 
+- Current entrypoint (pinocchio):
+
+| Name         | CUs | Delta |
+|--------------|-----|-------|
+| Account (1)  | 17  |  --   |
+| Account (2)  | 17  |  --   |
+| Account (3)  | 37  |  --   |
+| Account (4)  | 45  |  --   |
+| Account (8)  | 78  |  --   |
+| Account (16) | 143 |  --   |
+| Account (32) | 261 |  --   |
+| Account (64) | 504 |  --   |
+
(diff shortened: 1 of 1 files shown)


proposals/0449-direct-account-pointers-in-program-input.md
```diff
@@ -0,0 +1,134 @@
...
+As a result, the entrypoint consumes a constant number of compute units,
+independent of the number of accounts.
+
+## New Terminology
+
+- **accounts slice pointer**: A slice of 64-bit (8-byte) pointers, with
+  one pointer per instruction account, stored at the end of the program
+  input parameters.
+
+## Detailed Design
```
buffalojoec, line 39: Can you explicitly state somewhere in here this is only for ABI v1?

proposals/0449-direct-account-pointers-in-program-input.md
```diff
@@ -0,0 +1,134 @@
...
+parameters remain unchanged:
+
+```
+- 8 bytes: number of accounts (little-endian)
+- <variable>: accounts section
+- 8 bytes: length of instruction data (little-endian)
+- <variable>: instruction data bytes
+- 32 bytes: program ID
+- 0-7 bytes: padding bytes to align offset to 8-bytes
+- [u64; <number of accounts>]: slice of account pointers
```
buffalojoec, line 53: Will there be a length prefix for `number of accounts`?

proposals/0449-direct-account-pointers-in-program-input.md
```diff
@@ -0,0 +1,134 @@
...
+```
+
```
buffalojoec, line 55: It's implied, but it would be good to include a section here that briefly states all of these account pointers must be valid pointers within the input section of the VM.

buffalojoec: Maybe also mention the endianness of the pointers, too.

proposals/0449-direct-account-pointers-in-program-input.md
```diff
@@ -0,0 +1,134 @@
...
+| Account (1)  | 17  |  --   |
+| Account (2)  | 17  |  --   |
+| Account (3)  | 37  |  --   |
+| Account (4)  | 45  |  --   |
+| Account (8)  | 78  |  --   |
+| Account (16) | 143 |  --   |
+| Account (32) | 261 |  --   |
+| Account (64) | 504 |  --   |
+
+- Estimated entrypoint (after the changes proposed):
```
buffalojoec, line 112: Can you link/describe where these benchmarks came from?

buffalojoec (commented):
Nice work @febo! This looks like a slick way to improve ABI v1 entrypoint efficiency without breaking the ABI's backwards compatibility.

I left a few comments about beefing up the spec, but this seems reasonable to me.

alnoki:
@febo thanks for putting this together, direct pointers will be super helpful for ASM account parsing implementations as well

Since the serializer also has access to instruction data and program ID boundaries can we get pointers to those as well? Otherwise runtime logic still has to parse final account data length and instruction data length to get program ID when doing operations like PDA account creation, for example

Ideal would be the following configuration at runtime start:

| Register | Pointer to |
| - | - |
| r1, r9 | Virtual memory map input buffer |
| r8 | Array of direct account pointers |
| r7 | Instruction data |
| r6 | Program ID |

r9-r6 would be preferable for the pointers since r1-r5 are for syscalls and often you need to do something like this to keep input buffer on hand after a syscall: you could retain in r1 for legacy reasons but just also duplicate it in r9 to avoid the following:

```asm
    # Compute PDA.
    # ------------
    mov64 r9, r1 # Store input buffer pointer for later. <--------- This is annoying
    mov64 r1, r10 # Get stack frame pointer.
    # Update to point at user pubkey signer seed.
    sub64 r1, STK_INIT_SEED_0_ADDR_OFF 
    mov64 r2, 1 # Indicate single signer seed (user pubkey).
    mov64 r3, r9 # Get input buffer pointer.
    add64 r3, PROGRAM_ID_INIT_OFF # Update to point at program ID.
    mov64 r4, r10 # Get stack frame pointer.
    sub64 r4, STK_INIT_PDA_OFF # Update to point to PDA region on stack.
    mov64 r5, r10 # Get stack frame pointer.
    sub64 r5, STK_INIT_BUMP_SEED_OFF # Update to point to bump seed region.
    call sol_try_find_program_address # Find PDA. <------------ This forced me to mutate r1
    # Skip check to error out if unable to derive a PDA (failure to derive
    # is practically impossible to test since odds of not finding bump seed
    # are astronomically low):
    # ```
    # jne r0, SUCCESS, e_unable_to_derive_pda
    # ```
    mov64 r1, r9 # Restore input buffer pointer. <--------- This is annoying
```

cc @xbtmatt  @bidhan-a @clairechingching @deanmlittle

bidhan-a:
> @febo thanks for putting this together, direct pointers will be super helpful for ASM account parsing implementations as well
> 
> Since the serializer also has access to instruction data and program ID boundaries can we get pointers to those as well? Otherwise runtime logic still has to parse final account data length and instruction data length to get program ID when doing operations like PDA account creation, for example
> 
> Ideal would be the following configuration at runtime start:
> 
> Register	Pointer to
> r1, r9	Virtual memory map input buffer
> r8	Array of direct account pointers
> r7	Instruction data
> r6	Program ID
> r9-r6 would be preferable for the pointers since r1-r5 are for syscalls and often you need to do something like this to keep input buffer on hand after a syscall: you could retain in r1 for legacy reasons but just also duplicate it in r9 to avoid the following:
> 
> ```assembly
>     # Compute PDA.
>     # ------------
>     mov64 r9, r1 # Store input buffer pointer for later. <--------- This is annoying
>     mov64 r1, r10 # Get stack frame pointer.
>     # Update to point at user pubkey signer seed.
>     sub64 r1, STK_INIT_SEED_0_ADDR_OFF 
>     mov64 r2, 1 # Indicate single signer seed (user pubkey).
>     mov64 r3, r9 # Get input buffer pointer.
>     add64 r3, PROGRAM_ID_INIT_OFF # Update to point at program ID.
>     mov64 r4, r10 # Get stack frame pointer.
>     sub64 r4, STK_INIT_PDA_OFF # Update to point to PDA region on stack.
>     mov64 r5, r10 # Get stack frame pointer.
>     sub64 r5, STK_INIT_BUMP_SEED_OFF # Update to point to bump seed region.
>     call sol_try_find_program_address # Find PDA. <------------ This forced me to mutate r1
>     # Skip check to error out if unable to derive a PDA (failure to derive
>     # is practically impossible to test since odds of not finding bump seed
>     # are astronomically low):
>     # ```
>     # jne r0, SUCCESS, e_unable_to_derive_pda
>     # ```
>     mov64 r1, r9 # Restore input buffer pointer. <--------- This is annoying
> ```
> 
> cc @xbtmatt @bidhan-a @clairechingching @deanmlittle

As per SIMD-0321, a pointer to instruction data will be available in r2.

alnoki:
> As per SIMD-0321, a pointer to instruction data will be available in r2.

@bidhan-a thanks for flagging, hadn't seen that one yet

@febo any hope for a program ID pointer? Glad to submit a new SIMD if possible if should go there instead

proposals/0449-direct-account-pointers-in-program-input.md
febo: No, the number of accounts is already present in the input as the first value.

proposals/0449-direct-account-pointers-in-program-input.md
febo: Added a few explicit references to ABI v1.

proposals/0449-direct-account-pointers-in-program-input.md
febo: Done in https://github.com/solana-foundation/solana-improvement-documents/pull/449/commits/d14b4b4374f290d83f38d866905fc3430c5c4121

proposals/0449-direct-account-pointers-in-program-input.md
febo: Done in https://github.com/solana-foundation/solana-improvement-documents/pull/449/commits/d14b4b4374f290d83f38d866905fc3430c5c4121

febo:
> > As per SIMD-0321, a pointer to instruction data will be available in r2.
> 
> @bidhan-a thanks for flagging, hadn't seen that one yet
> 
> @febo any hope for a program ID pointer? Glad to submit a new SIMD if possible if should go there instead

This SIMD is intentionally simple, so it is easier to test/audit. ABIv2 ([SIMD-0177](https://github.com/solana-foundation/solana-improvement-documents/pull/177)) is a more extensive rework of the program input and it will include individual pointers to all parts of the input.

Agree that would be nice to have a direct pointer to the program ID, but the gains in CU will be very small (`~2 CUs`) since the majority of programs will be reading the instruction data length first. So perhaps it is a good compromise until we get ABIv2.

proposals/0449-direct-account-pointers-in-program-input.md
alnoki: > No, the number of accounts is already present in the input as the first value.

@buffalojoec for reference [here's the input buffer serialization implementation](https://github.com/anza-xyz/agave/blob/67963d68232074824fe0deb2476e9fd209d88f1d/program-runtime/src/serialization.rs#L531), which doesn't have a corresponding struct since it's dynamically sized a la n accounts, account data, instruction data length, etc.

More info about how it works: https://opcodes.dasmac.com/examples/transfer.html#account-layout-background

buffalojoec (dismissed):
Lgtm. As a next step we should make sure this change won't push Direct Mapping efforts further down the road. We should do this before merging & fully implementing, but we can test/fuzz with a draft implementation.

topointon-jump (dismissed):

topointon-jump (approved):

deanmlittle:
While it's not a *super* big difference at the end of the day, I think it probably makes sense to add `program_id` to r3 instead of using the ix data length from r2. This allows you to use purely static offsets. Since @buffalojoec already did the hard work of adding data flow graph analysis to [program-sync](https://github.com/blueshift-gg/program-sync) to check for uninitialized r2 usage, the work required to ensure that r3 is also safe to use is quite trivial now.

buffalojoec (approved):

febo:
> While it's not a _super_ big difference at the end of the day, I think it probably makes sense to add `program_id` to r3 instead of using the ix data length from r2. This allows you to use purely static offsets. Since @buffalojoec already did the hard work of adding data flow graph analysis to [program-sync](https://github.com/blueshift-gg/program-sync) to check for uninitialized r2 usage, the work required to ensure that r3 is also safe to use is quite trivial now.

Unfortunately it won't be purely static offsets since there are padding bytes between `program_id` and `accounts_pointer` to make it 8-bytes aligned. But it does simplify the logic to compute the offset, so I am happy either way. Another option is to set r3 to point directly to `accounts_pointer`.

deanmlittle:
> > While it's not a _super_ big difference at the end of the day, I think it probably makes sense to add `program_id` to r3 instead of using the ix data length from r2. This allows you to use purely static offsets. Since @buffalojoec already did the hard work of adding data flow graph analysis to [program-sync](https://github.com/blueshift-gg/program-sync) to check for uninitialized r2 usage, the work required to ensure that r3 is also safe to use is quite trivial now.
> 
> Unfortunately it won't be purely static offsets since there are padding bytes between `program_id` and `accounts_pointer` to make it 8-bytes aligned. But it does simplify the logic to compute the offset, so I am happy either way. Another option is to set r3 to point directly to `accounts_pointer`.

Ah, yes, if we are ensuring the pointers are aligned then I think it would actually be the same number of CUs. In that case, perhaps it does make sense for r3 to point to the first account in this section so you can construct the pointer array statically from the r1+0 and r3+0. Program ID isn't really *that* useful anyway as most people use a const ID and allow the runtime itself to fail things like PDA derivation. Well, I just wanted to raise this point. If you like the idea, I think it's probably not too controversial to include it. We can always do this later too if you think it's too much.

proposals/0449-direct-account-pointers-in-program-input.md
```diff
@@ -0,0 +1,138 @@
...
+    to `32 bytes`.
+  2. `accounts`: a `&[AccountView]` slice created from
+    `r2 + <length of the instruction data> + 32 + padding` with length
+    equal to `<number of accounts>`.
+  3. `instruction_data`: a `&[u8]` slice created from `r2` with
+    length equal to `<length of instruction data>`.
+
+## Alternatives Considered
+
+While ABIv2 proposes a new serialization layout that provides a similar
```
bw-solana, line 86: is the plan to still proceed with #177 and have two ways of getting account metadata? Or should we slim down #177 in response to this SIMD?

Lichtso: This SIMD does kinda backport this specific ABIv2 feature to ABIv1.

> have two ways of getting account metadata

This won't be the case because a program must choose between ABI v0, v1, or v2. It can't have multiple.

bw-solana (approved):

febo:
> Ah, yes, if we are ensuring the pointers are aligned then I think it would actually be the same number of CUs. In that case, perhaps it does make sense for r3 to point to the first account in this section so you can construct the pointer array statically from the r1+0 and r3+0. Program ID isn't really _that_ useful anyway as most people use a const ID and allow the runtime itself to fail things like PDA derivation. Well, I just wanted to raise this point. If you like the idea, I think it's probably not too controversial to include it. We can always do this later too if you think it's too much.

We can definitely consider to do it later.

PR #450, opened 2026-01-25
Add details on target group encoding and subgroup check to SIMD-0388

samkim-crypto:
I added additional details on:
- The target group encoding
- The fact that subgroup check should be skipped for addition and subtraction

cc: @0x0ece, @zz-sol, @Rexicon226


zz-sol (approved):

Rexicon226 (approved):
Just wondering if we should include a small example of an encoding of a GT, in both BE and LE.

0x0ece (approved):

samkim-crypto:
@t-nelson, @bw-solana can I get an approval on this? This is just making the EC spec a bit more precise. I already got approvals from all the cryptographers in anza and fd.

bw-solana (approved):

PR #451, opened 2026-01-25
Specify that subgroup check is skipped for add in SIMD-0302

samkim-crypto:
Specify the fact that the subgroup check should be skipped for the addition syscall for alt_bn128.

cc: @0x0ece, @zz-sol, @Rexicon226


zz-sol (approved):

Rexicon226 (approved):

0x0ece (approved):

samkim-crypto:
@t-nelson, @bw-solana can I get an approval on this? This is just making the EC spec a bit more precise. I already got approvals from all the cryptographers in anza and fd.

bw-solana (approved):

PR #452, opened 2026-01-26
SIMD-0452: Charge CUs for ABIv0/v1 Instructions

Lichtso:


proposals/0452-charge-cus-for-abi-v0-and-v1-instructions.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Security Considerations
+
+None
+
+## Drawbacks
+
+It will break existing TX building as the CUs required will be significantly
+increased. To counter this we either have to offer an alternative (ABIv2) or
+adjust the CU price / fees and block packing limits.
```
topointon-jump, line 100: We definitely need to do something like this. However, we should combine this with @igor56D's work on the cost model into a single change, so that we don't _keep_ breaking devs. We should do a big change, make certain usage patterns significantly more expensive, have proper comms around this, and do this once.

deanmlittle:
Conceptually, fixing usage-based pricing makes total sense and is a good idea. Practically speaking, if the goal is to fix CPI via direct mapping, it makes no sense to try and solve for distorted computational economics of behavior that we explicitly plan to remove. If we’re going to ship DM, we should do that first, eliminate the entire nested CPI cost section and reevaluate our cost model.

Lichtso:
> If we’re going to ship DM, we should do that first, eliminate the entire nested CPI cost section and reevaluate our cost model.

Absolutely. This SIMD has been written as a defensive fallback in case we can't ship direct mapping soon enough.

Yet, even after direct mapping there will still be incorrect CU metering around the account and instruction de/serialization. Thus, now that we are more confident in direct mapping I will rewrite this SIMD to match it.

PR #458, opened 2026-01-30
SIMD-0458: Stop using static SimpleVote transaction cost

tao-stones:
This proposal removes the use of statically defined compute unit (CU) costs for
Simple Vote transactions and instead accounts for them in the same way as normal
transactions when enforcing block-wide and vote-specific CU limits.

### DIFF
--- a/proposals/0458-stop-using-static-simplevote-transaction-cost.md
+++ b/proposals/0458-stop-using-static-simplevote-transaction-cost.md
@@ -48,6 +48,11 @@ N/A
   The impact is expected to be minimal, as the actual executed CU consumption of
 Simple Vote transactions is close to the previously statically defined value.
 
+  Detail: statically define CUs for simple vote includes 1 signature (720 CU), 2
+write locks (600 CU), 1 vote instruction which has 2,100 CU, and 8 CU to load
+small accounts, total 3428 CU. All components stay same except the CU for loaded
+accounts data size may change, which is a small part overall CUs.  
+
 - Block production:
   Under the current cost model, Simple Vote transactions may have higher
 estimated CUs due to inclusion of default account data loading costs. For
--- a/proposals/0458-stop-using-static-simplevote-transaction-cost.md
+++ b/proposals/0458-stop-using-static-simplevote-transaction-cost.md
@@ -54,7 +54,7 @@ summarized in table:
 
 - Cost tracking:
   The impact is expected to be minimal, as the actual executed CU consumption of
-Simple Vote transactions is close to the previously statically defined value.
+Simple Vote transactions is identical to the previously statically defined value for the vast majority of votes.  CU consumption will only change for unusual vote transactions that still classify as simple votes (e.g. two signatures, or additional writable accounts), or unusually large vote accounts.
 
   [ * ]statically define CUs for simple vote includes 1 signature (720 CU), 2
 write locks (600 CU), 1 vote instruction which has 2,100 CU, and 8 CU to load
--- a/proposals/0458-stop-using-static-simplevote-transaction-cost.md
+++ b/proposals/0458-stop-using-static-simplevote-transaction-cost.md
@@ -1,5 +1,5 @@
 ---
-simd: 'XXXX'
+simd: '0458'
 title: Stop using static SimpleVote transaction cost
 authors:
   - Tao Zhu
--- a/proposals/0458-stop-using-static-simplevote-transaction-cost.md
+++ b/proposals/0458-stop-using-static-simplevote-transaction-cost.md
@@ -1,6 +1,6 @@
 ---
 simd: '0458'
-title: Stop using static SimpleVote transaction cost
+title: Stop special-casing of Simple Vote in Cost Model
 authors:
   - Tao Zhu
 category: Standard
--- a/proposals/0458-stop-using-static-simplevote-transaction-cost.md
+++ b/proposals/0458-stop-using-static-simplevote-transaction-cost.md
@@ -1,6 +1,6 @@
 ---
 simd: '0458'
-title: Stop special-casing of Simple Vote in Cost Model
+title: Stop special-casing of Vote CU cost
 authors:
   - Tao Zhu
 category: Standard
--- a/proposals/0458-stop-using-static-simplevote-transaction-cost.md
+++ b/proposals/0458-stop-using-static-simplevote-transaction-cost.md
@@ -44,22 +44,30 @@ N/A
 
 ## Impact
 
+The impact on cost consuming/tracking, and CU reserving for simple vote can be
+summarized in table:
+
+| |Total consumed CUs	| Total reserved CUs|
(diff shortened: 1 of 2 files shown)


proposals/0458-stop-using-static-simplevote-transaction-cost.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Alternatives Considered
+
+N/A
+
+## Impact
+
+- Cost tracking:
+  The impact is expected to be minimal, as the actual executed CU consumption of
+Simple Vote transactions is close to the previously statically defined value.
```
bw-solana: Can we provide actual numbers somewhere?

proposals/0458-stop-using-static-simplevote-transaction-cost.md
```diff
@@ -0,0 +1,62 @@
...
+
+## New Terminology
+
+N/A
+
+## Detailed Design
+
+- Stop using statically defined CU values for Simple Vote transactions.
+- Calculate the cost of Simple Vote transactions using the same cost model and
+  accounting path as normal transactions.
```
bw-solana, line 39: will we change any associated client default behaviors (such as providing data budget ix)?

proposals/0458-stop-using-static-simplevote-transaction-cost.md
tao-stones: added @ 127b92fbbcfe0393c0f8efbbf1343b6dad8b6ce9

proposals/0458-stop-using-static-simplevote-transaction-cost.md
tao-stones: No changes on client side. "simple vote" transaction can only have 1 instruction, which must be vote program.

buffalojoec (commented):
Just out of curiosity: wouldn't it be better to just refine the process of filtering for "simple vote transactions"?

AFAICT, applying a static CU value is really for performance reasons, right? If so, you can just tighten up the vague criteria (number of signatures, message version, instruction count, etc.) to just also check the instruction discriminator (first four bytes) and only count instructions like `Vote` and `VoteSwitch`, depending on which ones we want in the fast-path.

Considering the current checks, glancing the first four bytes of the instruction doesn't seem too bad, and we don't have to lose any perf while we ship Alpenglow.

tao-stones:
>  wouldn't it be better to just refine the process of filtering for "simple vote transactions"?

Yea, it might be good to refine criteria used to tag "simple vote transaction", adding instruction discriminator check, maybe additional checks. tho it might belong to separate discussion. This simd proposes to sunset static CUs for transactions that's tagged as "simple vote".

> applying a static CU value is really for performance reasons, right?

Yes. iirc cost model used to have to find and iterate compute-budget ixs in order to calculate CUs for any transaction, that adds up to impact performance. Giving an estimated static CUs for "simple vote" helps. But it is implementation specific - i don't think firedancer does this, and agave no longer does that either. The idea of having static CU value for transaction tagged as "simple vote" isn't valid anymore, propose to stop using it in cost modeling and cost tracking.

proposals/0458-stop-using-static-simplevote-transaction-cost.md
bw-solana: Outside the scope of this SIMD, but if explicitly setting data budget is cheaper, we'll probably need to make that change to Agave (or it will become popular mod)

buffalojoec:
> tho it might belong to separate discussion. This simd proposes to sunset static CUs for transactions that's tagged as "simple vote".

Right, but I'm wondering - if we get perf gains from the simple-vote highway - if refining the criteria is a viable alternative to this proposal. I have only a basic understanding of what we even flag "simple vote transactions" for in the first place.

> Giving an estimated static CUs for "simple vote" helps. But it is implementation specific - i don't think firedancer does this, and agave no longer does that either.

Sorry I'm a little confused by this bit. If it's an implementation detail, why do we need a SIMD?

tao-stones:
> Sorry I'm a little confused by this bit. If it's an implementation detail, why do we need a SIMD?

If simple vote txs no longer have exact same static CUs, it'd change block CU tracking which is part of consensus, so it needs SIMD.

To clarify, the proposal doesn't change how simple-vote being fast tracked, it only says not to use static value when calculate CUs for simple-vote txs.

ptaffet-jump:
Seems reasonable. Could we get a simple table included like (filling in the ???):

|       |  Total reserved CUs | Total consumed CUs |
|----|-----------------------|------------------------|
| Current |    3428             |      3428 |
| Proposed |    19812 ??          |        ???  |

topointon-jump:
> Sorry I'm a little confused by this bit. If it's an implementation detail, why do we need a SIMD?

it's necessary to have a SIMD because the block CU limit is enforced by replay, which unfortunately means CUs have to be accounted exactly the same in each client instead of treated like a heuristic. this is another reason to remove the block CU limit and take CUs out of consensus

tao-stones:
> Seems reasonable. Could we get a simple table included like (filling in the ???):

| |Total reserved CUs	| Total consumed CUs|
|:--|:--|:--|
|Current	|3428	|3428|
|Proposed	|19812*	|3428**|

\* it would reserve additional ~16K CUs for default loaded_accounts_data_size, eg `Impact: BlockProduction` section
\*\* the actual consumed CUs would most likely be 3428, but could be slightly larger due to actual vote account size.  Accounts in simple-vote are small, usually less than 32K. eg `Impact: Cost tracking` section.

bw-solana (dismissed):
lgtm

ptaffet-jump:
I think the loaded account data size is rounded up, so the "proposed consumed CUs" in the table is not right? No?

tao-stones:
> I think the loaded account data size is rounded up, so the "proposed consumed CUs" in the table is not right? No?

it's 8CU per 32K (round up), most vote accounts are below 32K, but if it's 35K, then it'd round up to 16 CU.

proposals/0458-stop-using-static-simplevote-transaction-cost.md
```diff
@@ -0,0 +1,75 @@
...
+summarized in table:
+
+| |Total consumed CUs	| Total reserved CUs|
+|:--|:--|:--|
+|Current	|3428	|3428|
+|Proposed	|3428* |19812**	|
+
+- Cost tracking:
+  The impact is expected to be minimal, as the actual executed CU consumption of
+Simple Vote transactions is close to the previously statically defined value.
```
ptaffet-jump:
```suggestion
Simple Vote transactions is identical to the previously statically defined value for the vast majority of votes.  CU consumption will only change for unusual vote transactions that still classify as simple votes (e.g. two signatures, or additional writable accounts), or unusually large vote accounts.
```

ptaffet-jump (approved):

bw-solana (approved):

PR #459, opened 2026-01-30
SIMD-0459: Syscall Parameter Address Restrictions

Lichtso:


proposals/0459-syscall-parameter-address-restrictions.md
```diff
@@ -0,0 +1,87 @@
...
+---
+
+## Summary
+
+Syscall parameter related changes split off from SIMD-0219.
+
+## Motivation
+
+In a recent meeting between the Agave and Firedancer core developers it was
+decided that SIMD-0219 should be split into two feature gates. This SIMD covers
```
bw-solana, line 22: would be good to add the reasoning (items in this SIMD are not strictly needed to help de-risk direct mapping optimizations and therefore don't need to gate DM)

proposals/0459-syscall-parameter-address-restrictions.md
```diff
@@ -0,0 +1,87 @@
...
+the first half which is focused on reducing the attack surface and fix long
+standing issues with CPI in general.
```
bw-solana, line 24: would be good to copy+paste the relevant text from SIMD-0219 re: the motivations for fixing the CPI issues

proposals/0459-syscall-parameter-address-restrictions.md
```diff
@@ -0,0 +1,87 @@
...
+
+## New Terminology
+
+None.
+
+## Detailed Design
+
+### Additional changes
```
bw-solana, line 61: might make more sense if we flip the order of these sections ("Changes inherited from SIMD-0219" goes first and then "Additional changes")

proposals/0459-syscall-parameter-address-restrictions.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Alternatives Considered
+
+Leaving SIMD-0219 as is.
+
+## Impact
+
+Splitting SIMD-0219 should have no impact on dApp developers or validators.
+The additional changes have been tested to not cause any existing dApps to
+behave differently.
```
bw-solana: would be nice to mention that any dapps violating the new constraints above could have behavioral changes that are consensus breaking.

I'm assuming we profiled mainnet and determined no apps are currently violating the new constraints. It would be good to include any details about the window of activity that was profiled

bw-solana (commented):
@LucasSte @nagisa @buffalojoec - It would be good to get SME approval from at least 1 of you

proposals/0459-syscall-parameter-address-restrictions.md
```diff
@@ -0,0 +1,87 @@
...
+type: Core
+status: Review
+created: 2026-01-30
+feature: TBD
+supersedes: 0219
+---
+
+## Summary
+
+Syscall parameter related changes split off from SIMD-0219.
```
bw-solana: let's copy+paste the relevant summary portions from SIMD-0219 here

proposals/0459-syscall-parameter-address-restrictions.md
nagisa: Additionally, while this says that there was a decision taken in some behind-the-doors meeting, there is no description of what went into making that decision (i.e. what would be the actual underlying motivation to supersede 0219.)

Why is it necessary to split SIMD-0219 into two feature gates?

What is the reason to make the additional changes?

proposals/0459-syscall-parameter-address-restrictions.md
```diff
@@ -0,0 +1,87 @@
...
+of the account in the caller. This check currently allows for +10 KiB even in
+ABIv0 which is incorrect and must be reduced to 0 growth allowed in ABIv0.
+
+All sites in the CPI code which perform guest to host address translation first
+and then perform pointer arithmetic on the host must be swapped such that they
+perform pointer arithmetic in the virtual address space first followed by the
+address translation second. This specifically affects the account length field.
+
+The CU charging for the account length must be moved to occur after the address
+translation of the account length field.
```
nagisa: Is this specifically for CPI or a general change? What is the reason to do this? My intuition would be that we'd always want to charge CU before the relevant operation if possible to do so.

I'm also not entirely clear which charge is being moved around: the charge for address translation? Or the entire translation+access operation?

LucasSte: We were doing pointer arithmetics in the address space of the host and the guest interchangeably. After this change, pointer arithmetics will be done only in the guest address space, and only then we will translate the address.

Since that change affects how CUs are calculated, we are moving it to after the translation.

nagisa: Lets make sure that the wording in the SIMD is non-ambiguous in what is being moved and in what specific scenarios this new behaviour applies.

That said, I'm still not quite sure why would it be necessary to charge CUs *after* the translation. This conceptually diverges from the execution of regular BPF code where the CUs appear to be accounted for before the instruction(s) and any address translation as part of those instructions are processed. Is this some sort play to retain the functional behaviour in certain corner cases (maybe to prefer producing access violation instead of CU exhaustion error if both occur kinda at the same time?)

Lichtso: > Lets make sure that the wording in the SIMD is non-ambiguous in what is being moved and in what specific scenarios this new behaviour applies.

Agreed, will make the description of the change more detailed.

> This conceptually diverges from the execution of regular BPF code where the CUs appear to be accounted for before the instruction(s) and any address translation as part of those instructions are processed.

Not really, because the address translation is rather fast and the bulk of the work which we want to meter is still after the CU charging. We definitely have to do address translation of the length field first to know what to charge. 

>  Is this some sort play to retain the functional behavior in certain corner cases (maybe to prefer producing access violation instead of CU exhaustion error if both occur kinda at the same time?)

No, it is not about error priorities or error code stability in failing transactions. It is about maintaining consensus of which transactions succeed or fail because of the CU charging. See Lucas answer:

> Since that change affects how CUs are calculated, we are moving it to after the translation.

proposals/0459-syscall-parameter-address-restrictions.md
Lichtso: The changes added in this SIMD over SIMD-0219 are covered in https://github.com/anza-xyz/agave/pull/9995 and have been tested since then (a month ago).

proposals/0459-syscall-parameter-address-restrictions.md
Lichtso: > I'm also not entirely clear which charge is being moved around: the charge for address translation? Or the entire translation+access operation?

We never charged for the address translation, we charge for the length of the accounts as their data needs to be copied and compared.

nagisa (approved):

ripatel-fd (approved):

topointon-jump (approved):

bw-solana (approved):

proposals/0459-syscall-parameter-address-restrictions.md
```diff
@@ -0,0 +1,99 @@
...
+type: Core
+status: Review
+created: 2026-01-30
+feature: TBD
+supersedes: 0219
+---
+
+## Summary
+
+Syscall parameter related changes split off from SIMD-0219:
```
t-nelson, line 16: would have been ideal had this been written as a standalone document with only ref to 0219 being the supersedes field

PR #461, opened 2026-02-01
SIMD-0461: enabling falcon signature verification as a syscall

zz-sol:
#### Summary
This SIMD proposes adding a precompile for Falcon-512 (FN-DSA) signature verification, providing post-quantum cryptographic capability for Solana.

#### Scope
- Verification only — key generation and signing are off-chain
- Does NOT replace Ed25519 — provides an additional verification option
- Exploratory — gauging interest before committing to a PQC strategy

#### Related
- [EIP-8052](https://eips.ethereum.org/EIPS/eip-8052): Ethereum's Falcon precompile proposal
- [Draft FIPS 206](https://csrc.nist.gov/pubs/fips/206/ipd): NIST FN-DSA standard


zz-sol:
#### A note on PQC scheme selection:

The elephant in the room is whether we should also introduce Dilithium (ML-DSA) and SPHINCS+ (SLH-DSA), the other two NIST-selected PQC signature schemes. Proposing only Falcon might appear to be picking winners prematurely.

#### Why Falcon first:

Falcon-512 offers the best balance for blockchain use cases:

- Smallest signatures among the three (~666 bytes vs ~2,420 bytes for Dilithium2 vs ~17KB+ for SPHINCS+)
- Reasonable verification performance (~ 5 us, credit @Rexicon226)

That said, Dilithium has merits too—simpler implementation and no floating-point concerns. OTOH SPHINCS+ is probably a no-go as it has impractical signature sizes for on-chain use (if you disagree, please convince us).

_Ethereum's approach:_

For reference, Ethereum has separate EIPs:

- [EIP-8052](https://eips.ethereum.org/EIPS/eip-8052): Falcon precompile
- [EIP-8053](https://eips.ethereum.org/EIPS/eip-8053): Dilithium precompile
#### Options for Solana:

- Single **exploratory** SIMD (current approach) — start with Falcon, add others if demand emerges
- Parallel SIMDs — write separate SIMDs for Dilithium (and SPHINCS+) now (more comprehensive but more work for exploration)
- Umbrella SIMD — one SIMD covering all two/three schemes with separate feature gates

Looking for feedback on the preferred approach. This SIMD is exploratory—we're gauging interest and identifying implementation considerations before committing to a specific PQC strategy.

t-nelson:
please no more precompiles. we need to kill them conceptually. they're a scourge on optimizing the execution pipeline. can this not be done via syscall instead? ideally with primitive building blocks.

zz-sol:
> please no more precompiles. we need to kill them conceptually. they're a scourge on optimizing the execution pipeline. can this not be done via syscall instead? ideally with primitive building blocks.

yes. refactored into a syscall

zz-sol:
seem like I cannot edit the title of this PR. so admin please help to rename this PR:

~SIMD-0461: enabling falcon signature verification as a precompile~

SIMD-0461: enabling falcon signature verification as a **syscall**

proposals/0461-falcon-signature-syscall.md
```diff
@@ -0,0 +1,441 @@
...
+
+We propose a new syscall that verifies a single Falcon-512 signature per call:
+
+```rust
+define_syscall!(fn sol_falcon512_verify(
+    signature_addr: *const u8, // 666 bytes, fixed-length padded signature
+    public_key_addr: *const u8, // 897 bytes
+    message_addr: *const u8,
+    message_len: u64,
+    result_addr: *mut u8 // 1 = valid, 0 = invalid
```
0x0ece: Let's just return `Ok(0)` (success) or `Ok(1)` (failure)?
No need for an extra addr translation.

proposals/0461-falcon-signature-syscall.md
```diff
@@ -0,0 +1,441 @@
...
+## Backwards Compatibility
+
+Transactions using the Falcon syscall cannot be processed
+on Solana versions that do not implement this feature. A feature gate
+MUST be used to enable this feature when the majority of the cluster is
+running the required version.
+
+Transactions that do not use this feature are not impacted.
+
+Existing Ed25519 signatures and other precompiles (secp256k1, secp256r1)
```
0x0ece: Remove?

proposals/0461-falcon-signature-syscall.md
```diff
@@ -0,0 +1,441 @@
...
+
+### Syscall Integration
+
+- The syscall MUST be feature-gated and unavailable prior to activation.
+- The syscall MUST be registered in the runtime syscall table under the
+  name `sol_falcon512_verify`.
+- The ABI MUST follow existing syscall conventions:
+  - `0` return value indicates successful execution (validity is conveyed
+    only via `result_addr`).
+  - Non-zero return values indicate internal errors.
```
0x0ece: Remove?

proposals/0461-falcon-signature-syscall.md
```diff
@@ -0,0 +1,441 @@
...
+3. Write the verification result to `result_addr` as a single byte:
+   `1` for valid, `0` for invalid.
+
+The syscall MUST treat any signature or public key that fails decoding
+or format validation as invalid, and MUST return success with
+`result_addr` set to `0` (i.e., invalid signature is not a fatal error).
+
+The syscall returns `0` on successful execution (regardless of signature
+validity) after writing `result_addr`. Non-zero return values are
+reserved for internal errors.
```
0x0ece: No internal errors please 🙏 They make building conformant implementations unnecessarily hard.

0x0ece (approved):

proposals/0461-falcon-signature-syscall.md
```diff
@@ -0,0 +1,430 @@
...
+in draft FIPS 206 and may change in the final specification; this
+document and implementations MUST be updated to match the final FIPS
+206 before activation.
+
+### Signature Format
+
+Falcon-512 signatures MUST be in the fixed-length (padded) format as
+specified in draft FIPS 206. The signature consists of:
+
+1. A header byte (as specified in draft FIPS 206 for Falcon-512 padded)
```
Rexicon226: Could you specify what byte it is? (I believe it's 0x39?).

proposals/0461-falcon-signature-syscall.md
```diff
@@ -0,0 +1,430 @@
...
+The verification follows draft FIPS 206 Algorithm 18 (Verify):
+
+1. Decode the public key h from the encoded format
+2. Decode the signature (salt, s2) from the padded format
+3. Compute c = HashToPoint(salt || message)
+4. Compute s1 = c - s2 * h (mod q)
+5. Verify that ||(s1, s2)||^2 <= bound^2
+
+The HashToPoint function uses SHAKE-256 as specified in draft FIPS 206
+to hash the message into a polynomial in Z_q[x]/(x^n + 1).
```
Rexicon226: A very small nit, but if you're writing math, might as well wrap it in "$$" to make it render as latex?

proposals/0461-falcon-signature-syscall.md
```diff
@@ -0,0 +1,430 @@
...
+```
+falcon_verify_base
++ ceil(message_len / falcon_hash_bytes_per_cu)
+```
+
+Where `falcon_verify_base` and `falcon_hash_bytes_per_cu` are runtime
+parameters set by benchmarking. `MAX_FALCON_MESSAGE_LEN` is a fixed
+constant defined by this SIMD and enforced by the feature gate.
+**Tentative**: `MAX_FALCON_MESSAGE_LEN = 65_535`.
+Cost calculation MUST use overflow-checked arithmetic.
```
Rexicon226: Another semi-nit, I'm not sure about "overflow-checked" here. Currently, the implementations just use saturating arithmetic, given that you could never actually pay if they saturate. Just want to make sure this isn't interpreted as a possible failure condition.

proposals/0461-falcon-signature-syscall.md
```diff
@@ -0,0 +1,430 @@
...
+
+(**Tentative**)
+
+Falcon-512 verification is computationally more expensive than Ed25519
+due to the polynomial arithmetic involved. Benchmarking MUST be performed
+on representative hardware to determine appropriate compute costs.
+
+Based on preliminary estimates from reference implementations:
+
+- Falcon-512 verification: approximately 0.5-1ms on modern hardware
```
Rexicon226: I get this is tentative, but I just want to note that 0.5-1ms is much larger than the usual ~10-20us I've seen while benchmarking different implementations.

samkim-crypto (approved):

0x0ece (approved):

t-nelson:
i'm not necessarily opposed, but the motivation section doesn't really help me understand why we want to add this to the protocol. afaik we have winternitz vaults already, with no additional simds, so we have some degree of quantum resistant protection. is there external demand for the features that it enables? are we building stuff with it? if so is there demand for that? i have heard nothing about this in any prioritization conversations and we aren't generally keen to change the protocol for the sake of changing the protocol.

xAdamPos:
> i'm not necessarily opposed, but the motivation section doesn't really help me understand why we want to add this to the protocol. afaik we have winternitz vaults already, with no additional simds, so we have some degree of quantum resistant protection. is there external demand for the features that it enables? are we building stuff with it? if so is there demand for that? i have heard nothing about this in any prioritization conversations and we aren't generally keen to change the protocol for the sake of changing the protocol.

One thing that comes to mind is that Winternitz vaults are one-time use only, whereas with FN-DSA / ML-DSA / SLH-DSA you don’t need to rotate the public key after every spend, which can be much more practical for non-vault use cases. I agree that it’s not clear there is meaningful demand for these schemes today.

bw-solana:
can we edit the title to be:
`SIMD-0461: enabling falcon signature verification as a syscall`

0x0ece:
are we good with this simd then? @t-nelson? this way we cut it as part of 4.1

tigarcia:
@t-nelson Any questions on this?

t-nelson:
we're going to do group operation syscalls instead and sigverify as a library, right?

zz-sol:
we are going to pause this effort -- @deanmlittle has a great falcon implementation in SBF with reasonable cu cost (< 170k) and maybe further improved with #563 so we do not want to move forward further until we seed more demands on this

zz-sol:
> we are going to pause this effort -- @deanmlittle has a great falcon implementation in SBF with reasonable cu cost (< 170k) and maybe further improved with #563 so we do not want to move forward further until we seed more demands on this

closed for now. will reopen when there is more demand.

uNetworkingAB:
*Cough* Hawk *cough* (hawk-sign.info) 😉 🧌 

555 byte signatures and no floating point requirement 😄 The code is also extremely minimal and simple.

deanmlittle:
> _Cough_ Hawk _cough_ (hawk-sign.info) 😉 🧌
> 
> 555 byte signatures and no floating point requirement 😄 The code is also extremely minimal and simple.

Both already exist and work without any protocol changes today which is why we don't need to pursue a specific syscall just for this one algorithm:

https://github.com/blueshift-gg/solana-falcon512
https://github.com/blueshift-gg/solana-hawk512

We will instead introduce a keccak p1600 syscall that accelerates many such PQC DSAs without forcing you to choose one specifically:

https://github.com/solana-foundation/solana-improvement-documents/pull/563

uNetworkingAB:
That aged like milk... Hawk is broken 😬

PR #463, opened 2026-02-02
Update SIMD-406

LucasSte:
This PR amends the SIMD with the feature gate key and removes the error type from the detailed design section. During the implementation, we noticed we should not specify the error type in the SIMD (see https://github.com/anza-xyz/agave/pull/10223#discussion_r2744635089)


topointon-jump (approved):

LucasSte:
@t-nelson or @bw-solana any chance you can have a look here?

bw-solana (approved):

PR #457, opened 2026-01-30
amend SIMD-0387: extract `InitializeAccountV2` and list deps

buffalojoec:
Extracts `InitializeAccountV2` into its own SIMD and lists all dependencies.

PR #464, opened 2026-02-05
SIMD-0464: Vote Account Initialize V2

buffalojoec:


proposals/0464-vote-account-initialize-v2.md
```diff
@@ -0,0 +1,153 @@
...
+
+This proposal depends on the following previously accepted proposals:
+
+- **[SIMD-0387]: BLS Pubkey management in vote account**
+
+    Specifies BLS public key generation and proof of possession verification
+
+- **[SIMD-0458]: Stop Using Static SimpleVote Transaction Cost**
+
+    Removes statically defined compute unit costs for Simple Vote transactions
```
bw-solana: SIMD-0387 already calls this out as a dependency, so do we need to duplicate here?

proposals/0464-vote-account-initialize-v2.md
```diff
@@ -0,0 +1,153 @@
...
+
+Upon receiving the transaction, the vote program will perform a BLS
+verification on the submitted BLS public key and associated proof of
+possession, as described in [SIMD-0387]. The transaction will fail if the
+verification fails. Otherwise the new vote account is created with the given
+parameters.
+
+The BLS PoP verification will cost 34,500 CUs, as described in [SIMD-0387].
+
+### Disallow legacy InitializeAccount instruction
```
bw-solana: Do we need a hard cliff like this? Is there a universe where we could allow legacy and V2 at the same time (and deprecate legacy later)?

t-nelson:
how did the code land before the simd was merged?

AshwinSekar:
this instruction was initially specified in #387 (merged in nov 2025) . In order to decouple dependencies we decided to split it out for independent activation and keep the scope of 387 to just the bls pubkey #457

proposals/0464-vote-account-initialize-v2.md
buffalojoec: Sure, I can just make this one depend on 387.

proposals/0464-vote-account-initialize-v2.md
buffalojoec: I understand you're only talking about the cliff, but I started to wonder "why do we even need this instruction anyway". So, [I wrote this test](https://github.com/buffalojoec/solana-sdk/commit/ae959e863859d8478586d2eb5e022fd4de9d0202) to see if the entire post-SIMD-0123-and-0387 workflow would fit into one transaction. TL;DR yes for keypairs, no for derived vote accounts (`*WithSeed`).

As far as the cliff goes, I think as long as voters creating vote accounts without BLS keys isn't going to cause any disruption, we can do a slower deprecation strategy.

I believe the only reason for the hard cliff was the idea that SIMD-0387 should make BLS pubkey *mandatory* and the Vote program should no longer allow creation of vote accounts without them. However, maybe they just can't vote without them, and that incentive is enough?

bw-solana: yeah, my understanding is after SIMD-0387 gets enabled, it is still possible for someone to have a legacy style vote account that could be upgraded to have the VoteStateV4 fields with the upgrade commands. So in theory, it would be simple to support:

1. Current situation
2. InitializeAccountV2 available, InitializeAccount deprecated
3. InitializeAccount disabled

Of course, I can't imagine vote account initialization is a high velocity thing or that there are many workflows that would be disrupted, so jumping straight to disabling legacy InitializeAccount is probably fine.

topointon-jump: I think we should separate out the disabling of `InitializeAccount` from SIMD-0387. This is probably going to break CLIs and test cluster setup scripts etc, so we should de-couple this into a separate feature gate (like we have in this one).

Whether InitializeAccountV2 goes in SIMD-0387 or SIMD-0464 is inconsequential imo, whatever is easiest to ship this faster.

The main thing is not to couple deprecating `InitializeAccount` with SIMD-0387.

buffalojoec: Sounds good to me! I've added a commit to remove any mention of disabling the existing instruction. I figured mentioning in this proposal something like a "deprecated" log statement was gratuitous.

bw-solana (approved):
lgtm.

order of operations, I'm assuming we merge #457 first

AshwinSekar (approved):

topointon-jump (approved):

PR #470, opened 2026-02-25
amend SIMD-0464: add instruction discriminant, accounts, and SIMD-0232 checks

buffalojoec:
First updates the SIMD to specify the instruction discriminant and accounts. However, then we also needed to modify the input parameters as well as the accounts list in order to ensure that `InitializeAccountV2` imposes the same checks as `UpdateCommissionCollector` [as SIMD-0232 does](https://github.com/solana-foundation/solana-improvement-documents/blob/3713fb347ae02e3420a3f65f62900da831a5185a/proposals/0232-custom-commission-collector.md?plain=1#L105-L111).

### DIFF
--- a/proposals/0464-vote-account-initialize-v2.md
+++ b/proposals/0464-vote-account-initialize-v2.md
@@ -94,6 +94,8 @@ pub enum VoteInstruction {
     /// # Account references
     ///   0. `[WRITE]` Uninitialized vote account
     ///   1. `[SIGNER]` New validator identity (node_pubkey)
+    ///   2. `[WRITE]` Inflation rewards collector (or vote account if same)
+    ///   3. `[WRITE]` Block revenue collector (or vote account if same)
     InitializeAccountV2(VoteInitV2), // 16u32
 }
 ```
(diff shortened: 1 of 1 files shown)


proposals/0464-vote-account-initialize-v2.md
```diff
@@ -85,32 +85,55 @@ creation.
...
+A new instruction for initializing a vote account with all vote state v4 fields
+will be added to the vote program with the enum discriminant value of `16u32`
+little endian encoded in the first 4 bytes.
+
 ```rust
-InitializeAccountV2(VoteInitV2),
+pub enum VoteInstruction {
+    /// # Account references
+    ///   0. `[WRITE]` Uninitialized vote account
+    ///   1. `[SIGNER]` New validator identity (node_pubkey)
```
buffalojoec, line 96: I would have preferred to put the signer first, but this is the pattern for the rest of the program.

proposals/0464-vote-account-initialize-v2.md
```diff
@@ -85,32 +85,55 @@ creation.
...
+    ///   2. `[WRITE]` Inflation rewards collector (or vote account if same)
```
AshwinSekar, line 97: do these two need to be WRITE or is READ sufficient?

buffalojoec: Unfortunately SIMD-0232 requires them to be writable so as to prove they are not reserved account keys. Otherwise, yes I would have definitely kept them readonly.

https://github.com/solana-foundation/solana-improvement-documents/blob/8eeb25acd4537e11e3317d34ac481bb997a6575a/proposals/0232-custom-commission-collector.md?plain=1#L159-L161

AshwinSekar (approved):

joncinque (approved):
Looks good to me

bw-solana (approved):

topointon-jump (approved):

PR #484, opened 2026-03-05
feature(0001): add advisory simd type to process

t-nelson:
we would like to document some formal advice wrt to behavior and expectations surrounding some some sysvar fields that are ostensibly deprecated. however there is currently no canonical place in the ecosystem for such advice to live. given that the closest existing process would be explicit removal of the fields from the protocol via simd, it seems like this is as good a place as any.

introduce a new **Advisory** SIMD type

### DIFF
--- a/proposals/0001-simd-process.md
+++ b/proposals/0001-simd-process.md
@@ -48,7 +48,7 @@ more errors, etc.)
 
 ## Proposal Types
 
-There are two types of proposals:
+There are three types of proposals:
 
 - A **Standard** Proposal describes any change that affects most or all Solana
 implementations, such as a change to the network protocol, consensus, proposed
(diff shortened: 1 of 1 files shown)


topointon-jump (dismissed):

buffalojoec (commented):
Clarifying information about protocol features or notice of upcoming features makes sense, but for deprecation and errata the use case is less clear to me.

For deprecation, is the idea that you'd submit an Advisory to deprecate something, which will likely be removed via feature-gate in a later Standard SIMD?

For errata, does an errata Advisory only apply to protocol features that predate any SIMDs? Otherwise, we'd just want to amend the relevant SIMDs for any issues/adjustments.

It might be good to include some usage instructions on this new type to avoid it being a catch-all.

t-nelson:
> For deprecation, is the idea that you'd submit an Advisory to deprecate something, which will likely be removed via feature-gate in a later Standard SIMD?
> 

yeah for stuff that we want to simply stop using, not change in any way. like these rent sysvar field deprecations. we should document at the protocol level in some way so that we have something to point back to to justify adding deprecation warnings in sdks. however actually changing the protocol has very little urgency

> For errata, does an errata Advisory only apply to protocol features that predate any SIMDs? Otherwise, we'd just want to amend the relevant SIMDs for any issues/adjustments.

this is where we would add things that may not work as originally specified, but are not so broken as to justify a consensus breaking change to fix them

jacobcreech (approved):

buffalojoec (approved):

PR #486, opened 2026-03-06
SIMD-486: secp: Reject High-S Signatures

zz-sol:
## Summary

Adds SIMD-0486, which specifies switching the secp256k1 precompile's public-key recovery backend from `libsecp256k1` to `k256` (via `solana-secp256k1-recover`).

The only externally visible behavior change is that malleable (high-S) signatures — previously accepted — are rejected with
`PrecompileError::InvalidSignature` after the feature gate activates.

Feature gate: `secp256k1_precompile_use_k256` (`HgNAvKS5GrJbV57y1r8rF1FPmMet6WxEvSnChEeuy4QY`)


zz-sol:
prefer not to proceed.

PR #490, opened 2026-03-09
SIMD-490: Upgrade BPF Stake Program to v5.0.0

2501babe:


2501babe:
as far as im aware theres no precedent for what a core bpf upgrade simd is supposed to look like so i did what seemed reasonable. to my knowledge the only other core bpf upgrade is/will be the stake program with vote state v4 support, and that description was just a block in the vote state simd:

```
### Stake Program

The builtin stake program reads vote account state when creating, delegating,
and deactivating stake accounts. The program MUST be updated to support v4 vote 
accounts.
```

in particular structuring the design section as a changelog seems to make a lot more sense than laboriously describing implementation strategy for everything given that no other validators have to implement it. maybe even this is too much and we could have a simpler format for core bpf upgrades in the future

topointon-jump (approved):

ibhatt-jumptrading (approved):

proposals/0490-upgrade-stake-to-v5.md
```diff
@@ -0,0 +1,99 @@
...
+new stake accounts is held at `2_282_880`, the present rent-exempt reserve of a
+200-byte stake account, to minimize downstream breakage.
+* Make all sysvar account inputs optional. The Stake Program will continue to
+gracefully accept existing instructions that include sysvars.
+* Rewrite the implementation of `Split` to fix several longstanding bugs. Other
+that removing self-split, the new `Split` processor remains backwards
+compatible.
+
+In validator clients this feature will use existing code to effect a Core BPF
+Program upgrade. Otherwise, the only required validator support is to return the
```
t-nelson, line 51: do we have a simd that describes the core bpf upgrade process explicitly? might be nice to reference in these if so. might be good to have if not

t-nelson (approved):

joncinque (commented):
Looks good to me with Trent's comment resolved

buffalojoec:
> as far as im aware theres no precedent for what a core bpf upgrade simd is supposed to look like so i did what seemed reasonable. to my knowledge the only other core bpf upgrade is/will be the stake program with vote state v4 support, and that description was just a block in the vote state simd:
> 
> ```
> ### Stake Program
> 
> The builtin stake program reads vote account state when creating, delegating,
> and deactivating stake accounts. The program MUST be updated to support v4 vote 
> accounts.
> ```
> 
> in particular structuring the design section as a changelog seems to make a lot more sense than laboriously describing implementation strategy for everything given that no other validators have to implement it. maybe even this is too much and we could have a simpler format for core bpf upgrades in the future

Makes sense to me! I thought I at least outlined the upgrade method in [SIMD-0088](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0088-enable-core-bpf-programs.md) but nope! I guess I could write a retroactive SIMD for https://github.com/anza-xyz/agave/pull/1140 or just amend SIMD-0088.

proposals/0490-upgrade-stake-to-v5.md
```diff
@@ -0,0 +1,99 @@
...
+
+N/A
+
+## Detailed Design
+
+BPF Stake v5.0.0 introduces several changes:
+
+* Raise minimum delegation from 1 lamport to 1 sol. This is a medium-firm
+blocker on rent reduction because allowing the total lamports required to open a
+stake account to fall is highly undesirable.
```
buffalojoec, line 38: This means that the feature for minimum delegation is effectively activated upon successful upgrade of the Stake program, and *not* later activated in its own feature gate, right?

proposals/0490-upgrade-stake-to-v5.md
buffalojoec: Heh I can write one up. I thought I did a long time ago.

t-nelson:
i don't think we need to block here. would be best to have the standalone simd and link back to that wherever relevant. we can amend the simds that make sense

buffalojoec (approved):
LGTM

Here's the Core BPF upgrade SIMD: https://github.com/solana-foundation/solana-improvement-documents/pull/495

proposals/0490-upgrade-stake-to-v5.md
2501babe: thats right. minimum delegation was originally designed, and continues to be, a function in the stake program returning a hardcoded constant. for native stake that meant a feature gate to return a different one, for bpf stake there is no branch and updating the program itself changes the value everything else sees

buffalojoec: Sounds good, thanks. I also wrote this comment before reviewing the Agave change, so looks good on the validator side, too.

PR #495, opened 2026-03-18
SIMD-0495: Core BPF Upgrade Procedure

buffalojoec:


proposals/0495-core-bpf-upgrade-procedure.md
```diff
@@ -0,0 +1,204 @@
...
+   Upgradeable Loader (Loader v3).
+4. Generate a new keypair for the feature gate.
+5. Create a new feature gate to upgrade the target program with the source
+   buffer, according to your validator implementation.
+6. Follow the existing process for activating features.
+
+### Source Buffer Validation
+
+When the feature gate activates, the runtime must validate the source buffer
+account before proceeding. The source buffer account:
```
tlambertz, line 86: There should be a way to make sure that an upgrade is "set in stone". Like, you don't want a situation where there is some "random" buffer authority pubkey could potentially sneak in an update right before the epoch boundary in which the feature was activated. Two ideas come to mind: Burn the buffer authority before creating the feature/activating it, or hash the buffer before doing the update, with the hash part of the featuregate?

topointon-jump: Completely agree with this. Personally I like the buffer hash approach, as it makes it easy to replicate the upgrade in test clusters. We can just check that the hash of the buffer is equal to the expected hash before doing the upgrade, and if we change the buffer hash we re-key the feature gate.

This has the advantage that we can easily create test clusters and ledgers to test upgrades, with certainty that we are testing what will actually happen in production.

buffalojoec: Sounds good, I can add this. It also enforces use of verified builds. I believe the slashing program used a build hash. We can just cook that into the formal spec.

buffalojoec: https://github.com/solana-foundation/solana-improvement-documents/pull/495/changes/7f7c3a49afb9b750c916508315816ba9b1592db6

topointon-jump (approved):

mjain-jump (approved):

proposals/0495-core-bpf-upgrade-procedure.md
```diff
@@ -0,0 +1,238 @@
...
+
+Core BPF programs are Loader v3 (BPF Upgradeable Loader) programs, as
+specified in [SIMD-0088]. Upgrading a Core BPF program involves replacing
+the contents of an existing program data account with new ELF bytes from a
+source buffer account. The program account itself is not modified. A feature
+gate triggers the upgrade at activation time.
+
+### Preparation
+
+Before an upgrade can occur, the following steps must be completed:
```
t-nelson, line 62: these steps cannot be enforced in protocol. can we note that they should be part of the code review process for integrating the upgrade feature gate into validator client implementation?

proposals/0495-core-bpf-upgrade-procedure.md
```diff
@@ -0,0 +1,238 @@
...
+  adjusted by the difference between burned and funded lamports.
+
+- **Accounts data size delta**: Updated to reflect the difference between
+  the old total data size (old program data account + source buffer account)
+  and the new total data size (new program data account).
+
+### Program Availability
+
+In the slot immediately following the feature activation, the program will
+not be invocable. This status will last one slot, then the program will be
```
t-nelson: it's not one block? the one slot cool down should impact all forks with the upgrade slot as parent

proposals/0495-core-bpf-upgrade-procedure.md
```diff
@@ -0,0 +1,238 @@
...
+triggers the upgrade when activated.
+
+## Alternatives Considered
+
+### Standard Loader v3 Upgrade Instruction
+
+Core BPF programs with `None` as their upgrade authority cannot use the
+standard Loader v3 `Upgrade` instruction. Even if an upgrade authority were
+set, using the standard instruction would require a single signer to hold
+the authority, which is inappropriate for programs critical to network
```
t-nelson: `native_invoke_signed` would like a word 😉

proposals/0495-core-bpf-upgrade-procedure.md
```diff
@@ -0,0 +1,238 @@
...
+operations. The feature-gated approach ensures upgrades go through the same
+consensus process as other protocol changes.
+
+### Governance-Based Upgrade Authority
+
+An alternative would be to assign a multisig or governance program as the
+upgrade authority. This was rejected because multisig operations are slow
+and difficult to coordinate with the timing of feature gate activations,
+which must occur at the exact same slot across all validators. It also
```
t-nelson: fwiw, there's motion to move _all_ feature gate activations to multisig authority. so this may not be so much an issue anymore. or rather, will happen as a follow on effect of the social process changing

t-nelson (commented):
there is no explicit block marker for "the upgrade must not proceed" terminations, right? we end up with an otherwise no-op feature gate activation

proposals/0495-core-bpf-upgrade-procedure.md
buffalojoec: Comment added in https://github.com/solana-foundation/solana-improvement-documents/pull/495/changes/abedb286fe3bdbc23015ef20b846b2bb6071d762.

proposals/0495-core-bpf-upgrade-procedure.md
buffalojoec: Is this clearer? https://github.com/solana-foundation/solana-improvement-documents/pull/495/changes/abedb286fe3bdbc23015ef20b846b2bb6071d762

proposals/0495-core-bpf-upgrade-procedure.md
buffalojoec: Alright, redefined the motivation a bit in https://github.com/solana-foundation/solana-improvement-documents/pull/495/changes/abedb286fe3bdbc23015ef20b846b2bb6071d762

t-nelson (approved):
lgtm now thanks!

i can't resolve my own comments so assume 👍 equivalent

PR #497, opened 2026-03-18
Amend SIMD-0123: clarify block revenue distribution mechanism

buffalojoec:
Reopens #300


joncinque (approved):
Looks great, let's get this in!

bw-solana (approved):

topointon-jump (approved):

PR #499, opened 2026-03-18
SIMD-0499: Deactivate execution of loader-v1 and ABI-v0

Lichtso:
See the related [SIMD discussion](https://github.com/solana-foundation/solana-improvement-documents/discussions/483).


proposals/0499-deactivate-execution-of-loader-v1-and-abi-v0.md
```diff
@@ -0,0 +1,57 @@
...
+
+## New Terminology
+
+None.
+
+## Detailed Design
+
+After the activation of the associated feature key a validator must fail to
+execute programs owned by loader-v1, throwing the error message
+`InstructionError::UnsupportedProgramId`.
```
buffalojoec: I know it's a simple change, but do we want to be more specific about the pipeline here? Should validators still attempt to load their ELFs first, or should the error happen immediately during message processing, while processing the particular instruction for the program, before attempting to load the executable? I think we all prefer the latter.

proposals/0499-deactivate-execution-of-loader-v1-and-abi-v0.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Alternatives Considered
+
+- First bumping the CU cost of ABI-v0 significantly to get users off the
+Memo-v1 program.
+- Adapting the Memo Program v1 to ABI-v1 and redeploying it on loader-v3.
```
buffalojoec: This shouldn't be an alternative, since:
* This is a program-level decision, not an ABI-level decision
* We should still do this

proposals/0499-deactivate-execution-of-loader-v1-and-abi-v0.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Impact
+
+The only loader-v1 / ABIv0 program still in use today is Memo Program v1.
+Of which there has been a loader-v2 / ABIv1 replacement around for a long time.
```
buffalojoec: I would use [my research tooling](https://github.com/solana-foundation/solana-improvement-documents/discussions/483#discussioncomment-16039374) to flesh out this section and be as detailed as possible about who's going to break, what their transaction traffic looks like, and why we deem it's acceptable to break them. We should present a solid case for completely deactivating an ABI version, even if it's old.

buffalojoec: It's also worth making it explicitly clear that on-chain programs under Loader v1 will be bricked forever.

proposals/0499-deactivate-execution-of-loader-v1-and-abi-v0.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Security Considerations
+
+The activation of this feature should be straight forward but the later clean
+up together with the removal of alignment check skipping would be more complex.
```
buffalojoec: I don't think this statement adds much value. I agree the security implications are relatively low, but we don't need to mention cleanup overhead in a SIMD.

buffalojoec (commented):
Can you link the discussion thread? https://github.com/solana-foundation/solana-improvement-documents/discussions/483

Personally I think this SIMD is a bit premature. I think we should have had more discussion in #483 first, and reached a conclusion on what to do about programs like Memo and Serum. We should still do that over in the discussion and add updates here to the proposal accordingly!

Lichtso:
> Can you link the discussion thread? https://github.com/solana-foundation/solana-improvement-documents/discussions/483

Is already linked in the "Motivation" section, unless you want me to move it to the PR description instead.

proposals/0499-deactivate-execution-of-loader-v1-and-abi-v0.md
Lichtso: I only added it as a reminder to myself, because this SIMD is somewhat special: I expect the security audit of the removal to be far more involved than the implementation of the feature gate.

topointon-jump (dismissed):

proposals/0499-deactivate-execution-of-loader-v1-and-abi-v0.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Drawbacks
+
+Slight inconvenience to the last remaining users of the Memo Program v1 as well
+as bricking of any sleeper programs owned by loader-v1.
```
buffalojoec: Personally I'd just take this section out. If we decide to migrate Memo v1 to Loader V3 we can solve the problem. This SIMD should only reference Loader V1 programs in their entirety.

proposals/0499-deactivate-execution-of-loader-v1-and-abi-v0.md
buffalojoec: Yeah I still think it can just be removed. I don't see anything worth mentioning here for security.

proposals/0499-deactivate-execution-of-loader-v1-and-abi-v0.md
```diff
@@ -0,0 +1,67 @@
...
+
+## New Terminology
+
+None.
+
+## Detailed Design
+
+After the activation of the associated feature key a validator must fail to
+execute programs owned by loader-v1, throwing the error message
+`TransactionError::InvalidProgramForExecution` during transaction loading.
```
buffalojoec: Is this going to be a fees-only transaction? Can you specify?

proposals/0499-deactivate-execution-of-loader-v1-and-abi-v0.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Alternatives Considered
+
+Continuing to support this barely used functionality.
+
+## Impact
```
buffalojoec, line 44: Honestly I would dump all of this information about Memo v1 into the discussion thread and leave only the last paragraph here. What we decide to do about Memo v1 should get hashed out over there. The hammer of this SIMD is coming to the rest of Loader V1 programs anyway.

proposals/0499-deactivate-execution-of-loader-v1-and-abi-v0.md
buffalojoec: Also, since this is a transaction-level error, what about CPI? If I try to CPI to a Loader V1 -owned program, what happens?

I'm fine with top-level throwing `TransactionError::InvalidProgramForExecution`, but CPI should throw *instruction-level* `InstructionError::UnsupportedProgramId` and it should be specified here.

proposals/0499-deactivate-execution-of-loader-v1-and-abi-v0.md
buffalojoec: Reiterating that the tooling is available to add qualitative metrics to the Impact section.

proposals/0499-deactivate-execution-of-loader-v1-and-abi-v0.md
```diff
@@ -0,0 +1,67 @@
+---
+simd: '0499'
+title: Deactivate execution of loader-v1 and ABI-v0
+authors:
+  - Alexander Meißner (Anza)
+category: Standard
+type: Core
+status: Review
+created: 2027-03-16
```
buffalojoec:
```suggestion
created: 2026-03-16
```

proposals/0499-deactivate-execution-of-loader-v1-and-abi-v0.md
Lichtso: I added you to the authors and "Allow edits and access to secrets by maintainers" is enabled. You should be able to commit and push your analysis yourself.

buffalojoec: That's okay I'd rather still be able to review/approve.

I suggest placing the truncated statistics from [my comment here](https://github.com/solana-foundation/solana-improvement-documents/discussions/483#discussioncomment-16039374) in the Impact section and then also including a supporting doc with the full sparkline plot from running the [program-sync tool](https://github.com/blueshift-gg/program-sync):

```
cargo run --release -- rpc program-ids --loader 1
cargo run --release -- rpc usage --file <OUT_FILE>
```

It's imperative that we publish all of the data about Loader V1 programs before anyone can collectively agree to disable them forever.

buffalojoec (commented):
This is looking pretty good, but I'm adamant about [including the data](https://github.com/solana-foundation/solana-improvement-documents/pull/499#discussion_r3008332761) on Loader V1 traffic in this proposal. It should not be taken lightly.

Once we have all of the data included, it's probably best to mark this as a draft and iron out the what-if's in the discussion thread first. What do you think?

proposals/0499-deactivate-execution-of-loader-v1-and-abi-v0.md
```diff
@@ -0,0 +1,52 @@
...
+feature: TBD
+---
+
+## Summary
+
+Deactivate execution of loader-v1 and ABI-v0.
+
+## Motivation
+
+The trouble with ABIv0 is that it had no alignment padding in its serialization
```
mjain-jump, line 20: are there any actively-used v1 programs that are still out there?

Lichtso: See https://github.com/solana-foundation/solana-improvement-documents/discussions/483#discussioncomment-16683601

proposals/0499-deactivate-execution-of-loader-v1-and-abi-v0.md
```diff
@@ -0,0 +1,55 @@
...
+None.
+
+## Detailed Design
+
+After the activation of the associated feature key a validator must fail to
+execute programs owned by loader-v1 throwing the error message:
+
+- `TransactionError::InvalidProgramForExecution` for top level instructions at
+transaction loading time (before executing the first instruction)
+- `InstructionError::UnsupportedProgramId` for CPI calls
```
topointon-jump, line 38: Can you be a bit more specific about where these checks happen? So that we can implement it without reference to the Agave code.

Lichtso:
There were concerns raised about the amount of funds which would become locked / burnt if these programs would become unavailable forever. However, we are lacking a good way to estimate the value of these funds for all these different programs.

umbnat92:
Just answering here as well for completeness.

> However, we are lacking a good way to estimate the value of these funds for all these different programs.

The estimate is around $900k (full details [here](https://github.com/solana-foundation/solana-improvement-documents/discussions/483#discussioncomment-17672897)). There are ~130 mints associated to dead tokens, so left w/o price, but we don't expect the real value to be order of magnitude different from this.

Lichtso:
See discussion: https://github.com/solana-foundation/solana-improvement-documents/discussions/483#discussioncomment-17698115

@jacobcreech Think you said you are generally on board with this but wanted to try to find the last remaining users first, right?

PR #500, opened 2026-03-19
SIMD-0500: Disable deployment of SBPF-v0, SBPF-v1 and SBPF-v2

Lichtso:


proposals/0500-disable-deployment-of-sbpf-v0-v1-v2.md
```diff
@@ -0,0 +1,62 @@
...
+
+## New Terminology
+
+None.
+
+## Detailed Design
+
+After the activation of the associated feature key a validator must fail to
+execute programs with any SBPF version other than v3, throwing the error
+message `InstructionError::InvalidAccountData`.
```
buffalojoec: This section says "fail to execute" but the SIMD proposal is about rejecting *deployments*, not execution.

buffalojoec: Similar to my comment on [#499](https://github.com/solana-foundation/solana-improvement-documents/pull/499#discussion_r2963589527) we should detail *where* in the Loader V3 program that this rejection should occur.

proposals/0500-disable-deployment-of-sbpf-v0-v1-v2.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Alternatives Considered
+
+None.
```
buffalojoec: Allowing some or all versions to continue to be deployed?

proposals/0500-disable-deployment-of-sbpf-v0-v1-v2.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Impact
+
+dApp developers using outdated toolchains will have to update them and adjust
+their programs before they can re-/deploy them.
```
buffalojoec: This impact section only covers one camp of users: those who use the CLI and the platform-tools to compile programs before deployment. We have developers using their own custom solutions or upstream Rust/LLVM to compile program binaries as well, so we should include information here about specifically what they need to change.

proposals/0500-disable-deployment-of-sbpf-v0-v1-v2.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Security Considerations
+
+What security implications/considerations come with implementing this feature?
+Are there any implementation-specific guidance or pitfalls?
```
buffalojoec: Need to fill this or kill it. IMO this is a relatively low-security operation.

proposals/0500-disable-deployment-of-sbpf-v0-v1-v2.md
```diff
@@ -0,0 +1,62 @@
...
+## Dependencies
+
+This proposal depends on the following previously accepted proposals:
+
+- [SIMD-0178](https://github.com/solana-foundation/solana-improvement-documents/pull/178)
+- [SIMD-0189](https://github.com/solana-foundation/solana-improvement-documents/pull/189)
+- [SIMD-0377](https://github.com/solana-foundation/solana-improvement-documents/pull/377)
+
+which together form SBPFv3. There should be a new valid deployment target
+before we deactivate the older ones.
```
buffalojoec: Maybe you can reword this to say that the successful implementation of all three dependent SIMDs results in the new SBPFv3 target? Just so it's clear we actually have a new target, not suggestive.

proposals/0500-disable-deployment-of-sbpf-v0-v1-v2.md
```diff
@@ -0,0 +1,62 @@
...
+## Summary
+
+Disable deployment of SBPFv0, SBPFv1 and SBPFv2.
+
+## Motivation
+
+The number of programs using these older SBPF versions will decrease over
+time as dApp developers will have to target at least SBPFv3 for re-deployments.
+This will eventually allow us to deprecate the execution of these SBPF versions
+after their existence (and thus also usage) drops to near zero.
```
buffalojoec, line 23: This sounds less like a motivation and more like an impact consideration. Our motivation should be about the good things we've included in SBPFv3 that are missing in lower versions, such as static syscalls. It also should describe that SBPFv3 endeavors to be upstream BPF compatible, which IMO is the primary motivation for eliminating other versions.

Lichtso: > Our motivation should be about the good things we've included in SBPFv3 that are missing in lower versions

Well, that is making the case for **dApp devs** using SBPFv3, but that was already motivated in the three SIMDs of SBPFv3. Here, we are making the case for **validators** to stop supporting older SBPF versions.

proposals/0500-disable-deployment-of-sbpf-v0-v1-v2.md
Lichtso: I knew you were going to say this ;)
That is why it only states "toolchains" and not "platform-tools" or anything Anza specific.

proposals/0500-disable-deployment-of-sbpf-v0-v1-v2.md
LucasSte: We have considered as well increasing the CU costs of older programs to incentivize migration. Maybe this could be added to the section.

Are we still considering this?

Lichtso: This will only become relevant later, once we want to disable execution too.

topointon-jump (approved):

buffalojoec (commented):
Good enough for me, just one more thought. 

If we plan to eventually disallow *execution* of these SBPF versions, we should check the chain to see if they are all still upgradeable.

A nice addition to this SIMD - since it changes Loader V3 - would be to also disable **freezing** of programs that are not at least v3. That would save us some headache later.

proposals/0500-disable-deployment-of-sbpf-v0-v1-v2.md
```diff
@@ -0,0 +1,70 @@
...
+
+None.
+
+## Detailed Design
+
+After the activation of the associated feature key a validator must fail to
+deploy, upgrade or finalize programs with any SBPF version other than v3.
+Loader-v3 instructions `DeployWithMaxDataLen`, `Upgrade` and `Finalize` must
+throw the error `InstructionError::InvalidAccountData` when verifying the
+program data.
```
buffalojoec: Just a suggestion to make it more clear how the program is supposed to operate:

```suggestion
After the activation of the associated feature key a validator must fail to
deploy, upgrade or finalize programs with any SBPF version other than v3.

Loader V3 instructions `DeployWithMaxDataLen` and `Upgrade` must return
`InstructionError::InvalidAccountData` when verifying the buffer's program data
used to deploy or upgrade from.

Loader V3 instruction `Finalize` must return `InstructionError::InvalidAccountData`
when verifying the program data of the program attempting to be finalized.
```

proposals/0500-disable-deployment-of-sbpf-v0-v1-v2.md
```diff
@@ -0,0 +1,70 @@
...
+## Impact
+
+dApp developers using outdated toolchains will have to update them and adjust
+their programs before they can re-/deploy them.
+
+Furthermore, testing frameworks and mock ups will have to be adapted to either
+deactivate this feature or bypass the entire deployment in order to continue
+to test older SBPF versions.
+
+## Security Considerations
```
buffalojoec, line 88: It would be great to include stats in the impact section. I did the work in Blueshift's program-sync tool (https://github.com/blueshift-gg/program-sync/pull/9) to get the data we need. You can run it for all networks, but here's a suggested snippet to add for mainnet:

```suggestion
### Currently Deployed SBPF Versions per Loader (Mainnet)

Analysis performed using Blueshift's [program-sync] tool.

\```
⠏ >>> PROFILED 136 ACTIVE LOADER V1 PROGRAMS                                                                   BPFLoader v1  (136 programs)
  SBPF:      v0: 136  v1: 0  v2: 0  v3: 0  unknown: 0
  Finalized: v0: 136  v1: 0  v2: 0  v3: 0  unknown: 0

⠏ >>> PROFILED 315 ACTIVE LOADER V2 PROGRAMS                                                                   BPFLoader v2  (315 programs)
  SBPF:      v0: 315  v1: 0  v2: 0  v3: 0  unknown: 0
  Finalized: v0: 315  v1: 0  v2: 0  v3: 0  unknown: 0

⠏ >>> PROFILED 17332 ACTIVE BPFLoaderUpgradeable PROGRAMS                                                      BPFLoaderUpgradeable  (17332 programs)
  SBPF:      v0: 17279  v1: 12  v2: 41  v3: 0  unknown: 0
  Finalized: v0: 422  v1: 0  v2: 0  v3: 0  unknown: 0

⠏ >>> PROFILED 0 ACTIVE LoaderV4 PROGRAMS                                                                      LoaderV4:  0 programs (skipped)
\```

[program-sync]: https://github.com/blueshift-gg/program-sync

## Security Considerations
```

(sorry, github suggestions get clobbered when they include code snippets)
(besides, I think the formatting needs fixing anyway)

buffalojoec (commented):
Looks good! Just needs a bit more specifics and we can send it along.

buffalojoec (dismissed):
Thanks!

proposals/0500-disable-deployment-of-sbpf-v0-v1-v2.md
```diff
@@ -0,0 +1,85 @@
...
+before we deactivate the older ones.
+
+## New Terminology
+
+None.
+
+## Detailed Design
+
+After the activation of the associated feature key a validator must fail to
+deploy, upgrade or finalize programs with any SBPF version other than v3.
```
bw-solana: should probably wordsmith this to indicate there could be future loader versions that should not fail. E.g. call out the specific loaders that should fail

proposals/0500-disable-deployment-of-sbpf-v0-v1-v2.md
```diff
@@ -0,0 +1,85 @@
...
+Loader V3 instructions `DeployWithMaxDataLen` and `Upgrade` must return
+`InstructionError::InvalidAccountData` when verifying the buffer's program data
+used to deploy or upgrade from. Loader V3 instruction `Finalize` must return
```
bw-solana: there is no `Finalize` instruction for Loader v3 though, right?

proposals/0500-disable-deployment-of-sbpf-v0-v1-v2.md
```diff
@@ -0,0 +1,85 @@
...
+`InstructionError::InvalidAccountData` when verifying the program data of the
+program attempting to be finalized.
```
bw-solana: maybe we should specify buckets of behavior for every Loader v3 instruction. E.g. make it clear which ixs are still allowed for older SBPF versions such as `ExtendProgram`

proposals/0500-disable-deployment-of-sbpf-v0-v1-v2.md
```diff
@@ -0,0 +1,85 @@
...
+
+Core program migrations and upgrades are exempt from this,
```
bw-solana: we should probably clarify what this actually means. Maybe list out the set of core programs that are exempt and a brief explanation of what exempt means (e.g. All Loader v3 instructions may still be performed on core programs according to the rules specified in past SIMD XXX)

proposals/0500-disable-deployment-of-sbpf-v0-v1-v2.md
Lichtso: Good point, the finalize instruction is `SetAuthority`.

proposals/0500-disable-deployment-of-sbpf-v0-v1-v2.md
topointon-jump: This should read something like a list of all the components that need to be changed, and the new checks they should now do. Is it just in the loader v3 instructions? Or other places? This needs to be explicit.

bw-solana (approved):

proposals/0500-disable-deployment-of-sbpf-v0-v1-v2.md
```diff
@@ -40,15 +40,23 @@ None.
...
+Currently loader-v3 is the only loader with program management enabled, thus
+this means for each of the following loader-v3 instructions:
+
+- deployment: `DeployWithMaxDataLen`
+- redeployment: `Upgrade`
+- finalization: `SetAuthority` and `SetAuthorityChecked` with no new authority
+
+must return `InstructionError::InvalidAccountData` when the ELF contains
+a SBPFv0, SBPFv1 or SBPFv2 program. While `InitializeBuffer`, `Write`,
+`ExtendProgram` and `Close` remain unaffected.
```
buffalojoec, line 55: I don't mind this rewording, but can you just make it clear which ELF we're talking about with each instruction? Previously the paragraph stated program or buffer. It will be helpful to distinguish here.

buffalojoec (commented):
Small tweak for me, then I'll re-stamp.

buffalojoec (approved):

buffalojoec:
@jacobcreech is interested in amending this SIMD so we do not deny upgrades, instead only new deployments. For anyone whose program is using Anchor and an old version of SDK deps, the upgrade block forces them to overhaul the whole program, which can turn a small, incremental upgrade into a full-blown refactor. Since we can't ever truly get rid of deployed SBPF versions 0..2, seems like we can probably be less aggressive here if it helps the ecosystem.

PR #503, opened 2026-03-26
SIMD-0503: Static Sysvars

deanmlittle:


ptaffet-jump:
I don't think it's a good idea to overload memory translation like this. This leads to some strange behavior where

```
lddw r3, 0x494df715 // SOL_RENT_SYSVAR
ldxdw r3, [r3+0]
```
loads from a sysvar, but 
```
lddw r3, 0x494df714 // SOL_RENT_SYSVAR - 1
add  r3, 1
ldxdw r3, [r3+0]
```
OR
```
lddw r3, 0x494df714 // SOL_RENT_SYSVAR - 1
ldxdw r3, [r3+1]
```
gives a different value (if I'm understanding correctly).  An ISA where these snippets do not have identical behavior seems cursed to me.

I'm less opposed to putting these sysvars at a known, fixed, appropriately aligned location in an existing read-only data segment, but it doesn't seem better enough than passing in the account to justify a change.
Can't you just require the sysvar accounts to be passed in first, and then use pre-computed offsets into the input section? Checking the pubkey takes like 16 CUs, no?

deanmlittle:
You have perfectly demonstrated why this technique is safe and doesn't interfere with any existing APIs or instructions. It simply makes dereferencing certain 32bit addresses that currently point to nothing translate to a pointer to a global variable instead of throwing a memory access error.

I am not sure you have considered the performance benefits. We can load *all* sysvars once per slot for less than the cost of loading in the largest one just once, and consume them N times within a slot.

Beyond this, it makes writing onchain programs incredibly ergonomic and performant. There really are no downsides and some rather extreme performance and devex benefits. I am yet to see a developer whose mind wasn't blown by how easy it could be for them to write programs with this method.

alnoki:
> I am yet to see a developer whose mind wasn't blown by how easy it could be for them to write programs with this method

+1, please land this

ptaffet-jump:
How do you access the other fields of the sysvar? E.g. the epoch value from the clock sysvar?

Don't forget that making address translation in the VM more expensive makes every program slower, even if that's not measurable in CUs.

deanmlittle:
> How do you access the other fields of the sysvar? E.g. the epoch value from the clock sysvar?

It's actually super simple. As you've demonstrated, all existing behavior remains unmodified. The only change is that a previously invalid dereference to this address is now valid. Due to the fact that:

- We are utilizing an address in the 32-bit range (extended to 64 bits with upper 32 bits masked out in this case, as our wonderful Rust/LLVM/BPF doesn't currently have a good way to produce ALU32 instructions without inline assembly)
- All deferences for any address under 0x0100000000 are currently invalid 
- Write (stx) instructions remain unchanged and continue to fail as they do now
- All dereferences (ldx) require a relative offset
- The translated address is bound-checked

We simply treat any dereference to this address as the base, and its offset as a relative offset to that address.

In other words, if `0xff395088u64` is the start address of our `Clock`, then:

```asm
lddw r3, 0xff395088 // SOL_CLOCK_SYSVAR
ldxdw r4, [r3+0x00] // load current slot into r4
ldxdw r5, [r3+0x20] // load current timestamp into r5
```

By virtue of none of our existing sysvars having a length in excess of `i16::MAX` (32,767 bytes), relative offsets from a known base address is sufficient to encapsulate all existing sysvar account values.

> Don't forget that making address translation in the VM more expensive makes every program slower, even if that's not measurable in CUs.

This only happens globally once per slot. The overhead per transaction is a single fat pointer to the sysvar cache. Absolutely negligible considering how much we already copy just to process a transaction. Consider that by comparison today, a single sysvar account load in a single transaction (of which there are MANY more than one per slot) requires serializing >10kb of data, the majority of which is never even used in the VM, whereas the total aggregate of all sysvars today is <30kb – less than the cost of serializing our single largest sysvar account one time, or our smallest sysvar account 3 times in any given slot.

proposals/0503-static-sysvars.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Security Considerations
+
+1. We must ensure no intra-slot mutability of any exposed globals.
+2. We must ensure static sysvar pointers remain synchronized with SysvarCache 
+   at the slot boundary.
+3. Despite being 32-bit hashes, it is important that we cast to u64 first as
+   50% of 32-bit murmur hashes sign extend to negative 64-bit addresses. If 
+   our toolchain reliably generated `mov32` without requiring inline assembly, 
+   this would be more ideal, as it would save 8 bytes of binary size.
```
buffalojoec: I would make this explicit in the design section so it's very clear how to cast it.

proposals/0503-static-sysvars.md
```diff
@@ -0,0 +1,109 @@
...
+   resulting in degraded developer experience.
+2. Invoking a syscall to access a global requires both a stack allocation and 
+   halting execution – an immense amount of overhead just to read a value that 
+   is already readily available.
+3. Having to pass in an account to access a global results in a 10kb penalty 
+   to data serialization and has negative implications for composability.
+
+If these globals were simply exposed to the VM, as is common in kernel BPF, 
+and resolved JIT, we could dramatically improve developer experience, whilst 
+also reducing the runtime overhead of accessing them to just 2 CUs. This is
```
buffalojoec, line 40: The CU cost breakdown and rationale should have a dedicated section in the Detailed Design.

proposals/0503-static-sysvars.md
```diff
@@ -0,0 +1,109 @@
...
+category: Standard
+type: Core
+status: Idea
+created: 2026-03-25
+feature: (fill in with feature key and github tracking issues once accepted)
+---
+
+## Summary
+
+Leverage existing static linking infrastructure of JIT compilation to enable
```
Lichtso: > static linking

Nit on terminology: I know we called them "static syscalls" but they are not static linking (that would imply we would inline their implementation into the SBPF ELF at or before deployment), instead they are still dynamic linking but relocation-less.

> existing ... infrastructure

The infrastructure for "static syscalls" is for syscalls only, as the name implies. Linking of (in this case readonly) data is a different process and there is no existing infrastructure to utilize.

> JIT compilation

I know we have used JIT-only mode on MNB in Agave for years but with the next version we could have tiered compilation (Interpreter & JIT hybrid). Also Firedancer is using an Interpreter.

proposals/0503-static-sysvars.md
```diff
@@ -0,0 +1,109 @@
...
+Leverage existing static linking infrastructure of JIT compilation to enable 
+static resolution of sysvars.
+
+## Motivation
+
+In order to access Sysvar values, developers currently have three options:
+
+1. Invoke a specific getter syscall
+2. Invoke the get_sysvar syscall, or
+3. Include a Sysvar account in their program.
```
Lichtso: Obligatory https://xkcd.com/927/.

That said, yes we could improve upon sysvars and the sysvar syscalls are not great.

proposals/0503-static-sysvars.md
```diff
@@ -0,0 +1,109 @@
...
+lddw r3, 0x494df715 // SOL_RENT_SYSVAR
+ldxdw r3, [r3+0]
+```
+
+In this case, the murmur hash value of `0x494df715` is then resolved JIT to a 
+memory address containing the current `Rent` sysvar value. The same applies to 
+other sysvar accounts, such as the murmur hash of `0xff395088` for `Clock`.
+
+Ergo, we can safely and performantly expose any available global variable to 
+the VM without the overhead of additonal account loads or syscall invocation.
```
Lichtso: Again, this thinks about the JIT only, which is not protocol, it is one possible implementation. What about an interpreter? It would have to do this in address translation on every memory access.

Also, while maybe not that impactful, it would also make the JIT implementation more complex and thus compilation slower as this is a form of macro-op fusion which requires state to be tracked between the instructions. Currently we can compile every instruction independently.

proposals/0503-static-sysvars.md
```diff
@@ -0,0 +1,109 @@
...
+In order to access Sysvar values, developers currently have three options:
+
+1. Invoke a specific getter syscall
+2. Invoke the get_sysvar syscall, or
+3. Include a Sysvar account in their program.
+
+The downsides of all of these approaches are threefold:
+
+1. Sysvar values are globals that are always available to the validator, but 
+   aren't exposed as globals during execution. This is a clunky anti-pattern
```
Lichtso, line 30: ~Sysvar accounts are already loaded for all transactions even if their message does not mention the sysvar key.~

Scratch that, it only applies to the sysvar cache but it should be easy to expand to transaction accounts too.

proposals/0503-static-sysvars.md
```diff
@@ -0,0 +1,109 @@
...
+3. Include a Sysvar account in their program.
+
+The downsides of all of these approaches are threefold:
+
+1. Sysvar values are globals that are always available to the validator, but 
+   aren't exposed as globals during execution. This is a clunky anti-pattern 
+   resulting in degraded developer experience.
+2. Invoking a syscall to access a global requires both a stack allocation and 
+   halting execution – an immense amount of overhead just to read a value that 
+   is already readily available.
```
Lichtso, line 34: Agreed, sysvar syscalls are overkill for reading constant data.

proposals/0503-static-sysvars.md
```diff
@@ -0,0 +1,109 @@
...
+3. Having to pass in an account to access a global results in a 10kb penalty 
+   to data serialization and has negative implications for composability.
```
Lichtso, line 36: Additionally, depending on the design of the entrypoint a program must also find the instruction account of the sysvar, but that will be fixed by SIMD-0449.

> 10kb penalty to data serialization

That part will be gone after direct mapping and the CU charging adjustments of SIMD-0452, which I still have to rewrite.

> negative implications for composability

I think that is the remaining actual downside of sysvar accounts: They waste key space in the instruction invocation.

proposals/0503-static-sysvars.md
```diff
@@ -0,0 +1,109 @@
...
+memory address containing the current `Rent` sysvar value. The same applies to 
+other sysvar accounts, such as the murmur hash of `0xff395088` for `Clock`.
+
+Ergo, we can safely and performantly expose any available global variable to 
+the VM without the overhead of additonal account loads or syscall invocation.
+
+## Alternatives Considered
+
+- Leverage JIT intrinsics to provide similar syscall functionality.
+- Don't improve the existing design of sysvars/syscalls.
```
Lichtso: Think taking another look at solving this problem using sysvar accounts is worthwhile. The gap to bridge there might be relatively small (see https://github.com/solana-foundation/solana-improvement-documents/pull/503/changes#r3002020193).

Lichtso (commented):
> By opening this PR you are affirming that your SIMD has been thoroughly discussed and vetted in the [SIMD discussion section](https://github.com/solana-foundation/solana-improvement-documents/discussions/categories/ideas).

Did I miss anything?

Lichtso:
> All deferences for any address under `0x0100000000` are currently invalid

In [SIMD-0189](https://github.com/solana-foundation/solana-improvement-documents/pull/189) we moved the readonly-data down into the 32 bit address range as most 64 bit load immediate (`lddw`) only want to load addresses to that and can thus be shrunken to 32 bit load immediate.

So, the murmur hash and references to readonly-data could collide.

deanmlittle:
> > All deferences for any address under `0x0100000000` are currently invalid
> 
> In [SIMD-0189](https://github.com/solana-foundation/solana-improvement-documents/pull/189) we moved the readonly-data down into the 32 bit address range as most 64 bit load immediate (`lddw`) only want to load addresses to that and can thus be shrunken to 32 bit load immediate.
> 
> So, the murmur hash and references to readonly-data could collide.

It's always a fun time discovering new and exciting reasons why the runtime has continued to trend from "functional" to "borderline unusable" via SIMD trivia. I can guarantee you that devs would unanimously prefer to maintain the viability of the method laid out in this SIMD over curing you of your irrational hatred of lddw. That should be a sign to anyone who takes their end customers seriously to revert the change.

proposals/0503-static-sysvars.md
```diff
@@ -0,0 +1,109 @@
...
+
+```rust
+let lamports_per_byte: u64 = unsafe { *(SOL_RENT_SYSVAR as *const u64) };
+```
+
+This produces the following bytecode:
+
+```asm
+lddw r3, 0x494df715 // SOL_RENT_SYSVAR
+ldxdw r3, [r3+0]
```
Lichtso, line 77: This would benefit from explaining that sysvars are mapped in as byte slices starting at the murmur hash as address. Otherwise it invites the interpretation that all sysvars are supposed to be accessed via a single `u64` value, which would limit them to 8 bytes.

proposals/0503-static-sysvars.md
```diff
@@ -0,0 +1,109 @@
...
+2. Fewer stack allocations
+3. Reduced complexity of calculating rent exemption
+4. Reduced cost of resolving sysvar values to 2 CUs
+
+## Security Considerations
+
+1. We must ensure no intra-slot mutability of any exposed globals.
+2. We must ensure static sysvar pointers remain synchronized with SysvarCache 
+   at the slot boundary.
+3. Despite being 32-bit hashes, it is important that we cast to u64 first as
```
Lichtso: Alternatively one could also only use the 31 LSBs of the hash / mask out the 1 MSB and always load a `i32` immediate value.

LucasSte:
> We are utilizing an address in the 32-bit range (extended to 64 bits with upper 32 bits masked out in this case, as our wonderful Rust/LLVM/BPF doesn't currently have a good way to produce ALU32 instructions without inline assembly)

BPF upstream has the `+alu32` target feature which achieves that. If generating LDDW, instead of mov32, is a problem, please open an issue in our repo, and we'll fix that.

LucasSte:
> It's always a fun time discovering new and exciting reasons why the runtime has continued to trend from "functional" to "borderline unusable" via SIMD trivia.

I don't think this is an appropriate comment for this discussion. You have read and approved the SIMD yourself: https://github.com/solana-foundation/solana-improvement-documents/pull/189#pullrequestreview-3640405451.

Lichtso:
> It's always a fun time discovering new and exciting reasons why the runtime has continued to trend from "functional" to "borderline unusable" via SIMD trivia.

First of all, you approved the SIMD yourself, it did not pass you by unnoticed. You simply, like anybody else, did not have the foresight that there could be more interactions with future ideas.

> I can guarantee you that devs would unanimously prefer to maintain the viability of the method laid out in this SIMD

Devs would prefer any method that achieves the same low cost direct access to sysvars, independent of it being is this proposal or some alternative.

> over curing you of your irrational hatred of lddw.

Which you seem to be infected with too, reciting this very proposal:

"this would be more ideal, as it would save 8 bytes of binary size."

> That should be a sign to anyone who takes their end customers seriously to revert the change.

What if I told you there is an alternative which achieves the same interface and CU costs but does not need to revert SBPFv3?

deanmlittle:
> > We are utilizing an address in the 32-bit range (extended to 64 bits with upper 32 bits masked out in this case, as our wonderful Rust/LLVM/BPF doesn't currently have a good way to produce ALU32 instructions without inline assembly)
> 
> 
> 
> BPF upstream has the `+alu32` target feature which achieves that. If generating LDDW, instead of mov32, is a problem, please open an issue in our repo, and we'll fix that.

Yeah, we can do it with upstream already too. The issue is if we tried to force 32 bits, the feature being safe becomes inherently tied to a compiler version which doesn't sound great. If v3 was a feature branch until the feature gates activated and only shipped to master when it was fully baked, it would have been fine to assume alu32. In absence of that, lddw works just fine here. The main issue is just managing sign extension.

deanmlittle:
>First of all, you approved the SIMD yourself, it did not pass you by unnoticed. You simply, like anybody else, did not have the foresight that there could be more interactions with future ideas.

It has now become apparent why you wanted to sneak in this unrelated design choice into the header restrictions. I wish I had understood your motivation more clearly at the time instead of rubber stamping it after ripping out all of the bloat.

> What if I told you there is an alternative which achieves the same interface and CU costs but does not need to revert SBPFv3?

I'd say you're about to shill me on a new memory address prefix in the upper 32-bits, locking us into 64-bit targets for good. Sad if it has to come to that. Also, was absolutely not suggesting to revert all of V3, just the silly 32-bit mapping change. Ironically, in your quest to kill lddw, you've just made it more useful 🙃

proposals/0503-static-sysvars.md
deanmlittle: That's exactly where we get the memory pointers from.

proposals/0503-static-sysvars.md
deanmlittle: Keep it simple.

proposals/0503-static-sysvars.md
deanmlittle: An emulator of a target is not, in itself, a target. We are building for our JIT and the interpreter is attempting to pretend it is our JIT. There is only one target. I don't think this is worth discussing any further.

proposals/0503-static-sysvars.md
deanmlittle: A better term is probably "memory translation", but the two features are somewhat similar. The key idea is that they are no longer dynamically loaded into the program, but instead become globally available at a specified memory address in the JIT.

deanmlittle:
I have run preliminary benchmarks on an end-to-end implementation of this feature based on `solana-sbpf` version 0.17 in Agave 4.0.0 with a fork of Mollusk. In terms of both CUs and wall clock time averaged over 100,000 iterations using the JIT via Rosetta, with the feature gate disabled on the syscall and sysvar account tests, initial results show that static sysvars are consistently faster than the other two alternatives:

```
=== STATIC SYSVARS (direct memory read) ===
  CUs:     7
  Total:   8.178s
  Average: 81.784µs

=== SYSCALL (sol_get_rent_sysvar) ===
  CUs:     131
  Total:   8.307s
  Average: 83.067µs

=== SYSVAR ACCOUNT (pass rent account) ===
  CUs:     31
  Total:   8.897s
  Average: 88.973µs
```

To isolate the cost of sysvar access from other VM overhead, we can compare against an empty program baseline that simply returns `Ok(())`:

```
=== BASELINE (empty program) ===
  CUs:     2
  Total:   8.118s
  Average: 81.182µs
```

Deducting the baseline amount from the above benchmarks yields the following wall clock times for calculating rent from sysvar values:

| Method | Wall Clock | CUs | Time multiple |
|---|---|---|---|
| Static Sysvars | 0.602µs | 7 | 1x |
| Syscall | 1.885µs | 131 | 3.13x slower |
| Sysvar Account | 7.791µs | 31 | 12.94x slower |

Static sysvars deliver a **~3x wall clock speedup** over the rent syscall and a **~13x improvement** over passing the rent sysvar account while also consuming the fewest compute units at just 7 CUs (2 CUs per memory access) with significant additional benefits for devex, composability and unifying our existing fragmented sysvar APIs.

Lichtso:
Thanks for the experiments. Could you share your benchmark setup and the programs source / assembly? I would like to reproduce these on a Linux machine (to skip the Rosetta layer).

Specifically for the sysvar account case I am curious why it takes more CUs. Is it the pubkey check? Otherwise one could assume a static address (where the account is serialized at) and read from there directly.

Edit: I put together my own mockup and micro benchmark using the VM only, which while crude, should do the job. 6471078cb71fc7cb777a0aaf6c666abf3d44f68a

> unifying our existing fragmented sysvar APIs.

That, unfortunately, would only be the case if we also deprecated and disabled the others.

Lichtso:
I would assume them to look roughly like this:

### static address translation: 6 CUs
According to the proposal this would load a hash value in `r0` instead, but what ever.
```
    lddw r0, 0x000002A00
    ldxdw r1, [r0 + 0]
    ldxdw r2, [r0 + 8]
    ldxdw r3, [r0 + 16]
    ldxdw r4, [r0 + 24]
    exit
```

### dynamic address translation (sysvar account): 6 CUs
Directly loading from the input section only works under the wild assumption that we don't have to check the pubkey of the sysvar account, which might be possible to guarantee.
```
    lddw r0, 0x400002A00
    ldxdw r1, [r0 + 0]
    ldxdw r2, [r0 + 8]
    ldxdw r3, [r0 + 16]
    ldxdw r4, [r0 + 24]
    exit
```

### sysvar syscall: 120 CUs
Have the syscall write to the heap or stack and retrieve it from there.
```
    mov r1, 0x2A
    lddw r2, 0x200000000
    mov r3, 0
    mov r4, 32
    syscall sysvar
    ldxdw r1, [r2 + 0]
    ldxdw r3, [r2 + 16]
    ldxdw r4, [r2 + 24]
    ldxdw r2, [r2 + 8]
    exit
```

deanmlittle:
>Could you share your benchmark setup and the programs source / assembly? I would like to reproduce these on a Linux machine (to skip the Rosetta layer).

Yes, will open source everything after a bit of cleanup. This is just the first round of tests locally before I also run it on our Linux box. If anything, they show static sysvars are unlikely to come at any significant detriment, may have some reasonable performance benefits, and CUs are a ridiculously useless metric. 

> Specifically for the sysvar account case I am curious why it takes more CUs. Is it the pubkey check? Otherwise one could assume a static address (where the account is serialized at) and read from there directly.

The sysvar account takes more CUs because we cannot simply read it from a raw offset. You typically must use an entrypoint which serializes the account pointers somewhere. In the case of Solana's toolchain, that's always to the stack. This obviously isn't free. I didn't check the address at all in this case, but I should add that as a benchmark. I simply used Pinocchio's `Rent::from_account_view_unchecked` because I deemed this additional overhead would act as a confounding variable that make it impossible for the Sysvar account to compete. In reality, this interpretation is actually *far more charitable* than the real-world requirements of safely using the Sysvar account. It should perform even worse than it does.

> Also, did you use the default vm::Config?

I enabled all features and didn't modify anything.

> That, unfortunately, would only be the case if we also deprecated and disabled the others.

If this exists, the other two use cases have no reason to. We can slowly phase them out and stop making two versions of everything pertaining to Sysvars moving forwards. We can also immediately eliminate the syscall in every single existing usecase, or make an intrinsic that pulls it form the newly mapped memory region. We have a lot of options here! ;)

proposals/0503-static-sysvars.md
```diff
@@ -0,0 +1,178 @@
...
+This produces the following bytecode:
+
+```asm
+lddw r3, 0x5494df715 // SOL_RENT_SYSVAR
+ldxdw r3, [r3+0]
+```
+
+### Memory Region
+
+This proposal introduces a new read-only VM memory region for static sysvars:
```
buffalojoec, line 82: Is this memory region always available or is created only for programs who request it? Maybe only for certain SBPF versions?

proposals/0503-static-sysvars.md
```diff
@@ -0,0 +1,178 @@
...
+The following sysvars are supported through the static sysvar interface. All
+sysvars present in the `SysvarCache` are eligible for exposure.
+
+| Sysvar | Canonical Name | murmur32 | Size (bytes) |
+|--------|---------------|----------|--------------|
+| Rent | `SOL_RENT_SYSVAR` | `0x494df715` | 17 |
+| Clock | `SOL_CLOCK_SYSVAR` | `0xff395088` | 40 |
+| EpochSchedule | `SOL_EPOCH_SCHEDULE_SYSVAR` | TBD | 33 |
+| LastRestartSlot | `SOL_LAST_RESTART_SLOT_SYSVAR` | TBD | 8 |
+| EpochRewards | `SOL_EPOCH_REWARDS_SYSVAR` | TBD | 49 |
```
buffalojoec, line 127: Why bother with the last three? Simply including `Rent` and `Clock` should account for 99% of use cases, and they're much simpler data structures to start with.

buffalojoec (commented):
Thanks for the profiling data @deanmlittle. Going off your post-baseline numbers, it's clear we have program execution benefits from static sysvars compared to obtaining them from syscalls, therefore in principle this would be a CU-cheaper option for program devs.

My primary concern with this approach is that we're pushing a bit of work out of the VM and onto the validator by creating a new memory mapping. The work isn't really disappearing. The juicy datapoint would be to see how much work is involved with that extra memory mapping, and maybe consider limiting it to only those who ask.

We do a lot of silly things in the calling convention that we don't properly meter on CUs, so IMO as long as it's not an egregious amount of additional work, this seems like a reasonable feature to support.

proposals/0503-static-sysvars.md
deanmlittle: I agree that most of the benefits are unlocked by just Rent and Clock, however it just so happens that it's an *extremely* negligible amount of work to also support the others as we're not actually copying anything; we're literally just throwing in a pointer to a slice of owned memory once per slot. I don't see a good reason to not support this method globally when there's basically no cost involved. If we were copying, I would have a very different opinion.

proposals/0503-static-sysvars.md
deanmlittle: We can do this. I think it should go into justification though, not detailed design. 

It's quite comical. CUs are actually pretty irrelevant in this case, as it appears the "high CU" syscall outperforms the "low CU" sysvar account, and both lose to the "apparently super duper expensive memory translation" of the static sysvar. I think wall clock tells a much more accurate and interesting story here, but we'll need to tighten up my benchmarks to upgrade the preliminary findings from "hunch" to "evidence". :)

deanmlittle:
> We do a lot of silly things in the calling convention that we don't properly meter on CUs, so IMO as long as it's not an egregious amount of additional work, this seems like a reasonable feature to support.

+1. Looks like it's actually much cheaper. More evidence required.

deanmlittle:
Attached is a working PoC that applies the necessary changes to `agave`, `solana-sbpf` and `mollusk-svm` in order to implement the static sysvars feature: https://github.com/blueshift-gg/static-sysvars

Some specific caveats about benchmarks:

- All benchmarks ran over 100,000 iterations
- All Rent Sysvar accesses were compared to a control program that simply returns `Ok(())` to help weed out inherent VM overhead so we can have a better view on how much of the execution time was accessing the Rent values.
- In all instances where the feature was not used, it is disabled in the FeatureSet so as to not introduce any inherent overhead in the runtime that would tamper with the control cases of the Syscall and Sysvar account.
- In all instances where an account was not loaded in, the pinocchio entrypoint set accounts to 0
- There are two variants of Rent Sysvar account accesses, one using completely unsafe pointer casting that would never be used in reality, and the other using Pinocchio's `Rent::from_account_view` which checks the address of the Rent Sysvar account. Both are using the unreleased master branch of Pinocchio which introduces SIMD-0194-related improvements to Rent calculation so as to give them the fairest chance possible of competing.

You can see from all of the results across the interpreter, JIT (x86) and JIT (Rosetta) that every single benchmark paints the same very clear picture: Not only are static sysvars great for devex, there is not a single metric in which they fail to outperform existing methods of accessing sysvar values: https://github.com/blueshift-gg/static-sysvars/blob/master/BENCHMARKS.md

Would love to have some help from @alessandrod to benchmark this more professionally, as I am certainly not an expert in this area, but as far as I can tell, the preliminary results satisfy the requirements for this to both be IBRL and a major win for developers.

Lichtso:
First of all, thanks for publishing the benchmark setup.

It should be possible to do this without any changes in the VM (https://github.com/blueshift-gg/solana-sbpf/commit/1d0f6d56817514753901bbababaf25d114c65bc4).

@deanmlittle Can you instead add a memory mapping for each sysvar in `create_memory_mapping()` here:
https://github.com/anza-xyz/agave/blob/289aa4ea46889a1535962b727c0656d4d25527dc/program-runtime/src/vm.rs#L110

```Rust
    if invoke_context.get_feature_set().static_sysvars {
        for entry in invoke_context.get_sysvar_cache().get_static_sysvars() {
            regions.push(MemoryRegion::new_readonly(&entry.host_vec, entry.vm_addr));
        }
    }
```
And slightly adjust `StaticSysvarEntry` to `StaticSysvarEntry { vm_addr: u64, host_vec: Vec<u8> }`.

> there is not a single metric in which they fail to outperform existing methods of accessing sysvar values

That seems unrealistic, especially the win over unchecked sysvar accounts when using the interpreter is strange. It is very easy to come up with micro benchmarks to prove basically any argument, but they are often flawed, resulting in confirmation bias.

> We can slowly phase them out and stop making two versions of everything pertaining to Sysvars moving forwards.

Adding features is easy, removing them incredibly hard. Especially if there are already tons of deployed programs using them. Expect this to take years.

> for this to both be IBRL and a major win for developers.

I agree that direct access for sysvars is a win for developers. However, for this to be IBRL, it is not sufficient to just find a faster solution. There must be a problem in the first place. Thus, the interesting metric would be **how many percent of the time a validator spends on average MNB traffic for serving sysvars to programs** with the various existing paths.

I am sure we can find a lot of places where the code could be a lot faster. Yet, even if we can make a specific thing 1000x faster, it might not matter in the end if that thing was not a performance bottleneck in the first place. IBRL is about finding and fixing the bottlenecks.

proposals/0503-static-sysvars.md
deanmlittle: Also worth nothing, it's not like I "gave this a CU cost" so much as that's what it costs to read 64 bits of memory.

proposals/0503-static-sysvars.md
deanmlittle: Always available for V3 or newer. Only populated are feature activation. All reads to this region fail if the sysvar entry length is set to zero which is its default. There is no reason why it couldn't also be available to V0, but by the time this goes live that will probably be irrelevant.

deanmlittle:
>First of all, thanks for publishing the benchmark setup.

You're welcome. These things are always very hard to get right with so many moving parts. I would have published it immediately but I wanted to make sure it was easy for everyone to use to fact check my work.

> It should be possible to do this without any changes in the VM (https://github.com/blueshift-gg/solana-sbpf/commit/1d0f6d56817514753901bbababaf25d114c65bc4).
>
> @deanmlittle Can you instead add a memory mapping for each sysvar in create_memory_mapping() here:

Sure, I could, and I'm sure it would also work, but I don't think that's a better solution than what I'm proposing. Introducing a `Vec` to `create_vm` incurs a per `create_vm` cost (not just per top level instructions, but also per CPI), whereas introducing a pointer to the array of pointers in the SysvarCache only incurs a per slot cost. While small on the scale of a single invocation, this adds up to create the exact kind of overhead that this SIMD is trying to avoid. I am ambivalent about how this is implemented, but it is my strong preference for us to allocate once per slot and consume multiple times rather than allocate once per create_vm.

>That seems unrealistic, especially the win over unchecked sysvar accounts when using the interpreter is strange. It is very easy to come up with micro benchmarks to prove basically any argument, but they are often flawed, resulting in confirmation bias.

While I do agree that it's strange, I don't think it's unrealistic at all. The data doesn't lie. Objectively in this case: Static sysvar > syscall > unchecked sysvar account > checked sysvar account. I would argue that unchecked sysvar account is not even a valid benchmark considering checking the Pubkey is the bare minimum to safely consume it. If anything, these benchmarks actually point to a high likelihood of some low hanging fruit in our memory access/translation logic that could do with some optimization.

>I agree that direct access for sysvars is a win for developers. However, for this to be IBRL, it is not sufficient to just find a faster solution.

Sure it is! What a pessimistic way of thinking about engineering problems! Anything that lowers CU costs in general means more transactions in blocks. Just killing Rent/Clock syscall invocations alone probably saves us >50k CUs per block, which is ~1,000 additional prop AMM oracle updates. Small wins add up over time!

deanmlittle:
>It should be possible to do this without any changes in the VM (https://github.com/blueshift-gg/solana-sbpf/commit/1d0f6d56817514753901bbababaf25d114c65bc4).

>@deanmlittle Can you instead add a memory mapping for each sysvar in create_memory_mapping() here:
https://github.com/anza-xyz/agave/blob/289aa4ea46889a1535962b727c0656d4d25527dc/program-runtime/src/vm.rs#L110

~Okay, so I tried this. In interpreter, static sysvars still win, but as predicted above, in the JIT, the per `create_vm` changes results in a slight performance penalty making static sysvars slightly slower than the syscall, but still *much* faster than the sysvar account. JIT results below:~

~Static Sysvars: 0.168 (1.87x)~
~Syscall: 0.090 (1.00x) // new winner~
~Sysvar Account (unchecked): 0.463 (5.14x)~
~Sysvar Account (checked): 0.596 (6.62x)~

I have implemented it similarly to your suggestion and managed to optimize it down to essentially a rounding error difference between static sysvars and the syscall.

I would love to find a way to make this a per-slot cost, as I don't think it makes sense to have to do setup every single time we invoke `create_vm` for something that's only updated once per slot, but at least the bulk of the heavy lifting is done by SysvarCache.

proposals/0503-static-sysvars.md
```diff
@@ -0,0 +1,178 @@
...
+const MM_STATIC_SYSVARS: u64 = 0x05 << 32;
+// murmur3(SOL_RENT_SYSVAR) = 0x494df715
+const SOL_RENT_SYSVAR: *const u8 =
+    (0x494df715u64 & 0xffffffff | MM_STATIC_SYSVARS)
+    as *const u8;
+// murmur3(SOL_CLOCK_SYSVAR) = 0xff395088
+const SOL_CLOCK_SYSVAR: *const u8 =
+    (0xff395088u64 & 0xffffffff | MM_STATIC_SYSVARS)
+    as *const u8;
+```
```
topointon-jump, line 65: The `murmur3` scheme feels a bit too complicated - could we instead map each sysvar account into a predictable aligned offset in the 0x5 region? This would be inexpensive with direct mapping. Something like:

E.g.

```
  0x501000000 - Rent
  0x502000000 - Clock
  0x503000000 - EpochSchedule
  0x504000000 - LastRestartSlot
  0x505000000 - EpochRewards
```

This would make memory translation simple: `haddr = sysvar_haddr[ offset >> 24 ] + ( offset & 0xFFFFFF )`.

We should strive to make the memory mapping as simple and straightforward as possible - this will result in a hardware-friendly VM implementation with good performance characteristics.

proposals/0503-static-sysvars.md
```diff
@@ -0,0 +1,178 @@
...
+1. Sysvar values are globals that are always available to the validator, but 
+   aren't exposed as globals during execution. This is a clunky anti-pattern 
+   resulting in degraded developer experience and performance.
+2. Invoking a syscall to access a Sysvar requires both a stack allocation and 
+   halting execution – an immense amount of overhead just to read a value that 
+   is already readily available.
+3. Having to pass in an account to access a Sysvar results in a >10KiB penalty 
+   to data serialization and has negative implications for composability.
+
+If these globals were simply exposed to the VM, as is common in kernel BPF,
```
topointon-jump, line 38: Any solution which affects the memory mapping needs to be benchmarked against current (and synthetic) mainnet load - not just against the alternative methods of accessing. If the memory translation gets slower this will affect replay times, despite having no impact on CUs.

deanmlittle: The net effect is that this actually *lowers* CUs, however I'm sure you'd agree as even my microbenchmarks show, CUs are an almost completely arbitrary metric with regards to wall clock. You'll find no disagreements from me here re: benchmarking with real load!

proposals/0503-static-sysvars.md
deanmlittle: We can do that, but then you run into an ugly enum problem where any SVM fork that adds their own sysvar at 0x50600000 or beyond will get rugged if we were to ever add our own additional sysvar at address 0x50600000 which causes deployed programs on each chain to have subtle but potentially fatal divergences. I don't think this should be our number one priority, but if it doesn't cost us anything, we should consider it. We actually reached [the same conclusion with static syscalls](https://github.com/solana-foundation/solana-improvement-documents/pull/178#discussion_r1927605962) as well. The idea was to have a high probability that no two things would collide. I'm not sure whether alignment is important in this case since we're just using this address to translate a pointer. Isn't the important thing that the pointer itself is aligned?

riptl:
Simplifications to sysvar accesses are good but the proposed design is not, mainly due to the Murmur3 setup. Virtual-to-host address translation is _the most performance-critical_ operation in the validator (runs up to 10 billion times a second across the validator). Unfortunately any new region we add to the memory map slightly slows down the VM due to the way the current mapping scheme (ABIv1) works.

The current mapping scheme's worst aspect is that it allows _unaligned_ access privilege changes, whereas any modern CPU requires 4K alignment. 

This is not a good trade: >99% of memory accesses incur more complexity and CUs per block are reduced by less than <1%, and we haven't quantified the mainnet execution speed impact.

I think additional unaligned mapped data will almost certainly slow down real mainnet execution by a bit (memory accesses to everything but sysvars) and complicate future optimizations to Agave's and Firedancer's software MMU.
ABIv2 then forces regions to be aligned, which breaks Murmur3-based addressing regardless.

The existing microbenchmarks only measure the happy cases (the areas that this SIMD improves), as far as I can tell.
The best possible benchmark would be to run a representative ledger (mainnet-like program invocations) with and without the feature against an in-memory accounts DB and shred store.

The design would be better if you did these two changes, IMHO.
- Align sysvars to the mappign scheme's native alignment (2^32 for ABIv1, less for ABIv2) 
- Provide a syscall that tells users the address of a sysvar for future upgradability (instead of hardcoding offsets in programs)
Although I'm not sure whether this proposal is worth the implementation cost even with those.

Few more things to note:
- We can also reprice the get_sysvar syscall to make it cheaper
- Impact to ZK proving performance of SVM worth considering

PR #504, opened 2026-03-26
SIMD-0504: Stricter Shred Validation

ptaffet-jump:
Specifies shred validation criteria more precisely and eliminates some confusing corner cases.

### DIFF
--- a/proposals/0504-stricter-shred-validation.md
+++ b/proposals/0504-stricter-shred-validation.md
@@ -45,6 +45,10 @@ This proposal depends on the following previously accepted proposals:
 
 ## New Terminology
 
+This SIMD defines the constant `MAX_SHREDS_PER_SLOT` which is currently
+set to `32,768 = 2^15`.  This constant itself is not new, but this SIMD
+gives it a protocol-level name.
+
 
 ## Detailed Design
 
@@ -62,10 +66,10 @@ second family of constraints.
 The checks performed on a shred are revised to the following:
 
 1. The shred is validly signed
-2. `shred_index < 32,768`
+2. `shred_index < MAX_SHREDS_PER_SLOT`
 3. `fec_set_index <= shred_index < fec_set_index + 32`
 4. `fec_set_index` is a multiple of 32
-5. `fec_set_index <= 32,736`
+5. `fec_set_index <= MAX_SHREDS_PER_SLOT - 32`
 6. `variant` is chained or chained-resigned
 7. Merkle height MUST be equal to 6
 
@@ -92,13 +96,13 @@ immediately ignored or discarded.
 
 
 For reference, the existing checks are the following, with differences
-from the new checks at the end.
+from the new checks noted at the end.
 
 1. The shred is validly signed
-2. `shred_index < 32,768`
+2. `shred_index < MAX_SHREDS_PER_SLOT`
 3. `fec_set_index <= shred_index < fec_set_index + 32`
 4. `fec_set_index` is a multiple of 32
-5. `fec_set_index <= 32,736`
+5. `fec_set_index <= MAX_SHREDS_PER_SLOT - 32`
 6. `variant` is chained or chained-resigned
 
 For data shreds:
@@ -117,7 +121,7 @@ For coding shreds:
 13. `position < 32`
 14. `position <= shred_index`
 15. `num_data_shreds == 32` and `num_coding_shreds == 32`
-16. `shred_index - position <= 32,736`
+16. `shred_index - position <= MAX_SHREDS_PER_SLOT - 32`
 
 Note: A proof that the existing checks, excluding the signature check,
 are equivalent (necessary and sufficient) to those enforced by the
@@ -126,6 +130,8 @@ available [here](https://github.com/firedancer-io/agave/blob/ptaffet/shred-proof
 
 Check 7 is new.  Check 12 is just updated with the fixed Merkle height.
 Check 16 is new, but combined with check 5 implies the old check 16.
+Check 5 is implied by a combination of checks 2, 3, and 4, but is
+included explicitly for clarity.
 
 ### Family 2: FEC-level checks
 
@@ -155,12 +161,13 @@ repair or retransmitted via Turbine.
 
 ### A note on recovery
 
-The checks in Family 2 apply to both received and recovered
-shreds.  This implies that if a validator receives all needed data
-shreds it must still perform the reconstruction process and ensure that
-all shreds are valid before voting on the block.  This extra work is
-necessary to eliminate the possibility of divergence based on whether a
-validator received or recovered a shred.
+The checks in Family 2 apply to both received and recovered shreds.
+This implies that even if a validator receives all needed data shreds it
+MUST ensure the coding shreds are also valid according to this SIMD, and
+it MUST NOT vote on the block if the FEC set fails the check in family
+2.
+This extra work is necessary to eliminate the possibility of divergence
+based on whether a validator received or recovered a shred.
 
 
 ## Alternatives Considered
--- a/proposals/0504-stricter-shred-validation.md
+++ b/proposals/0504-stricter-shred-validation.md
@@ -154,7 +154,12 @@ is activated.
 5. All shreds in a FEC set MUST have the same shred version
 6. All shreds in a FEC set MUST have the same `fec_set_index`
 7. All data shreds in a FEC set MUST have the same `parent_offset`
-8. All shreds in a FEC set MUST satisfy all the constraints in Family 1
+8. If any data shred does not have the maximum value of the `size` field as
+   determined by check 12 in the first family, then all shreds in the
+   same FEC set with a larger shred_index MUST have the minimum value of
+   `size` (as determined by check 12), and the flags field of final
+   shred in the FEC set MUST have the batch complete bit set.
+9. All shreds in a FEC set MUST satisfy all the constraints in Family 1
 
 Under TowerBFT, a validator MUST NOT vote for a block containing any FEC
 set unless it satisfies all of these checks.  Under Alpenglow, a
--- a/proposals/0504-stricter-shred-validation.md
+++ b/proposals/0504-stricter-shred-validation.md
@@ -154,6 +154,7 @@ shreds for constructing equivocation proofs, and they MAY be served over
 repair or retransmitted via Turbine.
 
 ### A note on recovery
+
 The checks in Family 2 apply to both received and recovered
 shreds.  This implies that if a validator receives all needed data
 shreds it must still perform the reconstruction process and ensure that
--- a/proposals/0504-stricter-shred-validation.md
+++ b/proposals/0504-stricter-shred-validation.md
@@ -144,8 +144,7 @@ is activated.
 
 1. All shreds in a FEC set MUST have the same signature
 2. All shreds in a FEC set MUST have the same chained Merkle root
-3. All shreds in a FEC set MUST have the same {chained/unchained} and {retransmitter
-   signed/retransmitter unsigned} status
+3. All shreds in a FEC set MUST have the same shred type
 4. All shreds in a FEC set MUST have the same `slot`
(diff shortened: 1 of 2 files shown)


proposals/0504-stricter-shred-validation.md
```diff
@@ -0,0 +1,197 @@
...
+
+
+For reference, the existing checks are the following, with differences
+from the new checks at the end.
+
+1. The shred is validly signed
+2. `shred_index < 32,768`
+3. `fec_set_index <= shred_index < fec_set_index + 32`
+4. `fec_set_index` is a multiple of 32
+5. `fec_set_index <= 32,736`
```
AshwinSekar: I don't believe we actually check this in agave. It's enforced indirectly via (2) and (3)

proposals/0504-stricter-shred-validation.md
```diff
@@ -0,0 +1,197 @@
...
+MUST NOT vote for a block containing a FEC set failing to satisfy the
+second family of constraints.
+
+
+### Family 1: Shred-level checks
+
+The checks performed on a shred are revised to the following:
+
+1. The shred is validly signed
+2. `shred_index < 32,768`
```
AshwinSekar: we are planning on adjusting the shred limit in the future. It might be clearer to define a protocol constant MAX_SHREDS_PER_SLOT and say that it's = 32,768 so that future SIMDs can simply refer to the constant

proposals/0504-stricter-shred-validation.md
```diff
@@ -0,0 +1,197 @@
...
+satisfies all of these checks.  A validator SHOULD NOT replay the
+transacactions in a FEC set unless it satisfies all of these checks.
+Notwithstanding the Family 1 restrictions, a validator MAY use these
+shreds for constructing equivocation proofs, and they MAY be served over
+repair or retransmitted via Turbine.
+
+### A note on recovery
+
+The checks in Family 2 apply to both received and recovered
+shreds.  This implies that if a validator receives all needed data
```
AshwinSekar: To be clear here:
1) We receive 32 data shreds
2) we MUST recover the 32 coding shreds and verify that they are valid
3) Then we vote

If (2) fails then we must not cast our vote, is that correct?

proposals/0504-stricter-shred-validation.md
ptaffet-jump: Good idea!

proposals/0504-stricter-shred-validation.md
ptaffet-jump: I'll note this. It's actually a combination of (2), (3), and (4), at least for data shreds.  For coding shreds, it's enforced a bit more directly.

proposals/0504-stricter-shred-validation.md
ptaffet-jump: Correct. I'll clarify the text further.

AshwinSekar: cc @steviez @alexpyattaev this is new behavior that we'll need to address in our implementation.

Happy to chat in another venue but the TLDR:
- We never attempt to recover if we have 32 data sheds but < 32 coding shreds
https://github.com/anza-xyz/agave/blob/7d3289be4d2e8b1ed234724546527c3e4c948c0b/ledger/src/blockstore_meta.rs#L600-L602
- The reason for this is:
  - Recovered coding shreds are not inserted into blockstore (considered it wasteful). They are only used to retransmit to our turbine children
  - Since we don't insert anything, we have no idea if we've already done recovery in a prior iteration
  - If we turn on recovery for this case we could end up doing the recovery retransmit loop 32 times for each FEC set

I think to address this we'll want to store some information in blockstore that indicates that erasure recovery has been done.

AshwinSekar (approved):
looks good, having these clear rules will let us cleanup some existing code and reason about corner cases better.

proposals/0504-stricter-shred-validation.md
```diff
@@ -0,0 +1,204 @@
...
+   signed/retransmitter unsigned} status
+4. All shreds in a FEC set MUST have the same `slot`
+5. All shreds in a FEC set MUST have the same shred version
+6. All shreds in a FEC set MUST have the same `fec_set_index`
+7. All data shreds in a FEC set MUST have the same `parent_offset`
+8. All shreds in a FEC set MUST satisfy all the constraints in Family 1
+
+A validator MUST NOT vote for a block containing any FEC set unless it
+satisfies all of these checks.  A validator SHOULD NOT replay the
+transacactions in a FEC set unless it satisfies all of these checks.
```
lidatong:
```suggestion
transactions in a FEC set unless it satisfies all of these checks.
```

proposals/0504-stricter-shred-validation.md
```diff
@@ -0,0 +1,204 @@
...
+Notwithstanding the Family 1 restrictions, a validator MAY use these
```
lidatong: nit but maybe "Unlike the Family 1 restrictions" is clearer vocab?

proposals/0504-stricter-shred-validation.md
```diff
@@ -0,0 +1,204 @@
...
+shreds for constructing equivocation proofs, and they MAY be served over
+repair or retransmitted via Turbine.
```
lidatong:
```suggestion
Notwithstanding the Family 1 restrictions, a validator SHOULD use these
shreds for constructing equivocation proofs, and they SHOULD be served over
repair or retransmitted via Turbine.
```

perhaps SHOULD here is more appropriate instead of MAY to encourage reporting of a bad leader?

proposals/0504-stricter-shred-validation.md
```diff
@@ -0,0 +1,204 @@
...
+cases, that is equivalent to identifying a FEC set by a common
+signature, but a malicious leader with a specifically crafted key may be
+able to produce multiple valid signatures for the same Merkle root after
+[SIMD-0376](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0376-verify-strict.md)
+is activated.
+
+1. All shreds in a FEC set MUST have the same signature
+2. All shreds in a FEC set MUST have the same parent Merkle root
+3. All shreds in a FEC set MUST have the same {chained/unchained} and {retransmitter
+   signed/retransmitter unsigned} status
```
lidatong: status here meaning shred type?

proposals/0504-stricter-shred-validation.md
```diff
@@ -0,0 +1,204 @@
...
+superseded-by:
+extends:
+---
+
+## Summary
+
+Shred validation currently allows certain shreds that seem like they
+should be invalid.  This leads to lots of corner cases, where a new
+client base their behavior on what Agave implicitly allows rather than
+easily specified behavior.
```
lidatong:
```suggestion
Shred validation currently allows certain shreds that seem like they
should be invalid.  This leads to lots of corner cases when a new
client bases their behavior on what Agave implicitly allows rather than
easily specified behavior. This SIMD provides the spec.
```

proposals/0504-stricter-shred-validation.md
```diff
@@ -0,0 +1,204 @@
...
+
+Below, a FEC set is identified by the Merkle root.  In all standard
+cases, that is equivalent to identifying a FEC set by a common
+signature, but a malicious leader with a specifically crafted key may be
+able to produce multiple valid signatures for the same Merkle root after
+[SIMD-0376](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0376-verify-strict.md)
+is activated.
+
+1. All shreds in a FEC set MUST have the same signature
+2. All shreds in a FEC set MUST have the same parent Merkle root
```
lidatong:
```suggestion
2. All shreds in a FEC set MUST have the same chained Merkle root
```

proposals/0504-stricter-shred-validation.md
ptaffet-jump: Good point

proposals/0504-stricter-shred-validation.md
ptaffet-jump: I clarified this.

proposals/0504-stricter-shred-validation.md
```diff
@@ -0,0 +1,203 @@
...
+### Family 1: Shred-level checks
+
+The checks performed on a shred are revised to the following:
+
+1. The shred is validly signed
+2. `shred_index < MAX_SHREDS_PER_SLOT`
+3. `fec_set_index <= shred_index < fec_set_index + 32`
+4. `fec_set_index` is a multiple of 32
+5. `fec_set_index <= MAX_SHREDS_PER_SLOT - 32`
+6. `variant` is chained or chained-resigned
```
bw-solana: we should clarify what this means

proposals/0504-stricter-shred-validation.md
```diff
@@ -0,0 +1,203 @@
...
+7. Merkle height MUST be equal to 6
+
+For data shreds:
+
+8. `data_flags & 0xC0 != 0x80` (block complete implies batch complete)
+9. If `data_flags` has batch complete, then `shred_index % 32 == 31`
+10. `parent_offset <= slot`
+11. For slot 0 only, `parent_offset` MUST be equal to 0.  Otherwise,
+   `parent_offset` MUST NOT be equal to 0.
+12. For chained data shreds, `88 <= size <= 1051`.  For chained-resigned
```
bw-solana: clarify unit is bytes?

proposals/0504-stricter-shred-validation.md
```diff
@@ -0,0 +1,203 @@
...
+  data shreds, `88 <= size <= 987`.
+
+For coding shreds:
+
+13. `position < 32`
+14. `position <= shred_index`
+15. `num_data_shreds == 32` and `num_coding_shreds == 32`
```
bw-solana, line 92: yuck

proposals/0504-stricter-shred-validation.md
```diff
@@ -0,0 +1,203 @@
...
+
+For reference, the existing checks are the following, with differences
+from the new checks noted at the end.
+
+1. The shred is validly signed
+2. `shred_index < MAX_SHREDS_PER_SLOT`
+3. `fec_set_index <= shred_index < fec_set_index + 32`
+4. `fec_set_index` is a multiple of 32
+5. `fec_set_index <= MAX_SHREDS_PER_SLOT - 32`
+6. `variant` is chained or chained-resigned
```
bw-solana: maybe add an explicit empty 7 here? I thought it was a mistake at first

proposals/0504-stricter-shred-validation.md
```diff
@@ -0,0 +1,203 @@
...
+1. All shreds in a FEC set MUST have the same signature
+2. All shreds in a FEC set MUST have the same chained Merkle root
+3. All shreds in a FEC set MUST have the same shred type
+4. All shreds in a FEC set MUST have the same `slot`
+5. All shreds in a FEC set MUST have the same shred version
+6. All shreds in a FEC set MUST have the same `fec_set_index`
+7. All data shreds in a FEC set MUST have the same `parent_offset`
+8. All shreds in a FEC set MUST satisfy all the constraints in Family 1
+
+A validator MUST NOT vote for a block containing any FEC set unless it
```
bw-solana: maybe clarify MUST NOT vote in the affirmative* or something.
i.e. we want to allow validators to vote skip for Alpenglow

proposals/0504-stricter-shred-validation.md
ptaffet-jump: I was talking about the `size` field of the data header. It is measured in bytes, but for the purposes of this point, it's basically just a number. Do you think this clarifies it?

proposals/0504-stricter-shred-validation.md
ptaffet-jump: Yeah, it's getting close to time for a new shred format...

bw-solana (approved):

lidatong (approved):

PR #506, opened 2026-03-27
SIMD-506: Upgrade Agave secp256k1 syscall precompile to k256 (Agave-specific)

samkim-crypto:
Proposal to upgrade agave secp256k1 syscall and precompile implementation from using libsecp256k1, which is unmaintained, to the k256 crate.

This is an agave specific upgrade, which we hope to do behind a feature gate. We already had discussions with @0x0ece, @drubin-fd, and @jacobcreech. We'd like to express appreciation to the Firedancer team for their flexibility in accommodating an Agave-specific feature gate.

### DIFF
--- a/proposals/0505-upgrade-agave-secp256k1-to-k256.md
+++ b/proposals/0505-upgrade-agave-secp256k1-to-k256.md
@@ -1,6 +1,6 @@
 ---
-simd: "0505"
-title: Upgrade Agave secp256k1 syscall precompile to k256 (Agave-specific)
+simd: "0506"
+title: Upgrade secp256k1 to k256 (Agave-specific)
 authors:
   - Sam Kim
   - Zhenfei Zhang


samkim-crypto:
@t-nelson, @steviez, @bw-solana, can we get an approval on this from the anza side?

proposals/0505-upgrade-agave-secp256k1-to-k256.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Motivation
+
+Currently, Agave utilizes the `libsecp256k1` crate for the `secp256k1`
+precompile and syscall implementation. Because this crate is no longer actively
+maintained, a transition to the `k256` crate has been under consideration for
+some time.
+
+However, swapping out cryptographic dependencies carries inherent consensus
+risks; any differing behavior between the crates could result in a network fork.
```
t-nelson, line 28: i understand the sentiment but we have to find a better way... this is an agave implementation detail, not protocol. we should not be breaking consensus on all client implementations for our own convenience

samkim-crypto: Yeah I agree that we have to find a better way. I personally believe that the test vectors in `solfuzz-agave` are pretty comprehensive, but it is often not enough to convince others.

For this SIMD and technically for any cryptographic upgrades, it is not that we are not breaking consensus or making any protocol changes. The whole point is to preserve the protocol behavior. The feature gate here is just there as an extra safety measure.

bw-solana: we should put these 3 precompiles onchain (or into syscalls)

t-nelson: > we should put these 3 precompiles onchain (or into syscalls)

yes, but that doesn't solve this problem. we'd likely still be using the same dependencies regardless of how they're exposed to vm

topointon-jump: the solution here is to use the differential testing infrastructure that we have spent 2 years developing, that can get near total test coverage on upgrades like this

PR #508, opened 2026-03-31
Remove mentions to the verifier from SIMD-178

LucasSte:
We had decided to wind down the byte code verifier from the protocol, starting with sbpfv3, but I left a reference to it in the SIMD that should not exist.

That part is not implemented and is not part of the proposal.


Lichtso (approved):

0x0ece (approved):

LucasSte:
@bw-solana or @t-nelson would you mind checking this PR?

It doesn't change the functionality to end users in any way.

bw-solana (approved):

PR #512, opened 2026-04-03
SIMD-0512: Sha512 Syscall

deanmlittle:


t-nelson:
this seems fine for the sake of consistency with the other syscalls. i do wonder if they were short-sighted in api design (uncharacteristic, i know) and should have been implemented as separate `sol_hashfn_update` and `sol_hashfn_finalize` syscalls tho?

deanmlittle:
> this seems fine for the sake of consistency with the other syscalls. i do wonder if they were short-sighted in api design (uncharacteristic, i know) and should have been implemented as separate `sol_hashfn_update` and `sol_hashfn_finalize` syscalls tho?

There are *many* much bigger fish to fry with this kind of stupidity. Like, why did the `create_with_seed` function put the randomness into the *first* round of sha256 instead of the *second* round so grinding a vanity address could skip one entire round of Sha256 by grinding from a midstate. If you're looking for smart ideas regarding API, consider looking elsewhere. 😛

deanmlittle:
@samkim-crypto @0x0ece requesting stamps please 🙏

Rexicon226 (approved):
Looks good. Generally though, I do agree that we may want to generalize the API a bit more.

samkim-crypto (approved):

proposals/0512-sha512-syscall.md
```diff
@@ -0,0 +1,84 @@
...
+- Not all bytes in `[bytes, bytes + bytes_len * sizeof(SolBytes))` are
+  readable.
+- Not all bytes in each slice `[bytes[i].addr, bytes[i].addr + bytes[i].len)`
+  are readable.
+- Not all bytes in `[result, result + 64)` are writable.
+- `bytes_len` exceeds the configured maximum number of slices.
+
+### Compute Unit Usage
+
+Compute costs follow the same model and parameters as `sol_sha256`.
```
bw-solana: typically try to make SIMDs self contained. should inline CU cost here or link out to something

proposals/0512-sha512-syscall.md
```diff
@@ -0,0 +1,84 @@
...
+a single contiguous input, writing the 64-byte digest to `result`.
+
+The syscall aborts the virtual machine if any of these conditions are true:
+
+- Not all bytes in `[bytes, bytes + bytes_len * sizeof(SolBytes))` are
+  readable.
+- Not all bytes in each slice `[bytes[i].addr, bytes[i].addr + bytes[i].len)`
+  are readable.
+- Not all bytes in `[result, result + 64)` are writable.
+- `bytes_len` exceeds the configured maximum number of slices.
```
bw-solana: we should say more here about max bytes configuration

proposals/0512-sha512-syscall.md
```diff
@@ -0,0 +1,84 @@
...
+## New Terminology
+
+N/A
+
+## Detailed Design
+
+### Syscall Signature
+
+The syscall follows the same interface as `sol_sha256`, `sol_keccak256`,
+and `sol_blake3`.
```
bw-solana: let's specify or link out to ABI here

proposals/0512-sha512-syscall.md
```diff
@@ -0,0 +1,84 @@
...
+The syscall computes SHA-512 over the provided byte slices as if they were
+a single contiguous input, writing the 64-byte digest to `result`.
+
+The syscall aborts the virtual machine if any of these conditions are true:
+
+- Not all bytes in `[bytes, bytes + bytes_len * sizeof(SolBytes))` are
+  readable.
+- Not all bytes in each slice `[bytes[i].addr, bytes[i].addr + bytes[i].len)`
+  are readable.
+- Not all bytes in `[result, result + 64)` are writable.
```
bw-solana, line 52: how about 0 byte input? or overlapping input/output regions?

bw-solana (commented):
concept ack

I'm a fan of making SIMD self contained as opposed to assuming the interface from something existing. It's fine if this comes in the form of a link, but I like when the SIMD is the master index of everything needed to implement from scratch

proposals/0512-sha512-syscall.md
deanmlittle: the number is currently a completely arbitrary constant of 20,000. Would you rather me say "20,000" or link to the existing implementation?

proposals/0512-sha512-syscall.md
deanmlittle: 0 byte inputs are valid. The algorithm just runs against an array of zeroes. This is documented in the Sha512 standard itself. The main reason we specify anything to do with arrays of slices is because it's out way of avoiding having a hash::update and hash::finalize syscall. I feel like this would be self explanatory to any cryptographer. I have never tried to hash an overlapping input but I don't see why it wouldn't work as under the hood it is just calling update which is a copy.

proposals/0512-sha512-syscall.md
bw-solana: maybe just a (currently 20,000) at the end here

proposals/0512-sha512-syscall.md
deanmlittle: Done.

proposals/0512-sha512-syscall.md
deanmlittle: Done

proposals/0512-sha512-syscall.md
deanmlittle: Done

bw-solana (approved):

ripatel-fd:
Would have priced this higher than SHA256 since it's more expensive on x86

deanmlittle:
> Would have priced this higher than SHA256 since it's more expensive on x86

If we're making this argument, it actually makes more sense to make SHA256 cheaper than the others rather than making SHA512 more expensive, as Sha256 is uniquely faster due to SHA extensions. Keccak and Blake are priced the same as SHA256 despite lacking intrinsic support, and SHA512 is faster than non-intrinsic SHA256 despite producing a larger hash.

ripatel-fd:
> If we're making this argument, it actually makes more sense to make SHA256 cheaper than the others rather than making SHA512 more expensive, as Sha256 is uniquely faster due to SHA extensions. Keccak and Blake are priced the same as SHA256 despite lacking intrinsic support, and SHA512 is faster than non-intrinsic SHA256 despite producing a larger hash.

Very good point, will look into this

PR #514, opened 2026-04-08
SIMD-0186: Define fee-only semantics

2501babe:
this has been discussed internally and the concept has firedancer buy-in, we left fee-only behavior underspecified since this simd was originally written before fee-only transactions were enabled. there is presently a weird quirk where the raw (pre-186) sizes of rollback accounts are used for LTDS. this is a straightforwardly sane and normal improvement

i defined fee/nonce relaxation behavior as unspecified as these implementations are stalled pending nonce removal and it isnt immediately obvious what to do with them

### DIFF
--- a/proposals/0186-loaded-transaction-data-size-specification.md
+++ b/proposals/0186-loaded-transaction-data-size-specification.md
@@ -89,13 +89,16 @@ Transactions may include a
 a lower data size limit for the transaction. Otherwise, the default limit is
 64MiB (`64 * 1024 * 1024` bytes).
 
-If a transaction loading failure occurs due to exceeding its requested data size
-limit, its loaded transaction data size is defined as that requested data size
-limit.
+If transaction account loading fails by exceeding the requested data size limit,
+the final loaded transaction data size is defined as the requested data size
+limit itself. This ensures that account load order is *not* part of
+consensus: validator clients may load accounts in any order, in serial or
+parallel, and arrive at the same failure result.
 
-If transaction account loading completes successfully, but a transaction loading
-failure occurs subsequently due to the rules defined in "Sidebar: Program ID and
-Loader Validation," its calculated loaded transaction data size is used.
+*After* all transaction accounts have been loaded, the actual loaded size as
+defined by the the four rules of "Detailed Design" above is used for the
+transaction. This means that a post-account loading failure, such as described
+by "Sidebar: Program ID and Loader Validation," reports the actual loaded size.
 
 The loaded transaction data size of a committed no-op transaction, as would be
 introduced by
--- a/proposals/0186-loaded-transaction-data-size-specification.md
+++ b/proposals/0186-loaded-transaction-data-size-specification.md
@@ -89,11 +89,20 @@ Transactions may include a
 a lower data size limit for the transaction. Otherwise, the default limit is
 64MiB (`64 * 1024 * 1024` bytes).
 
-If a transaction exceeds its data size limit, a loading failure occurs. This
-SIMD does not change any aspect of how such a failure is handled. At time of
-writing, such a transaction would be excluded from the ledger. When
-`enable_transaction_loading_failure_fees` is enabled, it will be written to the
-ledger and charged fees as a processed, failed transaction.
+If a transaction loading failure occurs due to exceeding its requested data size
+limit, its loaded transaction data size is defined as that requested data size
+limit.
+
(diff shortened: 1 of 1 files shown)


2501babe:
cc @bw-solana and @ptaffet-jump as you were part of the discussion and have voting authority, also cc @mjain-jump and @apfitzge for being closer to the code in question

apfitzge (approved):

proposals/0186-loaded-transaction-data-size-specification.md
```diff
@@ -89,11 +89,20 @@ Transactions may include a
...
 64MiB (`64 * 1024 * 1024` bytes).
 
-If a transaction exceeds its data size limit, a loading failure occurs. This
-SIMD does not change any aspect of how such a failure is handled. At time of
-writing, such a transaction would be excluded from the ledger. When
-`enable_transaction_loading_failure_fees` is enabled, it will be written to the
-ledger and charged fees as a processed, failed transaction.
+If a transaction loading failure occurs due to exceeding its requested data size
+limit, its loaded transaction data size is defined as that requested data size
+limit.
```
bw-solana: I'm of the opinion that we should file a standalone SIMD for a change like this as opposed to amending one that was merged over a year ago.

but I also acknowledge the overhead that comes with this and am okay looking the other way here

bw-solana (approved):

proposals/0186-loaded-transaction-data-size-specification.md
2501babe: i wasnt entirely sure about this and did it this way not as a matter of expediency but because i believed keeping simds current with current behavior would be preferred over letting them be wrong unless the reader happens to know there are newer simds that supersede them. im fine closing and reopening if we think the other way is better

topointon-jump (approved):
Looks good to me (assuming @ptaffet-jump and @mjain-jump are ok with it!)

proposals/0186-loaded-transaction-data-size-specification.md
```diff
@@ -89,11 +89,20 @@ Transactions may include a
 a lower data size limit for the transaction. Otherwise, the default limit is
 64MiB (`64 * 1024 * 1024` bytes).
 
-If a transaction exceeds its data size limit, a loading failure occurs. This
-SIMD does not change any aspect of how such a failure is handled. At time of
-writing, such a transaction would be excluded from the ledger. When
-`enable_transaction_loading_failure_fees` is enabled, it will be written to the
-ledger and charged fees as a processed, failed transaction.
+If a transaction loading failure occurs due to exceeding its requested data size
```
mjain-jump: can we make this more broad?
```
If a transaction fails to load, its loaded transaction data size is defined as that requested data size limit.
```

2501babe: why? my view (and it seemed like the discussion was in line with this) was after we load all accounts, we can use the actual loaded size, because its clearly defined when this happens. it seems strictly worse to say, fully load a 100kb tx that requested the default 64mb and that count it as 64mb due to violating program id checks, since it only throws off the cost model

topointon-jump (changes requested):
Sorry just requesting changes whilst we hash this out

proposals/0186-loaded-transaction-data-size-specification.md
mjain-jump: it ties the exact iteration and loading order further into consensus. with this behavior, now we (and any other client) has to fail loading checks in the exact same spot with the exact same calculations as Agave, instead of having just fail <> fail being correct behavior.

2501babe: per the existing spec:

>After all accounts are loaded, if the transaction's size comes at or under its loaded accounts data size limit, the program IDs of each instruction, along with their program owners, are validated.

do you interleave loading with program id validation instead? the idea is specifically not to bring load order into consensus, which is why i chose to use the requested limit, instead of the actual amount loaded, for data size limit violation

but the idea is we load the accounts first at which point we have an order-independent canonical size, and then its sensible and straightforward to use that as the size instead

2501babe:
after talking with @mjain-jump, we had and have the exact same idea of how this is supposed to work and were just confused over the verbiage. ive rewritten the description to be more emphatic, but there are always difficulties translating plain english understanding into SIMD English

topointon-jump (approved):

mjain-jump (approved):

PR #515, opened 2026-04-10
template: updates to SIMD template

topointon-jump:
- Removes status, as all SIMDs that are "ideas" should actually live in https://github.com/solana-foundation/solana-improvement-documents/discussions
- Expand detailed design section
- Add conformance section

### DIFF
--- a/XXXX-template.md
+++ b/XXXX-template.md
@@ -55,7 +55,9 @@ document are to be interpreted as described in [RFC
 
 ### Edge Cases
 
-Which edge cases will this change affect? Enumerate these below.
+Which edge cases should the implementors of this change be aware of? Enumerate
+these below. This may includes topics such as extreme max/min/zero values,
+unusual states and scenarios that require particular attention.
 
 ### Validator Components Affected
 
--- a/XXXX-template.md
+++ b/XXXX-template.md
@@ -70,6 +70,7 @@ Which validator components are affected by this change?
 | Gossip                          |                                     |
 | Turbine                         |                                     |
 | Snapshots                       |                                     |
+| On-Chain Core BPF Programs      |                                     |
 | Other (please describe)         |                                     |
 
 ## Alternatives Considered
--- a/XXXX-template.md
+++ b/XXXX-template.md
@@ -5,7 +5,6 @@ authors:
   - (fill in with names of authors)
 category: Standard/Meta
 type: Core/Networking/Interface/Meta
-status: Idea/Review
 created: (fill me in with today's date, YYYY-MM-DD)
 feature: (fill in with feature key and github tracking issues once accepted)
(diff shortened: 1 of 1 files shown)


bw-solana (commented):
LGTM. Not clicking approve yet to give others a chance to review

XXXX-template.md
```diff
@@ -47,14 +46,32 @@ to another Solana core contributor. The generally means:
...
 "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this
 document are to be interpreted as described in [RFC
 2119](https://www.ietf.org/rfc/rfc2119.txt) and [RFC
 8174](https://www.ietf.org/rfc/rfc8174.txt).
 
+### Edge Cases
+
+Which edge cases will this change affect? Enumerate these below.
+
+### Validator Components Affected
```
buffalojoec, line 62: Should we have an optional section for SIMDs that pertain to on-chain programs? Mainly the core ones.

XXXX-template.md
```diff
@@ -77,3 +94,19 @@ Why should we not do this?
 
 Does the feature introduce any breaking changes? All incompatibilities and
 consequences should be listed.
+
+## Conformance
+
+How will all clients verify that they have correctly implemented this change?
+
```
buffalojoec, line 104: Should we just mandate use of SolFuzz-Agave et al to produce post-activation vectors? We could just cook that into the procedure.

XXXX-template.md
topointon-jump: great idea!

XXXX-template.md
topointon-jump: we should eventually, I think to get people familiar with the tooling we should do this on a case-by-case basis initially

buffalojoec (dismissed):

bw-solana (approved):

XXXX-template.md
```diff
@@ -47,14 +46,33 @@ to another Solana core contributor. The generally means:
...
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this
 document are to be interpreted as described in [RFC
 2119](https://www.ietf.org/rfc/rfc2119.txt) and [RFC
 8174](https://www.ietf.org/rfc/rfc8174.txt).
 
+### Edge Cases
+
+Which edge cases will this change affect? Enumerate these below.
```
t-nelson: we may need to give some examples of what we mean by edge case

ptaffet-jump (approved):

t-nelson (approved):

PR #520, opened 2026-04-21
SIMD-0392 amendment: Relax rent check for collector accounts

joncinque:
#### Problem

SIMD-0232 specifies conditions for collector accounts during reward deposit. These conditions, however, do not take into account a world where rent may fluctuate.

#### Summary of changes

Clarify an edge case for reward collector accounts.

### DIFF
--- a/proposals/0392-rent-increase-adaptations.md
+++ b/proposals/0392-rent-increase-adaptations.md
@@ -191,6 +191,15 @@ section for more details.
 - If the account owner changes, always enforce the current rent-exempt minimum
   for the post-exec size; the `min(pre_exec_balance, …)` clause does not apply.
 
+**Reward collector accounts:**
+
+- Inflation and block reward collector accounts no longer need to hold at least
+  the rent-exempt minimum balance during reward deposit. Instead, a collector
+  account with any lamports can be assumed to have been created by respecting
+  the rent-exempt minimum at that time. For that reason, the rent-exempt minimum
+  check must only be performed on a collector account with 0 lamports prior to
+  the reward deposit.
+
 ### Rent-Adjusted Stake Delegations
 
 During the epoch rewards calculation phase, a stake's updated delegation MUST be


proposals/0392-rent-increase-adaptations.md
buffalojoec, line 1: Can we also make this section a bit more clear on account of this particular relaxation?

https://github.com/solana-foundation/solana-improvement-documents/blob/5a43c706e7bc44e8a8978b1d9186278840006940/proposals/0392-rent-increase-adaptations.md?plain=1#L118-L124

proposals/0392-rent-increase-adaptations.md
```diff
@@ -191,6 +191,15 @@ section for more details.
...
 
+**Reward collector accounts:**
+
+- Inflation and block reward collector accounts no longer need to hold at least
+  the rent-exempt minimum balance during reward deposit. Instead, a collector
+  account with any lamports can be assumed to have been created by respecting
+  the rent-exempt minimum at that time. For that reason, the rent-exempt minimum
+  check must only be performed on a collector account with 0 lamports prior to
+  the reward deposit.
+
```
buffalojoec, line 202: Just to make sure I'm following, we plan to also amend SIMD-0232, right? Because as of right now, they are stated to be rent-exempt post-distribution.

https://github.com/solana-foundation/solana-improvement-documents/blob/5a43c706e7bc44e8a8978b1d9186278840006940/proposals/0232-custom-commission-collector.md?plain=1#L80

buffalojoec: Or are you suggesting we make SIMD-0232 a dependency of SIMD-0392, and thus overrule that piece through this feature?

proposals/0392-rent-increase-adaptations.md
joncinque: Oh nice! I didn't even realize this was spelled out, I'll amend that paragraph instead

proposals/0392-rent-increase-adaptations.md
joncinque: Yeah sure, whatever's clearest

proposals/0392-rent-increase-adaptations.md
```diff
@@ -191,6 +191,15 @@ section for more details.
 - If the account owner changes, always enforce the current rent-exempt minimum
   for the post-exec size; the `min(pre_exec_balance, …)` clause does not apply.
 
+**Reward collector accounts:**
```
igor56D, line 194: IIUC we're not really changing the spec here, right? Seems like we're just making more clear/explicit that 392 changes to RentState determination also apply when checking rent state transitions during reward distribution.

proposals/0392-rent-increase-adaptations.md
igor56D: I don't see why we need to add a dependency or change 232. I think there's confusion because of my poor choice of using pre/post *execution* rather than clarify that this applies to pre/post rent state transitions more generally (exec, fee debit, reward distribution, etc).

392 is basically agnostic to the specific transition rules. The main changes are to how we determine pre/post transition rent states:
1. every *existing* account is considered RentExempt pre-transition
2. expand the post-transition RentExempt category

proposals/0392-rent-increase-adaptations.md
igor56D: > For that reason, the rent-exempt minimum
  check must only be performed on a collector account with 0 lamports prior to
  the reward deposit.

This seems like an implementation detail to me and may be confusing because "**must**" implies there's a change to which RentState transitions are allowed, which shouldn't be the case. The reason we only need to check accounts that are zero-lamport pre-distribution is because every other account is RentExempt and incrementing the balance of a RentExempt account is already a valid transition.

proposals/0392-rent-increase-adaptations.md
igor56D: IOW this change seems both compatible with and independent of 232.

proposals/0392-rent-increase-adaptations.md
joncinque: That's correct, there's no change, but clarifying that collector account checks must take into account simd-0392

proposals/0392-rent-increase-adaptations.md
joncinque: I'm happy to do whatever is clearest. Maybe this is just my misunderstanding at work, and we don't need an amendment at all

proposals/0392-rent-increase-adaptations.md
igor56D: That's fine with me. I'd personally just remove the sentence I quoted to avoid confusion and keep everything else.

proposals/0392-rent-increase-adaptations.md
igor56D: my preference is to not add a dependency or modify 232. I'm okay with adding clarification here that the RentExempt definition change also applies to reward distribution but I'm curious what @buffalojoec thinks.

buffalojoec: I think it's really just the last sentence of this amendment that makes it confusing to me. The formula here already details zero-lamport accounts.

https://github.com/solana-foundation/solana-improvement-documents/blob/8b9e4b84fbbd881191ed83328c8e8fddb46608c0/proposals/0392-rent-increase-adaptations.md?plain=1#L106-L116

The paragraph I [linked here](https://github.com/solana-foundation/solana-improvement-documents/pull/520/changes#r3117456116) is actually already sufficient I think.

As far as 392 and 232 depending on each other, you're right @igor56D - 232 already states which accounts must be rent-exempt and when, and 392 changes the definition of "rent exempt". They can probably exist orthogonally. However, we just have to make sure they are intertwined properly in the implementation so that either one can be activated first without bungling the other.

Maybe with that being said, we don't need this amendment after all @joncinque. What do you think?

joncinque: Let's just close, this was more due to my incomprehension than anything else, it seems

PR #521, opened 2026-04-21
[SIMD-0388] Specify that the pairing length will be bounded by 8

samkim-crypto:
Add detail on the maximum bound on the number of BLS12-381 pairings that can be computed on a single call. To ensure bounded execution time and prevent excessive compute consumption, we should bound BLS12-381 syscall function to process at most 8 number of pairings at once.

### DIFF
--- a/proposals/0388-bls12-381-syscalls.md
+++ b/proposals/0388-bls12-381-syscalls.md
@@ -274,7 +274,9 @@ define_syscall!(fn sol_curve_pairing_map(
 ```
 
 The `num_pairs` parameter specifies the number of G1/G2 pairs to be processed.
-The `g1_points` and `g2_points` parameters are pointers to memory buffers.
+To ensure bounded execution time and prevent excessive compute consumption,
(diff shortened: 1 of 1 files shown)


bw-solana:
@zz-sol can you SME this?

zz-sol (approved):
LGTM. This is a follow up of https://github.com/anza-xyz/agave/pull/9759#discussion_r2695463752

Rexicon226 (approved):

samkim-crypto:
@t-nelson, @steviez, @bw-solana, can I get an approval for this? 🙏

steviez (approved):
Hi @samkim-crypto - Just making sure, nothing has changed with the implementation and this is just making the SIMD/docs more clearly match the implementation right ?

samkim-crypto:
> Hi @samkim-crypto - Just making sure, nothing has changed with the implementation and this is just making the SIMD/docs more clearly match the implementation right ?

Yes, nothing changed in terms of the implementation. This is just making the SIMD docs more precise.

PR #525, opened 2026-05-01
SIMD-0525: Shorter slot times

bw-solana:
Formal proposal for gradually reducing target slot times to 200ms.

For background information and context, see:
1. SIMD discussion for reducing slot times:
   https://github.com/solana-foundation/solana-improvement-documents/discussions/469
2. Agave exploratory implementation for halving slot times:
   https://github.com/anza-xyz/agave/pull/10740
3. Agave exploratory implementation for changing leader span and epoch length:
   https://github.com/anza-xyz/agave/pull/12154

### DIFF
--- a/proposals/0525-reduce-slot-times.md
+++ b/proposals/0525-reduce-slot-times.md
@@ -1,5 +1,5 @@
 ---
-simd: 'XXXX'
+simd: '0525'
 title: Reduce Slot Times
 authors:
   - Brennan Watt (Anza)
@@ -82,13 +82,21 @@ stored `slots_per_year` value by `old_slot_ms / new_slot_ms` at each feature
 activation rather than recomputing from SDK constants.
 `rent_collector.slots_per_year` MUST be updated with the same value.
 
-| Target slot time | Feature gate | `ns_per_slot` | Tick duration | 4-slot leader span | Epoch duration | `slots_per_year` |
-|------------------|--------------|---------------|---------------|--------------------|----------------|------------------|
-| 400ms | current | `400_000_000` | `6_250_000ns` | 1.6s | 48h | `78_892_314.984` |
-| 350ms | `reduce_slot_time_to_350ms` | `350_000_000` | `5_468_750ns` | 1.4s | 42h | `90_162_645.696` |
-| 300ms | `reduce_slot_time_to_300ms` | `300_000_000` | `4_687_500ns` | 1.2s | 36h | `105_189_753.312` |
-| 250ms | `reduce_slot_time_to_250ms` | `250_000_000` | `3_906_250ns` | 1.0s | 30h | `126_227_703.974` |
-| 200ms | `reduce_slot_time_to_200ms` | `200_000_000` | `3_125_000ns` | 0.8s | 24h | `157_784_629.968` |
+| Target | Feature gate |
+|--------|--------------|
+| 400ms | current |
+| 350ms | `reduce_slot_time_to_350ms` |
+| 300ms | `reduce_slot_time_to_300ms` |
+| 250ms | `reduce_slot_time_to_250ms` |
+| 200ms | `reduce_slot_time_to_200ms` |
+
+| Target | ns/slot | tick ns | span | epoch | slots/year |
+|--------|---------|---------|------|-------|------------|
+| 400ms | 400000000 | 6250000 | 1.6s | 48h | 78892314.984 |
+| 350ms | 350000000 | 5468750 | 1.4s | 42h | 90162645.696 |
+| 300ms | 300000000 | 4687500 | 1.2s | 36h | 105189753.312 |
+| 250ms | 250000000 | 3906250 | 1.0s | 30h | 126227703.974 |
+| 200ms | 200000000 | 3125000 | 0.8s | 24h | 157784629.968 |
 
 Note that none of the above table values (except for `slots_per_year` for
 inflation) are explicitly in protocol and may deviate from reality, but they are
@@ -100,17 +108,29 @@ Integer values are scaled by `floor(current_400ms_value * target_slot_ms /
 400)`. Flooring is used whenever the proportional value is fractional so the
 new value does not exceed the intended wall-clock budget.
 
-| Value | 400ms | 350ms | 300ms | 250ms | 200ms |
-|-------|-------|-------|-------|-------|-------|
-| Hashes per tick, non-Alpenglow | `62_500` | `54_687` | `46_875` | `39_062` | `31_250` |
-| Target signatures per slot | `20_000` | `17_500` | `15_000` | `12_500` | `10_000` |
-| Max block CUs | `60_000_000` | `52_500_000` | `45_000_000` | `37_500_000` | `30_000_000` |
-| Max writable account CUs | `24_000_000` | `21_000_000` | `18_000_000` | `15_000_000` | `12_000_000` |
-| Max vote CUs | `36_000_000` | `31_500_000` | `27_000_000` | `22_500_000` | `18_000_000` |
-| Max accounts data size delta | `100_000_000` | `87_500_000` | `75_000_000` | `62_500_000` | `50_000_000` |
-| Max data shreds per slot | `32_768` | `28_672` | `24_576` | `20_480` | `16_384` |
-| Max coding shreds per slot | `32_768` | `28_672` | `24_576` | `20_480` | `16_384` |
-| PER stake accounts written per slot | `4_096` | `3_584` | `3_072` | `2_560` | `2_048` |
+| Limit | 400ms | 350ms | 300ms |
+|-------|-------|-------|-------|
+| Hashes/tick | 62500 | 54687 | 46875 |
+| Target sigs/slot | 20000 | 17500 | 15000 |
+| Max block CUs | 60000000 | 52500000 | 45000000 |
+| Max writable acct CUs | 24000000 | 21000000 | 18000000 |
+| Max vote CUs | 36000000 | 31500000 | 27000000 |
+| Max data delta | 100000000 | 87500000 | 75000000 |
+| Max data shreds | 32768 | 28672 | 24576 |
+| Max coding shreds | 32768 | 28672 | 24576 |
+| PER stake writes | 4096 | 3584 | 3072 |
+
+| Limit | 250ms | 200ms |
+|-------|-------|-------|
+| Hashes/tick | 39062 | 31250 |
+| Target sigs/slot | 12500 | 10000 |
+| Max block CUs | 37500000 | 30000000 |
+| Max writable acct CUs | 15000000 | 12000000 |
+| Max vote CUs | 22500000 | 18000000 |
+| Max data delta | 62500000 | 50000000 |
+| Max data shreds | 20480 | 16384 |
+| Max coding shreds | 20480 | 16384 |
+| PER stake writes | 2560 | 2048 |
 
 The `hashes_per_tick` value MUST NOT be reduced by these feature gates when
 Alpenglow is active. Alpenglow's low-power PoH path should keep its active
@@ -190,17 +210,25 @@ Implementations should pay special attention to:
 
 ### Validator Components Affected
 
-| Validator Component | Impact |
-|---------------------|--------|
-| Transaction Execution (Runtime) | Bank timing fields, inflation slot-to-time conversion, fee-rate governor, cost tracker, block data size limits, and PER distribution limits change by feature gate. |
-| Virtual Machine | No direct VM semantic change. Programs may observe finer slot-time granularity through sysvars and RPC-derived timing. |
-| Block Packing | Leaders must pack smaller per-slot CU, account CU, vote CU, accounts-data, shred, and PER write budgets. |
-| Consensus | Feature-gated slot duration changes are consensus critical. Leader windows remain 4 slots but become shorter in wall-clock time. |
-| Gossip | Vote and epoch-slot traffic increase per wall-clock time as slots get faster. |
-| Turbine | The per-slot data and coding shred limits are reduced proportionally. |
-| Snapshots | Snapshot restore must reconstruct active runtime limits. Snapshot interval policy is not changed by this proposal. |
-| On-Chain Core BPF Programs | No direct program interface change. Programs that interpret slot distance as wall-clock time may need updates. |
-| Other | SDKs, RPC clients, explorers, and off-chain systems that use static slot-time constants will temporarily disagree with chain reality. |
+- Transaction Execution (Runtime): Bank timing fields, inflation slot-to-time
+  conversion, fee-rate governor, cost tracker, block data size limits, and PER
+  distribution limits change by feature gate.
+- Virtual Machine: No direct VM semantic change. Programs may observe finer
+  slot-time granularity through sysvars and RPC-derived timing.
+- Block Packing: Leaders must pack smaller per-slot CU, account CU, vote CU,
+  accounts-data, shred, and PER write budgets.
+- Consensus: Feature-gated slot duration changes are consensus critical. Leader
+  windows remain 4 slots but become shorter in wall-clock time.
+- Gossip: Vote and epoch-slot traffic increase per wall-clock time as slots get
+  faster.
+- Turbine: The per-slot data and coding shred limits are reduced
+  proportionally.
+- Snapshots: Snapshot restore must reconstruct active runtime limits. Snapshot
+  interval policy is not changed by this proposal.
+- On-Chain Core BPF Programs: No direct program interface change. Programs that
+  interpret slot distance as wall-clock time may need updates.
+- Other: SDKs, RPC clients, explorers, and off-chain systems that use static
+  slot-time constants will temporarily disagree with chain reality.
 
 ## Alternatives Considered
 
@@ -347,7 +375,7 @@ feature activations and include blocks near each activation boundary.
 
 1. Agave exploratory implementation for halving slot times:
    https://github.com/anza-xyz/agave/pull/10740
-2. SIMD discussion for reducing slot times:
-   https://github.com/solana-foundation/solana-improvement-documents/discussions/469
+2. SIMD discussion #469 in the SIMD repository:
+   https://github.com/solana-foundation/solana-improvement-documents
 3. Agave exploratory implementation for changing leader span and epoch length:
    https://github.com/anza-xyz/agave/pull/12154
--- a/proposals/0525-reduce-slot-times.md
+++ b/proposals/0525-reduce-slot-times.md
@@ -196,6 +196,27 @@ validator restarting after activation or after an effective transition must
 produce and validate the same limits as a validator that remained online across
 the transition.
 
+Snapshot serialization MUST also persist the effective timing values in the
+snapshot manifest. This proposal does not add new manifest fields, but the
+existing `ns_per_slot`, `hashes_per_tick`, `slots_per_year`, and
+`target_signatures_per_slot` manifest values MUST change in snapshots taken at
+or after a slot-time feature's effective slot. Snapshots taken before the
+effective slot, including during the one-epoch delay after activation, MUST keep
+the previous effective values. For non-Alpenglow banks, snapshot manifests MUST
+serialize the following values:
+
+| Effective target | `ns_per_slot` | `hashes_per_tick` | `slots_per_year` | `target_signatures_per_slot` |
+|------------------|---------------|-------------------|------------------|------------------------------|
+| 400ms | 400000000 | 62500 | 78892314.984 | 20000 |
+| 350ms | 350000000 | 54687 | 90162645.696 | 17500 |
+| 300ms | 300000000 | 46875 | 105189753.312 | 15000 |
+| 250ms | 250000000 | 39062 | 126227703.974 | 12500 |
+| 200ms | 200000000 | 31250 | 157784629.968 | 10000 |
+
+When Alpenglow is active, the `hashes_per_tick` manifest value MUST retain
+Alpenglow's active hashing behavior (set to `None`) instead of using the reduced
+table value.
+
 Inflation calculations MUST account for slot ranges that cross one or more
 effective slot-time transitions. Historical slots before an effective transition
 use the previous `slots_per_year`; slots at or after the effective transition
@@ -264,7 +285,9 @@ Implementations should pay special attention to:
 - Turbine: The per-slot data and coding shred limits are reduced
   proportionally after the one-epoch effectiveness delay, so shred filtering and
   bank execution enforce the same slot-aware limits.
-- Snapshots: Snapshot restore must reconstruct effective runtime limits. Snapshot
+- Snapshots: Snapshot manifests must persist the effective `ns_per_slot`,
+  `hashes_per_tick`, `slots_per_year`, and `target_signatures_per_slot` values,
+  and snapshot restore must reconstruct effective runtime limits. Snapshot
   interval policy is not changed by this proposal.
 - On-Chain Core BPF Programs: No direct program interface change. Programs that
   interpret slot distance as wall-clock time may need updates.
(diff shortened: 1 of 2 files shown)


proposals/0525-reduce-slot-times.md
```diff
@@ -0,0 +1,381 @@
...
+- Correct bank values for `ns_per_slot`, `slots_per_year`,
+  `rent_collector.slots_per_year`, non-Alpenglow `hashes_per_tick`, and
+  `target_signatures_per_slot`.
+- Correct block, writable-account, vote, accounts-data, shred, and PER limits
+  for every target slot time in the tables above.
+- Correct snapshot restore behavior after each feature gate is active.
+- Correct shred validation for slots before and after an activation boundary.
+- Correct inflation behavior for slot ranges and epochs that span activation.
+- Correct Alpenglow interaction, specifically that Alpenglow hashing behavior
+  is not reduced by these feature gates.
```
ripatel-fd, line 460: Instead of stating the obvious (a change should be tested), it would be nice to just include the actual test vectors and fixtures with the SIMD.

bw-solana: can you give me an example of what this would look like?

ripatel-fd: Link to a new repo simd-525 and upload files in whatever format Agave uses (or solfuzz fixtures)

proposals/0525-reduce-slot-times.md
```diff
@@ -0,0 +1,381 @@
...
+Reduce Solana's target slot time from 400ms to 200ms in four feature-gated
+steps:
+
+1. 350ms slots
+2. 300ms slots
+3. 250ms slots
+4. 200ms slots
+
+Each step keeps `ticks_per_slot` fixed at 64, keeps leader windows fixed at
+4 slots, and keeps epochs fixed at 432,000 slots. Per-slot work limits are
```
t-nelson: > keeps leader windows fixed at 4 slots

i'm still concerned that not fixing the leader window in time instead of slots _for this first phase_ is putting us at undue risk. any unforeseen issues that arise will do so on leader boundary slots. more slots per window gives us more boundary slots to sacrifice

proposals/0525-reduce-slot-times.md
```diff
@@ -0,0 +1,381 @@
...
+
+Shorter slots reduce confirmation and finalization latency for users. Any
+consensus or commitment threshold measured in slots takes less wall-clock time
+as slot time decreases.
+
+Keeping the leader span at 4 slots also shortens the wall-clock duration of a
+single leader window. A leader currently controls a nominal 1.6s window. At
+200ms slots that window is 800ms. This improves market structure by reducing
+the worst-case time a leader can delay, reorder, or selectively include
+transactions before the next leader has an opportunity to produce a block.
```
t-nelson, line 45: > reduce confirmation and finalization latency for users

> reducing the worst-case time a leader can delay, reorder, or selectively include transactions

while i agree that both of these features are desirable goals. i don't agree that either is so critical as to reach an emergency status whereby rushing them into prod simultaneously is justifiable

as stated elsewhere, my preference would be to reduce slot times first, then leader window span. that agave's implementation of this logic is a lazy shit show is no reason to impose undue risk on the ecosystem

proposals/0525-reduce-slot-times.md
```diff
@@ -0,0 +1,381 @@
...
+| 250ms | 250000000 | 3906250 | 1.0s | 30h | 126227703.974 |
+| 200ms | 200000000 | 3125000 | 0.8s | 24h | 157784629.968 |
+
+Note that none of the above table values (except for `slots_per_year` for
+inflation) are explicitly in protocol and may deviate from reality, but they are
+expected to be useful for ensuring expected timing is understood.
+
+### Scaled Per-Slot Values
+
+Integer values are scaled by `floor(current_400ms_value * target_slot_ms /
```
t-nelson: i don't think we have any negative numbers in the path, but we'd probably best use `trunc` (round to zero) rather than `floor` (round to negative infinity)

proposals/0525-reduce-slot-times.md
```diff
@@ -0,0 +1,381 @@
...
+
+This proposal refers to the following placeholder feature gates. Final feature
+IDs are TBD:
+
+- `reduce_slot_time_to_350ms`
+- `reduce_slot_time_to_300ms`
+- `reduce_slot_time_to_250ms`
+- `reduce_slot_time_to_200ms`
+
+## Detailed Design
```
t-nelson, line 70: do we anticipate that the scaling of dependent values occur literally in code or derived algebraically? we still have a bunch of ieee754 that needs to die and i think the latter is only going to necessitate introducing more. this needs to be descried explicitly and heavily tested for cross client compliance

proposals/0525-reduce-slot-times.md
```diff
@@ -0,0 +1,381 @@
...
+
+### Notable Non-Changes
+
+This proposal intentionally does not change:
+
+- Leader span: still 4 slots.
+- Ticks per slot: still 64.
+- Epoch length: still 432,000 slots.
+- `solana-clock` and SDK constant values. Agave may dynamically scale bank
+  values, but `solana-sdk` constants remain unchanged initially.
```
t-nelson, line 248: these constants will certainly bring about the most collateral damage. we need to enumerate which ones aren't available onchain and work with foundation devrel to ensure as much as possible that consumers know that using them is unreliable and will break imminently. we cannot wait for activation day to start on this

proposals/0525-reduce-slot-times.md
```diff
@@ -0,0 +1,381 @@
...
+  reduced target slot time.
+- Feature gates activated after snapshots, where runtime-only values must be
+  rebuilt during restore.
+- Multiple active slot-time gates, where the lowest active target slot time
+  wins.
+- Alpenglow activation, where `hashes_per_tick` must not be reduced by these
+  feature gates.
+- Integer rounding, especially the 350ms and 250ms `hashes_per_tick` values.
+- Inflation calculations for epochs and rewards that span feature activation
+  slots.
```
t-nelson: is this implying the inflation calculations count skipped slots at the new epoch boundary as previous epoch slots?

proposals/0525-reduce-slot-times.md
```diff
@@ -0,0 +1,381 @@
...
+time decreases, keeping the leader window close to the current 1.6s wall-clock
+duration. For example, a 200ms slot time would use 8 slots per leader span.
+
+This preserves leader-window duration but requires the leader schedule to become
+dynamic across feature activations. It affects code that computes leader-window
+boundaries, leader schedule offsets, next-leader routing, replay propagation,
+Alpenglow parent-ready behavior, and related client assumptions. The
+experimental Agave implementation that changes leader span and epoch length
+shows this introduces significant code complexity and therefore higher risk of
+mistakes.
```
t-nelson, line 325: implementation detail. it's time to pay the piper

proposals/0525-reduce-slot-times.md
```diff
@@ -0,0 +1,381 @@
...
+This proposal is not backwards compatible for validators. New feature gates
+change consensus-critical bank, block, and shred limits.
+
+The SDK constants will be out of sync with chain reality until the SDK is
+updated or the parameters are made available on chain. In particular, static
+constants for default slot duration, ticks per second, slots per year, and
+slot-to-time conversion may continue to describe the old 400ms assumption while
+the running cluster uses 350ms, 300ms, 250ms, or 200ms banks. Some short-term
+light breakage in clients that use these constants is tolerable in order to
+ship the staged reduction sooner.
```
t-nelson, line 433: is it? i've been told that an rpc outage is an outage

proposals/0525-reduce-slot-times.md
```diff
@@ -0,0 +1,381 @@
...
+- Activation of each feature gate from a 400ms baseline.
+- Correct bank values for `ns_per_slot`, `slots_per_year`,
+  `rent_collector.slots_per_year`, non-Alpenglow `hashes_per_tick`, and
+  `target_signatures_per_slot`.
+- Correct block, writable-account, vote, accounts-data, shred, and PER limits
+  for every target slot time in the tables above.
+- Correct snapshot restore behavior after each feature gate is active.
+- Correct shred validation for slots before and after an activation boundary.
+- Correct inflation behavior for slot ranges and epochs that span activation.
+- Correct Alpenglow interaction, specifically that Alpenglow hashing behavior
```
t-nelson, line 459: i think the point of this section is to define "correct"

t-nelson (commented):
directionally aligned. same opposition as usual. there is zero urgency to justify taking shortcuts for the sake of agave's technical debt. clean up the code first. expose the variable values first

vadorovsky:
Before I dig into the details of SIMD, I just wanted to say that the current epoch boundary times are problematic - the numbers I'm observing on mainnet with Agave are always above 200ms threshold. The lowest I've seen being 206ms, the highest more than 260ms, like:

before

At the same time, I have some good news - I've managed to prototype a change in stakes cache that pushes epoch boundary down to 116ms, which would hopefully stay below 200ms on different validators:

pool_after

PR coming soon, this week for sure, although I'm still figuring out how to split that change properly and minimize diff (for now I have a quite intrusive >3k LOC diff :skull: ).

The question I wanted to raise is - should we make this SIMD dependent on this work, and explicitly mention <200ms epoch boundary in all validator implementations as a requirement? Or even consider that work as an integral part of SIMD? One of Agave's achievements last year was pushing down epoch boundary below 400ms to avoid empty slots. It would be a pity to introduce the problem again.

proposals/0525-reduce-slot-times.md
bw-solana: anything specific that you're concerned about? More concerned about the client/targeting or validator side? 

The jump 200ms slot experiment on mainnet has alleviated some of the concerns here. Looks like it's packing more than 1/2 of the usual non-vote txs per slot on average, similar story w/ rewards, and vote latencies are pretty decent (though worse than the rest of the cluster)

proposals/0525-reduce-slot-times.md
bw-solana: we definitely shouldn't rush them into prod.

But I think `P(outage)` is significantly larger trying to change slot times, leader span in slots, and slots per epoch simultaneously - even with all the pre-plumbing work ahead of activation. We'll just have to agree to disagree on this one

proposals/0525-reduce-slot-times.md
bw-solana: it's a good question... the ambiguity is why I put the literals in the design section here in the SIMD.

Hard coding literals in the code is harder to screw up, but the ladder logic checking each feature to figure out which value should be used in some cases is pretty gross.

I used algebraic derivation in the MVP in most places, but part of this was so that we could arbitrarily add different slot time values and things would "just work"

Agree we should hammer this w/ some client conformance fuzzing

proposals/0525-reduce-slot-times.md
bw-solana:
```
DEFAULT_TICKS_PER_SECOND
MS_PER_TICK
DEFAULT_HASHES_PER_TICK
TICKS_PER_DAY
DEFAULT_MS_PER_SLOT
DEFAULT_S_PER_SLOT
MAX_HASH_AGE_IN_SECONDS
```

proposals/0525-reduce-slot-times.md
bw-solana: No, intention is to use epoch granularity for assigning target slot time for inflation/rewards calculations.

I can cleanup this language. It's trying to warn that clients will need to remember which slot time target applies to which epoch ranges for the rest of time basically

proposals/0525-reduce-slot-times.md
bw-solana: "light breakage" doing some heavy lifting here, but this was my takeaway from our meeting last month.

I prepared an MVP to look at moving some of these constants onchain: https://github.com/anza-xyz/agave/pull/12023

But not delaying on this feels like the kind of aggressive, "good risk" behavior we want more of. I don't see any increase in `P(outage)`. There might be some short impact to tx landing, but like... can't we just update and fix it in like 5 minutes? Feels like this is significantly de-risked by:

1. All the experiments using old client + new 200ms code landing a lot of transactions
2. Stepping down slot times by 50 ms at a time and therefore can reduce how far off client constants are for any single hop
3. Jump 200ms validator experiment landing a bunch of transactions
4. Not changing the blockhash expiration / status cache capacity

proposals/0525-reduce-slot-times.md
bw-solana: I started writing more here but realized I was basically duplicating information from the tables above in the detailed design section

proposals/0525-reduce-slot-times.md
bw-solana: Do I self-host the new repo? Seems like we would want files to be co-located in this SIMD repo, no?

Can you point me to the shred fuzzer? I'm not finding it

bw-solana:
> should we make this SIMD dependent on this work, and explicitly mention <200ms epoch boundary in all validator implementations as a requirement?

I think it would be great to drive epoch boundary crossing time below 200ms, but I vote against gating reduced slot times on this work. A skipped slot or 2 on boundary feels worth the tradeoff - and hey, each slot skipped is only half as painful 😄

proposals/0525-reduce-slot-times.md
topointon-jump: I think it's safest to explicitly hard-code the literals. We can't rely on fuzzing for this.

proposals/0525-reduce-slot-times.md
```diff
@@ -0,0 +1,381 @@
...
+values. Implementations MUST NOT repeatedly scale already-rounded values from a
+previous stage, because that can accumulate rounding error.
+
+### Timing Values
+
+The `slots_per_year` values below use the Mainnet-Beta bank value of
+`78_892_314.984`, as the 400ms baseline. Implementations SHOULD scale the bank's
+stored `slots_per_year` value by `old_slot_ms / new_slot_ms` at each feature
+activation rather than recomputing from SDK constants.
+`rent_collector.slots_per_year` MUST be updated with the same value.
```
topointon-jump: Can we instead specify that all implementations must use hardcoded values for `slots_per_year`? So that we can avoid having to recompute it. Explicitly hardcoding values for these constants is much safer.

proposals/0525-reduce-slot-times.md
```diff
@@ -0,0 +1,381 @@
...
+| Max writable acct CUs | 15000000 | 12000000 |
+| Max vote CUs | 22500000 | 18000000 |
+| Max data delta | 62500000 | 50000000 |
+| Max data shreds | 20480 | 16384 |
+| Max coding shreds | 20480 | 16384 |
+| PER stake writes | 2560 | 2048 |
+
+The `hashes_per_tick` value MUST NOT be reduced by these feature gates when
+Alpenglow is active. Alpenglow's low-power PoH path should keep its active
+hashing behavior unchanged, currently expected to be one hash per tick.
```
topointon-jump, line 158: > The `hashes_per_tick` value MUST NOT be reduced by these feature gates when Alpenglow is active.

Just so I understand, what is the reason for this?

proposals/0525-reduce-slot-times.md
bw-solana: I don't think we should declare how clients have to implement things in the SIMD, but we should be clear about what the values need to be to be compliant (enumerated in the table below).

* steps down from ivory tower *

We should both hardcode the values in our respective clients 😄

proposals/0525-reduce-slot-times.md
bw-solana: Alpenglow inserts a single "Alpentick" at the end of the block to signify it is the end. This tick contains a single hash. This is mostly a convenience thing because a bunch of state transitions and things leverage tick height to control when to transition.

I want to be clear that we should continue performing just a single hash for this Alpentick entry.

proposals/0525-reduce-slot-times.md
topointon-jump, line 70: We should specify whether or not these feature gates take affect at the epoch of activation, or are delayed by an epoch like the other turbine/shred/repair feature gates. Do we need this feature to be delayed by an epoch? My gut says that we do, because it involves changes to block packing, repair and gossip which could diverge if there are forks during the feature activation slots.

bw-solana: my thought was to apply it immediately because leader schedule and shred filtering are not impacted.

but either way, agree it should be called out clearly. I'll add that

AshwinSekar:
> > should we make this SIMD dependent on this work, and explicitly mention <200ms epoch boundary in all validator implementations as a requirement?
> 
> I think it would be great to drive epoch boundary crossing time below 200ms, but I vote against gating reduced slot times on this work. A skipped slot or 2 on boundary feels worth the tradeoff - and hey, each slot skipped is only half as painful 😄

Every "first" slot that happens on the boundary will be subject to the same thing, so we still need to ensure that we can get a slot through.

That being said there's still a buffer between block time and actual skip time to account for network delays. In Alpenglow we could even extend the skip timer for the first block at the boundary as well.

proposals/0525-reduce-slot-times.md
topointon-jump: We should specify how the _snapshot manifest_ fields will change, and that they will change in the snapshot. (`ticks_per_slot`, `ns_per_slot`, `hashes_per_tick`, `slots_per_year`, `target_signatures_per_slot`).

proposals/0525-reduce-slot-times.md
topointon-jump: > shred filtering are not impacted

it says elsewhere in this simd that the data and coding shred limits change. this means that the validator nodes (not the leader) need to apply different shred filtering rules depending on if the feature gate is active or not. I think this means that we'll have to delay by an epoch because otherwise we don't know what fork the shred was produced on, and thus if it should be considered valid or not.

bw-solana: ugh, you're right

proposals/0525-reduce-slot-times.md
```diff
@@ -0,0 +1,381 @@
...
+Integer values are scaled by `trunc(current_400ms_value * target_slot_ms /
+400)`. Truncating is used whenever the proportional value is fractional so the
+new value does not exceed the intended wall-clock budget.
+
+| Limit | 400ms | 350ms | 300ms |
+|-------|-------|-------|-------|
+| Hashes/tick | 62500 | 54687 | 46875 |
+| Target sigs/slot | 20000 | 17500 | 15000 |
+| Max block CUs | 60000000 | 52500000 | 45000000 |
+| Max writable acct CUs | 24000000 | 21000000 | 18000000 |
```
topointon-jump, line 137: We should specify how the changes to `Max writable acct CUs` and `Max block CUs` will interact with the `raise_block_limits_to_100m` feature gate, as it's likely that these will cross-over. Might be worth a separate table showing the matrix of values.

proposals/0525-reduce-slot-times.md
```diff
@@ -0,0 +1,381 @@
...
+- `fee_rate_governor.target_signatures_per_slot`
+- cost tracker block limits
+- block accounts data size delta
+- broadcast and shred-fetch shred limits
+- partitioned epoch rewards stake-account stores per block
+
+Snapshot restore and bank deserialization MUST reconstruct all non-persisted
+runtime values from the active feature set and the bank's slot. A validator
+restarting after activation must produce and validate the same limits as a
+validator that remained online across activation.
```
topointon-jump: This is why we need to specify the snapshot manifest values that change and how they change. (`ticks_per_slot`, `ns_per_slot`, `hashes_per_tick`, `slots_per_year`, `target_signatures_per_slot`)

proposals/0525-reduce-slot-times.md
```diff
@@ -0,0 +1,381 @@
...
+  conversion, fee-rate governor, cost tracker, block data size limits, and PER
+  distribution limits change by feature gate.
+- Virtual Machine: No direct VM semantic change. Programs may observe finer
+  slot-time granularity through sysvars and RPC-derived timing.
+- Block Packing: Leaders must pack smaller per-slot CU, account CU, vote CU,
+  accounts-data, shred, and PER write budgets.
+- Consensus: Feature-gated slot duration changes are consensus critical. Leader
+  windows remain 4 slots but become shorter in wall-clock time.
+- Gossip: Vote and epoch-slot traffic increase per wall-clock time as slots get
+  faster.
```
topointon-jump: Gossip has some usages of `DEFAULT_MS_PER_SLOT`:

- The eviction policy/retention window for CRDS entries
- The freshness threshold for accepting pull responses
- The scoring policy for which CRDS values to include in pull responses

We should specify which ones should change and which should stay the same.

proposals/0525-reduce-slot-times.md
ptaffet-jump: Yeah, the SIMD needs to clarify if the `slots/year` values in the table below are normative (i.e. hardcode that value literally) or just approximate. I think it only really matters for 250.

proposals/0525-reduce-slot-times.md
bw-solana: I vote for just keeping these all the same. Wdyt?

bw-solana: https://github.com/solana-foundation/solana-improvement-documents/pull/525/commits/2f40aab18f8d3b37fe7da6613a7192c77b744f6f

proposals/0525-reduce-slot-times.md
bw-solana: https://github.com/solana-foundation/solana-improvement-documents/pull/525/commits/1ecc2f73c2939d9ae3bcfbd8f59a7f36b3986735

proposals/0525-reduce-slot-times.md
bw-solana: https://github.com/solana-foundation/solana-improvement-documents/pull/525/commits/1ecc2f73c2939d9ae3bcfbd8f59a7f36b3986735

proposals/0525-reduce-slot-times.md
bw-solana: https://github.com/solana-foundation/solana-improvement-documents/pull/525/commits/e2b2a49b1c057b4af185c2a7f16f4bb881b067a0

proposals/0525-reduce-slot-times.md
bw-solana: https://github.com/solana-foundation/solana-improvement-documents/pull/525/commits/e2b2a49b1c057b4af185c2a7f16f4bb881b067a0

proposals/0525-reduce-slot-times.md
bw-solana: https://github.com/solana-foundation/solana-improvement-documents/pull/525/commits/e2b2a49b1c057b4af185c2a7f16f4bb881b067a0

bw-solana:
Blocked on the following:

1. @jacobcreech recommendation for status cache size / blockhash expiration time (151 vs. 301)
2. FD formal approval - we have concept ACK, but need to confirm finer details are agreed upon and stamp
3. Anza approval - waiting on @steviez to weigh in

Lmk if there are any updates expected from me that I may have missed

proposals/0525-reduce-slot-times.md
topointon-jump: yes lets keep these the same

topointon-jump (approved):

chnotley (approved):
Ship it

steviez (approved):
> Anza approval - waiting on @steviez to weigh in

Discussed in slack but will comment here - I'm in favor of the stated plan over alternatives that change epoch duration / slots-per-leader-span / etc.

Others have already reviewed thoroughly so I have nothing else to add than LGTM

jacobcreech:
> @jacobcreech recommendation for status cache size / blockhash expiration time (151 vs. 301)

Let's keep at 151 for this one. Any further lowering of slot time, we'll need to revisit this change.

rdev42:
I'm all for this, but this will add ~1 SOL/day in vote costs for validators, which will push the long tail of validators into losing money, potentially reducing the validator set quite a bit if this goes on too long. Is that an oversight or a feature? An easy fix to make the validator dynamics market-neutral would be to halve the voting cost

PR #129, opened 2024-03-19
SIMD-0129: Alt_BN128 Syscalls - Simplified Error Code

0x0ece:
## Summary

Simplify the return error code for the family of Alt_BN128 syscalls: 
`sol_alt_bn128_group_op`, `sol_alt_bn128_compression` and `sol_poseidon`.

A single error code is sufficient, in line with all other syscalls.

## Motivation

Syscalls in Solana can return:

- Success, e.g. represented in Rust as `Ok(0)` or `Ok(SUCCESS)`
- Error to the program, e.g. represented in Rust as `Ok(1)`
- Fatal error to the VM, aborting the transaction, e.g. represented in Rust
  as `Err(<something>)`

Most syscalls only have a single error code returned to the program, i.e. `Ok(1)`.
The family of Alt_BN128 syscalls, viceversa, has a richer set of error codes.

This proposal aims to simplify the error value for these syscalls, in line with
all the other syscalls, and simply return `Ok(1)` (in addition to fatal errors,
that are left unchanged).

We stress that multiple error codes cause a maintenance burden for the validators.
Moreover, if two different implementation were to return different error codes,
an attacker could exploit the different behavior to cause consensus failure.

PR #529, opened 2026-05-04
SIMD-0529: big integer modexp syscall

zz-sol:


proposals/0529-big-integer-modexp-syscall.md
```diff
@@ -0,0 +1,296 @@
+---
+simd: '0529'
+title: Big Integer ModExp Syscall
+authors:
+  - SK, ZZ
+category: Standard
+type: Core
+status: Idea
+created: 2026-05-01
+feature: expH2ppKPW2ANEdEmAjfhSEcnBQJfmoX4FjuNpe9ttg
```
samkim-crypto, line 10: Is this a key you own @zz-sol? We generally leave this blank until the feature is accepted (though this specific case is a weird case since this was implemented previously).

proposals/0529-big-integer-modexp-syscall.md
```diff
@@ -0,0 +1,296 @@
...
+document are to be interpreted as described in [RFC 2119] and [RFC 8174].
+
+### Syscall Interface
+
+Add the following syscall:
+
+```rust
+define_syscall!(fn sol_big_mod_exp(
+    endianness: u64,
+    params_addr: *const BigModExpParams
```
samkim-crypto: I see a case for absorbing everything into `BigModExpParams`, but my personal preference would be to have a separate `result: *mut u8` parameter for the following reasons:
- consistency with the other syscalls
- `*mut u8` is explicit about mutability/side-effects

Might be worth getting other people's thoughts here though.

proposals/0529-big-integer-modexp-syscall.md
```diff
@@ -0,0 +1,296 @@
...
+If `modulus_len > 0` and the numerical value of `modulus` is zero, the syscall
+MUST write `modulus_len` zero bytes and return success. This matches the
+Ethereum ModExp convention used by execution tests.
+
+If `modulus` is nonzero and `exponent` is zero, the result is
+`1 mod modulus`, encoded in exactly `modulus_len` bytes.
+
+### Compute Metering
+
+The syscall MUST charge compute before performing the exponentiation.
```
samkim-crypto: I think the performance of the exponentiation algorithms depends on a couple of factors:
- The Modulus Parity: Fast algorithms like Montgomery reduction requires an odd modulus and if the modulus is even, libraries may fall back to significantly slower algorithms (like Barrett reduction) or perform factorizations
- The Exponent Density: The standard square-and-multiply algorithm obviously takes longer on dense exponents (i.e. consisting of all 1s)

Briefly mentioning these performance variations and explicitly stating that the CU should be calculated with respect to the worst-case in the SIMD will be helpful I think.

samkim-crypto (commented):
I am in support of this!

A couple of comments below. Also, maybe we should note in the PR description or the SIMD that the big integer modular exponentiation was previously implemented before the SIMD process, and that we are writing the missing SIMD for it (with some modifications like endianness, etc.).

proposals/0529-big-integer-modexp-syscall.md
zz-sol: Yes I own this key

proposals/0529-big-integer-modexp-syscall.md
zz-sol: The parity can be checked easily.

It will be a bit hard to cater for the density case. I would suggest to measure the worst case and not care for the density. This aligns with, for example, #129, the cost is independent of the scalar.

proposals/0529-big-integer-modexp-syscall.md
```diff
@@ -0,0 +1,296 @@
...
+#[repr(C)]
+pub struct BigModExpParams {
+    pub base_addr: u64,
+    pub base_len: u64,
+    pub exponent_addr: u64,
+    pub exponent_len: u64,
+    pub modulus_addr: u64,
+    pub modulus_len: u64,
+    pub result_addr: u64,
+    pub result_len: u64,
```
deanmlittle: We don't need result_len as this can be inferred from modulus_len.

proposals/0529-big-integer-modexp-syscall.md
```diff
@@ -0,0 +1,296 @@
...
+### Length Limits
+
+The initial maximum supported size is:
+
+```rust
+pub const BIG_MOD_EXP_MAX_BYTES: u64 = 512;
+```
+
+Each of `base_len`, `exponent_len`, and `modulus_len` MUST be less than or
+equal to `BIG_MOD_EXP_MAX_BYTES`. This covers 4096-bit RSA moduli and keeps
```
deanmlittle: I'm not sure it makes sense to bind `base_len` and `modulus_len` this strictly when `exponent_len` is the thing that really needs to be bounded to avoid DoS, but it's also a pretty generous limit.

proposals/0529-big-integer-modexp-syscall.md
```diff
@@ -0,0 +1,296 @@
...
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this
+document are to be interpreted as described in [RFC 2119] and [RFC 8174].
+
+### Syscall Interface
+
+Add the following syscall:
+
+```rust
+define_syscall!(fn sol_big_mod_exp(
+    endianness: u64,
```
deanmlittle: BE would make sense if this were an RSA-specific syscall for ASN.1 DER compatibility. As it is a general purpose math syscall, I would suggest it makes more sense to just support LE for simplicity's sake. Imo, this is not a case like the altbn128 syscalls where we originally supported the Ethereum API that nobody doing ZK on Solana wanted, then changed to also support the Arkworks which everyone *did* want.

proposals/0529-big-integer-modexp-syscall.md
```diff
@@ -0,0 +1,296 @@
...
+```rust
+#[repr(C)]
+pub struct BigModExpParams {
+    pub base_addr: u64,
+    pub base_len: u64,
+    pub exponent_addr: u64,
+    pub exponent_len: u64,
+    pub modulus_addr: u64,
+    pub modulus_len: u64,
+    pub result_addr: u64,
```
deanmlittle: Wouldn't the address types here be `*const u8` for inputs and `*mut u8` for the result respectively?

zz-sol: oh i guess here it is the address of the memory which should be a constant. I see that this is confusing for everyone so I am going to change it back to *const u8 and *mut u8

proposals/0529-big-integer-modexp-syscall.md
zz-sol: cc @samkim-crypto

samkim-crypto: I don't have strong feelings either way, especially since developers can always handle byte-swapping outside of the syscall. I think the choice really depends on the primary motivation for adding this primitive:
- Ethereum compatibility? We likely want BE support to minimize friction for developers migrating from the EVM ecosystem.
- RSA support? The formal standards (like ASN.1 DER) strictly dictate BE.
- Curve cryptography? Official serialization standards for curves like BLS12-381 also require BE as well as popular libraries like `blst`.

That said, I am also all-in for simple APIs, so restricting to LE is fine too. For Agave, we plan on using `num-bigint` and since it already has native support for both LE and BE, supporting an endianness flag or restricting it to just one won't be much additional work for us either way.

Any thoughts @0x0ece @Rexicon226 ?

deanmlittle: To be clear, I also don't have strong feelings either way. I just think we should probably choose one in this case. If others strongly disagree, that's fine. I don't have any particularly strong reason for LE other than LE math typically being faster and thus there are likely some casts we can do for pre/post processing numbers in BPF. If BE were somehow faster, I would be just as happy to go with that. The reason I don't have such a strong preference here vs altbn128 is because the library we all used there was LE by default and reversing points was very cumbersome. In this case, it's just reversing an array which isn't such a big deal.

zz-sol:
I have updated the cost metric. It is based on worst case cost, i.e., it only cares for moduli's bit length. There is also a few buckets, and tailored for 

- 32 and 64 bits (starks, general computation)
- 256 bits, (general ec)
- 384 bits, (bls)
- 2048 and 4096 (RSA)


if you have any comments/suggestions/objections please LMK.

The only thing that left undecided for cost is odd or even moduli. Most of above use cases requires odd moduli. I am not sure how often a general computation will use an even moduli. I am in favor of just keeping moduli odd.

zz-sol:
> I have updated the cost metric. It is based on worst case cost, i.e., it only cares for moduli's bit length. There is also a few buckets, and tailored for
> 
> * 32 and 64 bits (starks, general computation)
> * 256 bits, (general ec)
> * 384 bits, (bls)
> * 2048 and 4096 (RSA)
> 
> if you have any comments/suggestions/objections please LMK.
> 
> The only thing that left undecided for cost is odd or even moduli. Most of above use cases requires odd moduli. I am not sure how often a general computation will use an even moduli. I am in favor of just keeping moduli odd.


Hi @deanmlittle @0x0ece @Rexicon226 @samkim-crypto can I please get your opinion on this CU model? I will fill in the actual cost if this model makes sense.

0x0ece (approved):

proposals/0529-big-integer-modexp-syscall.md
0x0ece: Let's do both endianesses, since we have it already in many syscalls. It'll make it easier to build libraries/apps.

samkim-crypto (approved):
Okay, I am happy with the issues of `endianness` and the CUs. Let's give @deanmlittle and @Rexicon226 if they have any additional thoughts.

deanmlittle:
> > I have updated the cost metric. It is based on worst case cost, i.e., it only cares for moduli's bit length. There is also a few buckets, and tailored for
> > 
> > * 32 and 64 bits (starks, general computation)
> > * 256 bits, (general ec)
> > * 384 bits, (bls)
> > * 2048 and 4096 (RSA)
> > 
> > if you have any comments/suggestions/objections please LMK.
> > The only thing that left undecided for cost is odd or even moduli. Most of above use cases requires odd moduli. I am not sure how often a general computation will use an even moduli. I am in favor of just keeping moduli odd.
> 
> Hi @deanmlittle @0x0ece @Rexicon226 @samkim-crypto can I please get your opinion on this CU model? I will fill in the actual cost if this model makes sense.

While logical from a use-case perspective, I'm not sure the buckets are the best approach from a threat-modelling perspective, which I would argue is our priority when budgeting CUs to avoid DoS. I think CUs should always try to emulate the worst case scenario. My methodology for measuring CUs for this syscall would be something like:

1. Benchmark the most pessimistic scenario of 4096 bit exponent with 4096 bit modulus
2. Benchmark the most optimistic realistic use case (probably 32 or 64 bit integers)
3. Create a formula like base_cost(exponent_len * modulus_len ^ n)
4. Figure out what `base_cost` and `n` should be to make the formula hold accurate from the optimistic scenario to the pessimistic scenario, accounting for both exponent and modulus, but more heavily indexing on modulus by factor n (square is probably good enough) as it disproportionately affects the incremental cost of each iteration.

zz-sol:
> I think CUs should always try to emulate the worst case scenario.

I think this is exactly why this bucket approach makes sense. Like for a fixed modulus size, we measure the worst case scenario, where the exp is basically `111...`, and the cu cost is for this.

zz-sol:
~I update the doc with a specific use case for RSA, where the exp is fixed to `65537`. One can use this syscall for cheaper cu, c.f. generic `big_int_mod_exp` syscall for arbitrary exp size.~

reverted. now using a derived formula similar to EIP-198

zz-sol:
Draft constants used:

```text
BIG_MOD_EXP_BASE_CU = 422
BIG_MOD_EXP_CU_DIVISOR = 189
BIG_MOD_EXP_MIN_EXPONENT_LENGTH = 75
measured CU = ceil(upper_bound_ns / 10)
```

## Balanced Dense Fit

| bits | high time | measured CU | mult_complexity | model CU | model/measured |
|---:|---:|---:|---:|---:|---:|
| 32 | 2.1909 us | 220 | 16 | 429 | 1.95x |
| 64 | 2.6903 us | 270 | 64 | 448 | 1.66x |
| 128 | 5.9499 us | 595 | 256 | 595 | 1.00x |
| 256 | 16.483 us | 1,649 | 1,024 | 1,804 | 1.09x |
| 384 | 35.231 us | 3,524 | 2,304 | 5,091 | 1.44x |
| 512 | 64.929 us | 6,493 | 4,096 | 11,497 | 1.77x |
| 768 | 188.97 us | 18,897 | 8,448 | 34,706 | 1.84x |
| 1024 | 376.82 us | 37,682 | 13,312 | 72,476 | 1.92x |
| 1536 | 1.1636 ms | 116,360 | 24,576 | 200,021 | 1.72x |
| 2048 | 2.6363 ms | 263,630 | 37,888 | 410,776 | 1.56x |
| 3072 | 8.6796 ms | 867,960 | 70,656 | 1,148,489 | 1.32x |
| 4096 | 20.475 ms | 2,047,500 | 111,616 | 2,418,769 | 1.18x |

## RSA-65537 Check

| bits | high time | measured CU | mult_complexity | model CU | model/measured |
|---:|---:|---:|---:|---:|---:|
| 32 | 2.3912 us | 240 | 16 | 429 | 1.79x |
| 64 | 2.5073 us | 251 | 64 | 448 | 1.78x |
| 128 | 3.2829 us | 329 | 256 | 524 | 1.59x |
| 256 | 4.6866 us | 469 | 1,024 | 829 | 1.77x |
| 384 | 6.7741 us | 678 | 2,304 | 1,337 | 1.97x |
| 512 | 9.4461 us | 945 | 4,096 | 2,048 | 2.17x |
| 768 | 18.472 us | 1,848 | 8,448 | 3,775 | 2.04x |
| 1024 | 28.356 us | 2,836 | 13,312 | 5,705 | 2.01x |
| 1536 | 57.134 us | 5,714 | 24,576 | 10,175 | 1.78x |
| 2048 | 97.876 us | 9,788 | 37,888 | 15,457 | 1.58x |
| 3072 | 213.31 us | 21,331 | 70,656 | 28,461 | 1.33x |
| 4096 | 376.88 us | 37,688 | 111,616 | 44,715 | 1.19x |


notes:
- `cu = time_in_nano_second / 10`. For the benchmark machine, the ed25519 sig verify took 22.686 µs and its cost is defined as 2,280–2,400 CU per signature

deanmlittle:
Thanks @zz-sol. I think these numbers look roughly like what I'd expect them to. I think this accurately prices out DoS whilst still happily accommodating the largest use case we currently expect to see. Lgtm!

proposals/0529-big-integer-modexp-syscall.md
```diff
@@ -0,0 +1,374 @@
...
+
+```rust
+pub fn sol_big_mod_exp(
+    endianness: u64,
+    base: *const u8,
+    exponent: *const u8,
+    exponent_len: u64,
+    modulus: *const u8,
+    modulus_len: u64,
+    result: *mut u8,
```
vadorovsky: 1. If `base` is a big number represented as an array, we need `base_len` as well.
2. Nit: we usually add `_addr` suffix to the pointer arguments.

```suggestion
    base_addr: *const u8,
    base_len: u64,
    exponent_addr: *const u8,
    exponent_len: u64,
    modulus_addr: *const u8,
    modulus_len: u64,
    result_addr: *mut u8,
```

vadorovsky (commented):
Overall looks good, one nit about the syscall interface.

proposals/0529-big-integer-modexp-syscall.md
zz-sol: I think we can assume `base_len` to be the same as `modulus_len`?

deanmlittle: well, the present version of the syscall doesn't do that. you can assume it's the same as the mod, but if you do then you'll have to zero pad the smaller array which is slightly slower.

0x0ece: I have an opinion here, since Poseidon we had to pay out a "bug" bounty because of extensions. No extensions.

The ring should be defined with a precise length, call it base_len, call it modulus_len, it doesn't matter to me. It's one len. Then base and modulus must be that amount of bytes, and the result will be the same number of bytes. No extensions, no zero padding.

Exponent is special because it can be significantly smaller, e.g. RSA. So we will enforce exponent_len <= len.

deanmlittle: @0x0ece All for reducing the surface area for bugs, but isn't the ring impl simply an implementation detail for the syscall itself, not its vm interface? We are copying these values out of the vm anyway. Wouldn't it make more sense to perform zeroing outside of the vm where it's cheaper? Doing a sized setup in SVM means we're likely performing multiple copies instead of just one. Also, is there a reason why we need to enforce exponent_len <= len? 🤔

samkim-crypto: I am neutral when it comes to adding a base length, but I think the check `exponent_len <= len` is good to have. By Fermat, we can always reduce the exponent, but I don't think this reduction is guaranteed by all libraries, which could result in DoS (at least under the CU formula in the current SIMD). If we strictly enforce `exponent_len <= len`, it prevents different validator clients from misimplementing it, thereby protecting against this risk.

zz-sol: > By Fermat, we can always reduce the exponent, but I don't think this reduction is guaranteed by all libraries

sorry my bad. I just realized that this requires `\phi(n)` which is not possible for RSA type moduli. I think it only works for prime numbers. 

OTOH, in terms of large exponent beyond modulus bits, the only application I can think of is RSA based VDF.

zz-sol: How about the following:

``` rust
    base_addr: *const u8,
    base_len: u64,
    exponent_addr: *const u8,
    exponent_len: u64,
    modulus_addr: *const u8,
    modulus_len: u64,
    result_addr: *mut u8,
```

with the following note:

- base_len must match modulus_len; or it will return an error?

deanmlittle: @zz-sol RSA accumulators

zz-sol:
Here is another concern.

- currently the cu cost is calculated based on the input parameters `modulus_len` and `exponent_len`. not based on the actual size of `modulus` and `exponent`
- the caller can, for whatever reasons, call the function with, say, `modulus_len = 2048` and `modulus = 11` (4 bits). For this case, **the caller is still charged as if modulus is 2048 bits**.
- the alternative method is to charge cu based on the actual modulus size, i.e., 4 bits.

I think the current method is more robust. the syscall still need to read 2048 bits of data, regardless of the actual modulus. so if we only charge for 4 bits it may be vulnerable to spamming.

deanmlittle:
> * currently the cu cost is calculated based on the input parameters `modulus_len` and `exponent_len`. not based on the actual size of `modulus` and `exponent`
> * the caller can, for whatever reasons, call the function with, say, `modulus_len = 2048` and `modulus = 11` (4 bits). For this case, **the caller is still charged as if modulus is 2048 bits**.

Length here is byte length, not bit length. 

> * the alternative method is to charge cu based on the actual modulus size, i.e., 4 bits.
> 
> I think the current method is more robust. the syscall still need to read 2048 bits of data, regardless of the actual modulus. so if we only charge for 4 bits it may be vulnerable to spamming.

I agree. This is not a real concern. It is our job to protect the chain from users trying to DoS it, not to protect users from having to do the bare minimum of taking 30 seconds to read a code comment to understand how to save CUs when consuming it.

proposals/0529-big-integer-modexp-syscall.md
```diff
@@ -0,0 +1,359 @@
...
+    modulus_addr: *const u8,
+    modulus_len: u64,
+    result_addr: *mut u8,
+);
+```
+
+Pointer fields are VM pointers to byte slices. `base`, `exponent`, and
+`modulus` point to readable input buffers. `result` points to a writable output
+buffer. Length fields are unsigned 64-bit values. The `base`, `modulus`, and
+`result` ranges are exactly `modulus_len` bytes. The `exponent` range is
```
samkim-crypto: I think this could be slightly ambiguous now that we have `base_len` field. Maybe we should explicitly specify that `base_len != modulus_len` is an abort condition for those who are looking at the simd for the first time without the PR discussion?

samkim-crypto (approved):
Just one minor comment, but I think this should be ready to go.

proposals/0529-big-integer-modexp-syscall.md
deanmlittle: Imo, we should either not enforce this constraint at all and just do what @vadorovsky suggested above (this is my preference, I think it's easy to clean this up in the syscall impl itself and doesn't have to happen in the vm), or follow @0x0ece's alternative suggestion of just having one `len` field for all of these fields with a separate one for exponent. Having two length fields *and* enforcing that they are equal is the worst of both worlds.

vadorovsky: I agree with @deanmlittle, I'm fine with either 1) not enforcing constraints; 2) having one `len` for base and modulus. From what I gather, the arguments for 2) and against 1) are:

* Worries about potential "bug" reports.
* The fact that arkworks users will likely use [`ark_ff::BigInt`](https://docs.rs/ark-ff/0.6.0/ark_ff/biginteger/struct.BigInt.html) with the same `N` for all numbers, by using one of the [type aliases](https://docs.rs/ark-ff/0.6.0/ark_ff/biginteger/index.html#types).

A counterargument for 1) and against 2) I can think of is that [`num_bigint::BigUint`](https://docs.rs/num-bigint/0.4.6/num_bigint/struct.BigUint.html#method.to_bytes_be) doesn't have any const generic and its [`BigUint::to_bytes_le()`](https://docs.rs/num-bigint/0.4.6/num_bigint/struct.BigUint.html#method.to_bytes_le) method doesn't add padding and will likely return vectors of different sizes for base and modulus.

No strong opinion from me, but perhaps mentioning these points and thinking of which library do we care more about (and whether we are going to use any of them in the syscall implementation), will make it easier for other people to decide.

vadorovsky: Anyway, in the current diff, the syscall function snippet doesn't match the text. The snippet uses separate `len`s, the text says `base` and `modulus` have the same `len`. Let's make it consistent regardless of the solution.

deanmlittle: It is my loosely held preference that we handle this with two different lengths and do not enforce equality in the API layer. We are copying a little endian value into a sized buffer. A length comparison must be done in the syscall anyway, so we may as well just max(a,b) instead. It is slightly more optimal in the VM this way.

0x0ece: Here's my final proposal:
- base, mod, result are all elements (of a ring), all the same len.
- exp is serialized as 2-byte exp_len followed by the actual bytes.

Because:
1. We have to stay in 5 params (including structs)
2. We want the syscall to do math and not worry too much about validating params (consensus risks, overhead)
3. In bpf we certainly want to reuse variables, for example it's common to save the result where we initially had the base, i.e. x := x^e mod q.

```
pub fn sol_big_mod_exp(
    len: u64,
    base_addr: *const u8,
    modulus_addr: *const u8,
    serialized_exponent_addr: *const u8,
    result_addr: *mut u8,
);
```

The alternative is that everything (or base, mod, exp... but then why not result?) is a (len, value) but handling all the cases makes everything unnecessarily messy (as per point 2 above).

And then, already agreed, everything is only little endian, and you pay what you declare to use as function of (len, exp_len).

deanmlittle: I believe we are discussing two different issues here, one being the API and the other being the implementation. With regards to the implementation, this is exactly correct:

>base, mod, result are all elements (of a ring), all the same len.

The main point of contention here is the API as it is consumed by the VM, and I think there are some things to unpack/justify here.

>We have to stay in 5 params (including structs)

It is true that syscalls must take in <=5 params, however a stack pointer to a struct containing our parameters only counts as a single param. This is exactly how the original syscall handled it, passing in a pointer to the BigModExpParams struct and unpacking it directly.

> exp is serialized as 2-byte exp_len followed by the actual bytes.

I think the existing u64 length makes sense so that each of the parameters of the struct itself remained 8-byte aligned (although pushing the length params to the end also achieves the same thing). I'm happy to concede if there is some kind of optimization here I am missing.

>The alternative is that everything (or base, mod, exp... but then why not result?) is a (len, value) but handling all the cases makes everything unnecessarily messy (as per point 2 above).

Firstly, we don't want result to have a length because it *must* be both large enough to house the modulus value, and cannot be larger than the modulus, so it makes sense for those two to share a length. 

I don't think this is messy at all. It is still a ring exactly as you stated, the only difference is that we do max(max(base,exp), mod) to determine the max length of all of the values before copying them to execute the syscall. Due to them all being little endian, this is actually really nice and simple.

Forcing users into a single length makes them unnecessarily blow up the stack with a bunch of superfluous zero allocations that we don't actually want to copy anyway. This is a pessimization to several common common use cases, including:

- forcing small exponent in RSA verification to match larger params
- forcing smaller params to match huge modulus length in RSA accumulators 
- forcing small base in Fermat/Miller-Rabin to match larger params

Imo, what @vadorovsky proposed above is the correct API if we go with 64-bit length values. If we do u16s, we should push them to the end to retain alignment. What @0x0ece proposed with the ring is the correct internal implementation, and what @zz-sol modeled appears to at least fall within the neighborhood of what I'd expect the correct CU costs to be.

samkim-crypto: Looking at this from the validator/consensus side, we have had issues in the past (like with the Poseidon syscall) where padding became a consensus hazard. Allowing different lengths for the base opens the door to more variations in subtle implementation details. Different backend libraries handle un-padded inputs differently, and we need to guarantee that a 1-byte `[3]` is treated identically to a 3-byte zero-padded `[3, 0, 0]` across all validator clients.

So IMO, the question boils down to whether we have a convincing use-case for the base number length to differ from the modulus length to justify the extra complication in the implementation.

- For RSA verification or RSA accumulators, the exponent length can be different but the base (the signature or accumulator state) will be the same length as the modulus.
- Primality testing (like Fermat/Miller-Rabin) is generally associated with private key generation, so I am not sure if we will ever need to execute it inside a smart contract (let me know if I am wrong though).
- For ECC arithmetic, we also always work in the underlying field. The base (e.g., an affine `X` or `Y` coordinate) is an element of that field, so it will always be serialized to the exact byte-length of the prime modulus.
- The only application that I can think of is maybe Wesolowski VDF verification where we hash 32 byte output as an integer to an RSA group with modulus 256 bytes. I am not sure if this is a strong enough application though as it is not really used in practice at the contract layer (everyone seems to be using vrfs instead).

samkim-crypto: As a side note, if we do ultimately allow the base to be greater than the modulus, then we should incorporate the CU model for the base length. The syscall would have to perform a preliminary reduction (`base % mod`) before exponentiation, which can be non-trivial if `base` is much greater than `mod` and I don't think the current CU formula accounts for this.

deanmlittle: > Different backend libraries handle un-padded inputs differently

I am proposing we *explicitly* use LE and allocate the max length of the largest parameter in the syscall implementation itself so that it is impossible to treat them differently. I actually do not see how the two proposed solutions differ in any way. All we're doing here is debating the merits of creating a bunch of zero padding in the VM where it's expensive and causes us to copy more information than required to execute the syscall, or in the syscall implementation itself where it's essentially free because we're copying anyway. Is there something I'm missing here?

> if we do ultimately allow the base to be greater than the modulus, then we should incorporate the CU model for the base length.

Isn't this both computationally insignificant and already paid for by the modulus price in the cost model?

samkim-crypto: > I am proposing we explicitly use LE and allocate the max length of the largest parameter in the syscall implementation itself so that it is impossible to treat them differently.

Okay, just to make sure that I understand what you are proposing, take the case `base > modulus`. Are we requiring that all validator client implementations to first pad the modulus by the length of the base in memory before the start of any computation? This seems a little counter intuitive to me and something that can easily be cheated by different validator implementations. Some validators can just compute `base % modulus` right off the bat and get a more efficient computation though they will technically be violating the SIMD standard.

I could be completely misunderstanding your point though.

> Isn't this both computationally insignificant and already paid for by the modulus price in the cost model?

If the base is very large bytes and the modulus is much smaller, then the computation will depend on the size of the base, not the modulus. Assuming that we cap all lengths to be 4096 bits, this computational is pretty small but non-trivial, so we should include the base length in the CU formula (if we don't, we will get a security advisory from someone I am sure 😅 ).

This is not an argument for or against having different lengths. I am just pointing out that we should account for the base length size if we do have different lengths.

0x0ece: We keep mentioning "a struct", but the simd has 7 params which is not good, and I've never seen the struct. Let's just write the struct in the simd, and describe in words all checks and other operations (zeroing, padding, whatever) that we have to do. Unless there's something completely wrong I'm ok with that solution too.

The reason I think the 5 params above are better is exactly because they represent the internal state.
I see the point on RSA accumulators in BPF, it's valid, but frankly I don't think they'll be the main application, and the extra cost in BPF doesn't mean that the cost is not somewhere else... the CPU still has to zero the 2-4kB or so of physical memory and then copy inside a bunch of bytes, it's faster just because our VM is slow, but it doesn't have to be this forever.

The overhead imo becomes an issue when we're trying to use the syscall to do small-ish multiplications e.g. for ECC or lattice or ZK. There, we're going to pay zeroing + copying (or padding, or whatever check we add) for every call. With the "internal representation" that I'm proposing, that cost goes away from the syscall.

deanmlittle: >We keep mentioning "a struct", but the simd has 7 params which is not good, and I've never seen the struct.

The struct is exactly what we're discussing in this thread. The syscall doesn't have 7 params. It currently has one, and should probably actually have two. That was intuitive to me because I'm familiar with the prior art, but if that was unclear to you, we should probably refine the SIMD language accordingly.

For reference, this is the structure of the existing (non-active) syscall in C: https://github.com/anza-xyz/agave/blob/4cd535ce52ac46892de4a964f943af1e4bd12d51/programs/sbf/c/inc/sol/inc/big_mod_exp.inc#L3

And this is the existing struct we are debating modifications to in this thread:
https://github.com/anza-xyz/solana-sdk/blob/14a725d6e9180e6cfbd98054473d61ef3aabde57/big-mod-exp/src/lib.rs#L4

As you can see, 2 parameters, both stack pointers – not 7.

>The overhead imo becomes an issue when we're trying to use the syscall to do small-ish multiplications e.g. for ECC or lattice or ZK.

Exactly. The 3 scenarios I outlined above are pretty much where this excessive zeroing makes no sense. RSA and Fermat are both very common use cases imo.

>With the "internal representation" that I'm proposing, that cost goes away from the syscall.

I believe the opposite is actually true. If we shift the cost into the VM, doesn't that mean we're paying for both slower zeroing *and* additional memory translation? Surely copying smaller numbers out of the VM is much faster here?

0x0ece: I'm pretty confident that if we debug the 2 impls:
- 5 params, you'll pay O(50 CUs) extra in the RSA accumulator case 
- struct, you'll pay O(10 CUs) extra for every syscall
(if things were measured perfectly in physical cpu cycles)

Anyway, as I said, I'm good with the struct as long as we write it in the simd, with the list of things I have to check when I parse the struct. The goal of simds is to have specs that can be impl independently, not that I have to go and copy agave code all the times :)

deanmlittle: Envelope math sounds right to me.

proposals/0529-big-integer-modexp-syscall.md
```diff
@@ -0,0 +1,378 @@
...
+document are to be interpreted as described in [RFC 2119] and [RFC 8174].
+
+### Syscall Interface
+
+Add the following program-facing syscall function:
+
+```rust
+pub fn sol_big_mod_exp(
+    params: *const BigModExpParams,
+    params_len: u64,
```
deanmlittle: You may have been referencing a similar syscall like Sha256 which has variable length inputs, but in the case of `sol_big_mod_exp`, we actually don't need the:

```rust
params_len: u64
```

`BigModExpParams` already has a known length. The second parameter here should actually be:

```suggestion
result: *mut u8
```

I'm not sure if there is a better reason as to why this is the convention we follow, but having the `result` pointer as a separate parameter creates a clean separation of inputs and outputs, and is how we have implemented all of our other syscalls that I'm aware of such as `sol_sha256` and `sol_curve_group_op`.

proposals/0529-big-integer-modexp-syscall.md
```diff
@@ -0,0 +1,378 @@
...
+
+Where `params` points to the following struct in VM memory:
+
+```rust
+#[repr(C)]
+pub struct BigModExpParams {
+    pub base: *const u8,
+    pub exponent: *const u8,
+    pub modulus: *const u8,
+    pub result: *mut u8,
```
deanmlittle: I would remove the result from `BigModExpParams` and make it a separate syscall parameter as is customary in all of our other syscalls

```suggestion
```

proposals/0529-big-integer-modexp-syscall.md
```diff
@@ -0,0 +1,378 @@
...
+    pub modulus_len: u64,
+    pub exponent_len: u64,
```
deanmlittle: Even if I were to ultimately be overruled and it is deemed that the efficiency gains in the handful of common use cases I've mentioned for differently-sized inputs are insufficient for us to want to pursue this functionality today, we should still include a `base_len` field here as a future-proofing measure. 

Without a separate `base_len` in the input parameters, the friction to remove such a constraint in the future should the need arise would be so great, it would render such a change more or less untenable. 

The length fields should also be listed in the same order as their respective input pointers:

```suggestion
    pub base_len: u64,
    pub exponent_len: u64,
    pub modulus_len: u64,
```

proposals/0529-big-integer-modexp-syscall.md
```diff
@@ -0,0 +1,378 @@
...
+}
+```
+
+`params` is a VM pointer to a readable `BigModExpParams` struct. `params_len`
+MUST equal `size_of::<BigModExpParams>()` (48 bytes); the syscall MUST abort if
+it does not.
```
deanmlittle:
```suggestion
`params` is a VM pointer to a readable `BigModExpParams` struct. `result` is 
a mutable pointer to an output buffer `modulus_len` bytes in size.
```

proposals/0529-big-integer-modexp-syscall.md
```diff
@@ -0,0 +1,378 @@
...
+
+All pointer fields inside the struct are VM pointers to byte slices. `base`,
+`exponent`, and `modulus` point to readable input buffers. `result` points to a
+writable output buffer. The `base`, `modulus`, and `result` ranges are exactly
+`modulus_len` bytes. The `exponent` range is exactly `exponent_len` bytes.
```
deanmlittle: I feel like making these a numbered list makes it a little more explicit and easier to understand:

```suggestion
All pointer fields inside the struct, along with the result pointer are VM 
pointers to byte slices of the following lengths:

1. `base` – `base_len` bytes
2. `exponent` – `exponent_len` bytes
3. `modulus` – `modulus_len` bytes
4. `result` – `modulus_len` bytes
```

proposals/0529-big-integer-modexp-syscall.md
```diff
@@ -0,0 +1,378 @@
...
+`modulus_len` MUST be greater than zero.
+
+### Output And Abort Behavior
+
+The function returns no value on success. There are no non-fatal error return
+values.
+
+The syscall MUST abort the virtual machine if any of the following are true:
+
+- `params_len` does not equal `size_of::<BigModExpParams>()` (48 bytes).
```
deanmlittle: We don't need `params_len`
```suggestion
```

proposals/0529-big-integer-modexp-syscall.md
```diff
@@ -0,0 +1,378 @@
...
+- The `params` pointer does not refer to a readable VM memory range of
+  `params_len` bytes.
+- `exponent_len` or `modulus_len` is greater than `BIG_MOD_EXP_MAX_BYTES`.
```
deanmlittle:
```suggestion
- `base_len`, `exponent_len` or `modulus_len` is greater than `BIG_MOD_EXP_MAX_BYTES`.
```

proposals/0529-big-integer-modexp-syscall.md
```diff
@@ -0,0 +1,378 @@
...
+- Any pointer plus length calculation overflows, including `base + modulus_len`,
```
deanmlittle:
```suggestion
- Any pointer plus length calculation overflows, including `base + base_len`,
```

proposals/0529-big-integer-modexp-syscall.md
```diff
@@ -0,0 +1,378 @@
...
+   `base + modulus_len`, `exponent + exponent_len`, `modulus + modulus_len`,
+   and `result + modulus_len`.
+5. Validate required input VM memory ranges are readable and the output VM
+   memory range is writable.
+6. Read all input bytes from VM memory, decode the modulus, and validate that it
+   is odd and greater than `1`.
+7. Determine the compute cost.
+8. Abort if the transaction does not have enough remaining compute units.
+9. Charge compute.
+10. Perform the exponentiation and write the result.
```
deanmlittle:
```suggestion
1. Validate and read the `params` struct from VM memory.
2. Validate all length fields, including maximum length checks and nonzero
   `modulus_len`.
3. Validate pointer plus length calculations for overflow, including
   `base + base_len`, `exponent + exponent_len`, `modulus + modulus_len`,
   and `result + modulus_len`.
4. Validate required input VM memory ranges are readable and the output VM
   memory range is writable.
5. Read all input bytes from VM memory, decode the modulus, and validate that it
   is odd and greater than `1`.
6. Determine the compute cost.
7. Abort if the transaction does not have enough remaining compute units.
8. Charge compute.
9. Perform the exponentiation and write the result.
```

deanmlittle (changes requested):
@zz-sol I've added a bunch more comments here. I think my rationale is sound. Please let me know what you think.

zz-sol:
Apology for going back and forth. @samkim-crypto and I discussed offline, and we find a good use case for which `base_len` makes sense. We can cast this `big_int_mod_exp` as a `mod` operation, when `base_len > modulus_len` and `exp = 1`. This can be quite helpful -- we are writing secp256r1 precompile and we can make some good use of this.

So the proposed API:
``` rust
pub fn sol_big_mod_exp(
    inputs: *const BigModExpParams,
    // inputs_len is a constant 
    result: *mut u8,
    // result_len is implicit == modulus_len
)
```



``` rust
pub struct BigModExpParams {
    pub base: *const u8,
    pub base_len: u64,
    pub exponent: *const u8,
    pub exponent_len: u64,
    pub modulus: *const u8,
    pub modulus_len: u64,
}
```
Will push a new version soon.

zz-sol:
Cost estimation for mod operations. so the cost formula seems consistent

| base/modulus bits | high time | measured CU | mult_complexity | model CU | model/measured |
| --- | ---: | ---: | ---: | ---: | ---: |
| 64/32 | 4.0628 us | 407 | 64 | 428 | 1.05x |
| 128/64 | 4.1465 us | 415 | 256 | 443 | 1.07x |
| 256/128 | 4.8712 us | 488 | 1,024 | 504 | 1.03x |
| 384/192 | 5.8135 us | 582 | 2,304 | 605 | 1.04x |
| 512/256 | 7.4192 us | 742 | 4,096 | 748 | 1.01x |
| 1024/512 | 11.426 us | 1,143 | 13,312 | 1,479 | 1.29x |
| 2048/1024 | 26.412 us | 2,642 | 37,888 | 3,429 | 1.30x |
| 4096/2048 | 85.762 us | 8,577 | 111,616 | 9,281 | 1.08x |

deanmlittle:
> Cost estimation for mod operations. so the cost formula seems consistent
> 
> base/modulus bits	high time	measured CU	mult_complexity	model CU	model/measured
> 64/32	4.0628 us	407	64	428	1.05x
> 128/64	4.1465 us	415	256	443	1.07x
> 256/128	4.8712 us	488	1,024	504	1.03x
> 384/192	5.8135 us	582	2,304	605	1.04x
> 512/256	7.4192 us	742	4,096	748	1.01x
> 1024/512	11.426 us	1,143	13,312	1,479	1.29x
> 2048/1024	26.412 us	2,642	37,888	3,429	1.30x
> 4096/2048	85.762 us	8,577	111,616	9,281	1.08x

lgtm

samkim-crypto (approved):

0x0ece (approved):

zz-sol:
> Thanks, [0x0ece](https://github.com/0x0ece)!
> 
> ⚠️ **Status:** Cannot merge yet
> 
> * Missing approval from: **Anza** ([t-nelson](https://github.com/t-nelson), [steviez](https://github.com/steviez), or [bw-solana](https://github.com/bw-solana))

Hi team, can I please get an approval on this

deanmlittle (dismissed):

samkim-crypto:
@t-nelson, @steviez, @bw-solana, can we get an approval on this please? 🙏

proposals/0529-big-integer-modexp-syscall.md
```diff
@@ -0,0 +1,419 @@
...
+    params: *const BigModExpParams,
+    result: *mut u8,
+);
+```
+
+Where `params` points to the following struct in VM memory:
+
+```rust
+#[repr(C)]
+pub struct BigModExpParams {
```
bw-solana, line 118: pedantic nitpick alert.

I think we should start defining these sorts of things like:
|Offset|Size|Field|
|-|-|-|
|0|8|base|

proposals/0529-big-integer-modexp-syscall.md
```diff
@@ -0,0 +1,419 @@
...
+
+Add a `sol_big_mod_exp` syscall that computes modular exponentiation over
+unsigned big integers:
+
+```text
+result = (base ^ exponent) mod modulus
+```
+
+The syscall is analogous to Ethereum's ModExp precompile specified by
+[EIP-198], but exposes a Solana-native syscall interface instead of the EVM
```
bw-solana, line 23: might be good to mention the even modulus restriction here given EIP-198 supports even, so it's not completely compatible

bw-solana (approved):
couple nits but I'm mostly good with this

crypt0miester:
some data from a second consumer of the 384-bit bucket: rfc 9380 hash_to_G1 for bls12-381, in-program (0388 has no hash-to-curve syscall). measured 392k CU end to end under mollusk, byte identical to blst, with 9 full-width modexps covering all the inversions/legendre/sqrts plus 2 of the exp=1 reductions with base_len > modulus_len from the june update. without this syscall each modexp becomes an in-program 381-bit exponentiation at ~2M CU and the hash goes from 392k to 20M+.

now that the spec is merged, is there any feeling for activation timeline? with 0388 already live this is the last missing primitive for the bls path.

zz-sol:
> now that the spec is merged, is there any feeling for activation timeline?

the spec is currently being implemented here https://github.com/anza-xyz/agave/pull/12321

PR #546, opened 2026-05-28
SIMD-0525: Remove signatures per slot change

bw-solana:
Amending #525 to remove the signatures per slot changes as part of the slot time reduction feature changes.

Reason is that these values are not actually used to change the base fee, so they don't need to change with slot time. Benefits of removing this is:

1. Less work to do / things to maintain
2. Avoids some confusion because the testnet bank value has diverged from genesis, so there's some weirdness about how certain things should be implemented in clients that can be avoided altogether

### DIFF
--- a/proposals/0525-reduce-slot-times.md
+++ b/proposals/0525-reduce-slot-times.md
@@ -132,7 +132,6 @@ new value does not exceed the intended wall-clock budget.
 | Limit | 400ms | 350ms | 300ms |
 |-------|-------|-------|-------|
 | Hashes/tick | 62500 | 54687 | 46875 |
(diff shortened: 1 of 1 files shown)


topointon-jump (approved):

bw-solana:
@apfitzge can you SME approve this one?

apfitzge (approved):
fee-rate governor is not tied to actual fees any longer - it is only used to maintain itself and should be removed separately.

There is no need for us to scale the target signatures per slot based on slot time, since it has no material effect.

bw-solana:
@steviez can you stamp?

steviez (approved):
> There is no need for us to scale the target signatures per slot based on slot time, since it has no material effect.

👌

PR #551, opened 2026-06-03
re-amend SIMD-0340: additional inter- and intra- validation

lidatong:
Amending SIMD-0340 again to be add additional cases (and be extremely explicit) about what FEC sets validly chain and what do not.

Both Firedancer and Agave already implement these checks (they are just not explicitly specified in the SIMD). Future clients should implement accordingly.


AshwinSekar (approved):
Looks good, as mentioned in the description both clients already fit this spec but better to be explicit

topointon-jump (approved):

steviez (approved):

PR #550, opened 2026-06-02
SIMD-0550: Double disinflation

lostintime101:

PR #553, opened 2026-06-04
SIMD-0553: Resource and Inclusion Fee

cavemanloverboy:
# Summary

(Note: Although it's reframed a bit, the design is identical to that in the [originating discussion](https://github.com/solana-foundation/solana-improvement-documents/discussions/547) with a higher `resource_fee_rate`, with the additional change of 5000->2500 to the base fee.)

SIMD-0553 splits today's 5000-lamport per-signature fee (2500 burned / 2500 to the leader) into:

- Base inclusion fee — 2500 lamports per signature, 100% to the leader (same leader pay as today, renamed).
- Resource fee — requested_cost_units × resource_fee_rate, 100% burned (replaces the flat burned half; scales with scheduler cost). Initial resource_fee_rate = 0.5.

Goal: tie SOL burn to real resource use. Low-resource-usage txs (votes, oracle updates) can pay slightly less than today's 5000 lamport base fee if we use a `resource_fee_rate` of 0.5; high-compute, zero-priority txs pay much more (see examples in the doc). 

Rough burn ~7.5–9k SOL/day vs ~648 SOL/day today.

### DIFF
--- a/proposals/0553-resource-fee-burn.md
+++ b/proposals/0553-resource-fee-burn.md
@@ -1,0 +1,431 @@
+---
+simd: '0553'
+title: Base Inclusion and Resource-based Fee
+authors:
+  - cavey
+category: Standard
+type: Core
+status: Draft
+created: 2026-06-03
+feature:
+  - BurnUeW8qFgJm3Y4uroMvqEQDrYWbhv7p231hHbrq8LV (`resource_fee_burn_1_10`)
+  - Burn2YJ9k9wTcyQCEMztMsbnBReV3FzXpW1aBB7MhwuY (`resource_fee_burn_1_4`)
+  - BurnSoLN7d4ASk4zVVxTg4ob3C8z71dN6cr8YsyEnaZ6 (`resource_fee_burn_1_2`)
+---
+
+## Summary
+
+We split the present static **5000-lamport per-signature fee** (2500 burned,
+2500 to the leader) into two components:
+
+- **Base inclusion fee** — static **2500 lamports** per transaction, paid
+  entirely to the leader (see **Base Inclusion Fee**).
+- **Resource fee** — dynamic, **100% burned**. Computed from
+  `requested_cost_units` and the resource fee rate (see **Resource Fee
+  Calculation**). The rate ramps in three feature-gated steps — **1/10**,
+  **1/4**, then **1/2** lamports per cost unit — replacing the burned half of
+  today's fee and scaling with scheduler cost.
+
+**Priority fees** are unchanged (100% to the leader, as per SIMD-0096).
+
+The mechanism is suitable for the network before and after Alpenglow activation.
+
+Originating discussion: [Discussion #547].
+
+## Motivation
+
+1. **Negligible SOL burn today.** At the current ~3000 TPS the network burns on
+   the order of ~648 SOL/day from signature fees alone — several orders of
+   magnitude below daily inflation (~60,000 SOL/day).
+
+2. **The current base fee does not price resources.** Users pay a single static
+   per-signature charge regardless of scheduler cost (compute, write locks,
+   loaded data, etc.). This leaves some users underpaying for compute-heavy
+   transactions.
+
+## Alternatives Considered
+
+**Uniform increase to the 5000-lamport per-signature fee.** Rejected.
+High-volume senders (market makers, validators pre-Alpenglow) pay mostly the
+flat charge; a uniform increase disproportionately harms them while leaving
+resource usage mispriced.
+
+**50/50 burn / leader split on the resource fee.** Rejected. The resource cost
+is borne by every node that replays and votes on the transaction, not only the
+leader. Paying the leader a share would create a much harder economic problem
+to model with second order effects on validator incentives and market structure
+(e.g. favoring inefficient programs that pay more resource fee to the leader,
+favoring compute-heavy swaps over cheap maker updates). Burning 100% keeps
+validator incentives and economics untouched for the most part.
+
+**Dynamic resource fee rate.** A controller that adjusts the resource fee rate
+each epoch from block utilization is a reasonable follow-up but adds
+predictability complexity. This SIMD specifies three static rates activated via
+separate feature gates (**1/10**, **1/4**, **1/2** lamports per cost unit); a utilization
+controller or further adjustments can come later.
+
+**SIMD-0110-style hot-account base fee.** Operates locally on short timescales
+from account hotness. This proposal is global, tied to per-transaction requested
+cost, and intended to move slowly (staged static rates via feature gates;
+dynamic tuning may follow).
+
+## New Terminology
+
+- **Base inclusion fee** — static **2500 lamports** per transaction to the
+  block leader; renamed from "base fee". Pays for inclusion, not resource
+  consumption. **Required** so the leader is paid even when the priority fee is
+  zero; otherwise game-theoretic incentives would lead leaders to drop zero-prio
+  txns.
+
+- **Resource fee** — lamports burned from `requested_cost_units` and the
+  resource fee rate (see **Resource Fee Calculation**).
+
+- **Requested cost units** — the pre-execution scheduler cost used for block
+  packing and cost tracking today. It is the sum of five cost-model components
+  (see **Resource Fee Calculation**). For program execution and loaded-account
+  data, the **requested** compute-unit limit and **requested** loaded-accounts
+  data size MUST be used (not consumed values).
+
+- **Resource fee rate** — lamports per requested cost unit, expressed as a
+  rational `resource_fee_rate_numerator / resource_fee_rate_denominator`.
+  Three staged values: **1/10**, **1/4**, and **1/2**. See **Feature Gates**.
+
+## Detailed Design
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this
+document are to be interpreted as described in [RFC
+2119](https://www.ietf.org/rfc/rfc2119.txt) and [RFC
+8174](https://www.ietf.org/rfc/rfc8174.txt).
+
+### Feature Gates
+
+Three feature gates SHALL be introduced:
+
+<!-- markdownlint-disable MD013 -->
+| Gate | Feature ID | Rate | Numerator | Denominator |
+| ---- | ---------- | ---- | --------- | ----------- |
+| `resource_fee_burn_1_10` | `BurnUeW8qFgJm3Y4uroMvqEQDrYWbhv7p231hHbrq8LV` | 1/10 | 1 | 10 |
+| `resource_fee_burn_1_4` | `Burn2YJ9k9wTcyQCEMztMsbnBReV3FzXpW1aBB7MhwuY` | 1/4 | 1 | 4 |
+| `resource_fee_burn_1_2` | `BurnSoLN7d4ASk4zVVxTg4ob3C8z71dN6cr8YsyEnaZ6` | 1/2 | 1 | 2 |
+<!-- markdownlint-enable MD013 -->
+
+When **any** of these gates is active for a bank, transaction fees MUST follow
+**Total Fee Calculation** below (base inclusion fee split plus resource fee).
+When **multiple** gates are active, implementations MUST use the **highest**
+effective resource fee rate from the table above.
+
+The gates SHOULD be activated in order (1/10 → 1/4 → 1/2). Activating them out
+of order is permitted; the effective rate is always the maximum among active
+gates. The staged rollout lets the network observe wallet, RPC, and application
+behavior at modest burn before reaching the terminal **1/2** rate.
+
+#### Activation timing
+
+Each gate follows the standard feature activation mechanism. If a gate is
+activated between epoch `E - 1` and epoch `E`, the corresponding rate change
+MUST apply starting at the **first slot of epoch `E`**. All earlier slots MUST
+use the previous effective fee model (pre-activation flat fee, or a lower
+resource fee rate if an earlier gate is already active).
+
+Transaction simulation and fee estimation at epoch boundaries therefore depend
+on which feature set the bank uses; clients SHOULD use the same feature
+activation rules as the runtime when simulating near a boundary.
+
+### Total Fee Calculation
+
+```text
+total_fee = base_inclusion_fee + priority_fee + resource_fee
+```
+
+| Component          | Formula                          | Burn | Leader |
+| ------------------ | -------------------------------- | ---- | ------ |
+| Base inclusion fee | 2500                             | 0%   | 100%   |
+| Priority fee       | CU price × limit (existing)      | 0%   | 100%   |
+| Resource fee (new) | see **Resource Fee Calculation** | 100% | 0%     |
+
+**Fee-only transactions.** In Agave, a transaction that fails during account
+loading may still be processed as **fee-only**
+(`ProcessedTransaction::FeesOnly`): the fee payer is debited the full
+**`total_fee`** even though the transaction does not execute. This behavior MUST
+be preserved after the first gate activates, now also including the resource
+fee.
+
+### Base Inclusion Fee (static component)
+
+The "base fee" / "transaction fee" SHOULD be called **base inclusion fee** in
+documentation, RPC fields, and user-facing surfaces.
+
+```text
+base_inclusion_fee = 2500
+```
+
+Leaders MUST receive a **nonzero reward for including a transaction even when
+the priority fee is zero**. Without it, game-theoretic fee maximization would
+leave zero-priority transactions unincluded. Present designs for multiple
+concurrent proposers introduce an inclusion fee; this split does not introduce a
+new fee category or additional complexity beyond what's to come.
+
+**Legacy per-signature base fee.** Today's 5000-lamport charge scales with
+signature count (`5000 × num_signatures`, split 50/50 burn / leader) because
+the original fee model priced each signature verification at the transaction
+level. The scheduler cost model also counts signatures in `signature_cost`, so
+signature work is priced twice today: once in the per-signature base fee and
+again (relatively cheaply) in `requested_cost_units`.
+
+**This SIMD.** The base inclusion fee is a flat **2500 lamports** per
+transaction — not `2500 × num_signatures`. Signature verification is priced
+only through `signature_cost` inside `requested_cost_units` and therefore the
+resource fee.
+
+**Follow-up path.**
+
+1. **This SIMD** — flat base inclusion fee; resource fee uses today's
+   (relatively cheap) `signature_cost` component in the cost model.
+2. **Follow-up SIMD** — if needed, after some investigation, raise signature
+   pricing in the cost model so multi-signer and precompile-heavy transactions
+   pay an appropriate resource fee after step 1. Specific constants are out of
+   scope here and belong in that proposal.
+
+Multi-signer transactions pay less fees than under today's per-signature model
+until that follow-up is activated.
+
+### Resource Fee Calculation (dynamic component)
+
+```text
+resource_fee = ceil_div(
+  requested_cost_units × resource_fee_rate_numerator,
+  resource_fee_rate_denominator)
+```
+
+Implementations MUST compute `resource_fee` using **integer arithmetic** only.
+
+The effective `resource_fee_rate_numerator` and
+`resource_fee_rate_denominator` MUST be taken from **Feature Gates** for the
+bank being executed or verified.
+
+`requested_cost_units` is the transaction's pre-execution scheduler cost — the
+same total computed for block packing and cost tracking today. Implementations
+MUST compute it as the saturating sum of:
+
+1. **Signature cost** (`signature_cost`) — cost of signature verification
+   (including precompile signatures accounted by the cost model).
+2. **Write-lock cost** (`write_lock_cost`).
+3. **Instruction-data cost** (`data_bytes_cost`).
+4. **Program-execution cost** (`programs_execution_cost`) — cost from the
+   transaction's **requested** compute-unit limit (Compute Budget).
+5. **Loaded-accounts-data-size cost** (`loaded_accounts_data_size_cost`) —
+   cost from the transaction's **requested** loaded-accounts data size.
+
+```text
+requested_cost_units = signature_cost
+                     + write_lock_cost
+                     + data_bytes_cost
+                     + programs_execution_cost      // requested CUs
+                     + loaded_accounts_data_size_cost  // requested bytes
+```
+
+(saturating u64 addition at each step, as in validator cost-model code today.)
+
+The resource fee MUST use this **requested** total, not post-execution
+consumed costs, for the same reasons as the priority fee:
+
+1. **Async execution.** Future async execution paths require a fee known before
+   execution completes.
+2. **Pre-execution validation.** The runtime can verify the fee payer can afford
+   the fee without executing the transaction.
+3. **Better UX.** Users know exactly how much they're paying before sending the
+   transaction.
+4. **Scheduling incentives.** Users should request accurate compute limits.
+
+Note that the sum of **requested** cost units across transactions in a block MAY
+exceed per-block packing limits.
+
+### Fee Collection and Burn
+
+Implementations MUST:
+
+1. Compute `base_inclusion_fee`, `resource_fee`, and `priority_fee`; burn the
+   full resource fee; route inclusion and priority fees to the collector.
+2. Deduct `total_fee` from the fee payer as today.
+
+### Validator Components Affected
+
+| Validator Component             | Impact                                |
+|---------------------------------|---------------------------------------|
+| Transaction Execution (Runtime) | Fee calculation, burn accounting      |
+| Virtual Machine                 | None                                  |
+| Block Packing                   | Fee payer check uses new total fee    |
+| Consensus                       | Burn totals in bank hash              |
+| Gossip                          | None                                  |
+| Turbine                         | None                                  |
+| TPU                             | Vote / Simple Vote fast-path updates  |
+| Snapshots                       | None                                  |
+| On-Chain Core BPF Programs      | None                                  |
+| Other (please describe)         | RPC fee responses, wallets, explorers |
+
+### RPC and client APIs
+
+Agave today exposes fees almost entirely as a **single lamport total** computed
+by `Bank::get_fee_for_message` / transaction simulation (`fee` crate →
+`FeeDetails::total_fee()`). Under this SIMD that total becomes
+`base_inclusion_fee + resource_fee + priority_fee`, but the JSON shape can
+stay scalar unless an implementation chooses to extend it.
+
+| RPC method                      | Field            | Change         |
+| ------------------------------- | ---------------- | -------------- |
+| `getFeeForMessage`              | `result`         | New total fee  |
+| `simulateTransaction`           | `fee`            | New total fee  |
+| `sendTransaction` (preflight)   | `fee` in error   | As simulation  |
+| `getTransaction`                | `meta.fee`       | Total charged  |
+| `getRecentPrioritizationFees`   | (priority only)  | Unchanged      |
+
+The **Banks** service
+(`get_fee_for_message_with_commitment_and_context`) follows the same bank path
+as `getFeeForMessage`.
+
+RPC implementations **SHOULD** ensure the above totals match **Total Fee
+Calculation**. They **MAY** add optional breakdown fields
+(`baseInclusionFee`, `resourceFee`, `priorityFee`) in a follow-up; this SIMD
+does not require a schema change. User-facing docs **SHOULD** stop treating
+`meta.fee` / simulation `fee` as “base fee only” and label them **total
+fee**.
+
+`getTransaction` already exposes `meta.costUnits` (requested scheduler cost);
+clients **MAY** use it with the resource fee rate to estimate resource fee, but
+**SHOULD** prefer `getFeeForMessage` / `simulateTransaction` for authoritative
+totals.
+
+## Impact
+
+**SOL holders.** Empirical network data (May 2026) suggests ~1,500–1,800,
+~3,750–4,500, and ~7,500–9,000 SOL/day from resource fees at **1/10**, **1/4**,
+and **1/2** lamports per requested cost unit respectively, each replacing the
+flat 2500-lamport-per-signature burn (~648 SOL/day at current throughput). The
+terminal **1/2** rate implies a meaningful net increase in total burn and roughly
+0.5% deflation against ~3.8% inflation; earlier stages burn less while the
+network validates the new fee model.
+
+**Retail / low-priority users.** Example transactions at the terminal
+`resource_fee_rate` of **1/2** (vote: 3765 requested cost units). At **1/10**
+or **1/4**, multiply the table's resource-fee column by **1/5** or **1/2**
+respectively (base inclusion and priority fees are unchanged):
+
+<!-- markdownlint-disable MD013 MD060 -->
+| Activity | Today | After | Δ |
+| -------- | ----- | ----- | -- |
+| Vote (w/ CB ixns) | 5000 | 2500 + 1883 | -12.3% |
+| [High Prio Swap (DFlow)][dflow-tx] | 5000 + 2002500 | 2500 + 197720 + 2002500 | +9.72% |
+| [Mid Prio Swap (OKX)][okx-tx] | 5000 + 130980 | 2500 + 412160 + 130980 | +301% |
+| [Zero prio Pumpfun swap][pump-tx] | 5000 + 0 | 2500 + 159910 + 0 | +3150% |
+| [Zerofi oracle update][zerofi-tx] | 5000 + 2568 | 2500 + 1220 + 2568 | -16.9% |
+<!-- markdownlint-enable MD013 MD060 -->
+
+At `resource_fee_rate` **1/2**, **low-resource-usage transactions win**: they pay
+less total fee than under today's flat 5000-lamport charge (see Vote and Zerofi
+rows). Burn tracks requested cost, rewarding efficient programs that request
+only the resources they need.
+
+**Compute-unit estimation.** Resource fee is charged on **requested** cost
+units, including the Compute Budget unit **limit**, not consumed compute.
+Applications that set a loose `SetComputeUnitLimit` (or never tighten it after
+simulation) pay resource fee on the full requested amount—the same basis as
+priority fees today, but now significant for **zero-priority** transactions
+that previously paid only the flat base fee. Over-estimating requested compute
+units increases `resource_fee` directly at `resource_fee_rate`.
+
+Note that legacy vote transactions as sent today request on the order of **54k**
+cost units — far more than the work a vote actually performs. At
+`resource_fee_rate` **1/2** that would raise pre-Alpenglow vote cost to nearly
+**6×** today's flat 5000-lamport fee. All client implementations **SHOULD**
+send votes with Compute Budget instructions that set requested compute units and
+loaded-accounts data size to reflect **actual vote work** (see the Vote row
+above, **3765** requested units). Omitting those instructions remains valid but
+is **NOT RECOMMENDED** under this fee model. [SIMD-0458] is live on
+mainnet-beta; vote transactions that include Compute Budget instructions no
+longer classify as Simple Vote, so implementations **SHOULD** update any TPU
+vote filter or fast-path logic that assumes the legacy Simple Vote shape.
+
+**Alpenglow.** Pre Alpenglow, validators submit **on-chain vote transactions**
+at high volume (on the order of ~1 SOL/day in vote fees today). The vote
+guidance above (requested cost, Compute Budget, TPU filters) matters directly
+for validator operating cost in that phase.
+
+Post Alpenglow, consensus **votes are no longer on-chain transactions**; they
+are exchanged directly between validators, and vote economics move to separate
+mechanisms (e.g. the validator admission ticket). **Total Fee Calculation**
+still applies to every on-chain transaction before and after Alpenglow, but the
+vote-specific impact above is primarily a (pre Alpenglow, migration) concern.
+
+**Wallets and SDKs.** MUST estimate fees using the same formula as
+**Resource Fee Calculation**, with the effective numerator and denominator for
+the bank epoch being simulated.
+
+## Security Considerations
+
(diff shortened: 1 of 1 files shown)


riptl (approved):

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,238 @@
...
+2500-lamport-per-signature burn (~648 SOL/day at current throughput) — a
+meaningful net increase in total burn and roughly 0.5% deflation against ~3.8%
+inflation.
+
+**Retail / low-priority users.** Example transactions at `resource_fee_rate`
+set to 0.5 (vote: 3765 requested cost units):
+
+| Activity | Today | After | Δ |
+| -------- | ----- | ----- | -- |
+| Vote (w/ CB ixns) | 5000 | 2500 + 1883 | -12.3% |
```
bw-solana, line 317: looks like we're implicitly suggesting we take the ceiling of `cost_units * resource_fee_rate`, we should specify that somewhere

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,238 @@
...
+
+### Total Fee Calculation
+
+```text
+total_fee = base_inclusion_fee + priority_fee + resource_fee
+```
+
+| Component          | Formula                                  | Burn | Leader|
+| ------------------ | -----------------------------------------| ---- | ------|
+| Base inclusion fee | 2500 lamports × num_signatures           | 0%   | 100%  |
```
bw-solana: pedantic: we should specify which signatures we're talking about here explicitly (just transaction level, or inclusive of precompiles)

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,238 @@
...
+
+### Resource Fee Calculation (dynamic component)
+
+```text
+resource_fee = requested_cost_units × resource_fee_rate
+```
+
+`resource_fee_rate` is **0.5** lamports per cost unit at activation.
+`requested_cost_units` is the transaction's pre-execution scheduler cost (same
+value used for block packing and cost tracking today).
```
bw-solana: might be good to be explicit about what goes into this. The sum of:

1. signature cost
2. write-lock cost
3. instruction-data cost
4. program-execution cost
5. loaded-account-data-size cost

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,238 @@
...
+| [Zerofi oracle update][zerofi-tx] | 5000 + 2568 | 2500 + 1220 + 2568 | -16.9% |
+
+At `resource_fee_rate` 0.5, **low-resource-usage transactions win**: they pay
+less total fee than under today's flat 5000-lamport charge (see Vote and Zerofi
+rows). Burn tracks requested cost, rewarding efficient programs that request
+only the resources they need.
+
+Note that simple votes as they exist today request 54k units! With a
+`resource_fee_rate` of 0.5 would skyrocket pre-Alpenglow vote cost to nearly
+6x current values! But, votes can request appropriate loaded account data
```
bw-solana: this seems like a de facto MUST when it comes to default client implementation.

Should we also mention SIMD-0387 as a dependency? I believe that is the one that creates the change in how vote costs are classified.

What is the vote cost after adding CBP ixs to set max CUs/data?

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,238 @@
...
+
+- **Resource fee** — `requested_cost_units × resource_fee_rate`, entirely
+  burned.
+
+- **Requested cost units** — pre-execution scheduler cost (signatures, write
+  locks, loaded accounts data, compute budget limit, etc.). Not consumed CUs.
+
+- **Resource fee rate** — lamports per requested cost unit; **0.5** at activation.
+
+## Detailed Design
```
bw-solana, line 93: might be good to specify exactly when the new fee model should take place. Presumably if it goes active on the transition from Epoch E -> E+1, you want the new fees in place starting with the first slot in E+1.

But this could create an awkward transition for anyone using transaction sim on the boundaries.

(I'm still fine with doing the obvious thing, just think we should spell it out)

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,238 @@
...
+increases meaningfully but remains well below inflation. Higher
+`resource_fee_rate` values increase burn further and hit **high-compute,
+low-priority** traffic hardest (see Pump.fun row).
+
+## Backwards Compatibility
+
+This is a consensus-breaking change gated behind a feature activation. Nodes
+that do not implement the resource fee will diverge on activation. All validator
+client implementations (Agave, Firedancer, Sig, etc.) MUST support this feature
+gate before mainnet-beta activation.
```
bw-solana: should also mention how fee changes could break existing flow. E.g. "low cost" txs that aren't setting CBP today that would incur much higher fees and could see transactions fail

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,238 @@
...
+**Minimum Cost Decreases** from 5000 to 3010 lamports.
+
+## Drawbacks
+
+**Modest burn vs inflation.** At 0.5 lamports per cost unit, aggregate burn
+increases meaningfully but remains well below inflation. Higher
+`resource_fee_rate` values increase burn further and hit **high-compute,
+low-priority** traffic hardest (see Pump.fun row).
+
+## Backwards Compatibility
```
bw-solana, line 383: we might also want to mention any RPC commands/return data that will need to change as part of this.

bw-solana (commented):
Left some comments.

That said, I'm a little concerned we don't have enough alignment on:

1. Pre/post Alpenglow and how to deal with votes
2. Using cost units as the ruler
3. The 0.5 multiplier

and thus talking about bits and bytes is time wasting

proposals/0553-resource-fee-burn.md
cavemanloverboy: yes, will update

proposals/0553-resource-fee-burn.md
cavemanloverboy: isn't it this one? already live?

image

bw-solana: @tao-stones can you chime in?

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,238 @@
...
+type: Core
+status: Draft
+created: 2026-06-03
+feature: 
+ - BurnSoLN7d4ASk4zVVxTg4ob3C8z71dN6cr8YsyEnaZ6
+---
+
+## Summary
+
+We split the present static **5000-lamport per-signature fee** (2500 burned, 2500
```
tao-stones: I understand the stated "split the present static (base fee) ..." as two separate proposals:
1. lower signature fee in half, reward 100% to leader
2. introducing requested-cu based fee, 100% burned. This better align burn and resource usage.

They are related, but don't have to depend on each other.

proposals/0553-resource-fee-burn.md
tao-stones: SIMD-0458 is live in mnb. The cost-model charges simple vote transaction same as normal transaction, it's compute units count is: 
1,440 signatures + 600 write locks + 19 instruction data + 200,000 execution allocation + 16,384 loaded-account data (for default 64MiB) = 218,443 units

tao-stones: If to use CBP ixs, client would request 1 page (32KiB) for loaded accounts, then request ~4,500CU for compute-units, which includes:
- 2,100 CU for Vote ix, 
- 2*150 =300 CU for 2 CBP ixs,
- 8 for loaded accounts size (for 1 page size)
- ~2,000 for signatures, write locks and instruction data 

Note, by adding CBP ixs to legacy vote transaction, it not longer qualifies as "Simple Vote" transaction. The concept of "Simple Vote" is going away, but the flag still exists and have effects on current execution path.

Client could also send vote with v1 transaction, setting compute-units and loaded-accounts-data-size without CBP.

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,238 @@
...
+
+The mechanism is suitable for the network before and after Alpenglow activation.
+
+Originating discussion: [Discussion #547].
+
+## Motivation
+
+1. **Negligible SOL burn today.** At the current ~3000 TPS the network burns on
+the order of ~648 SOL/day from signature fees alone — several orders of magnitude
+below daily inflation (~60,000 SOL/day). This makes the tokenomics of SOL unsound.
```
tao-stones: From original discussion, where condition's opposite: tons of traffic, lots of heavy transactions; with proposed 100% burn Resource Fee (based on often over-requested compute units), would daily burns exceed inflation significantly?

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,238 @@
...
+
+## New Terminology
+
+- **Base inclusion fee** — static **2500 lamports** per signature to the block
+  leader; renamed from "base fee". Pays for inclusion, not resource consumption.
+  **Required** so the leader is paid even when the priority fee is zero;
+  otherwise game-theoretic incentives would lead leaders to drop zero-prio txns.
+
+- **Resource fee** — `requested_cost_units × resource_fee_rate`, entirely
+  burned.
```
tao-stones: to clarify - in case of fee-only, tx will be charged by `Base inclusion fee` + `Resource fee`?

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,238 @@
...
+2. **The current base fee does not price resources.** Users pay a single static
+per-signature charge regardless of scheduler cost (compute, write locks, loaded
+data, etc.). This leaves some users underpaying for compute-heavy transactions.
+
+## New Terminology
+
+- **Base inclusion fee** — static **2500 lamports** per signature to the block
+  leader; renamed from "base fee". Pays for inclusion, not resource consumption.
+  **Required** so the leader is paid even when the priority fee is zero;
+  otherwise game-theoretic incentives would lead leaders to drop zero-prio txns.
```
tao-stones: Seems `number of signatures` are accounted in both `Base inclusion fee` and `Resource fee`. Why can't `Base inclusion fee` be a simple static number, so `number of signatures` are not double counted?

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,238 @@
...
+do a little more work, we should remove the multiplicand and update the cost
+model in a follow-up SIMD (which would require more research and fine-tuning)
+to more accurately price the signatures (along with everything else in the cost
+model).
+
+### Resource Fee Calculation (dynamic component)
+
+```text
+resource_fee = requested_cost_units × resource_fee_rate
+```
```
tao-stones: `Requested cost units` are weakly corresponds to `Resource usage`. Can `resource_fee` be defined with `requested_resource * resource_fee_rate`, where `requested_resource` is `requested_cost_units` upon activation, but can be expanded to other dimensions?

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,238 @@
...
+each epoch from block utilization (e.g. `resource_fee_rate = utilization`) is a
+reasonable follow-up but adds predictability complexity. This SIMD specifies
+**0.5** at activation; incremental complexity or adjustments can come later.
+
+**SIMD-0110-style hot-account base fee.** Operates locally on short timescales
+from account hotness. This proposal is global, tied to per-transaction requested
+cost, and intended to move slowly (static rate at activation; dynamic tuning
+may follow).
+
+## Impact
```
tao-stones, line 299: nit: should `Impact` section includes:
- changes economics for estimation cu limit, over-estimating = over paying, which may lead to large changes on existing app.

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,238 @@
...
+
+Note that simple votes as they exist today request 54k units! With a
+`resource_fee_rate` of 0.5 would skyrocket pre-Alpenglow vote cost to nearly
+6x current values! But, votes can request appropriate loaded account data
+and compute units to circumvent this
+
+**Wallets and SDKs.** MUST estimate
+`2500 × num_signatures + requested_cost_units × resource_fee_rate` (+ priority).
+
+## Security Considerations
```
tao-stones, line 364: Another possible consideration: potentially makes it cheaper (as much as half) to construct dos attack

proposals/0553-resource-fee-burn.md
cavemanloverboy: and priority fee, as is the case now. right?

proposals/0553-resource-fee-burn.md
cavemanloverboy: there's a note about this down below, ctrl+f "legacy"

proposals/0553-resource-fee-burn.md
tao-stones: right, was focusing on the base fee split part.

proposals/0553-resource-fee-burn.md
tao-stones: yea, was thinking setting `base inclusion fee` to a static value would make things a bit easier

cavemanloverboy: i agree but if you were to do that now **before** update cost model to make sigs more expensive, the cost per txn for multiple sigs would be very low

cavemanloverboy: so i think the move should be:

1. inclusion fee * sigs (for legacy reasons) with cheaper sig-component of resource based fee
2. increase sig-component of resource based fee by updating cost model
3. remove *sig multiplicand on inclusion fee

proposals/0553-resource-fee-burn.md
cavemanloverboy: this is here? i can explicitly mention dos but i assumed it would be implied

Image

proposals/0553-resource-fee-burn.md
cavemanloverboy: the reason why they must depend on each other is precisely because if done separately either

1. vote costs are way too high (if base stays the same with new resource fee)
2. dos attacks are super cheap (if base halves before new resource fee)

proposals/0553-resource-fee-burn.md
cavemanloverboy: no it is bounded and does not scale with traffic in the same way that priority fee does. it will certainly increase but only by a factor of 2 or so on average is my hunch... until block limits are removed...

proposals/0553-resource-fee-burn.md
cavemanloverboy: but my point i guess is that fee-only behavior stays the same (user still pays their full fee).

i agree however that the simd should be updated to state this explicitly. i will do this

proposals/0553-resource-fee-burn.md
cavemanloverboy: ok i will do this.

proposals/0553-resource-fee-burn.md
tao-stones: thanks for clarifying, it makes sense

proposals/0553-resource-fee-burn.md
cavemanloverboy: will do this

proposals/0553-resource-fee-burn.md
tao-stones: block limits only applies to *actual* compute units. *requested* compute units technically are not bounded by it (but do by tx CU limit, and indirectly by number of large txs can be packed in one block), heavy/complex transactions are likely to over-requesting CUs, potentially leads to over-burning.

MaxResnick:
I'm in favor but would prefer to keep the base signature fee where it is at 5000 lamports. I think its too low already at about $0.0003 per tx.

gwalen:
Writing this as an everyday Solana developer.

I’m not sure this is the right place for this kind of discussion, so forgive me if it isn’t. I found this while reading the latest Solana Changelog post on twitter.

We already have the base fee, priority fee, and, on top of that, tips such as Jito/Nozomi.

Now this proposal adds a resource fee, with a fairly complicated charging scheme.

While I understand the need to increase SOL burn, I think adding another fee component could be bad for DX and add more complexity.

From a developer’s point of view, this would mean dealing with 3 protocol-level fee components, plus tips on top, for a total of 4 separate payment components. Usually, blockchain developers only need to think about 2 protocol-level fee components.
This proposal would push Solana to 3 protocol-level fee components, plus external tips in practice. That makes the fee model harder to explain and reason about.

On top of that, many people will read this as: “transactions on Solana are getting more expensive.” Even if that is not true for every transaction, I think that will still be the public perception, which could be bad optics.

My main concern is day-to-day developer experience and the ease of onboarding new people into the Solana ecosystem.

That said, I know I have only a fraction of your knowledge about Solana internals, but I just wanted to share my “regular” Solana developer’s point of view.

arpoas:
I agree that keeping it as simple as possible is important. But in the same time I don´t have any better idea than the one presented by the author of this SIMD.

1) The burn has to increase somehow. It is not optional but absolutely necessary otherwise the current tragic supply-demand imbalance will inevitably keep pushing SOL price down over the long-term. Since the burning of 50% of the whole fee was abandoned, SOL started strongly underperforming things like BNB and TRX which may have not as good tech but maintain consistent deflationary policy. If you look at monthly charts it is clearly visible that the chart started behaving differently after that change. I am not advocating for pumping quickly SOL token by implementing some quick tokenomics fix and then cashing out. But Solana is not Linux. SOL can not be free or depreciating long-term, because the token value needs to guarantee security and attract capital and developers. So extra burn is absolutely essential for long-term viability of Solana.

2) What are the alternatives to resource-usage fees? Fees based on value of transaction are legally not possible. Flat fees would make simple transactions too costly. Resource-usage based fees are complicated but still they seem like the least bad option after all.

3) I did a quick calculation and even after the fee increase Solana average fee would be cheaper than Arbitrum or Base. This is based on the estimates mentioned by the author of this proposal and the average Solana, Base and Arbitrum fees last week according to Google AI. 

4) BNB and Tron have much higher fees and users are not leaving, on the contrary the number of users is increasing. A slight suggested increase of Solana fee in my opinion will not discourage new users. Fees is not everything. The liquidity, ecosystem, security, speed etc. ... users will be still getting the top tech and ecosystem for one of the lowest fees in the industry.

5) Solana at the moment is undercharging, because the fees do not cover the cost of security. Validators are still subsidised by issuance. And with the current burn mechanism even if there were many times more transactions Solana would still be inflationary. So fees and burn have to increase somehow for Solana to become economically viable. There is simply no way around it, and the sooner it is done, the better. This is not the case just for Solana but almost all pubic decentralized blockchains, so eventually they will also have to increase fees or fade into irrelevance due to ongoing long-term token price depreciation caused by structural supply surplus. Solana has the advantage that it is one of very few blockchains that can realistically achieve economic viability in very near future by combining slight increase of burn (this proposal), slight reduction of issuance (#0550) and general increase of network activity.

gwalen:
In principle, I agree with your arguments.
In particular, the point about deflation stands.
But, as always, there is a trade-off.

We are increasing complexity, and under the current design some low-fee, zero-priority transactions would become ~32x more expensive. That is a large jump.

Some kind of deflation mechanism may be needed, but I wonder whether this is the best way to achieve it.

We already have a [default CU limit per instruction](https://solana.com/docs/core/fees) when `SetComputeUnitLimit` is not explicitly set. Could we instead add a _default mandatory priority_ fee when none is specified, or when the specified fee is below some minimum threshold?
That would be less precise than the current proposal, but it could avoid introducing a third fee component (resource fee) and preserve the simpler fee model people are already used to, with only a small adjustment.

cavemanloverboy:
> I'm in favor but would prefer to keep the base signature fee where it is at 5000 lamports. I think its too low already at about $0.0003 per tx.

the long term plan outlined by the simd is to 

1. eventually turn the base fee into an inclusion fee (required for MCP)
2. increase cost of signatures by fixing signature cost (and other things while we are at it) in cost model.

so the base cost for 1-signature txns can be made higher and argued in the follow-up simd.

cavemanloverboy:
> In principle, I agree with your arguments. In particular, the point about deflation stands. But, as always, there is a trade-off.
> 
> We are increasing complexity, and under the current design some low-fee, zero-priority transactions would become ~32x more expensive. That is a large jump.
> 
> Some kind of deflation mechanism may be needed, but I wonder whether this is the best way to achieve it.
> 
> We already have a [default CU limit per instruction](https://solana.com/docs/core/fees) when `SetComputeUnitLimit` is not explicitly set. Could we instead add a _default mandatory priority_ fee when none is specified, or when the specified fee is below some minimum threshold? That would be less precise than the current proposal, but it could avoid introducing a third fee component (resource fee) and preserve the simpler fee model people are already used to, with only a small adjustment.

some low-fee, zero-priority **compute-heavy*** transactions will become 32x more expensive. this introduces an incentive for those inefficient programs in question to optimize their programs, which is a healthy incentive long term imo. there might be a 32x sudden jump for some, but after optimization (for the case of pump/helium, as simple as anchor v2 migration), the jump will be lower.

cavemanloverboy:
ok me and my clanker addressed everything here pretty sure. @tao-stones @bw-solana.

arpoas:
> In principle, I agree with your arguments. In particular, the point about deflation stands. But, as always, there is a trade-off.
> 
> We are increasing complexity, and under the current design some low-fee, zero-priority transactions would become ~32x more expensive. That is a large jump.
> 
> Some kind of deflation mechanism may be needed, but I wonder whether this is the best way to achieve it.
> 
> We already have a [default CU limit per instruction](https://solana.com/docs/core/fees) when `SetComputeUnitLimit` is not explicitly set. Could we instead add a _default mandatory priority_ fee when none is specified, or when the specified fee is below some minimum threshold? That would be less precise than the current proposal, but it could avoid introducing a third fee component (resource fee) and preserve the simpler fee model people are already used to, with only a small adjustment.

I am certainly up for any simplification. But making some minimum priority fee mandatory would not solve the burn problem, because all priority fees are paid as reward and not burned.  Vote on re-introducing again 50% priority fee burn would probably not pass if in the past super-majority voted for abandoning it. I am not sure that the understanding, that removing the burn actually reduced the profit for stakers by creating large supply/demand imbalance that pushes SOL price down long-term, has reached the majority of stakers yet ...

cavemanloverboy:
> > In principle, I agree with your arguments. In particular, the point about deflation stands. But, as always, there is a trade-off.
> > We are increasing complexity, and under the current design some low-fee, zero-priority transactions would become ~32x more expensive. That is a large jump.
> > Some kind of deflation mechanism may be needed, but I wonder whether this is the best way to achieve it.
> > We already have a [default CU limit per instruction](https://solana.com/docs/core/fees) when `SetComputeUnitLimit` is not explicitly set. Could we instead add a _default mandatory priority_ fee when none is specified, or when the specified fee is below some minimum threshold? That would be less precise than the current proposal, but it could avoid introducing a third fee component (resource fee) and preserve the simpler fee model people are already used to, with only a small adjustment.
> 
> I am certainly up for any simplification. But making some minimum priority fee mandatory would not solve the burn problem, because all priority fees are paid as reward and not burned. Vote on re-introducing again 50% priority fee burn would probably not pass if in the past super-majority voted for abandoning it. I am not sure that the understanding, that removing the burn actually reduced the profit for stakers by creating large supply/demand imbalance that pushes SOL price down long-term, reached the majority of stakers yet ...

in a future where there is protocol-enforced ordering (before and after MCP), resource fee vs priority fee are independent axes. i think this three-component fee is basically what we will end up at, so it introduces no complexity beyond what will eventually be introduced. This is beyond the scope here but MCP may introduce a fourth (success) fee, in fact:

1. inclusion fee paid to proposer (not leader)
2. priority fee, fully burned (used in sorting)
3. success fee, fully burned (used to break sorting tie. can be thought of as in-protocol "tip". figuring out why this doesn't just go into priority fee is left as an exercise to the reader :P)
3. resource fee, fully burned (not used to sort; does not affect in-protocol ordering at all)

under MCP, and more generally under any system with in-protocol ordering, it does not make sense for the leader to get any revenue otherwise the leader has an advantage on their own slots which is the whole thing in-protocol ordering is attempting to avoid.

This would be a post-mcp discussion, but if the full prio fee burn is enough after MCP, then i think we can make a plan to remove the resouce based fee after that. But that's about 18-24 months away and this SIMD can do quite a bit of good from now until then (and also hedge against further delays).

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,418 @@
...
+Wallets, SDKs, and transaction senders **SHOULD** set Compute Budget
+instructions with requested limits that reflect actual transaction work before
+activation, and **SHOULD** re-estimate `total_fee` under **Total Fee
+Calculation** rather than assuming the legacy flat fee. Programs **SHOULD**
+optimize on-chain compute so callers can request lower compute-unit limits and
+pay less resource fee. (Vote transactions are covered in **Impact** above.)
+
+## Conformance
+
+Implementations SHOULD verify:
```
bw-solana: MUST*

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,418 @@
...
+The mechanism is suitable for the network before and after Alpenglow activation.
+
+Originating discussion: [Discussion #547].
+
+## Motivation
+
+1. **Negligible SOL burn today.** At the current ~3000 TPS the network burns on
+   the order of ~648 SOL/day from signature fees alone — several orders of
+   magnitude below daily inflation (~60,000 SOL/day). This makes the tokenomics
+   of SOL unsound.
```
bw-solana: > This makes the tokenomics of SOL unsound.

advocacy language doesn't belong in the SIMD

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,418 @@
...
+**2** at activation; incremental complexity or adjustments can come later.
+
+**SIMD-0110-style hot-account base fee.** Operates locally on short timescales
+from account hotness. This proposal is global, tied to per-transaction requested
+cost, and intended to move slowly (static rate at activation; dynamic tuning
+may follow).
+
+## New Terminology
+
+- **Base inclusion fee** — static **2500 lamports** per signature to the block
```
apfitzge: I think the idea here is that zero-prio txns shouldn't just be dropped since they do not reward the leader.

Is there a reason this base reward to leader should scale with number of signature, aside from current behavior? Why not just use an actual static inclusion like 2500 lamports per txn?

My motivation in this comment is that it gives less of a special-case to signatures in terms of economic cost. Just treat them more similarly to how we treat other compute.
Additionally, we *already* model the computational cost of the different signatures differently...we have pre-compiles at nearly an order of magnitude ore expensive than top-level signatures, yet the cost to user is exactly the same! I'd rather we just move to something that is more consistent, this proposal is a fantastic opportunity to do so.

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,418 @@
...
+transactions before the cost model prices signatures more accurately in resource
+fee. A follow-up path:
+
+1. **This SIMD** — keep `× num_signatures` on inclusion fee; resource fee
+   uses the existing (relatively cheap) signature component in the cost model.
+2. **Follow-up** — raise signature pricing in the cost model.
+3. **Follow-up** — move to a flat base inclusion fee and remove the
+   per-signature multiplicand.
+
+Steps 2 and 3 are out of scope here.
```
apfitzge: see my previous comment - I do not think this is out of scope. Let's just move straight to a consistent model of flat_fee + resource_fee + priority_fee.

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,418 @@
...
+                     + programs_execution_cost      // requested CUs
+                     + loaded_accounts_data_size_cost  // requested bytes
+```
+
+(saturating addition at each step, as in validator cost-model code today.)
+
+The resource fee MUST use this **requested** total, not post-execution
+consumed costs, for the same reasons as the priority fee:
+
+1. **Async execution.** Future async execution paths require a fee known before
```
apfitzge, line 233: no disagreement, but even our current synchronous execution requires the fee to be known upfront since we take the fee *before* actual vm execution.

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,418 @@
...
+  (`BurnSoLN7d4ASk4zVVxTg4ob3C8z71dN6cr8YsyEnaZ6`).
+
+#### Activation timing
+
+Activation follows the standard feature activation mechanism. If the gate is
+activated between epoch `E - 1` and epoch `E`, **Total Fee Calculation** MUST
+apply starting at the **first slot of epoch `E`**. All earlier slots MUST use
+the pre-activation fee model.
+
+Transaction simulation and fee estimation at epoch boundaries therefore depend
```
apfitzge, line 131: Doesn't belong in this section, but do not see it mentioned elsewhere.

IMO we should explicitly call out nonces. I do not think they should get any special treatment here, after activation they will get charged more. However, this does introduce the case where a txn signed a very long time ago costs more if executed after the activation.

apfitzge (commented):
Overall very supportive of the SIMD. Compute on the network should not be free. Would like to see a couple clarifications and I feel pretty strongly that something you listed as out of scope should be brought within scope. See those comments.

I will also comment here about the general compute/cost unit contention. Cost units are the meter we have now, if we want to make a change now, they are the tool we have. IF we have a better measure in the future, I'm fully in support of moving to that, but we don't have it **now**.

arpoas:
> I'm in favor but would prefer to keep the base signature fee where it is at 5000 lamports. I think its too low already at about $0.0003 per tx.

I agree to keep the base fee as it is at 5000 of which 2500 is burned and 2500 goes to validators and just add a new resource-based fee.

So the total fee would be 
+ base fee (keep as it is)
+ resource-based fee (new, burn 100%)
+ priority fee (keep as it is)

I think it is not necessary to reduce the base fee when SIMD-0296 expands the maximum transaction packet size 4x.

proposals/0553-resource-fee-burn.md
cavemanloverboy: will rewrite this

proposals/0553-resource-fee-burn.md
cavemanloverboy: addressed

proposals/0553-resource-fee-burn.md
cavemanloverboy: 👍

proposals/0553-resource-fee-burn.md
bw-solana: agree. Use inclusion cost to:
1. Nudge things towards the 5k lamport "base fee" concept we're familiar with today w/o having to overhaul cost model.
2. Incentivize leader to pack
Separately, encompass signature cost as just another dimension of overall resource cost

bw-solana:
after chatting w/ @cavemanloverboy and @apfitzge my understanding is we'll rework to be something like the following:

1. Flat base inclusion fee to leader - decoupled from signature counts
2. Prio Fee unchanged (goes to leader)
3. Cost unit (resource) multiplier burn.

We will use a multi feature gate ladder up to 0.5 multiplier (likely 0.1, 0.25, 0.5). We will be conservative in ramping up multiplier so we can give people a chance to experience increased fees and optimize programs before their costs shoot up immediately. We also don't want to drastically increase cost for cheap operations like oracle updates

Part of the appeal of this approach is:

1. Not drag cost model updates into this SIMD, which would blow it up substantially
2. Not drastically overhaul market structure and builder incentives

Then in future SIMD(s), we get the cost model more accurate so that the resource costs align more closely to what they're costing the cluster.

I'm fine with 2.5k, 3k, 5k, or whatever as the base fee. I lean towards something like 4k lamports because it ends up fairly close to the 5k lamport min fee of today for most cheap operations and increasing operational cost of, e.g. oracle updates, feels like a silly side effect of what we're trying to do here and not the place to stall out for getting this across the line.

cavemanloverboy:
> after chatting w/ @cavemanloverboy and @apfitzge my understanding is we'll rework to be something like the following:
> 
> 1. Flat base inclusion fee to leader - decoupled from signature counts
> 2. Prio Fee unchanged (goes to leader)
> 3. Cost unit (resource) multiplier burn.
> 
> We will use a multi feature gate ladder up to 0.5 multiplier (likely 0.1, 0.25, 0.5). We will be conservative in ramping up multiplier so we can give people a chance to experience increased fees and optimize programs before their costs shoot up immediately. We also don't want to drastically increase cost for cheap operations like oracle updates
> 
> Part of the appeal of this approach is:
> 
> 1. Not drag cost model updates into this SIMD, which would blow it up substantially
> 2. Not drastically overhaul market structure and builder incentives
> 
> Then in future SIMD(s), we get the cost model more accurate so that the resource costs align more closely to what they're costing the cluster.
> 
> I'm fine with 2.5k, 3k, 5k, or whatever as the base fee. I lean towards something like 4k lamports because it ends up fairly close to the 5k lamport min fee of today for most cheap operations and increasing operational cost of, e.g. oracle updates, feels like a silly side effect of what we're trying to do here and not the place to stall out for getting this across the line.

keep in mind that the minimum cost model cost of a txn is 1080, so at the terminal resource fee rate of 1/2, the min txn fee is `inclusion_fee + 540`. This is why i don't think we should use 5000. 3000-4000 is reasonable. for oracles which use about ≈2000 cost units, it brings min fee to about 4000-5000 lamports (with zero priority).

however, note that moving this fee will impact leader incentives, as the validator income from the base fee increases by 20-60% (so the priority fee market structure will change). 3000 seems to be the right choice for optimizing these two things (minimizing impact to builder incentive, keeping txn fees close to original fee for small txns).

apfitzge (commented):
A few lints to fix, but I am happy with the proposal as is. Changes to cost-model can be done separately to make sure that signatures are not under-charged.

cavemanloverboy:
> A few lints to fix, but I am happy with the proposal as is. Changes to cost-model can be done separately to make sure that signatures are not under-charged.

fixed lints (skip line length for tables with pubkeys)

apfitzge (approved):

bw-solana (approved):

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,452 @@
...
+   pay an appropriate resource fee after step 1. Specific constants are out of
+   scope here and belong in that proposal.
+
+Multi-signer transactions pay less fees than under today's per-signature model
+until that follow-up is activated.
+
+### Resource Fee Calculation (dynamic component)
+
+```text
+resource_fee = ceil(
```
ptaffet-jump: I think `ceil` is a typo here. Do you mean `ceil_div` or do you mean `ceil` with a `/` instead of a `,`?

cavemanloverboy: yes, typo, thanks for catching. `ceil_div`

cavemanloverboy: i think i messed up when cleaning this up... there's another block below

cavemanloverboy: fixed this

arpoas:
> after chatting w/ @cavemanloverboy and @apfitzge my understanding is we'll rework to be something like the following:
> 
> 1. Flat base inclusion fee to leader - decoupled from signature counts
> 2. Prio Fee unchanged (goes to leader)
> 3. Cost unit (resource) multiplier burn.
> 
> We will use a multi feature gate ladder up to 0.5 multiplier (likely 0.1, 0.25, 0.5). We will be conservative in ramping up multiplier so we can give people a chance to experience increased fees and optimize programs before their costs shoot up immediately. We also don't want to drastically increase cost for cheap operations like oracle updates
> 
> Part of the appeal of this approach is:
> 
> 1. Not drag cost model updates into this SIMD, which would blow it up substantially
> 2. Not drastically overhaul market structure and builder incentives
> 
> Then in future SIMD(s), we get the cost model more accurate so that the resource costs align more closely to what they're costing the cluster.
> 
> I'm fine with 2.5k, 3k, 5k, or whatever as the base fee. I lean towards something like 4k lamports because it ends up fairly close to the 5k lamport min fee of today for most cheap operations and increasing operational cost of, e.g. oracle updates, feels like a silly side effect of what we're trying to do here and not the place to stall out for getting this across the line.


So under this model the only thing that gets burned is the new resource fee?

What is the estimated burn if it gets implemented with the suggested multipliers?

cavemanloverboy:
> > after chatting w/ @cavemanloverboy and @apfitzge my understanding is we'll rework to be something like the following:
> 
> > 
> 
> > 1. Flat base inclusion fee to leader - decoupled from signature counts
> 
> > 2. Prio Fee unchanged (goes to leader)
> 
> > 3. Cost unit (resource) multiplier burn.
> 
> > 
> 
> > We will use a multi feature gate ladder up to 0.5 multiplier (likely 0.1, 0.25, 0.5). We will be conservative in ramping up multiplier so we can give people a chance to experience increased fees and optimize programs before their costs shoot up immediately. We also don't want to drastically increase cost for cheap operations like oracle updates
> 
> > 
> 
> > Part of the appeal of this approach is:
> 
> > 
> 
> > 1. Not drag cost model updates into this SIMD, which would blow it up substantially
> 
> > 2. Not drastically overhaul market structure and builder incentives
> 
> > 
> 
> > Then in future SIMD(s), we get the cost model more accurate so that the resource costs align more closely to what they're costing the cluster.
> 
> > 
> 
> > I'm fine with 2.5k, 3k, 5k, or whatever as the base fee. I lean towards something like 4k lamports because it ends up fairly close to the 5k lamport min fee of today for most cheap operations and increasing operational cost of, e.g. oracle updates, feels like a silly side effect of what we're trying to do here and not the place to stall out for getting this across the line.
> 
> 
> 
> 
> 
> So under this model the only thing that gets burned is the new resource fee?
> 
> 
> 
> What is the estimated burn if it gets implemented with the suggested multipliers?

https://burn.cavey.xyz

proposals/0553-resource-fee-burn.md
```diff
@@ -0,0 +1,431 @@
...
+
+```text
+requested_cost_units = signature_cost
+                     + write_lock_cost
+                     + data_bytes_cost
+                     + programs_execution_cost      // requested CUs
+                     + loaded_accounts_data_size_cost  // requested bytes
+```
+
+(saturating addition at each step, as in validator cost-model code today.)
```
ptaffet-jump: `saturating u64`, though all these values are bounded elsewhere so that overflow even of int32 can't happen IIRC

topointon-jump (approved):

arpoas:
Is there some discussion going on under the SGP or somewhere else after passing the 15% interest? Or this here is still the primary place to discuss the proposal?

cavemanloverboy:
> Is there some discussion going on under the SGP or somewhere else after passing the 15% interest? Or this here is still the primary place to discuss the proposal?

sgp probably best place for discussion

PR #561, opened 2026-06-10
SIMD-0560: Mandatory Delegation of Surplus Lamports (#1)

staccDOTsol:
## Summary

This SIMD proposes that the runtime sweep every lamport above an account's
rent-exempt minimum into approved liquid staking positions, unwinding
automatically on debit and on account close. Holding undelegated SOL above
the rent floor becomes a state the protocol no longer represents.

Idea discussion (per SIMD-0001):
<https://github.com/solana-foundation/solana-improvement-documents/discussions/560>

## What this PR is actually for

The proposal is submitted in the expectation of rejection, and says so in its
own terms: its success criterion is not feature-gate activation but a
rejection rationale articulated at the same rigor as the objections the
document itself enumerates (enshrined-cartel layering, float conservation and
bank-run dynamics, yield convergence to zero at forced full participation,
and the rebasing reduction). The protocol currently holds an implicit,
unwritten position on what idle SOL is for; this PR is a forcing function to
get that position written down.

Reviewers are accordingly asked to reject it well.

## Notes

- Currently numbered 0560 to match the discussion; will renumber to this
  PR's number once assigned, per SIMD-0001.
- Passes markdownlint and the custom SIMD linter with 0 errors.
- Appendix A (the wSOL extension) is deliberately severable; Appendix B
  lists open empirical items with exact-figure methodology.
- Alternatives Considered, item 1 (inactivity-gated sweep of terminal
  program state) is the salvageable artifact if the body dies as intended.

### DIFF
--- a/proposals/0560-mandatory-delegation-of-surplus-lamports.md
+++ b/proposals/0560-mandatory-delegation-of-surplus-lamports.md
@@ -1,0 +1,303 @@
+---
+simd: '0560'
+title: Mandatory Delegation of Surplus Lamports
+authors:
+  - Jarett Dunn (Stacc)
+category: Standard
+type: Core
+status: Idea
+created: 2026-06-10
+feature: TBD (simd_0560_mandatory_delegation)
+---
+
+## Summary
+
+Every lamport above an account's rent-exempt minimum is delegated. At
+feature-gate activation, and thereafter at every lamport credit, the runtime
+converts an account's surplus lamports into an approved liquid staking
+position held against the account, and unwinds that position automatically on
+transfer-out and on account close. Holding undelegated SOL above the rent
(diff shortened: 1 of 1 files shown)


staccDOTsol:
I hadn’t read “thoroughly discussed and vetted”. There is a discussion, now there’s pr. If there’s thoroughness required, I beg you to be honest from the start.

jacobcreech:
> I hadn’t read “thoroughly discussed and vetted”. There is a discussion, now there’s pr. If there’s thoroughness required, I beg you to be honest from the start.

Hey @staccDOTsol I'm going to go ahead and close this for now. Let's get some concept acks and discussion going before making a formal simd

PR #562, opened 2026-06-13
Remove mentions of SIMD-0161

plafer:
[SIMD-0161](https://github.com/solana-foundation/solana-improvement-documents/pull/161) was proposed but later closed. Subsequent SIMDs still refer to it; this commit removes all such mentions.

To my knowledge, given that SIMD-0161 no longer exists, there are no document describing the repurposing of the ELF header's `e_flags` field to encode the SBPF version. Hence, I opted to simply remove the references to SIMD-0161.

### DIFF
--- a/proposals/0166-dynamic-stack-frames.md
+++ b/proposals/0166-dynamic-stack-frames.md
@@ -131,8 +131,8 @@ retrieving the caller’s frame pointer address to access those parameters.
 
 ### Identification of programs
 
-As per the description in SIMD-0161, programs compiled with dynamic stack 
-frames must contain the `0x01` flag on their ELF header `e_flags` field.
+Programs compiled with dynamic stack frames must indicate the SBPF version 
+`0x01` or higher in their ELF header `e_flags` field.
 
 ## Impact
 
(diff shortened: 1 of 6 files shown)


bw-solana:
@Lichtso can you SME?

Lichtso (approved):

bw-solana:
Removing reference to 161 is good 👍

But several of these affected SIMDs are version-gated, and SIMD-0161 was the only cited place that explained how the SBPF version is derived from ELF e_flags.

After this PR, 173/174/189 say “program header,” which is ambiguous/wrong because e_flags is in the ELF file header, not an ELF program header.

I'm thinking we should either:

1. replace the deleted references with explicit definition of SBPF version resolution from ELF e_flags
2. add a small reference doc and point these SIMDs to it

Lichtso:
> add a small reference doc and point these SIMDs to it

Ironically that was exactly what SIMD-0161 was. A reference point for other SIMDs to build on.

bw-solana:
this is where a canonical spec is useful

plafer:
>I'm thinking we should either:
>
>   1. replace the deleted references with explicit definition of SBPF version resolution from ELF e_flags
>   2. add a small reference doc and point these SIMDs to it

Can we hope to ever reopen and merge SIMD-0161? Indeed I find SIMD-0161 to be way better since it gives all the appropriate context. Otherwise I'll do (1).

PR #563, opened 2026-06-15
SIMD-0563: Keccak-p1600 syscall

zz-sol:
Introduce a `sol_keccak_p1600` syscall that applies the half 12-round or full 24-round Keccak-p[1600] permutation to a caller-provided 1600-bit state. The syscall is lower level than a SHAKE or SHA3 hash syscall: programs and SDKs build the
desired sponge mode around it by applying the correct rate, padding, domain separation suffix, and output handling.

Exposing the permutation instead of one fixed mode supports SHAKE128, SHAKE256, SHA3-224, SHA3-256, SHA3-384, SHA3-512, legacy Keccak-256, and protocol-specific Keccak sponge or duplex constructions with one consensus ABI.

------

the cost is derived as follows:

- a single `keccak-f[1600]` takes 132.85 ns
- a single `keccak(135 bytes)` takes 159.37 ns; the cost we are currently charging for this is `85 + max(10, 135 / 2 * 1) = 152 CU`
- the `keccak-f[1600]` is therefore budget at `152 * 132.85 / 159.37 = 126.71 ~= 130` CU
- for all the derived functions (SHAKE128, SHAKE256, SHA3-224, SHA3-256, SHA3-384, SHA3-512), the costs are determined by the number of keccak-f calls


samkim-crypto:
Hey @deanmlittle, do you have any thoughts on this?

deanmlittle:
> Hey @deanmlittle, do you have any thoughts on this?

@0x0ece's idea of letting you choose full or half number of rounds to support turbo variants is smart imo. other than that, i think it makes sense.

zz-sol:
I refactored to keccak_p1600 syscall. So it supports the turbo variants.
Note that the round parameters are limited to 12/24 to avoid accidental misuse.

deanmlittle (approved):

0x0ece (approved):

samkim-crypto (approved):

samkim-crypto:
@t-nelson, @steviez, @bw-solana, can we get an approval for this?

zz-sol:
prefer #579

PR #565, opened 2026-06-16
SIMD-0565: Secp256r1 curve syscalls

samkim-crypto:


deanmlittle:
ACK from me. This looks good. The only thing I would suggest is perhaps checking out my work on Secp256k1 here: https://github.com/blueshift-gg/solana-secp256k1

So long as we have add/mul, everything else is actually already very cheap. MSM is definitely a nice addition though. Some operations like point negation/compression (I see you omitted compression already, that's actually the biggest offender) are either cheaper, or at least cheap enough to do directly in BPF.

If you still want to have those operations in the syscalls for consistency's sake across APIs, I have no strong objections though.

0x0ece (approved):

zz-sol (approved):

proposals/0565-secp256r1-syscalls.md
```diff
@@ -0,0 +1,232 @@
...
+### Addition and Scalar Multiplication
+
+The existing `sol_curve_group_op` syscall will be extended to support
+`secp256r1`.
+
+```rust
+define_syscall!(fn sol_curve_group_op(
+    curve_id: u64,
+    group_op: u64,
+    left_input_addr: u64, // interpreted as *const u8
```
deanmlittle: Why do we want to put an ambiguous type here?

samkim-crypto: My bad, some of the syscall libraries in the solana-sdk are meant to be run on multiple platforms and we try to use `u64` instead of pointers since pointers are not abi-stable. For syscall functions themselves, we will ever run this in the vm, so it is fine to use pointer types. I reverted the changes to use pointers as before 🙏

samkim-crypto:
@t-nelson, @steviez, @bw-solana, can we get an approval for this?

proposals/0565-secp256r1-syscalls.md
```diff
@@ -0,0 +1,232 @@
...
+    left_input_addr: *const u8,
+    right_input_addr: *const u8,
+    result_point_addr: *mut u8
+) -> u64);
+```
+
+The syscall will use the newly defined `SECP256R1_{LE,BE}` constants. The group_op
+parameter will use the existing `ADD`, `SUB`, and `MUL` constants.
+
+Validation rules for `secp256r1` operands:
```
bw-solana, line 135: I'm wondering if we need more detail around the validation rules. Like what happens if scalars are out of range? Which error do we report?

proposals/0565-secp256r1-syscalls.md
```diff
@@ -0,0 +1,232 @@
...
+
+To provide flexibility across different application architectures, we define
+both little-endian (LE) and big-endian (BE) variants for scalar and field
+element byte-ordering within these fixed-length arrays.
+
+We note that this deviates from the standard SEC1 encoding. SEC1 is a
+variable-length format (e.g., representing the identity point as a single
+`0x00` byte), making it more error-prone for validator clients to handle.
+
+The exception to this rule is the `sol_curve_decompress` syscall. This
```
bw-solana: is decompression BE only?

proposals/0565-secp256r1-syscalls.md
```diff
@@ -0,0 +1,232 @@
...
+`secp256r1`. The required operations are:
+
+1. Group operations (addition, subtraction, and scalar multiplication)
+2. Multiscalar multiplication (MSM)
+3. Point validation
+4. Decompression operations
+
+### Curve Specification and Encoding
+
+For the curve definition, the standard FIPS 186-4 specification for NIST P-256
```
bw-solana: 186-5*?

proposals/0565-secp256r1-syscalls.md
```diff
@@ -0,0 +1,232 @@
...
+generalized curve operations follows the established architectural direction of the
+network.
+
+## Impact
+
+This will support the strategic migration of cryptographic precompiles to Core BPF
+and enable hardware-secured WebAuthn and Passkey applications to be built natively
+on Solana.
+
+## Security Considerations
```
bw-solana, line 264: might want to call out cost tracker accuracy here and any details around CU costs if we have them already

proposals/0565-secp256r1-syscalls.md
samkim-crypto: Okay added more detail on input validation including scalar range. If you think more detail is needed, let me know.

proposals/0565-secp256r1-syscalls.md
samkim-crypto: Added clarification.

proposals/0565-secp256r1-syscalls.md
samkim-crypto: Ah good catch. It seems like FIPS 186-4 was broken up into: 
- FIPS 186-5 for the ECDSA algorithm spec
- NIST SP 800-186 for the curve spec

Since we just need the curve spec, I replaced FIPS 186-4 with NIST SP 800-186.

proposals/0565-secp256r1-syscalls.md
samkim-crypto: Good point. Added.

bw-solana (approved):

bw-solana (commented):
LGTM

but looks like some lints to resolve

samkim-crypto:
Okay, all lints seem to pass now so let me merge!

PR #566, opened 2026-06-16
Clarify SIMD-0392 delegation adjustment timing

joncinque:
#### Problem

During partitioned epoch rewards distribution, although it isn't possible to invoke the stake program to modify stakes, it is possible to credit lamports to stake accounts.

The current stake adjustment logic in SIMD-0392 performs the adjustment at calculation time, but it is possible for a stake account to be credited lamports prior to distribution, potentially removing any need to adjust the stake delegation.

#### Summary of changes

Update SIMD-0392 to specify that adjustment happens at distribution time, and not at calculation time.

### DIFF
--- a/proposals/0392-rent-increase-adaptations.md
+++ b/proposals/0392-rent-increase-adaptations.md
@@ -193,8 +193,8 @@ section for more details.
 
 ### Rent-Adjusted Stake Delegations
 
-During the epoch rewards calculation phase, a stake's updated delegation MUST be
-calculated with the following formula:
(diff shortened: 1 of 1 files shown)


joncinque:
@igor56D can you take a look?

buffalojoec (commented):
Changes LGTM, but I still think we should *not* be doing this at all. 😊

bw-solana (approved):

topointon-jump (approved):

PR #567, opened 2026-06-17
SIMD-0567: CU-optimized ATA Program (`p-ATA`)

grod220:
Proposal to replace the current version of the Associated Token Account Program (`ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL`) with a CU-optimized Pinocchio version: `p-ATA`.

### DIFF
--- a/proposals/0567-cu-optimized-ata-program.md
+++ b/proposals/0567-cu-optimized-ata-program.md
@@ -1,0 +1,240 @@
+---
+simd: '0567'
+title: "p-ATA: CU-optimized ATA Program"
+authors:
+  - Gabe R. (Anza)
+  - Peter K. (Ergonia)
+category: Standard
+type: Core
+status: Review
+created: 2026-06-17
+feature: TBD
+---
+
+## Summary
+
+This proposal outlines a plan to replace the current Associated Token Account
+program (`ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL`) with a Pinocchio-based,
+CU-optimized implementation (`p-ATA`) at the same program address.
+
+The replacement is a drop-in implementation for the existing `Create`,
+`CreateIdempotent`, and `RecoverNested` instructions with the same ABI. It also
+adds a new `CreateWithArgs` instruction for callers that can provide optional
+optimization hints.
+
+## Motivation
+
+ATA creation is one of the most frequently executed instructions on Solana. This
+program governs the standard that derives canonical token addresses and handles
+the CPIs to create the token accounts. This makes the ATA program a
+high-leverage target for compute reduction.
+
+On a 30k confirmed-block stratified mainnet sample:
+
+- ATA appears in 11.85% of transactions
+- ATA share of total CU usage: 13.28%
+- Top invoked programs (ranked):
+    1. spl token (p-token already live),
+    2. compute budget (Transaction V1 makes this ix unnecessary)
+    3. vote (removed in Alpenglow)
+    4. system (a runtime built-in)
+    5. ATA (the next best candidate to pinocchio-ize)
+
+[P-Token][SIMD-0266] was a strong validation of the pinocchio philosophy:
+zero-copy + no-std + no-alloc + low deps. The CU usage of the token program
+effectively collapsed (-93%) with the usage as active as before. No userflow
+changes, but simply meaningfully freed block capacity on the network. The same
+results can be had with the ATA program.
+
+## Impact
+
+The average CU reduction (weighted by instruction frequency) of `p-ATA`
+instructions is 80.9% against the legacy ATA program baseline. Given ATA's share
+of total CU usage, we can expect a cluster-wide CU reduction of ~10-11%.
+
+| Case                    | Share of use | Legacy | p-ATA | Reduction |
+|-------------------------|-------------:|-------:|------:|----------:|
+| idempotent new SPL      |       39.86% | 22,940 | 4,171 |    -81.8% |
+| idempotent existing SPL |       30.32% |  3,710 |   548 |    -85.2% |
+| idempotent existing T22 |       18.60% |  8,210 | 1,634 |    -80.1% |
+| create SPL              |        5.65% | 18,433 | 3,083 |    -83.3% |
+| idempotent new T22      |        3.65% | 15,474 | 5,496 |    -64.5% |
+| create T22              |        1.91% | 13,967 | 5,132 |    -63.3% |
+| recover_nested SPL/SPL  |        0.00% | 26,806 | 5,196 |    -80.6% |
+
+`CreateWithArgs` is a new proposed instruction where callers can cheaply provide
+the bump, account length, and rent sysvar. Ordered by observed use where
+available, it saves the following on creation and idempotent paths:
+
+| Case                    |  Base | WithArgs | Delta |
+|-------------------------|------:|---------:|------:|
+| idempotent new SPL      | 4,171 |    3,984 |  -187 |
+| idempotent existing SPL |   548 |      606 |   +58 |
+| idempotent existing T22 | 1,634 |    1,695 |   +61 |
+| create SPL              | 3,083 |    2,891 |  -192 |
+| idempotent new T22      | 5,496 |    5,303 |  -193 |
+| create T22              | 5,132 |    4,934 |  -198 |
+| create prefunded SPL    | 3,083 |    2,891 |  -192 |
+| create prefunded T22    | 5,132 |    4,934 |  -198 |
+| create T22 known mint   | 6,164 |    5,750 |  -414 |
+
+Note: idempotent paths still being optimized (at the moment, CUs higher against
+baseline).
+
+## New Terminology
+
+n/a
+
+## Detailed Design
+
+`p-ATA` is a no-std [Pinocchio] program. It avoids heap allocation, uses
+zero-copy instruction/account parsing, and minimizes dependencies. The
+[implementation] (tracked in [associated-token-account#196]) splits into a
+public `interface` crate (instructions, PDA derivation, errors) and the on-chain
+`program` crate. It maintains strict compatibility with the existing ATA
+program:
+
+- Unchanged Program ID: `ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL`
+- Backward compatible instruction interface: discriminants, required accounts,
+  and signer statuses are identical
+- Identical address and bump derivation
+- Same semantics with error codes and state transitions
+
+### New instruction: `CreateWithArgs`
+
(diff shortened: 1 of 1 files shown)


proposals/0567-cu-optimized-ata-program.md
```diff
@@ -0,0 +1,240 @@
...
+
+| Index | Role                  | Description                             |
+|-------|-----------------------|-----------------------------------------|
+| 0     | `[writeable]`         | Nested ATA, owned by the owner ATA      |
+| 1     | `[]`                  | Nested token mint                       |
+| 2     | `[writeable]`         | Wallet's ATA for the nested mint        |
+| 3     | `[]`                  | Owner ATA                               |
+| 4     | `[]`                  | Owner token mint                        |
+| 5     | `[writeable, signer]` | Wallet or token multisig wallet (NEW)   |
+| 5     | `[writeable]`         | Wallet or token multisig wallet (NEW)   |
```
bw-solana: is this bad copypasta? Or meant to indicate this account can be either just writeable or writeable/signer?

bw-solana (commented):
LGTM in general. Can we get SME review from someone. @joncinque or @febo ?

proposals/0567-cu-optimized-ata-program.md
```diff
@@ -0,0 +1,240 @@
...
+  will save CUs by using the new direct-account-pointer entrypoint.
+
+- **[token-2022-interface#1231]: allocation-free Token-2022 account sizing**.
+  Waiting on release. Adds a no-std, no-allocation helper for sizing token-2022
+  accounts locally instead of a `GetAccountDataSize` CPI.
+
+- **New Pinocchio crate release**. `p-ATA` uses Pinocchio APIs (the
+  `CreateAccountAllowPrefund` builder, token `Batch`, Token-2022
+  `StateWithExtensions` and extension state, `GetAccountDataSize`) that are
+  merged to `main` but not all in current releases.
```
joncinque: nit: I don't think we need to spell all of this out

joncinque (commented):
Looks great to me! Just a nit to go with Brennan's comment

deanmlittle:
Ship it!

proposals/0567-cu-optimized-ata-program.md
grod220: Removed

proposals/0567-cu-optimized-ata-program.md
grod220: True, this is confusing. It's only a signer if not a multisig. Going to collapse to one row and clarify.

proposals/0567-cu-optimized-ata-program.md
```diff
@@ -0,0 +1,223 @@
...
+## Motivation
+
+ATA creation is one of the most frequently executed instructions on Solana. This
+program governs the standard that derives canonical token addresses and handles
+the CPIs to create the token accounts. This makes the ATA program a
+high-leverage target for compute reduction.
+
+On a 30k confirmed-block stratified mainnet sample:
+
+- ATA appears in 11.85% of transactions
```
bw-solana: all transactions or non-vote only?

proposals/0567-cu-optimized-ata-program.md
```diff
@@ -0,0 +1,223 @@
...
+- ATA share of total CU usage: 13.28%
+- Top invoked programs (ranked):
```
bw-solana: by count or by CUs?

proposals/0567-cu-optimized-ata-program.md
```diff
@@ -0,0 +1,223 @@
...
+
+## New Terminology
+
+n/a
+
+## Detailed Design
+
+`p-ATA` is a no-std [Pinocchio] program. It avoids heap allocation, uses
+zero-copy instruction/account parsing, and minimizes dependencies. The
+[implementation] (tracked in [associated-token-account#196]) splits into a
```
bw-solana, line 91: pedantic: SIMD is supposed to be strictly protocol level and omit implementation details.

but I also get it that the fact you've already done the work on the backend makes it really easy to provide a detailed design and quantify impact 😄 , so I'm fine with it

proposals/0567-cu-optimized-ata-program.md
```diff
@@ -0,0 +1,223 @@
...
+| 2     | `[]`                  | Wallet address for the ATA                 |
+| 3     | `[]`                  | Token mint                                 |
+| 4     | `[]`                  | System program                             |
+| 5     | `[]`                  | SPL Token or Token-2022 program            |
+| 6     | `[]` optional         | Rent sysvar (NEW)                          |
+
+The rent sysvar is optional and trades off against the caller's expected case.
+Supplying it lets account creation read rent from the sysvar account instead of
+invoking the rent syscall, so the create and prefunded paths are cheaper with
+it. On the idempotent no-op path the account already exists and creation is
```
bw-solana: On the sysvar vs. syscall tradeoff, I'm assuming cheaper here is in terms of CUs, right? Do you know if the CU difference is correlated with actual wall clock difference?

Just want to make sure we're not CU golfing the wrong thing

proposals/0567-cu-optimized-ata-program.md
```diff
@@ -0,0 +1,223 @@
...
+- The nested mint may use a different token program than the owner mint
+- The wallet may be a token multisig
+- The nested mint may carry a transfer hook
+
+These additions are backward compatible. The new accounts are optional and
+trailing, so a caller passing the legacy account layout gets identical behavior.
+
+The instruction supports the following account layout:
+
+<!-- markdownlint-disable MD013 -->
```
bw-solana, line 178: I've made so many ugly, split tables due to this lint... I'm on board w/ the disable

proposals/0567-cu-optimized-ata-program.md
```diff
@@ -0,0 +1,223 @@
...
+buffer.
+
+## Dependencies
+
+- **[SIMD-0312]: CreateAccountAllowPrefund** Recently activated on mainnet.
+  Preserves prefunded-ATA support while avoiding the legacy `Transfer` +
+  `Allocate` + `Assign` sequence.
+
+- **[SIMD-0418]: Loader v2 to v3 Program Migrations**. `p-ATA` will re-use the
+  same pattern exercised by `p-token` for the program upgrade.
```
bw-solana, line 244: we should probably call out this means the ATA program will not be invocable for one slot after migration. Maybe in the Impact section or something

proposals/0567-cu-optimized-ata-program.md
```diff
@@ -0,0 +1,223 @@
...
+token multisig so that trailing signer accounts are unambiguous.
+
+### Upgrade path
+
+Since the current mainnet ATA program is a Loader v2 program, the migration will
+follow [SIMD-0418]. A Loader v3 buffer containing the verified `p-ATA` ELF will
+be deployed before activation. When the feature activates, the runtime creates
+the derived program data account, rewrites the existing ATA program account as a
+Loader v3 program account with no upgrade authority, and closes the source
+buffer.
```
topointon-jump, line 204: The upgrade process should check that the hash of the buffer account is correct. This hash should also be specified in this SIMD.

proposals/0567-cu-optimized-ata-program.md
```diff
@@ -0,0 +1,223 @@
...
+rejected them, which broke those callers. Overloading `Create` and
+`CreateIdempotent` to read a trailing rent sysvar or extra hint bytes would
+reinterpret data those callers already send and risk the same breakage.
+
+## Security Considerations
+
+- **Comprehensive Testing**: All existing test fixtures will be ported and
+  passed. A differential testing framework will execute instructions against
+  both program versions and assert that the resulting state and status are
+  identical.
```
topointon-jump, line 264: Can you provide instructions in this SIMD as to how to build and create the .so file with the correct hash, so that we can create test ledgers.

proposals/0567-cu-optimized-ata-program.md
deanmlittle: Worth noting loading from account is cheaper in CUs, but more expensive in terms of wall clock 🙃

benchmarked here: https://github.com/blueshift-gg/static-sysvars/blob/master/BENCHMARKS.md#sysvar-access-methods-1

Worth comparing against the checked variant as syscall doesn't require checking an account key to use safely.

proposals/0567-cu-optimized-ata-program.md
grod220: That includes vote txs. Excluding those, it's ~31%. Will update.

proposals/0567-cu-optimized-ata-program.md
grod220: Count. Added clarification.

proposals/0567-cu-optimized-ata-program.md
grod220: Good call, added. Though likely to change with audit remediations, fuzz findings, etc. I suppose I can just update this SIMD when the program is finalized.

proposals/0567-cu-optimized-ata-program.md
grod220: My local mollusk benching is consistent with @deanmlittle's. Providing the rent sysvar account lowers CUs but is marginally slower in wall-clock time. Though, at the transaction level here, the wall-clock regression is small (+1%). And compared to the wall clock reduction from legacy to p-ata (-35%) think we can probably disregard it. I'd imagine the CU versus wall-clock mismatch is a broader cost-model discussion.

proposals/0567-cu-optimized-ata-program.md
grod220: We have an upgrade section, for sure will include that

proposals/0567-cu-optimized-ata-program.md
grod220: Done 👍

topointon-jump (approved):

proposals/0567-cu-optimized-ata-program.md
bw-solana: thanks for checking! Agree this sounds small enough we can punt to broader cost model discussions

bw-solana (approved):

proposals/0567-cu-optimized-ata-program.md
alnoki: > I've made so many ugly, split tables due to this lint... I'm on board w/ the disable

@bw-solana or whoever is in charge of linting I suggest this in the `markdownlint.yml` config file:

```yaml
MD013:
  strict: true
  tables: false
```

It escapes tables

bw-solana:
Note that this SIMD will only be merged after formal verification of the new program (starting imminently)

PR #568, opened 2026-06-17
SIMD-0568: Deprecate Precompiles

Lichtso:
In a discussion on Discord [^1] consensus was reached to proceed with option
4 "Drop protocol-owned verifiers entirely and let normal programs handle this".

[^1]: https://discord.com/channels/428295358100013066/1164657208613683280/1516459607726358672

### DIFF
--- a/proposals/0568-deprecate-precompiles.md
+++ b/proposals/0568-deprecate-precompiles.md
@@ -17,7 +17,7 @@ Remove support for the secp256k1, secp256r1 and ed25519 precompiles.
 ## Motivation
 
 Precompiles are rigid, require special handling in validator implementations
-and their cross instruction referencing (in ed25519) is hacky.
+and their inter-instruction data referencing is hacky.
 
 ## Dependencies
 
@@ -60,10 +60,9 @@ These changes are limited to the program runtime component
 
 ### Migrating them to core programs on-chain
 
-The ed25519 one requires top-level instruction introspection, which is not
-available to programs on-chain without the Instructions sysvar. Thus it would
-require either breaking changes to the precompile or adding features to the
-SVM:
+Precompiles can use top-level instruction introspection, which is not available
+to programs on-chain without the Instructions sysvar. Thus it would require
+either breaking changes to the precompile or adding features to the SVM:
 
 - Require users to pass in all bytes to verify in the precompile instruction,
 forbidding cross referencing other instructions.
@@ -110,18 +109,23 @@ Breaking change: All precompiles will become unavailable.
 
 ## Conformance
 
-The change will be tested by twelve transactions, each transaction calling one
+The change will be tested by 18 transactions, each transaction calling one
 of the three precompiles before and after the feature activation, at top level
-and in CPI.
+with a valid signature, at top level with an valid signature and in CPI with a
+valid signature.
 
-The excpected results **before** the feature activation are:
+The expected results **before** the feature activation are:
 
-- at top level: success
-- in CPI: `CpiError::ProgramNotSupported`
+- at top level with a valid signature: success
+- at top level with an invalid signature: `PrecompileError::InvalidSignature`
+- in CPI with a valid signature: `CpiError::ProgramNotSupported`
 
-The excpected results **after** the feature activation are:
+The expected results **after** the feature activation are:
 
-- at top level: `TransactionError::InvalidProgramForExecution`
-- in CPI: `InstructionError::UnsupportedProgramId`
+- at top level with a valid signature:
+`TransactionError::InvalidProgramForExecution`
+- at top level with an invalid signature:
+`InstructionError::UnsupportedProgramId`
+- in CPI with a valid signature: `InstructionError::UnsupportedProgramId`
 
 [to be filled in before merging]
--- a/proposals/0568-deprecate-precompiles.md
+++ b/proposals/0568-deprecate-precompiles.md
@@ -19,6 +19,16 @@ Remove support for the secp256k1, secp256r1 and ed25519 precompiles.
 Precompiles are rigid, require special handling in validator implementations
 and their cross instruction referencing (in ed25519) is hacky.
 
+## Dependencies
+
+This proposal depends on the following proposals:
+
+- **[SIMD-0565]: Secp256r1 curve syscalls**
+
+    To provide an efficient alternative to the secp256r1 precompile.
+
+[SIMD-0565]: https://github.com/solana-foundation/solana-improvement-documents/pull/565
+
 ## New Terminology
 
 None.
@@ -35,6 +45,9 @@ With the activation of the feature gate their accounts must be deleted.
 
 After the activation of the feature gate:
 
+- Fee calculation must ignore them, thus not charge `lamports_per_signature`
+for them.
+- The cost tracker and other cost estimation heuristics should ignore them.
 - Every transaction invoking them with top-level instructions must
 throw `TransactionError::InvalidProgramForExecution` during transaction
 loading time.
@@ -71,6 +84,21 @@ level.
 44% of ed25519 invocations access sibling top-level instruction data. Meaning
 these would not be possible to fill in with an exact replacement right now
 (see [alternatives considered](#migrating-them-to-core-programs-on-chain)).
+
+With the cleanup, after the feature gate is activated on MNB, precompiles
+should also be removed from:
+
+- test validator genesis
+  - the program accounts of the precompiles
+- ledger tool:
+  - `fix_testnet_ed25519_precompile_account`
+- solana-sdk:
+  - `SanitizedMessage`
+  - `SVMStaticMessage`
+  - `TransactionSignatureDetails`
+  - `solana_sdk_ids`
+  - `PrecompileError`
+- tests, mockups, examples, documentation
 
 ## Security Considerations
 
--- a/proposals/0568-deprecate-precompiles.md
+++ b/proposals/0568-deprecate-precompiles.md
@@ -48,12 +48,8 @@ After the activation of the feature gate:
 - Fee calculation must ignore them, thus not charge `lamports_per_signature`
 for them.
 - The cost tracker and other cost estimation heuristics should ignore them.
-- Every transaction invoking them with top-level instructions must
-throw `TransactionError::InvalidProgramForExecution` during transaction
-loading time.
-- Every instruction invoking them with in CPI must
-throw `InstructionError::UnsupportedProgramId` instead of
-`CpiError::ProgramNotSupported` as they currently do.
+- The callee CPI restrictions for precompiles which throw
+`CpiError::ProgramNotSupported` must be removed.
 
 ### Validator Components Affected
 
@@ -114,7 +110,18 @@ Breaking change: All precompiles will become unavailable.
 
 ## Conformance
 
-The change will be tested by three transactions, each transaction calling one
-of the three precompiles before and after the feature activation.
+The change will be tested by twelve transactions, each transaction calling one
+of the three precompiles before and after the feature activation, at top level
+and in CPI.
+
+The excpected results **before** the feature activation are:
+
+- at top level: success
+- in CPI: `CpiError::ProgramNotSupported`
+
+The excpected results **after** the feature activation are:
+
+- at top level: `TransactionError::InvalidProgramForExecution`
+- in CPI: `InstructionError::UnsupportedProgramId`
 
 [to be filled in before merging]
--- a/proposals/0568-deprecate-precompiles.md
+++ b/proposals/0568-deprecate-precompiles.md
@@ -1,0 +1,109 @@
+---
+simd: '0568'
+title: Deprecate Precompiles
+authors:
+  - Alexander Meißner (Anza)
+category: Standard
+type: Core
+status: Review
+created: 2026-06-17
+feature: [to be filled in before merging]
+---
+
+## Summary
+
+Remove support for the secp256k1, secp256r1 and ed25519 precompiles.
+
+## Motivation
+
+Precompiles are rigid, require special handling in validator implementations
+and their cross instruction referencing (in ed25519) is hacky.
+
+In a discussion on Discord [^1] consensus was reached to proceed with option
+4 "Drop protocol-owned verifiers entirely and let normal programs handle this".
+
+[^1]: https://discord.com/channels/428295358100013066/1164657208613683280/1516459607726358672
+
+## New Terminology
+
+None.
+
+## Detailed Design
+
+After the activation of the feature gate every instruction attempting to invoke
+any of the following programs must fail with
+`InstructionError::UnsupportedProgramId`:
+
+- secp256k1: `KeccakSecp256k11111111111111111111111111111`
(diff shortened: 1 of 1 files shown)


proposals/0568-deprecate-precompiles.md
```diff
@@ -0,0 +1,109 @@
...
+any of the following programs must fail with
+`InstructionError::UnsupportedProgramId`:
+
+- secp256k1: `KeccakSecp256k11111111111111111111111111111`
+- secp256r1: `Secp256r1SigVerify1111111111111111111111111`
+- ed25519: `Ed25519SigVerify111111111111111111111111111`
+
+As a consequence the check in CPI which prevents precompiles from being called
+and throws `CpiError::ProgramNotSupported` becomes pointless and must be
+skipped as well.
```
buffalojoec: It would be more clear if you said this case was "replaced" by the new gate. You could also argue it's implicit, also an implementation detail, and this paragraph can be removed altogether.

proposals/0568-deprecate-precompiles.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Edge Cases
+
+The three program accounts will continue to exist and be owned by the native
+loader (`NativeLoader1111111111111111111111111111111`) thus they will continue
+to count as programs during transaction loading and only be filtered out during
+instruction execution.
```
buffalojoec: Can you clarify "count as programs during transaction loading" here? `Secp256k1` is 17, `Secp256r1` is zero, and `Ed25519` is 15 bytes.

I'd say just include their sizes and mention that they will still count as *loaded transaction accounts*.

proposals/0568-deprecate-precompiles.md
```diff
@@ -0,0 +1,109 @@
...
+
+- secp256k1: `KeccakSecp256k11111111111111111111111111111`
+- secp256r1: `Secp256r1SigVerify1111111111111111111111111`
+- ed25519: `Ed25519SigVerify111111111111111111111111111`
+
+As a consequence the check in CPI which prevents precompiles from being called
+and throws `CpiError::ProgramNotSupported` becomes pointless and must be
+skipped as well.
+
+### Edge Cases
```
buffalojoec: Will these programs continue to remain reserved account keys? Why?

proposals/0568-deprecate-precompiles.md
```diff
@@ -0,0 +1,109 @@
...
+`TransactionError::InvalidProgramForExecution` when any top-level instruction
+for a precompile is detected. This however would still leave the CPI callee
+special case relevant.
+
+## Impact
+
+The community will have to develop their own on-chain replacements.
+However, this is also an opportunity as solutions using crypto syscalls can be
+cheaper compared to the lamports charged for signature verification at the
+transaction level.
```
buffalojoec: I think this SIMD should do the community a service and describe how the three verification methods can be completely replaced by on-chain syscalls. It should be self-evident in the proposal that alternatives are available, working, and reasonable.

proposals/0568-deprecate-precompiles.md
```diff
@@ -0,0 +1,109 @@
...
+
+According to Sam Kim about 44% of ed25519 invocations access sibling top-level
+instruction data. Meaning these would not be possible to fill in with an
+exact replacement right now
+(see [alternatives considered](#migrating-them-to-core-programs-on-chain)).
```
buffalojoec: This is weird. I wouldn't shout someone out by name in a SIMD.

proposals/0568-deprecate-precompiles.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Security Considerations
+
+Implementation-specific guidance will be required for transaction builder
+clients and dApp devs who develop the on-chain replacements. Preferably we
+should provide canonical replacements, which is different from a core program
+as these replacements would life under different addresses and have different
+interfaces.
```
buffalojoec: I would probably drop this second sentence.

proposals/0568-deprecate-precompiles.md
```diff
@@ -0,0 +1,109 @@
...
+
+## New Terminology
+
+None.
+
+## Detailed Design
+
+After the activation of the feature gate every instruction attempting to invoke
+any of the following programs must fail with
+`InstructionError::UnsupportedProgramId`:
```
buffalojoec: Is there a reason why we're doing all three in one feature gate? I suppose it doesn't really matter much, but curious what the rationale is.

proposals/0568-deprecate-precompiles.md
```diff
@@ -0,0 +1,109 @@
...
+
+These changes are limited to the program runtime component
+(Transaction Execution).
+
+## Alternatives Considered
+
+### Migrating them to core programs on-chain
+
+The ed25519 one requires top-level instruction introspection, which is not
+available to programs on-chain out of the box right now. Thus it would require
```
buffalojoec: I would clarify it's not **natively available without the Instructions sysvar**.

proposals/0568-deprecate-precompiles.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Motivation
+
+Precompiles are rigid, require special handling in validator implementations
+and their cross instruction referencing (in ed25519) is hacky.
+
+In a discussion on Discord [^1] consensus was reached to proceed with option
+4 "Drop protocol-owned verifiers entirely and let normal programs handle this".
+
+[^1]: https://discord.com/channels/428295358100013066/1164657208613683280/1516459607726358672
```
buffalojoec: I don't think linking Discord discussions in SIMD bodies is canon.

proposals/0568-deprecate-precompiles.md
deanmlittle: Kill them

proposals/0568-deprecate-precompiles.md
Lichtso: I will put it in the PR description instead. After all we are now required to show "that your SIMD has been thoroughly discussed".

proposals/0568-deprecate-precompiles.md
Lichtso: Reworded it, did not mean the size / data length.

proposals/0568-deprecate-precompiles.md
Lichtso: The check has to be explicitly removed, otherwise the error codes are different. Will reword it.

proposals/0568-deprecate-precompiles.md
Lichtso: We already are hitting feature gate activation queue capacity with our policy of one activation per epoch. Also, these three are related, it makes sense to disable them simultaneously.

proposals/0568-deprecate-precompiles.md
Lichtso: Sounds good, question is just what level of detail is appropriate here:

- only mention the specific syscalls
- provide some pseudo code
- link actual reference implementations

buffalojoec: IMO just list the syscalls they can use to perform the same signature verification as the precompiles. That's probably enough.

buffalojoec: We just also need to make sure they actually are available, working, and *reasonable* (CUs) 😅

proposals/0568-deprecate-precompiles.md
```diff
@@ -19,35 +19,28 @@ Remove support for the secp256k1, secp256r1 and ed25519 precompiles.
...
-The three program accounts will continue to exist and be owned by the native
-loader (`NativeLoader1111111111111111111111111111111`) thus they will continue
-to count as programs during transaction loading and only be filtered out during
-instruction execution.
+- Every transaction invoking them with top-level instructions must
+throw `TransactionError::InvalidProgramForExecution` during transaction
+loading time.
+- Every instruction invoking them with in CPI must
+throw `InstructionError::UnsupportedProgramId` instead of
+`CpiError::ProgramNotSupported` as they currently do.
```
buffalojoec: One more bullet for "these will no longer be reserved account keys"?
https://github.com/solana-foundation/solana-improvement-documents/pull/568#discussion_r3432322537

proposals/0568-deprecate-precompiles.md
buffalojoec: Heh, looks like you just took them out instead. Works for me.

proposals/0568-deprecate-precompiles.md
deanmlittle: All good here once we get secp256r1 curve ops syscalls imo.

proposals/0568-deprecate-precompiles.md
deanmlittle: Why do we need to say what error will occur when invoking them? Why not simply say we're unloading the programs and let the VM respond with its normal behavior?

Lichtso: `TransactionError::InvalidProgramForExecution` happens as a natural consequence of deleting the accounts, yes. Mentioning it here explicitly makes it easier to write conformance tests, that is all.

The transition from `CpiError::ProgramNotSupported` to `InstructionError::UnsupportedProgramId` in CPI (to bring it in line with other non existing programs) is an actual change in the code because right now precompiles get special handling in CPI.

deanmlittle: Right, my point is, we are not *implementing* anything to create this error message, we are simply removing the implementation of CpiError, resulting in the default error message for a non-existent program. That's why I don't think it makes sense to write it here, as it sounds like we're adding some functionality to cause this error message to appear instead of the existing one.

buffalojoec: You at least want to call out the `TransactionError`, since it will be a load-error and thus a fees-only transaction, which is protocol. However, I agree, the instruction error code specifically doesn't need to be enshrined, and could just get detailed under the conformance/fixtures section towards the end of the newer template.

deanmlittle: The SIMD is to document what changes must be made to the protocol. If unloading the program results in the default VM/Runtime error behavior for a program that doesn't exist, we are not *making* it do that, we are removing the condition that causes it to *not* do that today. Making statements like: "Invoking these programs must return such and such error", it implies we're adding some kind of behavior to make this the case, which is  the exact opposite of what we're doing here. I think if you want to explicitly mention what would happen when invoking these programs after they're removed, you can throw that into the backwards compatibility section.

proposals/0568-deprecate-precompiles.md
deanmlittle: I think the best rationale is that we can deprecate all 3 from fee calculation at the same time which is much less invasive than changing it 3 times.

proposals/0568-deprecate-precompiles.md
Lichtso: I will move it to the "Conformance" section as it is mostly relevant to testing.

proposals/0568-deprecate-precompiles.md
```diff
@@ -0,0 +1,127 @@
...
+---
+
+## Summary
+
+Remove support for the secp256k1, secp256r1 and ed25519 precompiles.
+
+## Motivation
+
+Precompiles are rigid, require special handling in validator implementations
+and their cross instruction referencing (in ed25519) is hacky.
```
deanmlittle, line 20: All 3 precompiles do cross-instruction referencing.

```suggestion
and their inter-instruction data referencing is hacky.
```

proposals/0568-deprecate-precompiles.md
```diff
@@ -0,0 +1,127 @@
...
+### Validator Components Affected
+
+These changes are limited to the program runtime component
+(Transaction Execution).
+
+## Alternatives Considered
+
+### Migrating them to core programs on-chain
+
+The ed25519 one requires top-level instruction introspection, which is not
```
deanmlittle, line 63: Same as above, all 3 do this, not just Ed25519.

proposals/0568-deprecate-precompiles.md
```diff
@@ -0,0 +1,127 @@
...
+
+Breaking change: All precompiles will become unavailable.
+
+## Conformance
+
+The change will be tested by twelve transactions, each transaction calling one
+of the three precompiles before and after the feature activation, at top level
+and in CPI.
+
+The excpected results **before** the feature activation are:
```
deanmlittle, line 117:
```suggestion
The expected results **before** the feature activation are:
```

proposals/0568-deprecate-precompiles.md
```diff
@@ -0,0 +1,127 @@
...
+
+- at top level: success
+- in CPI: `CpiError::ProgramNotSupported`
+
+The excpected results **after** the feature activation are:
+
+- at top level: `TransactionError::InvalidProgramForExecution`
+- in CPI: `InstructionError::UnsupportedProgramId`
```
deanmlittle, line 125: I am assuming by 12 transactions, you meant 3x:

1. Invoke pre-activation
2. CPI pre-activation
3. Invoke post-activation
4. CPI post-activation

I think you should be explicit about *how* you're invoking it, as different inputs could lead to a result other than the one you have written here. I would suggest to instead be explicit that we will:

1. Invoke with valid signature pre-activation
2. Invoke with invalid signature pre-activation
3. CPI with valid signature pre-activation
4. Invoke with valid signature post-activation
5. Invoke with invalid signature post-activation
6. CPI with valid signature post-activation

Lichtso:
@deanmlittle I think I addressed everything, can you give it another review?

proposals/0568-deprecate-precompiles.md
```diff
@@ -0,0 +1,131 @@
...
+The expected results **before** the feature activation are:
+
+- at top level with a valid signature: success
+- at top level with an invalid signature: `PrecompileError::InvalidSignature`
+- in CPI with a valid signature: `CpiError::ProgramNotSupported`
+
+The expected results **after** the feature activation are:
+
+- at top level with a valid signature:
+`TransactionError::InvalidProgramForExecution`
```
deanmlittle, line 126: is this correct? seems odd that the error would change based upon program inputs for an invalid program

proposals/0568-deprecate-precompiles.md
```diff
@@ -0,0 +1,131 @@
...
+### Validator Components Affected
+
+These changes are limited to the program runtime component
+(Transaction Execution).
+
+## Alternatives Considered
+
+### Migrating them to core programs on-chain
+
+Precompiles can use top-level instruction introspection, which is not available
```
deanmlittle, line 63: I found the wording here a little confusing as it feels like it conflates instrospection from the instructions sysvar account and how it *actually* accesses it from the TransactionContext. Consider making this clear by saying something like:

"As they are built into the runtime, precompiles have the ability to access instruction data from other instructions from TransactionContext without having to use the SysvarInstructions account. Replicating this in a Core BPF program requires breaking changes, either to the API, or by adding this functionality to the SVM."

deanmlittle:
> @deanmlittle I think I addressed everything, can you give it another review?

Sure. Just a few nits. Looks pretty good overall!

PR #571, opened 2026-06-24
SIMD-0571: Soft Deprecation of Durable Nonce Transactions

MaxResnick:
## Summary

This SIMD proposes a **soft deprecation of durable nonce transactions**: behind
a feature gate, a durable nonce transaction's requested compute unit price is
treated as `0` for both fee collection and scheduler prioritization. The
transaction stays valid and can still land it just cannot pay to be
prioritized, which removes the incentive to use durable nonces as a low-cost
spam vector while custodians and other users migrate ahead of full nonce
removal.

## Context / Discussion

Based on and discussed in
solana-foundation/solana-improvement-documents#564.

The motivation, supporting data, and the option chosen here (zero the priority
fee rather than reject the transaction or burn the fee) come directly from that
discussion thread.

### DIFF
--- a/proposals/0571-soft-deprecate-nonce-priority-fees.md
+++ b/proposals/0571-soft-deprecate-nonce-priority-fees.md
@@ -1,0 +1,193 @@
+---
+simd: '0571'
+title: Soft Deprecation of Durable Nonces
+authors:
+  - Max Resnick
+category: Standard
+type: Core
+status: Idea
+created: 2026-06-24
+feature: (fill in with feature key and github tracking issues once accepted)
+---
+
+## Summary
+
+Treat a durable nonce transaction's requested compute
+unit price as `0` for both fee collection and scheduler prioritization. The
+transaction remains valid and is not rejected; it simply cannot pay to be
+prioritized, which removes the incentive to route fee-bidding traffic
+through durable nonces.
+
+## Motivation
+
+Durable nonces are expected to be deprecated and eventually removed, but that
+removal must wait for custodians and other non-time-sensitive users to
+migrate. While durable nonce transactions still exist, they cause two
+problems:
+
+1. They are heavily used as a fee-bidding channel, generating high-volume
+   retry traffic that congests the scheduler
+2. Because that traffic pays competitive priority fees, it is coupled to the
+   regular landing game and is expensive to ignore client-side alone.
+
+Zeroing their priority fee removes the ability to outbid other traffic, and
+acts as a forcing function: once durable nonces cannot be prioritized,
+high-volume workloads have an incentive to migrate to the recent-blockhash
+path ahead of full nonce removal.
+
+Supporting data (last epoch, 10,000-block sample): the marginal non-vote,
+non-Jito-bundle transaction paid a priority fee of `0`; durable nonce
+transactions were ~3.1% of transactions but ~30% of priority-fee revenue
+(1,975 / 6,621 SOL); 86% of them paid a priority fee versus ~28% network-wide;
+and 0.5% of nonce fee-payers (555 of 103,672) drove 97.5% of nonce
+transactions and 70% of nonce priority fees. The bulk of nonce priority-fee
+revenue comes from a handful of high-volume bidding operations, not custodial
+use — a rational response to the nonce path's favorable auction mechanics
+(made precise below), but far from the custodial cold-signing use case the
+mechanism exists to serve.
+
+This spending on durable nonces is expected to mostly migrate over to
+non-durable nonce transactions. Bidding through a durable nonce is closer to a
+first-price auction (the priority fee is only paid on the attempt that lands),
+while bidding through the standard path is closer to an all-pay auction
+(failed attempts still land and pay). First-price bids are higher than all-pay
+bids for the same opportunity; however, by the [revenue equivalence
+theorem](https://en.wikipedia.org/wiki/Revenue_equivalence), the expected
+total payment for the same underlying opportunity is the same across the two
+formats: bidders shade their bids to offset the difference in payment rules.
+Since this proposal does not change the size of the underlying opportunities,
+expected spend should migrate to the standard path roughly one-for-one,
+spread across more, lower-priced transactions.
+
+## New Terminology
+
+No new terminology. **Durable nonce transaction** is defined in Detailed
+Design. **Prioritization fee** (priority fee) is the fee component derived
+from the requested compute unit price and limit, and is distinct from the
+per-signature base fee.
+
+## Detailed Design
+
+This change is gated behind a feature; the rules below apply to durable nonce
+transactions once the feature is activated.
+
+Durable nonce transactions have been defined more than once already: by the
+shape of the first instruction (SIMD-0242), and by the runtime's full nonce
+check over account state, with slight variations across implementations and
+layers. This proposal attempts to resolve the inconsistencies by introducing
+a new definition that should be treated as canonical going forward.
+
+[![How standards proliferate (xkcd 927)][xkcd-img]][xkcd-927]
+
+[xkcd-img]: https://imgs.xkcd.com/comics/standards.png
+[xkcd-927]: https://xkcd.com/927/
+
+
+> A transaction is a **durable nonce transaction** if and only if its
+> `recent_blockhash` field is not a valid recent blockhash of the bank
+> processing it.
+
+This simple rule is sufficient because the account-dependent parts of the
+runtime's existing nonce check gate validity, not classification. A
+transaction whose `recent_blockhash` is not a valid recent blockhash can
+only be committed by passing that check: its first instruction must advance
+a System-owned, initialized nonce account whose stored durable nonce equals
+the `recent_blockhash` field, signed by the nonce authority, with the nonce
+still advanceable. If it fails the check, the transaction can never land and
+never pays fees. So for every transaction that can land, "stale
+`recent_blockhash`" and "processed via the durable nonce path" coincide.
+This proposal does not modify the nonce check itself.
+
+Conversely, a transaction with a valid recent blockhash is processed via the
+normal path with normal fee and priority semantics — even if its first
+instruction is `AdvanceNonceAccount`. Such a transaction can still advance
+the nonce (and fails if the nonce was already advanced in that slot), but it
+lands and pays fees like any other transaction. This is no different from
+what any user-defined nonce program can already do, so it introduces no new
+congestion or fee-evasion vector.
+
+For a durable nonce transaction, after activation:
+
+- Its prioritization fee MUST be computed as `0`, irrespective of the CU
+  price requested in the message. The fee charged to
+  the fee-payer therefore includes the base fee and any other
+  non-prioritization components, but MUST NOT include a prioritization fee.
+  This affects bank state and is consensus-critical: it MUST be applied
+  identically by all validators under the feature gate, wherever the fee
+  enters consensus-relevant computation — including fee-payer balance
+  validation, pre-execution fee checks, and fee debit and collection. In
+  particular, a fee-payer able to cover only the base fee (and other
+  non-prioritization components) is valid after activation.
+- For scheduling, a leader SHOULD treat the transaction's CU price as `0`
+  when deriving scheduling priority. This is leader-local behavior and is not
+  consensus-critical; applying it is what removes the incentive to bid
+  through the nonce path.
+
+Durable nonces remain valid and can land in blocks that have excess capacity.
+
+## Alternatives Considered
+
+- **Client-side only (no protocol change).** Rejected: a revenue-seeking
+  client would fork to reverse it mid-migration, restoring the incentive to
+  bid through the nonce path.
+
+- **Burn the priority fee instead of not charging it.** Harder to implement.
+
+- **Increase fees for durable nonce transactions instead.** Raising the base
+  fee or the cost of `AdvanceNonceAccount` offers similar functionality in
+  principle. It would encourage MEV-related traffic to migrate off of the
+  durable nonce path; however, it is not compatible with bankless leader DoS
+  prevention. A bankless leader admits and packs transactions without account
+  state, so it cannot verify at admission time that a fee-payer can cover an
+  elevated nonce-specific fee; underfunded durable nonce transactions could
+  still consume block space while ultimately paying nothing. Zeroing the
+  priority fee means that nonced transactions can never outbid normal
+  transactions that pay at least some priority fee, which lines up nicely
+  with the requirements for bankless. Note that increasing the fee does not
+  do this, because what we are worried about is transactions that don't pay
+  fees: increasing the fee for a transaction that does not pay the fee does
+  nothing to deter this behavior.
+
+## Impact
+
+- **Legitimate durable nonce users (e.g. custodians):** mostly unaffected.
+  The marginal priority fee in blocks is almost always `0` and ~94% of blocks
+  are not full, so a priority-fee-`0` transaction lands within a slot or two
+  with very high probability — even more so after the 100M CU block limit
+  activation. End-to-end latency for a custodial durable-nonce flow is
+  therefore, with extremely high confidence, dominated by the cold-signing
+  operation itself, not by waiting for a non-full block.
+- **High-volume bidding workloads:** lose the ability to prioritize durable
+  nonce transactions, incentivizing migration to the standard path ahead of
+  full removal.
+- **Validators / leaders:** lose some fees in the short term,
+  since the fee-payer is not charged (cf. SIMD-0096); however,
+  this should mostly migrate to standard path since opportunity size has not
+  changed.
+- **Out-of-protocol consumers of compute unit price:** consumers that derive
+  fees or priority from the CU price (RPC fee estimation, transaction
+  simulation, packet admission or priority filtering above the scheduler)
+  SHOULD use the effective CU price (`0` for durable nonce transactions).
+  Transaction metadata / RPC SHOULD serve the priority fee and effective CU
+  price as `0`, and SHOULD additionally expose the requested CU price as a
+  new field, analogous to Ethereum serving `effectiveGasPrice` alongside the
+  signed gas fields. The signed transaction data itself is unchanged.
+
+## Security Considerations
+
+- **Consensus determinism.** The fee charged to the fee-payer changes, so the
+  fee-computation portion affects bank state and MUST be applied identically
+  by all validators under the feature gate; divergence would cause a consensus
+  failure.
+- **Transaction classification.** Implementations MUST classify durable
+  nonce transactions exactly by the recent-blockhash rule defined above.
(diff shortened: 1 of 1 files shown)


apfitzge (approved):
This is simple to implement and gives direction to remove nonce-spam before we can remove nonces.

Very much in support of this.

cavemanloverboy:
can also raise advance nonce ixn cost by a lot and make people pay more, in conjunction with simd 553 :D

proposals/0571-soft-deprecate-nonce-priority-fees.md
```diff
@@ -0,0 +1,116 @@
...
+Supporting data (last epoch, 10,000-block sample): the marginal non-vote,
+non-Jito-bundle transaction paid a priority fee of `0`; durable nonce
+transactions were ~3.1% of transactions but ~30% of priority-fee revenue
+(1,975 / 6,621 SOL); 86% of them paid a priority fee versus ~28% network-wide;
+and 0.5% of nonce fee-payers (555 of 103,672) drove 97.5% of nonce
+transactions and 70% of nonce priority fees. The bulk of nonce priority-fee
+revenue is spam, not custodial use.
+
+This spending on durable nonces is expected to mostly migrate over to
+non-durable nonce transactions.
```
bw-solana: any concerns that this will not happen and that bidding w/ nonces is higher given there is some amount of execution protection?

proposals/0571-soft-deprecate-nonce-priority-fees.md
```diff
@@ -0,0 +1,116 @@
...
+- **Legitimate durable nonce users (e.g. custodians):** mostly
+  unaffected, since their use is not time-sensitive and the marginal priority
+  fee is almost always `0`. They will land in the next non full block
+  (94% of blocks are not full).
+- **Spammers:** lose the ability to prioritize durable nonce transactions,
+  incentivizing migration ahead of full removal.
+- **Validators / leaders:** lose some fees in the short term,
+  since the fee-payer is not charged (cf. SIMD-0096); however,
+  this should mostly migrate to standard path since opportunity size has not
+  changed.
```
bw-solana, line 158: Not sure which section this belongs in (maybe impact), but it might be worth mentioning all of the other (out of protocol) places this impacts in some form:

- packet admission if people are establishing priority filtering above scheduler
- RPC fee estimations
- Tx simulation
- Tx metadata / indexing / RPC data

proposals/0571-soft-deprecate-nonce-priority-fees.md
```diff
@@ -0,0 +1,116 @@
...
+
+- **Client-side only (no protocol change).** Rejected: a revenue-seeking
+  client would fork to reverse it mid-migration, restoring the spam incentive.
+
+- **Burn the priority fee instead of not charging it.** Harder to implement.
+
+## Impact
+
+- **Legitimate durable nonce users (e.g. custodians):** mostly
+  unaffected, since their use is not time-sensitive and the marginal priority
```
bw-solana: have we vetted this claim that none of these are time sensitive?

cavemanloverboy:
here to say again that charging ppl more via resource fee is better option. setting prio fee to zero just incentivizes out of band solutions (jito, nozomi, rakurai tip)

bw-solana:
> here to say again that charging ppl more via resource fee is better option.

If we crank up CUs significantly to increase resource fee, won't that break a lot of dormant nonce txs (CUs exceed requested)?

> setting prio fee to zero just incentivizes out of band solutions (jito, nozomi, rakurai tip)

That seems fine. Dedicated builder/landing services are better equipped to deal with the spam

cavemanloverboy:
> That seems fine. Dedicated builder/landing services are better equipped to deal with the spam

define better equipped. you are just putting this on someone else's plate lol. out of sight, out of mind

bw-solana:
> define better equipped.

for example:
can amortize horizontal scaling across a larger amount of stake.
can use API key and revoke the key for malicious actors or heavily throttle unknown identities to prevent abuse.

> you are just putting this on someone else's plate

correct, I'm sure the entities running custom schedulers and receiving out of protocol priority tips will be able to solve their own issues

2501babe:
my main concern with invariably setting nonce priority to 0 is if spam from nonce-using bad actors can crowd out what we consider to be "legitimate" nonce txns. do we ever come close to saturating the priority queue? 

>can also raise advance nonce ixn cost by a lot and make people pay more

ive always liked the concept of just making nonce txns way more expensive, in theory. im not sure what an appropriate number would be though for what our goals are (price out 50% of nonce usage? 80%? etc). based on resnicks data in the discussion, the average priority fee across all landed nonce transactions is 142k lamps, but the top 10 average is 30m lamps (and they drag the average up a lot obviously so maybe the median is negligible), but also "pays a high fee" is presumably strongly correlated with "spams the network" because theyre doing this specifically to abuse the discard mechanism to dodge fees for all the _other_ transactions they submit

in other words i dont know if theres a number high enough to dissuade spammers but also low enough to not price out all normal users, since the spammers are willing to pay well above what any other user would consider an egregious price. before resnicks numbers i liked saying "just 100x, 1000x the signature cost on nonce transactions" but that would be a rounding error compared to some of these large priority fees

>define better equipped. you are just putting this on someone else's plate lol

isnt it more like giving them dealflow? these users spam nonce transactions for some better inclusion property (more reliability? faster? more likely top-of-block?), exploiting an unfixable flaw in nonce txns to avoid paying for most of their submissions. if we nerf their ability to get priority with spam, its not like they can just spam the landing services to get this property, because "nonce txns pay no priority fee" would be part of consensus. theyd presumably stop spamming and start properly buying whatever property theyre aiming for

>If we crank up CUs significantly to increase resource fee, won't that break a lot of dormant nonce txs (CUs exceed requested)?

nonce accounts have a field `lamports_per_signature` that i have always imagined using for this. its original idea was if we change the global signature cost, your next nonce txn is grandfathered into the old cost from the previous txn they ran for the account. but we can change its meaning to effect _any_ fee change we want, giving every account one "freebie" so no existing transactions are invalidated

maybe one idea is 10x the signature cost for nonce txns and add a 10x multiplier to priority fee for nonce txns. this means anyone who cares about transaction economy (pays no priority) will switch to blockhash _unless_ they actually need the nonce mechanism. one of the flaws of nonce transactions is that there is no reason _not_ to use them. and 50k isnt that expensive for our legitimate users. meanwhile users paying _high_ priority will also be nudged toward blockhash because the costs will scale there too, whereas 5k vs 50k per signature means nothing if youre paying 30m for priority anyway

im not well-versed in the exact mechanics of the "signature cost" to "half signature plus resource cost" change but we can presumably accomplish the same thing by calling using a nonce instead of a blockhash a "resource" and adding an extra fixed cost plus an extra priority cost multiplier to it

cavemanloverboy:
> do we ever come close to saturating the priority queue?

not outside of adversarial conditions, although perhaps possible during a huge market move like 10/10

proposals/0571-soft-deprecate-nonce-priority-fees.md
MaxResnick: Bidding through a durable nonce is closer to a first-price auction (you only pay the priority fee on the attempt that lands), while bidding through the standard path is closer to an all-pay auction (failed attempts still land and pay). First-price bids are always higher than all-pay bids for the same opportunity; however, by the [revenue equivalence theorem](https://en.wikipedia.org/wiki/Revenue_equivalence), the expected total payment for the same underlying opportunity is the same across the two formats: bidders shade their bids to offset the difference in payment rules. Since this proposal doesn't change the size of the underlying opportunities, expected spend should migrate to the standard path roughly one-for-one, just spread across more, lower-priced transactions.

proposals/0571-soft-deprecate-nonce-priority-fees.md
MaxResnick: The marginal priority fee in blocks is almost always 0 (the marginal non-vote, non-bundle transaction in our 10,000-block sample paid 0), and ~94% of blocks are not full, so a priority-fee-0 transaction lands within a slot or two with very high probability. That gets even stronger after the 100M CU block limit activation. Meanwhile, end-to-end landing time for a custodial durable-nonce flow is, with extremely high confidence, dominated by the cold-signing operation itself, not by waiting for a non-full block. I've updated the wording in the SIMD to make this argument explicit.

proposals/0571-soft-deprecate-nonce-priority-fees.md
MaxResnick: Added these to the Impact section.

bw-solana:
> do we ever come close to saturating the priority queue?
like Cavey said, not really.

> my main concern with invariably setting nonce priority to 0 is if spam from nonce-using bad actors can crowd out what we consider to be "legitimate" nonce txns.
Practically speaking, I think we should be fine. A couple things to consider:

1. An infinite number of transactions willing to pay a prio fee would be able to crowd these nonce txs out forever. But practically speaking, majority of blocks are not full, and we just increased block space 67%. We will always be IBRLing to ensure there is excess capacity
2. So the question becomes if non-payers could crowd out the nonce tx such that we never even try to execute it ever. In theory I suppose this is possible, but practically speaking, this seems improbable to pull off in a sustained way. We've been pushing improvements to load shed nonce spam earlier in the pipeline. There's also a non-zero cost in the SWQoS world to spamming leaders with transactions at a sustained rate that probabilistically blocks out the nonce tx. There's also the ability to use a landing service to gain priority by tipping out of protocol

proposals/0571-soft-deprecate-nonce-priority-fees.md
```diff
@@ -0,0 +1,136 @@
...
+  since the fee-payer is not charged (cf. SIMD-0096); however,
+  this should mostly migrate to standard path since opportunity size has not
+  changed.
+- **Out-of-protocol consumers of compute unit price:** any component that
+  reads the requested compute unit price will observe an effective price of
+  `0` for durable nonce transactions after activation. This includes packet
+  admission or priority filtering applied above the scheduler, RPC fee
+  estimation, transaction simulation, and transaction metadata / indexing /
+  RPC responses. These SHOULD treat durable nonce transactions consistently
+  with the scheduler (effective compute unit price `0`).
```
bw-solana: This reads a little confusing to me. I would imagine if someone is pulling down the raw transaction information for some nonce, we want to show what the user actually put for the compute unit price, right? Or are you proposing actually overwriting this field for the read layer?

more generally, it seems like we have 2 or 3 distinct concepts:

1. Requested CU price: encoded in the signed message
2. Effective fee CU price: zero for fee computation
3. Scheduler CU price: zero when a conforming scheduler derives priority from CU price

2/3 are mostly the same thing, but obviously very different from 1. 2 is the MUST. 3 is the SHOULD (but not actually enforced in consensus).

proposals/0571-soft-deprecate-nonce-priority-fees.md
```diff
@@ -0,0 +1,136 @@
...
+and is distinct from the per-signature base fee.
+
+## Detailed Design
+
+This change is gated behind a feature; the rules below apply to durable nonce
+transactions once the feature is activated.
+
+A **durable nonce transaction** is a transaction whose first instruction has
+program id `System` and deserializes to `AdvanceNonceAccount`, matching the
+existing runtime definition (SIMD-0242).
```
bw-solana: What do we do if a transaction uses a recent, valid blockhash but has the first instruction is `AdvanceNonceAccount`? I believe current Agave code will take the normal tx path for this 😭 

I don't really care what we do as long as it's defined and consistent between leader/replay/clients

proposals/0571-soft-deprecate-nonce-priority-fees.md
```diff
@@ -0,0 +1,136 @@
...
+  treated as `0`), irrespective of any
+  `ComputeBudgetInstruction::SetComputeUnitPrice`. The fee charged to the
+  fee-payer therefore includes the base fee and any other non-prioritization
+  components, but MUST NOT include a prioritization fee. This affects bank
+  state and is consensus-critical: it MUST be applied identically by all
+  validators under the feature gate.
+- For scheduling, a leader SHOULD treat the effective compute unit price as
+  `0` when deriving scheduling priority. This is leader-local behavior and is
+  not consensus-critical; applying it is what realizes the spam-disincentive
+  effect.
```
bw-solana, line 133: I'm thinking we should explicitly call out some of the places where interpreting PF as 0 has consensus implications:

- fee-payer balance validation
- pre-execution fee checks
- fee debit and collection

proposals/0571-soft-deprecate-nonce-priority-fees.md
MaxResnick: Addressed in the [three-price taxonomy](https://github.com/MaxResnick/solana-improvement-documents/blob/f95849b8e0a65941ff945cf07c25513efa9615af/proposals/0571-soft-deprecate-nonce-priority-fees.md#L85-L93) and the [reworked Impact bullet](https://github.com/MaxResnick/solana-improvement-documents/blob/f95849b8e0a65941ff945cf07c25513efa9615af/proposals/0571-soft-deprecate-nonce-priority-fees.md#L136-L143): the read layer serves the priority fee / effective CU price as `0`, and the requested CU price is exposed as a new field — analogous to Ethereum serving `effectiveGasPrice` next to the signed gas fields. The signed message itself is unchanged.

proposals/0571-soft-deprecate-nonce-priority-fees.md
MaxResnick: Defined [here](https://github.com/MaxResnick/solana-improvement-documents/blob/f95849b8e0a65941ff945cf07c25513efa9615af/proposals/0571-soft-deprecate-nonce-priority-fees.md#L74-L83): a transaction is subject to this proposal iff it is processed via the durable nonce path, so a valid-recent-blockhash transaction with `AdvanceNonceAccount` first stays on the normal path with normal semantics and fees. Since nonce values are domain-separated from blockhashes, the two paths are mutually exclusive for any signed transaction — classification is deterministic and consistent across leader/replay/clients.

proposals/0571-soft-deprecate-nonce-priority-fees.md
MaxResnick: Called out [here](https://github.com/MaxResnick/solana-improvement-documents/blob/f95849b8e0a65941ff945cf07c25513efa9615af/proposals/0571-soft-deprecate-nonce-priority-fees.md#L97-L106): fee-payer balance validation, pre-execution fee checks, and fee debit and collection.

proposals/0571-soft-deprecate-nonce-priority-fees.md
bw-solana: my initial reaction was that the read layer should just show what's in the transaction. But I've come around to the idea it should show the effective prio fee paid in some cases.

The read layer has always done some amount of translation from raw tx bytes to some human display

bw-solana (dismissed):

proposals/0571-soft-deprecate-nonce-priority-fees.md
ptaffet-jump: You should remove the part that says it matches the SIMD-0242 definition. I think the problem is that SIMD-0242 is actually wrong, and this is the right definition.

MaxResnick: Done

proposals/0571-soft-deprecate-nonce-priority-fees.md
```diff
@@ -0,0 +1,156 @@
...
+per-signature base fee.
+
+## Detailed Design
+
+This change is gated behind a feature; the rules below apply to durable nonce
+transactions once the feature is activated.
+
+A **durable nonce transaction** is a transaction that is processed via the
+durable nonce path: its `recent_blockhash` field is not a valid recent
+blockhash and the nonce check succeeds. A transaction with a valid recent
```
ptaffet-jump: Sorry to keep hammering on this. Can you define the nonce check so that the SIMD is fully specified?

MaxResnick: Done: https://github.com/solana-foundation/solana-improvement-documents/blob/bf38d5f6311d8d21759f7a128292ca86973e742b/proposals/0571-soft-deprecate-nonce-priority-fees.md#L74-L106

bw-solana (dismissed):

proposals/0571-soft-deprecate-nonce-priority-fees.md
MaxResnick: Reworked: classification is now purely recent-blockhash based, no account state consulted. https://github.com/solana-foundation/solana-improvement-documents/blob/d5a5e04182e1500768d78214624702920683e33f/proposals/0571-soft-deprecate-nonce-priority-fees.md#L70-L114

ptaffet-jump: Thanks! Sometimes with changes like these, I don't know whether to laugh or cry, but I think this is the right move.

ptaffet-jump (approved):

bw-solana (dismissed):

ptaffet-jump (changes requested):
Although the design is still fine, after some additional discussion, we'd like to talk more internally about the goal of the SIMD.

proposals/0571-soft-deprecate-nonce-priority-fees.md
```diff
@@ -0,0 +1,193 @@
...
+  this should mostly migrate to standard path since opportunity size has not
+  changed.
+- **Out-of-protocol consumers of compute unit price:** consumers that derive
+  fees or priority from the CU price (RPC fee estimation, transaction
+  simulation, packet admission or priority filtering above the scheduler)
+  SHOULD use the effective fee CU price (`0` for durable nonce transactions).
+  Transaction metadata / RPC SHOULD serve the priority fee and effective CU
+  price as `0`, and SHOULD additionally expose the requested CU price as a
+  new field, analogous to Ethereum serving `effectiveGasPrice` alongside the
+  signed gas fields. The signed transaction data itself is unchanged.
```
joncinque, line 174: Not a huge deal, but I wanted to flag this: on-chain programs can use instruction introspection to derive the priority fee paid in a transaction. With this proposal, they'll also need to check for an advance-nonce instruction.

tx v1 already breaks this functionality, so on-chain programs have to figure out an alternative anyway.

MaxResnick:
> Zeroing the priority fee means that nonced transactions can never outbid normal transactions that pay at least some priority fee, which lines up nicely with the requirements for bankless. Note that increasing the fee does not do this, because what we are worried about is transactions that don't pay fees: increasing the fee for a transaction that does not pay the fee does nothing to deter this behavior.

proposals/0571-soft-deprecate-nonce-priority-fees.md
MaxResnick: I think this is ok. We should probably build a way for this introspection to happen for transaction v1 tho

PR #582, opened 2026-07-14
SIMD-0582: Early detection of instruction trace overflow

LucasSte:


proposals/0582-instruction-cpi-limit.md
```diff
@@ -0,0 +1,107 @@
...
+feature: (fill in with feature key and github tracking issues once accepted)
+---
+
+## Summary
+
+The protocol only allows a transaction to execute 64 instructions, accounting 
+for both top level and CPIs. Runtime counts executed instructions to determine 
+when to fail a transaction. This SIMD proposes counting the top level 
+instructions (executed or not) plus the executed CPIs. Consequently, 
+transactions that exceed the instruction trace length will fail early.
```
buffalojoec, line 19: I think you could clarify this second half of the summary paragraph by saying "This SIMD proposes _first_ counting _all_ top level instructions and _then_ counting each CPI as the runtime attempts to add each CPI frame."

Something like that would help elucidate the design change.

proposals/0582-instruction-cpi-limit.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Motivation
+
+The primary motivator for this change is ABIv2 
+[SIMD-177](https://github.com/solana-foundation/solana-improvement-documents/pull/177).
```
buffalojoec, line 24: I think speeding up transactions is a more pertinent motivation than a feature we haven't even finished designing yet.

proposals/0582-instruction-cpi-limit.md
```diff
@@ -0,0 +1,107 @@
...
+
+## New Terminology
+
+No new terminology introduced in this document.
+
+## Detailed Design
+
+When pushing the instruction to be processed onto the execution stack, the 
+runtime must verify if the total number of top level instructions plus the 
+number of CPIs does not exceed the maximum allowed instructions in the trace.
```
buffalojoec, line 44: I think you could expand on this paragraph a bit, and say something like "Currently, the total number of instructions (top-level + CPI) is checked against the limit as each instruction is executed. This means that a transaction with 64 top-level instructions, whose _first_ instruction performs a CPI, will still have the 62 out of the remaining 63 instructions executed, before finally exceeding the limit on its 64th."

Something like that. Then after it you can introduce the new top-level-eager approach.

proposals/0582-instruction-cpi-limit.md
```diff
@@ -0,0 +1,107 @@
...
+
+Below is a basic snippet of how this verification is recommended to look like:
+
+```
+if number_of_top_level_instructions_in_tx + number_of_cpis > 64 {
+  return Err(InstructionError::InstructionTraceLengthExceeded);
+}
+```
```
buffalojoec, line 52: Works for me, but just mention earlier that the number of CPIs is tallied each time a CPI is invoked. It can't be tracked ahead of time, and won't be by this SIMD.

proposals/0582-instruction-cpi-limit.md
```diff
@@ -0,0 +1,107 @@
...
+
+It is worth pointing that the number of CPIs aforementioned already accounts 
+for the CPI that is about to be pushed onto the execution stack, if we are not 
+dealing with a top level instruction.
+
+### Edge Cases
+
+No edge case appears prominent enought to listed.
```
buffalojoec, line 60: I would list the 64-instruction edge case that motivated this SIMD, and that I described a bit in my [earlier comment](https://github.com/solana-foundation/solana-improvement-documents/pull/582/changes#r3603309560).

proposals/0582-instruction-cpi-limit.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Alternatives Considered
+
+We have considered revising the address layout for ABIv2 
+[SIMD-177](https://github.com/solana-foundation/solana-improvement-documents/pull/177)
+to acommodate extra spaces for CPIs executed in transactions with 64 top level 
+instructions or similar high numbers. We eliminated such alternative because 
+these transactions would always fail. The address layout acommodation wouldn't 
+benefit developers and would only increase complexity both for runtime and 
+programs.
```
buffalojoec, line 86: nit: this can probably just be `None`. We aren't depending on ABIv2 for this SIMD and the PR is not anywhere close to being accepted and merged. This is also an ABIv1 feature and it has sufficient motivation to stand on its own regardless of any ABIv2 encouragement.

proposals/0582-instruction-cpi-limit.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Impact
+
+There must be no difference to dapp developers, except for the number of CUs 
+consumed by transactions that fail with instruction trace length exceeded.
```
buffalojoec, line 91: Some of the dApp impact is more subtle here. If somebody was firing off a transaction with tons of instructions and failures were happening earlier in the stack, they'd be seeing those instruction errors, and the transaction's execution would exit and return them before the limit was reached. Now, they'd see the limit exceeded before even hitting potentially failure-prone instructions (perhaps midway through their transaction).

Again, it's a subtle case, but it's worth mentioning I think. You could also mention the transaction logs.

buffalojoec (commented):
The change itself makes sense to me! Let's just tighten up the design language and the motivation, then get it in.

This is a win for performance.

mjain-jump (approved):
lgtm!

topointon-jump (approved):

proposals/0582-instruction-cpi-limit.md
LucasSte: Well, it can be None, but changing the ABIv2 is still an alternative.

proposals/0582-instruction-cpi-limit.md
```diff
@@ -14,13 +14,18 @@ feature: (fill in with feature key and github tracking issues once accepted)
...
+instructions (executed or not) plus the executed CPIs, as the runtime attempts 
+to add each instruction frame. Consequently, transactions that exceed the 
+instruction trace length will fail early.
 
 ## Motivation
 
-The primary motivator for this change is ABIv2 
+A motivation would be slightly speeding up transactions that are going 
+to fail regardless. This SIMD proposes failing offending transactions earlier, 
+increasing block space for well formed ones.
```
buffalojoec: Can we make this more definitive? "The primary motivation is".

proposals/0582-instruction-cpi-limit.md
```diff
@@ -29,19 +34,22 @@ two CPIs in a transaction with 63 top level instructions, would cause runtime
...
 
 No new terminology introduced in this document.
 
 ## Detailed Design
 
-When pushing the instruction to be processed onto the execution stack, the 
-runtime must verify if the total number of top level instructions plus the 
-number of CPIs does not exceed the maximum allowed instructions in the trace.
+Currently, the total number of instructions (top-level + CPI) is checked 
+agains the limit as each instruction is executed. Consequently, a transaction
```
buffalojoec: nit

```suggestion
against the limit as each instruction is executed. Consequently, a transaction
```

proposals/0582-instruction-cpi-limit.md
```diff
@@ -29,19 +34,22 @@ two CPIs in a transaction with 63 top level instructions, would cause runtime
...
 
 ## Detailed Design
 
-When pushing the instruction to be processed onto the execution stack, the 
-runtime must verify if the total number of top level instructions plus the 
-number of CPIs does not exceed the maximum allowed instructions in the trace.
+Currently, the total number of instructions (top-level + CPI) is checked 
+agains the limit as each instruction is executed. Consequently, a transaction 
+with 64 top-level instructions, whose first instruction performs a CPI, will 
+still have 62 out of the remaining 63 instruction to be executed before
```
buffalojoec: nit:

```suggestion
still have 62 out of the remaining 63 instructions executed before
```

buffalojoec: Might have to touch up the line width. The idea is to clearly show that we're forcing the validator to execute 63 instructions when, after one CPI, it can already understand the transaction will break the limit, therefore executing them is wasteful.

proposals/0582-instruction-cpi-limit.md
```diff
@@ -53,11 +61,15 @@ if number_of_top_level_instructions_in_tx + number_of_cpis > 64 {
...
-dealing with a top level instruction.
+dealing with a top level instruction. The number of CPIs is tallied whenever a 
+CPI is invoked, and can't be tracked ahead of time.
 
 ### Edge Cases
 
-No edge case appears prominent enought to listed.
+Transactions containing exactly 64 top-level instructions will fail as soon as 
+the first CPI is pushed onto the execution stack, instead of executing all 64 
+instruction budget the protocol provides. Likewise, any transaction with
```
buffalojoec: nit (I think):

```suggestion
instructions in the budget the protocol provides. Likewise, any transaction with
```

proposals/0582-instruction-cpi-limit.md
```diff
@@ -53,11 +61,15 @@ if number_of_top_level_instructions_in_tx + number_of_cpis > 64 {
...
+dealing with a top level instruction. The number of CPIs is tallied whenever a 
+CPI is invoked, and can't be tracked ahead of time.
 
 ### Edge Cases
 
-No edge case appears prominent enought to listed.
+Transactions containing exactly 64 top-level instructions will fail as soon as 
+the first CPI is pushed onto the execution stack, instead of executing all 64 
+instruction budget the protocol provides. Likewise, any transaction with 
+64-X instructions, will fail as soon as X+1 CPIs are pushed onto the stack.
```
buffalojoec: Might just be me, but it wasn't clear to me what the `X` was at first.

```suggestion
`64 - X` instructions, will fail as soon as `X + 1` CPIs are pushed onto the stack.
```

The line itself is good, though.

proposals/0582-instruction-cpi-limit.md
buffalojoec: Yeah, but that would be discussed in the ABIv2 SIMD. It's actually weird to mention it here in this SIMD at all, since this change has pretty solid motivation on its own. I maintain we shouldn't even mention it anywhere in here.

buffalojoec (dismissed):
Looks great, thanks for addressing my feedback!

LucasSte:
@bw-solana and @t-nelson would you mind reading through? This SIMD helps us clear an edge case in runtime.

proposals/0582-instruction-cpi-limit.md
```diff
@@ -0,0 +1,118 @@
...
+None.
+
+## Impact
+
+There is a potential impact on the error order. Transactions that would 
+exceed the instruction trace length, but are failing, for instance, because
+of an error in a CPI before reaching the instruction limit, will now fail 
+without such an error. As a consequence, developers may see different logs 
+in transactions. In addition, CU consumption will change for transactions 
+now halt earlier.
```
bw-solana, line 102: this reads more like a backwards compatibility or security consideration. maybe push to the next section

proposals/0582-instruction-cpi-limit.md
```diff
@@ -0,0 +1,118 @@
...
+
+Validators should see more block space as offending transactions will fail 
+earlier.
+
+## Security Considerations
+
+This proposal suggests a change basic enough to have a low security risk.
```
bw-solana, line 109: would be good to mention the need for error matching between clients to ensure precise CU accounting

proposals/0582-instruction-cpi-limit.md
```diff
@@ -0,0 +1,118 @@
...
+
```
bw-solana, line 110: isn't there usually a "Backwards Compatibility" section? Would be good to add that section back and note that execution can diverge in terms of when failure occurs

proposals/0582-instruction-cpi-limit.md
```diff
@@ -0,0 +1,118 @@
...
+### Validator Components Affected
+
+Which validator components are affected by this change?
+
+| Validator Component             | Impact                              |
+|---------------------------------|-------------------------------------|
+| Transaction Execution (Runtime) | A condition is changing place       |
+| Virtual Machine                 | None                                |
+| Block Packing                   | None                                |
+| Consensus                       | None                                |
```
bw-solana, line 84: Block Packing and Consensus impacted in terms of CU accounting

proposals/0582-instruction-cpi-limit.md
```diff
@@ -0,0 +1,118 @@
...
+In the new design, when pushing the instruction to be processed onto the 
+execution stack, the runtime must verify if the total number of top level 
+instructions plus the number of CPIs does not exceed the maximum allowed 
+instructions in the trace.
+
+Below is a basic snippet of how this verification is recommended to look like:
+
+```
+if number_of_top_level_instructions_in_tx + number_of_cpis > 64 {
+  return Err(InstructionError::InstructionTraceLengthExceeded);
```
bw-solana, line 58: `MaxInstructionTraceLengthExceeded`*

proposals/0582-instruction-cpi-limit.md
```diff
@@ -0,0 +1,118 @@
...
+Currently, the total number of instructions (top-level + CPI) is checked 
+against the limit as each instruction is executed. Consequently, a transaction 
+with 64 top-level instructions, whose first instruction performs a CPI, will 
+still have 62 out of the remaining 63 instructions executed before 
+exceeding the limit.
+
+In the new design, when pushing the instruction to be processed onto the 
+execution stack, the runtime must verify if the total number of top level 
+instructions plus the number of CPIs does not exceed the maximum allowed 
+instructions in the trace.
```
bw-solana, line 52: probably good to be pedantic about CU consumption: 
1. CPI base charge?
2. ix & account data charges
3. callee setup and execution after the push

similarly, specifying exact precedence of errors during push. Presumably CPI prep and re-entrancy restrictions would take precedence. I don't think this matters for consensus tho

topointon-jump (approved):

proposals/0582-instruction-cpi-limit.md
LucasSte: I added a comment about this to the conformance section. Do you think it belongs in here as well?

proposals/0582-instruction-cpi-limit.md
LucasSte: Since you mentioned security considerations [here](https://github.com/solana-foundation/solana-improvement-documents/pull/582#discussion_r3963266766), I wrote that in security considerations.

Let me know how that sounds.

proposals/0582-instruction-cpi-limit.md
LucasSte: I tried to be more explicit about the base CPI charges not changing, and the error precedence.

In the case of push alone, the error precedence won't change from what we have today, and, although I included in the SIMD, it is not documented anywhere besides the code itself, so I'm not sure it fits correctly.

proposals/0582-instruction-cpi-limit.md
```diff
@@ -0,0 +1,134 @@
...
+dealing with a top level instruction. The number of CPIs is tallied whenever a 
+CPI is invoked, and can't be tracked ahead of time.
+
+During the instruction push onto the instruction stack, the error precendence 
+must be as follows:
+
+1. `InstructionError::UnsupportedProgramId`.
+2. `InstructionError::ReentrancyNotAllowed`.
+3. `InstructionError::UnbalancedInstruction`.
+4. `InstructionError::MacInstructionTraceExceeded`.
```
bw-solana: Max*

proposals/0582-instruction-cpi-limit.md
bw-solana: I think what you have now looks good

bw-solana (commented):
one typo then I think I'm good with this

proposals/0582-instruction-cpi-limit.md
LucasSte: I read through everything again, fixed typos and changed some words for better style.

https://github.com/solana-foundation/solana-improvement-documents/pull/582/commits/2f24ec040a5131ba3810b7399969d1314c4a71f4

bw-solana (approved):

PR #585, opened 2026-07-20
Amend 0529

zz-sol:
- In https://github.com/anza-xyz/agave/pull/12321 we slightly changed the order of cu is charged, as per comment https://github.com/anza-xyz/agave/pull/12321#discussion_r3571299123 , as a result it does not match the spec.
- this PR consolidated the changes so the spec and implementation matches


zz-sol:
manually tag @samkim-crypto for review

fdz2zw4gv4-cpu (approved):

PR #586, opened 2026-07-22
SIMD-0511: On-Chain Epoch Stakes

sam0x17:
Store the epoch stakes (the per-vote-account stake mapping) in a dedicated on-chain account per epoch, keyed by epoch number, owned by a new native program that rejects all instructions.

Advanced from the [SIMD discussion](https://github.com/solana-foundation/solana-improvement-documents/discussions/511), where the design converged on epoch-keyed PDAs and the scope was narrowed to epoch stakes only (leader schedule deferred to a future SIMD).

The schema carries the full vote account v4 (SIMD-0185) income state per vote account: both commission collector addresses and both commission rates in basis points, as required for SIMD-0232 compatibility and to give the state flagged in [discussion #583](https://github.com/solana-foundation/solana-improvement-documents/discussions/583) a verifiable on-chain home under the bank hash.

A preliminary Agave implementation is available [here](https://github.com/anza-xyz/agave/compare/master...sam0x17:agave:simd-leader-schedule).


proposals/0511-on-chain-epoch-stakes.md
```diff
@@ -0,0 +1,601 @@
...
+feature: (fill in with feature tracking issues once accepted)
+---
+
+## Summary
+
+Store the epoch stakes (the mapping from vote account to delegated stake) in
+a dedicated on-chain account per epoch, keyed by epoch number. All epochs
+from feature activation forward remain accessible as persistent account
+state, enabling on-chain programs, indexers, and snapshot-consuming nodes
+to read the stake distribution directly without relying on RPC.
```
topointon-jump, line 19: Do we need historical epochs? Can we just store the most recent 3 epochs in a sysvar?

topointon-jump:
I think that this would be much simpler if we just stored the 3 most recent epochs in a sysvar. That's all the information the runtime needs. We don't need to maintain PDAs or have special native programs, and it avoids the state growth problem as well.

sam0x17:
I agree that a bounded sysvar would simplify this proposal.

Since this was opened, the [UpcomingLeader sysvar discussion](https://github.com/solana-foundation/solana-improvement-documents/discussions/558) has converged on putting the full leader schedule in a sysvar. That covers the immediate on-chain leader use cases, and Brennan has said it is ready to move from discussion to a SIMD.

Given that direction, I think we should close this one for now and move #558 forward. If the snapshot-manifest or RPC-removal work still needs epoch stakes independently, we can revisit a bounded version.

PR #593, opened 2026-07-31
simd-0504: remove identical signature requirement

ptaffet-jump:
After studying this further, we realized the requirement to have identical signatures created potential consensus issues.

### DIFF
--- a/proposals/0504-stricter-shred-validation.md
+++ b/proposals/0504-stricter-shred-validation.md
@@ -140,26 +140,20 @@ included explicitly for clarity.
 
 ### Family 2: FEC-level checks
 
-Below, a FEC set is identified by the Merkle root.  In all standard
-cases, that is equivalent to identifying a FEC set by a common
-signature, but a malicious leader with a specifically crafted key may be
-able to produce multiple valid signatures for the same Merkle root after
-[SIMD-0376](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0376-verify-strict.md)
-is activated.
+Below, a FEC set is identified by the Merkle root.
 
-1. All shreds in a FEC set MUST have the same signature
-2. All shreds in a FEC set MUST have the same chained Merkle root
-3. All shreds in a FEC set MUST have the same shred type
-4. All shreds in a FEC set MUST have the same `slot`
-5. All shreds in a FEC set MUST have the same shred version
-6. All shreds in a FEC set MUST have the same `fec_set_index`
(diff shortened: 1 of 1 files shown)


proposals/0504-stricter-shred-validation.md
```diff
@@ -140,26 +140,20 @@ included explicitly for clarity.
...
-
-1. All shreds in a FEC set MUST have the same signature
-2. All shreds in a FEC set MUST have the same chained Merkle root
-3. All shreds in a FEC set MUST have the same shred type
-4. All shreds in a FEC set MUST have the same `slot`
-5. All shreds in a FEC set MUST have the same shred version
-6. All shreds in a FEC set MUST have the same `fec_set_index`
-7. All data shreds in a FEC set MUST have the same `parent_offset`
-8. If any data shred does not have the maximum value of the `size` field as
+Below, a FEC set is identified by the Merkle root.
```
AshwinSekar, line 143: in our implementation a FEC set is identified by shreds with the same `fec_set_index`

e..g if there are 4 shreds with `fec_set_index: 5` and `MR1` and then a 5th shred with `fec_set_index: 5` and `MR2` arrives, the whole block is disqualified (marked dead)

lidatong: yeah i think this is the way for alpenglow + DMR, but for tower + CMR we keyed FECs by merkle root (also what FD is doing today). that predated both this SIMD and alpenglow DMR design, which may have made it possible to key by (slot, fec_set_idx)

for the purposes of Agave implementing this SIMD, i think it works to swap:

`Below, a FEC set is identified by the Merkle root.`

with:

`3. All shreds in a FEC set MUST have the same `slot`
`5. All shreds in a FEC set MUST have the same `fec_set_index`

(ie. `FEC sets are keyed by (slot, fec_set_idx) and must all have the same merkle root)

It does lead to slightly different impls between FD and Agave (FD can maintain multiple FECs for the same slot, fec_set_idx that Agave can't), but i think the convergence effect is the same?

ptaffet-jump: Yeah, I spent some time thinking about this, and I think it is fine. The wording is just for the SIMD and doesn't require Agave to key by Merkle root. The case you're talking about (shreds with the same fec_set_index and different Merkle roots) is outside the scope of this SIMD, but Agave does the right thing there of course.

topointon-jump (approved):

PR #599, opened 2026-08-13
SIMD-0599: Remove inactive stakes from partitioned epoch rewards

joncinque:
#### Problem

Partitioned epoch rewards considers inactive stakes.

#### Proposed solution

Remove them.


topointon-jump (approved):

brooksprumo (approved):
LGTM.

@2501babe can you review as well?

proposals/0599-uncache-inactive-stakes.md
```diff
@@ -0,0 +1,101 @@
...
+type: Core
+status: Idea
+created: 2026-08-10
+feature: RMsTKfD6hZnBhhNvgGBeKNrqCNkeoP3DYYxNtcuWtRg
+---
+
+## Summary
+
+The stakes cache contains all stake accounts, including inactive stakes. The
+inactive stakes must be removed.
```
bw-solana: I think this should focus more on what must go into the rewards calculation/payout.

Stakes cache is client implementation detail (but mentioning it in the motivation section feels appropriate)

2501babe:
tbh i dont have the same misgivings wrt talking explicitly about the stakes cache in the simd, assuming fd found it sensitive enough they had to implement it completely to ensure parity (@topointon-jump can confirm / deny?) the code that is tied to this is very much, patch specific issues with the stakes cache producing a better stakes cache...

i would hope the new rules could be described by the rewards mechanic on actives plus the delegation rewriting on actives plus the alpenglow credits pumping on actives, but i dont know if thats truly the case. in any event it is overdue for a top-down analysis, it has been kind of a no-mans land for years

topointon-jump:
The stakes cache is very much part of consensus unfortunately, as this is data that is not persisted on-chain, but used in consensus-critical operations. So we have to have the same semantics as Agave.

proposals/0599-uncache-inactive-stakes.md
joncinque: No problem, I've added a line specifying what must be included. Otherwise, the table summarizes the situation

joncinque:
I've removed the cache language, and instead focus on just what should / shouldn't be included in epoch rewards.

Happy to also revert that if better

proposals/0599-rm-inactive-stakes-from-per.md
```diff
@@ -0,0 +1,100 @@
...
+feature: RMsTKfD6hZnBhhNvgGBeKNrqCNkeoP3DYYxNtcuWtRg
+---
+
+## Summary
+
+The partitioned epoch rewards process considers all stake accounts, including
+inactive stakes. Inactive stakes do not contribute to stake weight or rewards,
+so must not be included.
+
+Only activating, active, and deactivating stake accounts must be included.
```
bw-solana: Summary seems a bit hard to parse for someone unfamiliar with this issue because it moves from current behavior to proposed behavior pretty abruptly. I would probably wordsmith to something like so:
```
The partitioned epoch rewards process currently requires all stake accounts,
including inactive stakes, to be considered in order to compute the correct
lattice hash and rewards partition count. But inactive stakes do not contribute
to stake weight or rewards, so they shouldn't be required.

This proposal changes rewards to only require activating, active, and deactivating
stake accounts to be included.
```

bw-solana (commented):
One small change requested on the summary.

I much prefer this current version that talks about what rewards requires (inactive stakes to be considered), why they're required (partition count and lattice hash computation, which are consensus critical), and stays away from client implementation detail (the backing cache where these stakes are tracked)

proposals/0599-rm-inactive-stakes-from-per.md
joncinque: Fine by me! I adapted the language a little bit, but I mostly took what you had

bw-solana (approved):

PR #600, opened 2026-08-14
simd-0553 amendment: adjust inclusion fee

cavemanloverboy:
The original inclusion fee parameter was set at 2500 lamports because it was seen as the parameter which would minimize changes to the protocol and validator income. After discussing with validators and looking at traffic on mainnet-beta (approximate average of ≈1.04 signatures per transaction), it was realized that this would result in a minor -4% impact to validator base fee income. Therefore, we amend the SIMD and adjust the inclusion fee to 2600 lamports to offset the impact on the base fee income of validators.


ripatel-fd (approved):

bartenbach:
Ship it

PR #602, opened 2026-08-18
SIMD-0602: Disallow Nonce Account as Program ID

2501babe:
fingers crossed this is the final protocol change we need to cleanly implement ALT relaxation

i considered adding "require nonce accounts be 80 bytes" in here but decided this should stand alone because we _need_ it and we can do qol things separately

### DIFF
--- a/proposals/0602-disallow-nonce-account-as-program-id.md
+++ b/proposals/0602-disallow-nonce-account-as-program-id.md
@@ -99,7 +99,8 @@ writable on the transaction.
 4. The nonce account must exist, be owned by System Program, parse successfully
 as a Current Initialized nonce account, and the stored nonce value must match
 the transaction lifetime specifier.
-5. The nonce account authority must be a signer on the transaction.
+5. The nonce account authority must be a signer on the `AdvanceNonceAccount`
+instruction.
 
 The order of the above steps is non-strict: the failure of any condition results
 in the same outcome, an invalid nonce transaction that must not be committed to
--- a/proposals/0602-disallow-nonce-account-as-program-id.md
+++ b/proposals/0602-disallow-nonce-account-as-program-id.md
@@ -32,20 +32,38 @@ ALT resolution, because the nonce must be marked writable on the transaction,
 and ALT resolution can change the outcome of write-lock demotion by bringing
 LoaderV3 into scope.
 
-The purpose of this SIMD is to sever that relationship fully:
+The purpose of this SIMD is to sever that relationship fully.
 
-* Nonce transaction nonce accounts already cannot come from ALTs.
-* Signers cannot come from ALTs, so we can verify authority without resolving.
+The relevant ruleset is presently as follows:
+
+* Nonce transaction nonce accounts cannot come from ALTs.
+* Signers cannot come from ALTs, so we can verify nonce authority without
+resolving.
 * Program IDs cannot come from ALTs.
-* Presently, a nonce account used as a program ID would be demoted to read-only,
-producing an invalid nonce transaction. However, an ALT with LoaderV3 would
-prevent that demotion, creating a consensus-relevant coupling between nonce
-transaction validation and ALT resolution.
+* Program IDs marked as writable are demoted to read-only, _unless_ LoaderV3 is
+a transaction account, including via an ALT, in which case this demotion is not
+performed.
 
-By banning nonces from being used as program IDs, we can determine whether a 
-transaction is a semantically valid nonce transaction with just the transaction
-bytes and the currently valid blockhashes, and we can fully validate the nonce
-transaction with both plus the nonce account, without needing to resolve ALTs.
+If the nonce account pubkey defined by the `AdvanceNonceAccount` instruction on
+a nonce transaction is also used as a program ID by any instruction on the
+transaction, there are two possible end-states via three conceptual paths:
+
+* LoaderV3 is not provided. In this case, the account is evaluated as read-only,
+and nonce validation fails no matter what the account state is. The transaction
+is discarded.
+* LoaderV3 is provided, but some other step in nonce validation fails, for
+instance if the account is not a valid nonce account, or the nonce authority is
+not a signer, etc. The transaction is discarded identically to the above case.
+* LoaderV3 is provided, _and_ all other nonce validation steps succeed. In this
+case, assuming a valid fee-payer, we will proceed to account loading and fail at
+(if not before) program ID validation, because no valid nonce account can also
+be a valid program. The transaction is committed and charged as fee-only.
+
+The final case is the one this SIMD aims to tackle. By banning nonces from being
+used as program IDs, we can determine whether a transaction is a semantically
+valid nonce transaction with just the transaction bytes and the currently valid
+blockhashes, and we can fully validate the nonce transaction with both plus the
+nonce account, without needing to resolve ALTs.
 
 There is a notional separation between nonce validation that depends only on the
 transaction bytes, versus nonce validation that depends on the stored on-chain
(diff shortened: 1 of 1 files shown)


2501babe:
cc @apfitzge @bw-solana @topointon-jump @mjain-jump

proposals/0602-disallow-nonce-account-as-program-id.md
```diff
@@ -0,0 +1,158 @@
...
+in the same outcome, an invalid nonce transaction that must not be committed to
+chain or else violate the nonce block constraint.
+
+We can very easily check that nonce account is not a program ID during point 3.
+The new condition would likewise result in an invalid nonce transaction. The
+restriction cannot be part of sanitization because whether a transaction is a
+"nonce transaction" depends on the recent blockhashes at the time it is
+processed.
+
+## Alternatives Considered
```
apfitzge, line 123: Another alternative is that we remove lock demotion entirely. I'd had some initial discussions with Tom & Phillip about doing that just as a means to simplify the protocol (it should have never existed).

It seems that change would accomplish the same goal of unblocking 192, but by simplification rather than an additional special-case.

Maybe we revive that discussion and actually prioritize it?

2501babe: 65f238a added this to alternatives considered as well as some remarks on reserved keys. if we remove program id demotion it would also unblock ALT relaxation yes. im not sure how long we would need to discuss it or how complex it would be, one advantage of this proposal is the whole impl is two new boolean checks in `get_durable_nonce()`

the only relevant half of demotion ALT relaxation cares about is program id demotion (really, only the loaderv3 _execption_ to it). i think removing program id demotion is totally fine since it doesnt really accomplish anything. i do worry about the consequences if we remove reserved key demotion though, because it does fully protect what it aims to protect, is relatively more simple, and guards accounts that are used omnipresently (tbh i think we should add token to it)

apfitzge: Yeah I think the token case adds to the argument for moving this out of consensus tbh. Scheduler can just reject, or allow them with the assumption of failure to change such immutable programs.

It being under consensus actively makes it harder to widen the scope to include token in that protective class.

2501babe:
after some discussion we concluded we cant remove write-lock demotion from consensus at this time, but may be able to after alpenglow

2501babe:
i was meditating on what a true gateless workaround to unblock ALT relaxation would look like and i think this would do it:
* write-lock the nonce if requested writable regardless of program id demotion. no consensus change, just for proper dispatch
* use requested writable for runtime nonce validation, a slightly looser constraint. this layer has not been meaningful for consensus since we started single-batching replay
* if nonce is a program id and loaderv3 is not a static key, load the ALT in svm. do not error if invalid, if valid check for loaderv3 in resolved keys to determine true writable nature of nonce
* svm nonce validation, adding a proper writable check. this now behaves in line with current consensus
* svm fee-payer validation, unchanged
* at this point we are free to process transactions as fee-only. load and validate the ALT for real according to all SIMD-0192 rules

(plus all the right magic with reserved keys, but that part is not a problem)

the new workaround design is actually less bad than my old one, which is what made me throw in the towel and open this SIMD. previously i had been imagining more like:
* weaken nonce validation to allow unclear-writable through _past_ the svm nonce checks
* fee-payer validation
* ALT validation and catch the weird nonce case here to discard, mixing our nonce discard and ALT fee-only code

so id still prefer this SIMD to minimize error surface, but i think i can call it a medium-firm blocker instead of a hard blocker on ALT relaxation

topointon-jump:
I'm sure this has been discussed, but what do we think about incentivising people to use txnv1 over the old transaction format?

2501babe:
im definitely for aggressively pushing people toward txv1 by whatever means, but im not sure if there is a timeline for completely removing txv0, which is what we would need to moot ALT relaxation

mjain-jump (approved):

MaxResnick:
> I'm sure this has been discussed, but what do we think about incentivising people to use txnv1 over the old transaction format?

+1 for sure. The hope is we could do something similar to SIMD 571 where we just set the prio fees to 0

2501babe:
after reviewing the options for removing program id demotion, i am more confident this SIMD is the right thing to do to unblock ALT relaxation. in fact, ALT relaxation makes program id demotion easier to remove after the fact, because we wont need to demote when resolving on ingress and then force resanitization across the epoch boundary we activate in

also it would churn across all three repos because demotion is inherent in every transaction type constructor. so we would need to add a feature bool to `LegacyMessage`, `SanitizedMessage`, `LoadedMessage`, `CachedMessage`, and `SanitizedTransaction`, or add a mutable un-demote function to them. since several of these dont compute the property off the message, but cache it internally

on the other hand this SIMD is basically trivial. when we determine the nonce pubkey after failing to find a valid blockhash, we check if the pubkey is a program id, and discard the transaction if so. we dont need to check reserved keys (until SIMD-0297...) because we will necessarily fail the nonce account state checks from any reserved key

removing reserved keys is not possible and unrelated to this SIMD, to ALT relaxation, or to removing program id demotion. it would require a deeper change like restricting what accounts can be credited lamports. actually maybe instead of "these keys are magic for consensus, mungle writablity" we could have a post-execution rule "accounts owned by native / sysvar / loaderv1 / loaderv2 are functionally sol incinerators." then we burn transfers into them and can unconditionally read-lock a magic list of keys with the list being _not_ in consensus. anyway... not related to this

proposals/0602-disallow-nonce-account-as-program-id.md
```diff
@@ -0,0 +1,176 @@
...
+ALT resolution, because the nonce must be marked writable on the transaction,
+and ALT resolution can change the outcome of write-lock demotion by bringing
+LoaderV3 into scope.
+
+The purpose of this SIMD is to sever that relationship fully:
+
+* Nonce transaction nonce accounts already cannot come from ALTs.
+* Signers cannot come from ALTs, so we can verify authority without resolving.
+* Program IDs cannot come from ALTs.
+* Presently, a nonce account used as a program ID would be demoted to read-only,
```
bw-solana: nit: I think we should spell out what "used as a program ID" means. Something like:

"If the nonce-account index, as identified by instruction 0, equals the program ID index of any transaction instruction, the transaction is protocol-invalid"

proposals/0602-disallow-nonce-account-as-program-id.md
```diff
@@ -0,0 +1,176 @@
...
+processed.
+
+This allows us to fully determine whether the nonce account is writable or read-
+only without resolving ALTs as a consequence of the fact that it would never be
+subject to program ID demotion. As an implementation note, we should still check
+that the nonce is not a reserved key or perform demotion on the static
+transaction; we _must_ do this if we decouple static validation from nonce-
+account-dependent validation. Otherwise it is an implementation detail because
+no reserved key account may ever be a valid nonce account.
+
```
bw-solana: I'm trying to think about how we should treat a tx like this:

1. stale blockhash
2. first ix is `AdvanceNonceAccount`
3. static nonce key as a program ID
4. contains unresolvable ALT

My understanding according to SIMD-0192 is that 4 would make this tx fee-only, but this SIMD and 3 would make it invalid. Would be good to specify precedence of checks (or maybe it's in here and I missed it).

proposals/0602-disallow-nonce-account-as-program-id.md
```diff
@@ -0,0 +1,176 @@
...
+the System Program.
+3. The nonce account pubkey must be a static account key and must be marked
+writable on the transaction.
+4. The nonce account must exist, be owned by System Program, parse successfully
+as a Current Initialized nonce account, and the stored nonce value must match
+the transaction lifetime specifier.
+5. The nonce account authority must be a signer on the `AdvanceNonceAccount`
+instruction.
+
+The order of the above steps is non-strict: the failure of any condition results
```
bw-solana, line 123: Might be good to state explicitly that while the order of the above is not strict, they must be checked before other checks such as ALT resolution that would result in fee paying

bw-solana (commented):
couple small things. Looks pretty good to me

proposals/0602-disallow-nonce-account-as-program-id.md
2501babe: 8e70926

proposals/0602-disallow-nonce-account-as-program-id.md
2501babe: 8e70926 i elaborated more on what all is going on, and clarified "is also used as a program ID by any instruction on the transaction." im not sure how bringing account indexes in would make it clearer. if "used" is a problem im not sure if there is a better verb for "it is the program id"... i had to pare back a few attempts at sentences that ended up sounding like "the account that which is designated by the as that which is considered to be therefore art whence thereof" hyper-specific nonsense

bw-solana (approved):

apfitzge (approved):

PR #616, opened 2026-09-01
Amend simd 0376 ed25519-zebra verification

zz-sol:
This PR amends the 0376 signature verification. 

Prior to this PR, simd-0376 uses purely zebra verification as per zip215. It is [reported](https://github.com/anza-xyz/agave/security/advisories/GHSA-28xr-hfwq-22r6) that zip215 deliberately allows for zero-encoded public key (this is not an issue for zcash as they do not use public keys for other purposes) which means, an attacker can forge signatures for any messages for a zero key (i.e. `Pubkey::default()`), widely used as a "disabled authority" sentinel.

We propose to use "reject small-order A" variant from the [Taming the many EdDSAs](https://eprint.iacr.org/2020/1244) paper instead of pure ZIP-215. Speficially

- reject non-canonical encodings of A and R,
- reject non-reduced S,
- reject small-order A and R,
- keep the cofactored equation as in zip-215.

### DIFF
--- a/proposals/0376-verify-strict.md
+++ b/proposals/0376-verify-strict.md
@@ -83,8 +83,7 @@ eight such points, including the identity.
 
 Switch to using the cofactored verification equation described in
 [ZIP-215](https://zips.z.cash/zip-0215) for Ed25519 EdDSA signature
-verification, retaining the encoding and small-order checks performed by
-`verify_strict`.
+verification, while making the encoding and small-order checks explicit.
 
 ### Algorithm:
 
@@ -108,13 +107,17 @@ $\bmod L$ to get scalar $h$.
 8(S \cdot B) - 8R - 8(h \cdot A) = \mathcal{O}
 ```
 
-Steps 1–5 are exactly the checks that RFC-8032 and `verify_strict` require of
-the inputs; the only relaxation introduced by this proposal is the
-multiplication of the verification equation in step 7 by the cofactor. That
-relaxation makes verification insensitive to a torsion component *added* to an
-otherwise valid $A$ or $R$, which is what permits a compatible batched
-verification, while step 5 continues to reject points that consist of nothing
-but torsion.
+Steps 1–5 define the input-validation policy for this proposal; they are not all
+inherited from a single existing verification rule. RFC-8032 requires canonical
+point decoding and a fully reduced $S$, but does not separately reject
+small-order $A$ or $R$. Existing `verify_strict` paths reject a non-reduced $S$
+and small-order $A$ and $R$, while their treatment of non-canonical point
+encodings depends on the `ed25519-dalek` version and call path.
+
+Step 7 is the principal relaxation: multiplying the verification equation by
+the cofactor makes verification insensitive to a torsion component *added* to
+an otherwise valid $A$ or $R$. This permits compatible batched verification,
+while step 5 continues to reject points that consist of nothing but torsion.
 
 Because steps 1 and 2 have already rejected non-canonical encodings, step 5 may
 be implemented either as a cofactor multiplication or as a comparison of the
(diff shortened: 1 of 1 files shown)


zz-sol:
tagging @robre @samkim-crypto @Rexicon226  for review

robre:
lgtm

samkim-crypto (approved):
Looks good to me.

samkim-crypto:
@Rexicon226 @0x0ece can you take a look at this please 🙏

0x0ece (approved):

proposals/0376-verify-strict.md
```diff
@@ -34,10 +46,20 @@ being part of consensus, this proposal suggests using a better proven
 verification equation, which is better described,
 rather than the implicit behaviour found in the `ed25519-dalek` library.
 
-2. `verify_strict` rejects small-order `R`, which makes batch verification
-impossible. Batched verification can reduce costs by ~40% for large
-signatures batches, which is significant at Solana's scale, where validators
-process hundreds of thousands of signatures every block.
+2. `verify_strict` uses the *cofactorless* verification equation
+$`S \cdot B - h \cdot A = R`$. Batch verification is inherently cofactored: it
```
drubin-fd: not true, there can be cofactorless batch verification. they just cannot reproduce the current cofactorless single-verification acceptance set.

drubin-fd:
the idea is fine, i agree with it, thank you for writing it up @zz-sol, i like the test vectors
i'd prefer a bit more cleanup on the whole SIMD, but i can handle that at a later time, no need for it here

proposals/0376-verify-strict.md
zz-sol: Thanks! fixed in 4c4854b59997b62a655dca0c6d5f31dfdae304f5

zz-sol:
hi @drubin-fd does the fix look good to you?

drubin-fd:
Sorry for the delay; Yes!

zz-sol:
hi @t-nelson @steviez @bw-solana can we please get an approval on this PR? thanks!

proposals/0376-verify-strict.md
bw-solana, line 138: I know this isn't in scope, but this caught my eye. Protocol changes for shreds are always tricky because of how receiving/verifying is async w/ respect to execution. Historically, we have applied a 1-epoch delay between activation/application for shred related changes to make sure:

1. every node has a chance to observe the feature execution before making it effective
2. any shreds that try to sneak in for the effective period before feature activation observation would be well beyond the accepted limits (50k slots currently vs. 432k slot epochs, I believe)

as long as we're amending this SIMD, we should spec this out

proposals/0376-verify-strict.md
```diff
@@ -128,21 +195,106 @@ scheme is SUF-CMA secure if an attacker cannot create *any* new valid signature,
...
-be accepted by the new verification equation as well:
+## Backwards Compatibility
+
+This upgrade will require one feature gate. Once this feature gate is active,
+the equation and checks described above will be used for all EdDSA signature
+verifications, instead of `verify_strict`.
+
+- All signatures accepted by `verify_strict` with canonical $A$ and $R$
+encodings remain valid.
+- A small class of signatures previously rejected may now be accepted: those
```
bw-solana, line 276: can we add test vector(s) to cover these?

zz-sol: added in df3ed23c06f9065bac903ae30bc98ba19e6a9dfe

the corresponding tests are here https://github.com/anza-xyz/cryptography/pull/121/changes/6b728f83c6be225e4cb55c1f0f11f3e40164bea7 within https://github.com/anza-xyz/cryptography/pull/121

proposals/0376-verify-strict.md
zz-sol: added d6198cbe0107900bf6effc8ccce8896e234ea802

proposals/0376-verify-strict.md
```diff
@@ -88,29 +142,103 @@ Section 3.2 of [Taming the many EdDSAs](https://eprint.iacr.org/2020/1244.pdf)
...
+Agave currently bounds this at $`\max(500, \text{slots\_per\_epoch} / 2)`$
+slots. For any epoch of at least 1,000 slots, including the 432,000-slot
+epochs of mainnet-beta, testnet, and devnet, this bound is strictly less than
+one epoch. A node close enough to epoch $E + 1$ to accept its shreds therefore
+has a root in epoch $E$ and has already observed the activation. Conversely,
+any shred for epoch $E + 1$ that reaches a node before it has observed the
+activation lies beyond that distance and is dropped regardless of which rule
+would have been applied.
+
+For epochs shorter than 1,000 slots, which Agave supports for test clusters,
```
bw-solana: let's avoid any Agave/client specific details.

I think it's fine to abstract this out to saying something along the lines of "in the event shreds for blocks in epoch E+1 are admitted under old admission rules before root advances to epoch E, expectation is these blocks will never be part of the canonical chain. In all public cluster configurations, shred filtering parameters and epoch durations guarantee this to be true."

proposals/0376-verify-strict.md
```diff
@@ -88,29 +142,103 @@ Section 3.2 of [Taming the many EdDSAs](https://eprint.iacr.org/2020/1244.pdf)
...
+the feature set of that block's bank. Implementations that verify transaction
+signatures ahead of execution, such as a leader's ingest pipeline, must ensure
+that every transaction they include satisfies the rule of the slot in which it
+is included. A transaction verified under the pre-activation rule and still
+buffered when the activation slot is reached must be re-verified under the new
+rule or discarded. Otherwise a leader could include a signature with a
+non-canonical $A$ encoding, which `verify_strict` accepts and this proposal
+rejects, and produce a block that replay rejects.
+
+#### Gossip
```
bw-solana: if we're going to call out all of these applications of signature verification, we should probably make it comprehensive and include repair signatures (and maybe duplicate proof signatures)

proposals/0376-verify-strict.md
```diff
@@ -128,21 +256,186 @@ scheme is SUF-CMA secure if an attacker cannot create *any* new valid signature,
...
+| ------ | --------------- | ----------------------- | ------------- |
+| 0      | accept          | accept                  | accept        |
+| 1      | reject          | reject                  | accept        |
+| 2      | reject          | reject                  | accept        |
+| 3      | reject          | reject                  | accept        |
+
+The `verify_strict` and `verify` columns were confirmed against
+`ed25519-dalek` 2.2.0, and vector 0 matches that library's signer for the same
+seed and message. Vectors 1–3 must be rejected until the feature gate is active
+and accepted afterwards.
```
bw-solana: given the nuance on shred allowance, we should probably wordsmith this to something like "rejected until the new rule is effective for that verification path"

PR #621, opened 2026-09-03
SIMD-0558 - Leader Info Syscall

soundsonacid:
Based on discussion at https://github.com/solana-foundation/solana-improvement-documents/discussions/558


t-nelson:
i see once again no one in the discussions is using their head

ripatel-fd (approved):

ripatel-fd:
> i see once again no one in the discussions is using their head

image

proposals/0558-leader-info-syscall.md
```diff
@@ -0,0 +1,115 @@
+---
+simd: '0558'
```
bw-solana, line 2: we generally have the SIMD match the PR number (not the discussion number), but I'll let it slide

proposals/0558-leader-info-syscall.md
```diff
@@ -0,0 +1,115 @@
...
+swaps coming in on specific leaders (e.g. a leader is malicious, very remote,
+or has historically unreliable scheduling characteristics).
+
+## New Terminology
+
+No new terminology is introduced by this proposal.
+
+## Detailed Design
+
+### Returned Value
```
bw-solana: nit: Either call this `Returned Result` or specify both the results and the value. e.g. returns 0 on success. If any validation condition fails, the syscall aborts VM execution without returning to the calling program

proposals/0558-leader-info-syscall.md
```diff
@@ -0,0 +1,115 @@
...
+}
+```
+
+The `LeaderInfo` struct is written to the `result` formal by the syscall.
+
+The `leader_identity` and `leader_vote` fields
+must be the block producer's identity pubkey and vote account pubkey for the
+current slot. The `next_leader_identity` and `next_leader_vote` fields must be
+the block producer's identity pubkey and vote account pubkey for slot
+`slot + 1`.
```
bw-solana: nit: current slot + 1

proposals/0558-leader-info-syscall.md
```diff
@@ -0,0 +1,115 @@
...
+created: 2026-08-28
+feature:
+development:
+---
+
+## Summary
+
+Create a new syscall that returns the leader for the current & the next slot.
+
+`fn sol_get_leader(result: *mut u8) -> u64`
```
bw-solana, line 19: we should specify what happens if invalid pointer is provided

proposals/0558-leader-info-syscall.md
```diff
@@ -0,0 +1,115 @@
...
+### CU Cost
+
+This syscall copies data into a caller-provided memory address similar to the
+sysvar-specific getter syscalls (`SolGetClockSysvar`,
+`SolGetLastRestartSlotSysvar`, etc.).
+
+We suggest pricing this syscall in line with the cost model used by those
+syscalls (`100 + size_of::<T>() as u64`).
+
+Under this model, `sol_get_leader` would cost `100 + 32 * 4 = 228 CU`.
```
bw-solana: no suggestions. this is the time to be bold, comrade

proposals/0558-leader-info-syscall.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Leader & Vote Pubkeys
+
+For the duration of an epoch, the `leader_identity`, `leader_vote`,
+`next_leader_identity`, and `next_leader_vote` fields must correspond to the
+canonical leader schedule generated for that epoch. This is to address
+mid-epoch `UpdateValidatorIdentity` changes and the fact that multiple vote
+accounts may correspond to the same identity.
+
```
bw-solana, line 87: should consider adding a "corner cases" section to discuss epoch boundary crossing behavior (current in epoch E, next in epoch E+1)

proposals/0558-leader-info-syscall.md
```diff
@@ -0,0 +1,115 @@
...
+## Alternatives Considered
+
+- Add fields to the current `Clock`. This was rejected as many programs assert
+  invariants like `clock.data().len() == 40`, so adding fields would break these
+  programs.
+- Only include the current leader. We opted to include both current and next
+  because knowledge of the next leader enables programs to compensate for, for
```
bw-solana, line 100: given we only look at current slot and next slot, we will be looking at the same leader 75% of the time. Might be good to call out if/why that's okay

proposals/0558-leader-info-syscall.md
```diff
@@ -0,0 +1,115 @@
...
+  example, the next leader having a history of outright censorship. A program
+  can then take actions to ensure that its state remains valid even if it is
+  censored for the entire next leader window.
+- A full `LeaderSchedule` sysvar containing the entire schedule for the current
+  epoch. This was rejected to avoid variable-length sysvars and because the
+  account would be well over 200,000 bytes. Additionally, getting the current
+  slot from `Clock` and indexing by window offset makes the current-leader
+  lookup cost `O(500)` CU instead of `O(150)` CU here.
```
bw-solana: should this be `O(228)`?

proposals/0558-leader-info-syscall.md
```diff
@@ -0,0 +1,137 @@
...
+
+The `LeaderInfo` struct is written to the `result` formal by the syscall.
+
+The `leader_identity` and `leader_vote` fields
+must be the block producer's identity pubkey and vote account pubkey for the
+current slot. The `next_leader_identity` and `next_leader_vote` fields must be
+the block producer's identity pubkey and vote account pubkey for slot
+`current_slot + 1`.
+
+### Returned Result
```
bw-solana: this should be Returned Value and the section above should be Returned Result

proposals/0558-leader-info-syscall.md
```diff
@@ -0,0 +1,137 @@
...
+
+No new terminology is introduced by this proposal.
+
+## Detailed Design
+
+This syscall is intended to be an account-less sysvar accessor
+and therefore use `sol_get_sysvar` internally similar to the
+sysvar-specific getter syscalls (`SolGetClockSysvar`,
+`SolGetLastRestartSlotSysvar`, etc.). Therefore an
+invalid `result` pointer will fail & halt execution.
```
bw-solana: I think we can re-phrase this to avoid getting into implementation details.

Something like

"This syscall is intended to be an account-less sysvar accessor. It validates the complete 128 byte result pointer and returns a result in accordance with the details below."

bw-solana (approved):

deanmlittle:
I think this SIMD is now in a much better shape than the original version, but I don't think it makes any sense to  have merged it while it is proposing a new syscall instead of using the existing sol_get_sysvar. Is there some reason why we thought that was a good idea?

buffalojoec:
> I think this SIMD is now in a much better shape than the original version, but I don't think it makes any sense to have merged it while it is proposing a new syscall instead of using the existing sol_get_sysvar. Is there some reason why we thought that was a good idea?

I don't think we'd want to overload that one. Technically it's designed to read from _sysvars_, and this new syscall states in the motivation it is independent from sysvars. It actually more closely aligns with [`sol_get_epoch_stake`](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0133-syscall-get-epoch-stake.md), which is a runtime callback just like this one, and doesn't associate with sysvars.

That being said I didn't get a chance to review this one before it was merged.

EDIT: after discussing Static Sysvars a bit more, I'm not actually against the `sol_get_sysvar` route if we are committed to including the leader info in Static Sysvars in the near future.

proposals/0558-leader-info-syscall.md
```diff
@@ -0,0 +1,135 @@
...
+swaps coming in on specific leaders (e.g. a leader is malicious, very remote,
+or has historically unreliable scheduling characteristics).
+
+## New Terminology
+
+No new terminology is introduced by this proposal.
+
+## Detailed Design
+
+This syscall is intended to be an account-less sysvar accessor.
```
buffalojoec, line 34: I wouldn't have even mentioned sysvars. This isn't a sysvar. Instead it's a runtime callback that gets plumbed through the VM's calling context and can retrieve information the broader runtime caches.

I wrote a SIMD for a callback just like this in [SIMD-0133](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0133-syscall-get-epoch-stake.md).

proposals/0558-leader-info-syscall.md
```diff
@@ -0,0 +1,135 @@
...
+It validates the complete 128 byte result pointer and returns
+a result in accordance with the details below.
```
buffalojoec, line 36: What does this mean "validates" the pointer? Does it assert it's non-null? Does it assert it's a stack or heap pointer like [SIMD-0459](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0459-syscall-parameter-address-restrictions.md)? Is this SIMD dependent on SIMD-0459, or can validation of said pointer happen before SIMD-0459 is activated?

I'm also assuming this means the syscall will assert `[result, result + 128)` is not read-only? It might seem gratuitous, but these docs are supposed to be specs.

proposals/0558-leader-info-syscall.md
```diff
@@ -0,0 +1,135 @@
...
+}
+```
+
+The `LeaderInfo` struct is written to the `result` formal by the syscall.
+
+The `leader_identity` and `leader_vote` fields
+must be the block producer's identity pubkey and vote account pubkey for the
+current slot. The `next_leader_identity` and `next_leader_vote` fields must be
+the block producer's identity pubkey and vote account pubkey for slot
+`current_slot + 1`.
```
buffalojoec: We probably have to clarify just how accurate these fields are. Since this information is going to come from the already-established leader schedule, we should note that validators may change their addresses during the current epoch - the epoch during which a program may query this information - but this syscall always reflects what's in the already-produced leader schedule.

proposals/0558-leader-info-syscall.md
```diff
@@ -0,0 +1,135 @@
...
+### CU Cost
+
+This syscall copies data into a caller-provided memory address similar to the
+sysvar-specific getter syscalls (`SolGetClockSysvar`,
+`SolGetLastRestartSlotSysvar`, etc.).
+
+We price this syscall in line with the cost model used by those
+syscalls (`100 + size_of::<T>() as u64`).
+
+Under this model, `sol_get_leader` costs `100 + 32 * 4 = 228 CU`.
```
buffalojoec, line 78: nit: I'm assuming this 100 is `sysvar_base_cost`? This isn't a sysvar, so I wouldn't price it like one. I'd use the variable name `syscall_base_cost` (which is also 100) here instead, in case we ever change `syscall_base_cost`.

soundsonacid:
> I think this SIMD is now in a much better shape than the original version, but I don't think it makes any sense to have merged it while it is proposing a new syscall instead of using the existing sol_get_sysvar. Is there some reason why we thought that was a good idea?

Screenshot 2026-09-09 at 1 04 00 AM

proposals/0558-leader-info-syscall.md
buffalojoec: I guess linking back to the discussion I can see why this was framed this way. If we want to make this call available via Static Sysvars, then maybe the framing is okay. I would have liked to hash this out a bit, though.

bw-solana:
@buffalojoec that's my fault. I thought we already had SME review on this one for some reason ☠️ 

let's not let my process fuckup leave us with a suboptimal design. if we just need to tweak things, let's open a PR to amend. if there are big disagreements on the design, let's hash those out in the discussion, abandon this, and re-create

deanmlittle:
> @buffalojoec that's my fault. I thought we already had SME review on this one for some reason ☠️ 
> 
> 
> 
> let's not let my process fuckup leave us with a suboptimal design. if we just need to tweak things, let's open a PR to amend. if there are big disagreements on the design, let's hash those out in the discussion, abandon this, and re-create

No need to abandon. Let's nail it down here and PR later. It's almost there imo. It is my strong preference that we put this in the sysvar cache and use sol_get_sysvar, and as Trent suggested, not assign it an address or an account.

It does not make sense to have a special behavior just for handling this one thing. If we migrate to static sysvars later, not having a special syscall also leaves a smaller longterm footprint on the protocol.

Beyond that, sol_get_sysvar is more efficient as it doesn't necessitate the full 128 byte stack allocation. For most use cases, a single pubkey (or even a truncated pubkey) is sufficient.

soundsonacid:
> > @buffalojoec that's my fault. I thought we already had SME review on this one for some reason ☠️ 
> > 
> > 
> > 
> > let's not let my process fuckup leave us with a suboptimal design. if we just need to tweak things, let's open a PR to amend. if there are big disagreements on the design, let's hash those out in the discussion, abandon this, and re-create
> 
> No need to abandon. Let's nail it down here and PR later. It's almost there imo. It is my strong preference that we put this in the sysvar cache and use sol_get_sysvar, and as Trent suggested, not assign it an address or an account.
> 
> It does not make sense to have a special behavior just for handling this one thing. If we migrate to static sysvars later, not having a special syscall also leaves a smaller longterm footprint on the protocol.
> 
> Beyond that, sol_get_sysvar is more efficient as it doesn't necessitate the full 128 byte stack allocation. For most use cases, a single pubkey (or even a truncated pubkey) is sufficient.

strong agree that we should do sol_get_sysvar. similar pattern to epoch stakes or something that isn't an account or pubkey but uses sysvar cache and sol_get_sysvar for access. i would be a bit sad if this was impl'd as special syscall

buffalojoec:
I'm on board for `sol_get_sysvar` as well. @soundsonacid if you don't mind PR'ing an amendment that would be great.

PR #630, opened 2026-09-08
SIMD-0630: Slot Time Compensation for Alpenglow Fast Leader Handover

qkniep:


umbnat92:
I'm supportive of this SIMD. However there are few things that are puzzling me:

1. I'm a bit worried of hardcoding a parameter that is dependent on the stake distribution and network conditions. This seems to require continuous monitoring to ensure that the chosen value remains representative over time.
2. A mean optimistic prefix time observed on mainnet could be a bit misleading, as it risks transforming a heterogeneous distribution of handover latencies into systematic differences in the effective execution budget across leaders. I understand tho that the purpose is to preserve the expected 200ms slot time. If this is the case and we want to accept possible adverse externalities, I'm fine with it.

qkniep:
1. I agree with the sentiment. However, the correct number for this inherently depends on the stake distribution and latencies. It might be feasible to move part of this calculation into the protocol, e.g. by having all validators register their geo location on chain (which might make this a better fit for a follow-up SIMD we will write anyways about a geography-aware leader schedule).
2. I am not entirely sure I understand your specific point, but generally I would say: The amount of time different leader handovers take and thus the time the new leader has for their first block will vary anyways. The question is just whether the mean is significantly above 200 ms or closer to 200 ms.

umbnat92:
1. Couldn't this be gamed?
2. What I mean is that if the distribution has a very long right tail, there exists a subset of validators with expected value $$\mathcal{E}[t_{start} - t_{ParentReady}] \sim 0 ms$$. For all of them the expected slot time would be $$\mathcal{E}[T_{slot}]=\mathcal{E}[t_{start} - t_{ParentReady}]+200 ms-60 ms=140 ms$$. This clearly makes the validators' effective slot time production worst (i.e. less revenues), forcing them to modify the validator client to adjust this behaviour. Very roughly speaking, if I understand it correctly, this SIMD may implicitly incentivize client modifications.

qkniep:
1. If a large number of validators were lying about their location, they could influence this parameter. The only effect of this would be that the first slot of each leader window becomes a little bit shorter or longer than 200 ms. Also, the geo-aware leader schedule itself gives an incentive to not lie about your location because you will be scheduled worse and pay more time for handovers.
2. This SIMD should be neutral regarding incentives to modify behavior: Maliciously delaying your slot by X ms has exactly the same risks of being skipped as before, as the timeouts are adapted accordingly.

umbnat92:
I completely agree that delaying your slot by X ms has exactly the same risks of being skipped as before. However the reward component has changed, meaning the risk to reward ratio has changed. So, what I'm saying is that to preserve an expected value for slot time due to handoff times, we are introducing new dynamics; and all of this because we want to absorb the long tail of the distribution during leader handoffs

topointon-jump (commented):
Concept ack from us, will review the text one more time before approving.

proposals/0630-flh-slot-time-compensation.md
```diff
@@ -0,0 +1,276 @@
...
+
+This proposal reduces the post-`ParentReady` production budget for the first
+slot of a leader window by a constant `HANDOVER_COMPENSATION` (60 ms) from 200
+ms (after full activation of SIMD-0525) to 140 ms.
+Likewise, the timeout duration after `ParentReady` for the first slot is reduced
+by `HANDOVER_COMPENSATION` and timeouts for other slots shift accordingly.
+As we show below, with `HANDOVER_COMPENSATION` the *effective slot time* roughly
+averages `Δblock` also for the first slot.
+
+Further, in order to reduce idle time, especially in the long tail of long
```
bw-solana, line 30: we should create a separate SIMD for this.

also, Agave already does this, but it would be good to capture it as part of expected interop behavior

proposals/0630-flh-slot-time-compensation.md
```diff
@@ -0,0 +1,276 @@
...
+again be longer than `Δblock` on average, and shorter if it’s above the mean.
+
+## Backwards Compatibility
+
+Block validity is unaffected: blocks produced under either rule are accepted by
+upgraded and non-upgraded validators alike, and no wire format, marker, or
+certificate changes.
+Ledger replay of historical blocks is unaffected.
+
+A feature gate is REQUIRED.
```
bw-solana, line 250: we should get pedantic on the rules during the activation epoch. For example, if we are transitioning from epoch E --> E+1 and E+1 is when this feature goes active, do we apply the new timeouts for any slot created in E+1? Or do we wait for a parent that resides in E+1?

proposals/0630-flh-slot-time-compensation.md
```diff
@@ -0,0 +1,276 @@
...
+- **Effective slot time**: the time interval between the moment a leader starts
+  producing a block and the moment it sends out the last shred. For every slot
+  other than the first of a window this is already supposed to equal `Δblock`.
+  Another validator can only approximately observe compliance with this time
+  interval, but cannot verify it.
+- **Optimistic prefix**: for the first slot of a leader window, the wall-clock
+  interval between the start of block production and the moment `ParentReady(s,
+  ·)` enters `state[s]`.
+
+## Detailed Design
```
bw-solana, line 99: somewhere in here, I think it would be nice to map out a concrete example of when validators will vote skip for a full leader span.

something like this:
|slot|skip vote timeout|
|-|-|
|N|parent_ready + 140ms|
|N+1|parent_ready + 340ms|
|N+2|parent_ready + 540ms|
|N+3|parent_ready + 740ms|

bw-solana (commented):
Concept ACK, but left a few requests

proposals/0630-flh-slot-time-compensation.md
rogerANZA: If this was a separate SIMD, then it should be an earlier SIMD, because the math for timeout bounds depend on having this in place. Since it's only a little thing, we can maybe keep it here?

bw-solana: my personal preference would be to just remove it (it's already de facto in by virtue of being in the Agave code).

but I won't block the SIMD on this point.

PR #634, opened 2026-09-10
SIMD-0558 Amendment: Use sysvar instead of new syscall

soundsonacid:


soundsonacid:
cc @buffalojoec @deanmlittle @bw-solana for review per discussion on https://github.com/solana-foundation/solana-improvement-documents/pull/621

buffalojoec:
I think the general idea we were courting is that it wouldn't be a new syscall that _uses_ `sol_get_sysvar`, but rather the SIMD would just say this data is now available via `sol_get_sysvar`. We'd need to give it address if we do that, though.

soundsonacid:
> I think the general idea we were courting is that it wouldn't be a new syscall that _uses_ `sol_get_sysvar`, but rather the SIMD would just say this data is now available via `sol_get_sysvar`. We'd need to give it address if we do that, though.

hmmm. my understanding was that we were gonna do something similar to like, `SolGetClockSysvar`. if the consensus is that we should do `sol_get_sysvar` entirely that is cool as well but yea we would need address (which is fine if that's what we decide on)

if we want to do `sol_get_sysvar` entirely & address that is fine, but this was basically the original proposal which trent seemed to be strongly against. the only diff was that original proposal had an account

buffalojoec:
> > I think the general idea we were courting is that it wouldn't be a new syscall that _uses_ `sol_get_sysvar`, but rather the SIMD would just say this data is now available via `sol_get_sysvar`. We'd need to give it address if we do that, though.
> 
> hmmm. my understanding was that we were gonna do something similar to like, `SolGetClockSysvar`. if the consensus is that we should do `sol_get_sysvar` entirely that is cool as well but yea we would need address (which is fine if that's what we decide on)
> 
> if we want to do `sol_get_sysvar` entirely & address that is fine, but this was basically the original proposal which trent seemed to be strongly against. the only diff was that original proposal had an account

Ah, I see what you're going for. I was a bit confused since we're conflating implementation details and interfaces here a bit. In Agave, [`SolGetClockSysvar`](https://github.com/anza-xyz/agave/blob/6dd9d38771e46103b9680357a855804165612602/syscalls/src/sysvar.rs#L50) calls into a *helper* [`get_sysvar`](https://github.com/anza-xyz/agave/blob/6dd9d38771e46103b9680357a855804165612602/syscalls/src/sysvar.rs#L7), which is a separate thing (and not used by) [`SyscallGetSysvar`](https://github.com/anza-xyz/agave/blob/6dd9d38771e46103b9680357a855804165612602/syscalls/src/sysvar.rs#L188).

If we want this new data structure to be available via `sol_get_sysvar` (`SyscallGetSysvar`), it will require generating a 32-byte address for the `sysvar_id_addr` interface parameter. `sol_get_sysvar` also charges for that 32-byte input.

I think it might be weird and unnecessary to create an address for this data. If it's going to be available in Static Sysvars, it won't need a 32-byte address; it will just use an 8-byte memory address. If others really want it to be available in `sol_get_sysvar`, perhaps we can do so *additionally*, but we can maintain this standalone syscall to avoid the unnecessary charge for the 32-byte input parameter.

I'd appreciate it if some of my comments ([here](https://github.com/solana-foundation/solana-improvement-documents/pull/621#discussion_r3965022287), [here](https://github.com/solana-foundation/solana-improvement-documents/pull/621#discussion_r3965086312)) about the rest of the proposal could be included here, too!

deanmlittle:
> if the consensus is that we should do `sol_get_sysvar` entirely that is cool as well but yea we would need address (which is fine if that's what we decide on)

This is my vote. I'll let others chime in.

deanmlittle:
> I think it might be weird and unnecessary to create an address for this data. If it's going to be available in Static Sysvars, it won't need a 32-byte address; it will just use an 8-byte memory address. If others really want it to be available in `sol_get_sysvar`, perhaps we can do so _additionally_, but we can maintain this standalone syscall to avoid the unnecessary charge for the 32-byte input parameter.

The thing is, regardless of this, it's still going to be cheaper in 99% of cases to do sol_get_sysvar than it is to roll a custom syscall that copies 128 bytes each time. 

Tbh, the most ideal solution would be to just immediately rip static sysvars, but I would assume that's going to take a little longer to reach consensus, and I don't think it's fair to gate a useful proposal on a bigger feature like that landing first.

soundsonacid:
ok. new proposal has been updated using `sol_get_sysvar`, representing this as an accountless sysvar, no new syscall, and includes an address.

i am in agreement with dean that static sysvars would be ideal but we shouldn't block this small & useful proposal on getting a much larger feature to prod.

if this iteration is something we can all agree upon i think we should send it. let me know if this is copacetic and we can start harassing people for approvals

proposals/0558-leader-info-sysvar.md
```diff
@@ -31,9 +32,13 @@ No new terminology is introduced by this proposal.
 
 ## Detailed Design
 
-This syscall is intended to be an account-less sysvar accessor.
-It validates the complete 128 byte result pointer and returns
-a result in accordance with the details below.
+This sysvar is intended to be accountless and therefore is only
+accessible via `sol_get_sysvar`.
```
buffalojoec: Can you add a link somewhere to https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0127-get-sysvar-syscall.md?

proposals/0558-leader-info-sysvar.md
```diff
@@ -31,9 +32,13 @@ No new terminology is introduced by this proposal.
...
+
+There is no new syscall introduced by this proposal.
+
+The `var_addr` pointer is validated by `sol_get_sysvar`. An
+invalid pointer halts execution without returning to the caller.
```
buffalojoec: nit: this is fine, but you could also just say "control flow is specified in SIMD-0127".

proposals/0558-leader-info-sysvar.md
```diff
@@ -47,35 +52,19 @@ pub struct LeaderInfo {
 }
 ```
 
-The `LeaderInfo` struct is written to the `result` formal by the syscall.
+The `LeaderInfo` struct is written to the `var_addr` formal by `sol_get_sysvar`.
```
buffalojoec: So, you might need to rework some of the design section, since the returned result depends on the offset and length that the caller asked for in `sol_get_sysvar`.

You can probably just define the available layout and call it a day. No need to say what's returned.

proposals/0558-leader-info-sysvar.md
```diff
@@ -47,35 +52,19 @@ pub struct LeaderInfo {
...
-The syscall itself returns `0` on success. If any validation condition fails,
-the syscall aborts VM execution without returning to the calling program.
-
 ### Program Access
 
-Programs may access the `LeaderInfo` information via the `sol_get_leader` syscall.
-
-No sysvar is introduced.
-
-### CU Cost
```
buffalojoec, line 69: nit: can also say something like "CU cost is specified in SIMD-0127".

buffalojoec (commented):
Let's touch it up a bit, but works for me.

> The thing is, regardless of this, it's still going to be cheaper in 99% of cases to do sol_get_sysvar than it is to roll a custom syscall that copies 128 bytes each time.

True, because you can ask for less than the 128 bytes. I still think it's a bit strange that you could ask for one address and still pay for two, but I'm willing to chalk that optimization up to static sysvars.

buffalojoec (dismissed):

proposals/0558-leader-info-sysvar.md
```diff
@@ -14,9 +14,10 @@ development:
 
 ## Summary
 
-Create a new syscall that returns the leader for the current & the next slot.
+Create a new accountless sysvar that returns the leader for the current &
+the next slot.
 
-`fn sol_get_leader(result: *mut u8) -> u64`
+`SysvarLeader1nfo111111111111111111111111111`
```
deanmlittle: Perhaps just SysvarLeader11111111111?

proposals/0558-leader-info-sysvar.md
deanmlittle, line 46: If users would typically ID by both keys, id,vote,id,vote makes sense. If they would typically key by just one, it makes more sense to do id,id,vote,vote.

soundsonacid:
ok. i think we are ready here

cc @topointon-jump @bw-solana @buffalojoec @deanmlittle for final approval?

deanmlittle (dismissed):
lgtm. thanks for all the hard work @soundsonacid getting this out so quickly.

buffalojoec (dismissed):
Ship it

bw-solana (dismissed):

t-nelson:
no. jesus fucking christ will you people stop? sysvars _bad_. sol_get_sysvar ONE OF THE STUPIDEST FUCKING THINGS WE"VE DONE

topointon-jump:
From our side, we're good with leaving the SIMD as it was originally proposed. That said we don't want to block this or bikeshed this to death. I do think that having an "accountless sysvar" is the worst of all worlds - it's not in the bank hash, and the semantics are more complicated than a syscall. I think that either a normal sysvar is fine, or a syscall as originally proposed.

soundsonacid:
@buffalojoec @deanmlittle @topointon-jump @bw-solana this now specifies the use of a regular sysvar. please approve. lol

ripatel-fd:
Just noting that Firedancer will abstain from voting on this. Once Anza decides what to do, we will follow their vote. We already shared our opinions and just want to see the proposed functionality on-chain. It does not matter whether we go the syscall (original SIMD) or sysvar (amendment) route.

soundsonacid:
closed. moving forward with syscall as merged. https://github.com/anza-xyz/agave/pull/15306

PR #607, opened 2026-08-24
Remove floating point arithmetic from Runtime critical inflation and rent path

tao-stones:

PR #642, opened 2026-09-17
Amend SIMD-0607: Derive per-slot decay and increase intermediate precision

apfitzge:
### Motivation
- #607 defined a fixed set of constants associated with slot times and taper fractions that we intend to use on live networkers
	- Not all local clusters or tests use those standard times, and there's no good reason to limit ourselves to a table of permitted cluster settings
- The scaled integer arithmetic uses an easy to reason about scale, but it leads to higher than desired innaccuracies over long periods of time or with low slot times

### Summary of Changes
- Add methodolgy to derive per-slot decay constants given the slot duration in nanoseconds and the taper
- Change scale for intermediate calculations to improve accuracy


proposals/0607-remove-runtime-float-ops.md
```diff
@@ -126,62 +127,186 @@ activation must use legacy calculation.
...
-      if n > 0:
-          base = mul_scaled_floor(base, base)
-```
 
-Because `mul_scaled_floor` floors after each multiplication, clients must use
-this association order exactly.
+`None` denotes invalid inputs or failure of checked arithmetic. Derivation is
+expected to succeed for the protocol's chosen slot durations and tapers; checked
+arithmetic exposes unsupported choices during testing. Clients MUST NOT wrap,
+saturate, clamp inputs, or substitute a floating-point result.
```
topointon-jump, line 247: Nit, but should we specify here what the behaviour should be if the calculation does error out?

topointon-jump (approved):

proposals/0607-remove-runtime-float-ops.md
apfitzge: not a nit, thanks for callout.

There's a few places we can overflow and I think we may want to treat them differently.

#### logarithm approximation:

As currently specified the logarithm approximation could fail - we can actually just fix this by accepting num, den as u32s instead of u128. That doesn't really change the fractions we can reasonably want to represent but just guarnatees that the full calculation, widened to a u128, cannot overflow.

#### per-slot-decay derivation:

When calculating the per-slot-decay we can overflow IF we have an excessively long slot time. With 2^88 scale, this is not THAT long with times on order of seconds...i'll double check how error rates change but we could maybe simplify our scales and just also use 2^64 here which would put the slot-times required for overflow on order of years.

IF we overflow, I think it is reasonable to just return DECAY_SCALE. This in effect means there is no taper or decay at all.

#### Final Reward Calculation

This can overflow with excessively long slots.
This one is less clear on the right choice - I lean toward just having zero inflation rewards as the handling, since any other number is essentially arbitrary or leads to overflow in the distribution of the rewards.

--------

If these make sense to you I'll update the amendment.

PR #643, opened 2026-09-17
SIMD-0643: Move fee burn rounding from per slot to per transaction

apfitzge:


proposals/0643-per-transaction-fee-burn-rounding.md
```diff
@@ -0,0 +1,73 @@
...
+## New Terminology
+
+None.
+
+## Detailed Design
+
+For each transaction charged fees, calculate the following in lamports:
+
+```text
+burned = floor(base_fee / 2)
```
t-nelson, line 37:
```suggestion
burned = trunc(base_fee / 2)
```

same thing in this case, since lamports are unsigned, but more explicitly the desired operation and transparent to the eventual conversion of this calculation to fixed point

proposals/0643-per-transaction-fee-burn-rounding.md
```diff
@@ -0,0 +1,73 @@
...
+This supports future changes that burn priority fees in certain circumstances
+without complicating slot-level fee accounting.
+
+## New Terminology
+
+None.
+
+## Detailed Design
+
+For each transaction charged fees, calculate the following in lamports:
```
t-nelson, line 34: this happens at the time of the fee-payer balance check?

t-nelson (commented):
short, simple footgun removal. easy win

proposals/0643-per-transaction-fee-burn-rounding.md
apfitzge: wdym eventual conversion to fixed point? these are integers

proposals/0643-per-transaction-fee-burn-rounding.md
apfitzge: yes - will clarify in the text

proposals/0643-per-transaction-fee-burn-rounding.md
apfitzge: probably would be more clear if i just state that this is integer devision - can see how floor is confusing

PR #644, opened 2026-09-17
amend SIMD-0558: fix CU cost

ripatel-fd:
sol_get_leader is more expensive than sol_get_sysvar would have been, which does not make sense, considering it does less work


buffalojoec (approved):

topointon-jump (approved):

frank-temporal (approved):

bw-solana:
@ripatel-fd confirming you are okay w/ closing this out in favor of #651

PR #645, opened 2026-09-17
SIMD-0645: SVM JIT intrinsics sol_multi3

clairechingching:
Propose JIT intrinsics to IBRL 128-bit multiplication and open up opportunities for other wide arithmetic and vector operations in the future. This approach has been discussed across multiple channels, including a talk at the NYC Forum.


Lichtso:
I think most of the confusion comes from the overloaded word "syscall". It can be used for both, to descripe the encoding or the dispatch mechanism. A call instruction with a `src=0` encoding in SBPFv3 is what we call a syscall instruction or external call instruction. It typically is implemented as a syscall dispatch and a calling convention switch, which has a lot of overhead. However, all of these are implementation details. Even before this proposal, validator implementations were already free to not do a syscall dispatch and inline codegen instead. Same goes for how you got to the ELF: The explanaition about how you can use intrinsics to achieve this in your toolchain is still off-chain, out-of-protocol. It does not belong in the SIMD either, only the actual protocol change does.

All in all it is still encoded as just another syscall (or "external call" if you dislike the word "syscall" that much) from the protocol perspective and what implementations make out of it is their own thing. Thus, all these implementation details should be stripped out of the SIMD. You can still reference your blog post in the PR description for the full background / context.

proposals/0645-svm-jit-intrinsics.md
```diff
@@ -0,0 +1,250 @@
...
+identify static syscalls. This allows JIT intrinsics to use the existing BPF
+call encoding without introducing new instructions or changing the bytecode
+format.
+
+When loading a program, a static call whose immediate matches a registered JIT
+intrinsic is treated as an intrinsic call. Static calls that do not match a
+registered JIT intrinsic continue to use the existing syscall resolution and
+dispatch behavior.
+
+### Initial Intrinsic: `sol_multi3`
```
Lichtso: Didn't see the CU cost of this syscall mentioned.
You probably want to set it to 1, but it should still be explicit.

proposals/0645-svm-jit-intrinsics.md
```diff
@@ -0,0 +1,250 @@
...
+| Turbine                         | None                                       |
+| Snapshots                       | None                                       |
+| On-Chain Core BPF Programs      | None                                       |
+| Other                           | None                                       |
+
+## Alternatives Considered
+
+- New BPF instructions could represent accelerated operations directly, but
+  would extend the BPF ISA and require changes across the compiler and tooling
+  ecosystem
```
Lichtso, line 207: Another difference between the encoding as an external call instruction or its own opcode is the register binding: When it is its own opcode it can bind any registers, but fewer of them, and as an external call instruction the argument and result operands are fixed. Thus one might need additional mov / load / store instructions before and after the external call instruction to shuffle the registers into position and back.

proposals/0645-svm-jit-intrinsics.md
```diff
@@ -0,0 +1,250 @@
...
+
+When the JIT encounters a registered JIT intrinsic, it emits equivalent native
+host-architecture instructions directly instead of generating the normal
+syscall dispatch sequence.
+
+The initial JIT implementation targets x86-64. It lowers `sol_multi3` to native
+64-bit unsigned `MUL` and `ADD` instructions that compute the low 128 bits of
+the product. The precise native instruction sequence is implementation-defined,
+but it must produce the register results and compute-unit consumption specified
+above.
```
Lichtso, line 164: > compute-unit consumption specified above

Where? Probably obmitted accidentially. I assume you want this to cost 1 CU right?

proposals/0645-svm-jit-intrinsics.md
```diff
@@ -0,0 +1,250 @@
...
+
+## Backwards Compatibility *(Optional)*
+
+Existing programs are unaffected. Existing source code can adopt `sol_multi3`
+by recompiling with a Solana compiler-builtins implementation that lowers
+`__multi3` to the intrinsic.
+
+## Conformance
+
+Conformance tests must verify identical `r0` and `r2` results and compute-unit
```
Lichtso, line 245: Should just say all registers, since we want to make sure that the inputs are not touched either.

proposals/0645-svm-jit-intrinsics.md
```diff
@@ -0,0 +1,250 @@
...
+LLVM lowers 128-bit integer multiplication to the `__multi3` libcall. Compiler
+tooling can implement `__multi3` using the `sol_multi3` JIT intrinsic instead
+of a BPF software implementation. This allows existing programs to benefit
+from the intrinsic simply by recompiling with a compiler-builtins library
+optimized for SVM, without source-level rewrites.
+
+### Prototype results
+
+We prototyped `sol_multi3` in
+[sbpf](https://github.com/anza-xyz/sbpf/commit/4ce37fad6c4773730c4a2445674c9e9b55621b09).
```
Lichtso, line 67: We have an interface to customize codegen here: https://github.com/anza-xyz/sbpf/blob/main/src/program.rs#L386

that way you don't need to tinker with ‎src/interpreter.rs and src/jit.rs and can maybe implement it without any changes in the SBPF repo at all.

proposals/0645-svm-jit-intrinsics.md
```diff
@@ -0,0 +1,250 @@
...
+- **[SIMD-0178]: SBPF Static Syscalls**
+
+  JIT intrinsics reuse the static syscall encoding and hash-based call
+  resolution introduced by SIMD-0178.
+
+[SIMD-0178]: https://github.com/solana-foundation/solana-improvement-documents/pull/178
+
+## New Terminology
+
+**JIT intrinsic:** A protocol-defined operation represented as a static BPF
```
Lichtso, line 88: I still don't see what any of this has to do with the protocol. You might as well introduce the concept of colorful syscalls. Red syscalls and green syscalls and blue syscalls ... But, why?

Lichtso (commented):
I am fine with the proposed solution in itself, just think the proposal should be restructured and a lot shorter.

proposals/0645-svm-jit-intrinsics.md
```diff
@@ -0,0 +1,250 @@
+---
+simd: '0645'
+title: SVM JIT Intrinsics
```
buffalojoec: I think this title should reflect the specific intrinsic you want to add here, right? Something like `SVM JIT Intrinsic: sol_multi3`.

proposals/0645-svm-jit-intrinsics.md
```diff
@@ -0,0 +1,250 @@
...
+of a BPF software implementation. This allows existing programs to benefit
+from the intrinsic simply by recompiling with a compiler-builtins library
+optimized for SVM, without source-level rewrites.
+
+### Prototype results
+
+We prototyped `sol_multi3` in
+[sbpf](https://github.com/anza-xyz/sbpf/commit/4ce37fad6c4773730c4a2445674c9e9b55621b09).
+In a benchmark performing 10,000 `u128` multiplications, the intrinsic method
+consumed approximately four times fewer compute units (110k versus 450k CU) and
```
buffalojoec, line 69: nit: can we say "~75% less" instead of "four times fewer"? I think it's more clear.

proposals/0645-svm-jit-intrinsics.md
```diff
@@ -0,0 +1,250 @@
...
+- Are lowered directly to host-architecture instruction sequences
+- Execute inline at runtime without exiting the JIT
+
+**Host architecture:** The physical instruction set on which an SVM
+implementation executes JIT-compiled code. The initial host architecture
+targeted by this proposal is x86-64.
+
+## Detailed Design
+
+JIT intrinsics use the static call encoding defined by SIMD-0178:
```
buffalojoec, line 102: Is this intrinsic specifically for SBPFv3, or should it apply to all bytecode formats?

proposals/0645-svm-jit-intrinsics.md
buffalojoec, line 207: Don't you still need compiler and tooling changes to support (or stop disallowing) the intrinsic as well?

proposals/0645-svm-jit-intrinsics.md
buffalojoec: The idea is that `murmur32("sol_multi3")` needs to be a recognized symbol by the VM. Right now it would flag it as an unrecognized instruction and throw `EbpfError::UnsupportedInstruction`, bubbling up to `ProgramFailedToComplete`.

https://github.com/anza-xyz/sbpf/blob/1be400928a046729cf55595c033933553cf89e06/src/jit.rs#L814-L839

Here's a test for this: https://github.com/buffalojoec/sbpf/commit/5144ef720aa57310a7d2de4dd399e1ce41411048

We also have to make sure syscall hashes don't collide with supported intrinsics.

buffalojoec (commented):
Sounds like a good idea to me!

proposals/0645-svm-jit-intrinsics.md
Lichtso: That is still the exact same behavior as syscall registration is in general.

> needs to be a recognized symbol by the VM

See https://github.com/solana-foundation/solana-improvement-documents/pull/645#discussion_r4039641153

buffalojoec: Sure but you need to define the interface as well as the backing behavioral ruleset, right? And you need to specify that it's not a host-side syscall implementation. Maybe I'm just missing _which_ part of the proposal you were suggesting isn't protocol.

proposals/0645-svm-jit-intrinsics.md
```diff
@@ -0,0 +1,250 @@
...
+|---------------------------------|--------------------------------------------|
+| Transaction Execution (Runtime) | Intrinsic registration and compute metering |
+| Virtual Machine                 | Interpreter and JIT implementation         |
+| Block Packing                   | None                                       |
+| Consensus                       | None                                       |
+| Gossip                          | None                                       |
+| Turbine                         | None                                       |
+| Snapshots                       | None                                       |
+| On-Chain Core BPF Programs      | None                                       |
+| Other                           | None                                       |
```
ripatel-fd, line 201: this seems a bit redundant to mention

ripatel-fd (commented):
At the protocol level this is just a syscall. "JIT compiler" is an Agave implementation detail. Although this research is all very cool, and I support it, SIMDs are more for protocol changes.

This could just be 50 lines of syscall name, murmur3 hash, math, and the fact that it doesn't charge any CUs. And in the motivation section, I don't think you have to say anything but "it's cheap to implement -> 0 CU" and "useful for compilers" 

(not asking you to implement these changes, just suggestions)

proposals/0645-svm-jit-intrinsics.md
deanmlittle: By the time this lands, it probably won't matter. I'd suggest it's less painful to simply let it work on all versions instead of having a special case.

proposals/0645-svm-jit-intrinsics.md
deanmlittle: Nope. It just overloads CALL_IMM which reuses the existing calling convention for syscalls but is inherently known to the VM, treated as "just another instruction" and resolved before syscall resolution.

proposals/0645-svm-jit-intrinsics.md
clairechingching: done

proposals/0645-svm-jit-intrinsics.md
clairechingching: done

proposals/0645-svm-jit-intrinsics.md
clairechingching: done

proposals/0645-svm-jit-intrinsics.md
clairechingching: done

proposals/0645-svm-jit-intrinsics.md
clairechingching: done

clairechingching:
@Lichtso @buffalojoec @ripatel-fd 

Regarding whether JIT intrinsics belong in the protocol: IMO they do, because this introduces VM-level semantics, not just a JIT implementation detail.

The intent is for the runtime to recognize a JIT intrinsic before normal syscall resolution and treat it more like another sBPF instruction than a syscall. This also directly affects the cost model, which is consensus-critical and therefore part of the protocol.

I don’t think we should define a separate cost model for JIT intrinsics. Conceptually, a JIT intrinsic is just another instruction: although it reuses the external-call encoding, it doesn’t have the dispatch/host-call semantics that motivate syscall costs. Its cost should therefore be treated as an instruction cost and decoupled from the syscall cost model.

ripatel-fd:
@clairechingching Well, Firedancer does not have a JIT. We will treat it as a regular syscall and use tricks to make it fast (syscalls have less overhead in an interpreter than a JIT). The syscall would charge 0 CU and the syscall instruction is charged as usual. 

That would work just fine right? 

Firedancer will have a JIT eventually (maybe next year) and we will make sol_multi3 fast there, too.

clairechingching:
@ripatel-fd yeah that's fine agave needs to handle `sol_multi3` in the interpreter as well

proposals/0645-svm-jit-intrinsics.md
Lichtso: > And you need to specify that it's not a host-side syscall implementation.

Still need the host-side Rust implementation as well, see Claires response:

> Agave needs to handle sol_multi3 in the interpreter as well

> Sure but you need to define the interface as well as the backing behavioral ruleset

The ruleset is the same as syscalls already have today. The only thing this SIMD actually needs to specify is the new syscall `sol_multi3` and its specific behavior.

> Maybe I'm just missing which part of the proposal you were suggesting isn't protocol.

The entire "New Terminology" section should be removed: There is no need to, and we should absoultely not enshrine the concept of JIT compilation and especially not host architectures (like x86) into the protocol. It is **not even possible** to do so as there is no way to prove that you actually JIT compiled on x86 in replay anyway.

`sol_multi3` can be done without any of that and still be implemented with inlined codegen instead of a syscall dispatch to make it fast at execution time.

Lichtso:
> This also directly affects the cost model, which is consensus-critical and therefore part of the protocol ... Its cost should therefore be treated as an instruction cost

Syscalls also cost 1 CU by default, thus it is already compatible.

> The intent is for the runtime to recognize a JIT intrinsic **before normal syscall resolution**

They can be resolved like any other syscall **during** syscall resolution, no need for an extra phase before or after. Also, no need to modify the actual interpreter or JIT compiler either, syscalls support customizing codegen, keeping the abstraction layers intact.

Here, a very simple example to show you what I mean:

```Rust
use solana_sbpf::{jit::{OperandSize, REGISTER_MAP}, x86::X86Instruction};

declare_builtin_function!(
    SyscallSumU64,

    fn rust(
        _context_object: &mut TestContextObject,
        summand_a: u64,
        summand_b: u64,
        _arg3: u64,
        _arg4: u64,
        _arg5: u64,
    ) -> Result<u64, Box<dyn std::error::Error>> {
        Result::Ok(summand_a.wrapping_add(summand_b))
    }

    fn codegen(jit: &mut JitCompiler<TestContextObject>) {
        jit.emit_ins(X86Instruction::mov(OperandSize::S64, REGISTER_MAP[1], REGISTER_MAP[0])); // r0 = r1
        jit.emit_ins(X86Instruction::alu(OperandSize::S64, 0x01, REGISTER_MAP[2], REGISTER_MAP[0], None)); // r0 += r2
    }
);

#[test]
fn test_sum_u64_syscall() {
    test_syscall_asm!(
        "
        mov64 r1, 39
        mov64 r2, 3
        syscall sum_u64
        exit",
        NO_INPUT,
        ("sum_u64" => SyscallSumU64),
        TestContextObject::new(4), // CUs comsumed
        ProgramResult::Ok(42), // Result in R0
    );
}
```

This too, does avoid the syscall dispatch and instead results in fast inlined machinecode. The test runs both the interpreter and JIT compiler and makes sure they agree on everything.

PR #646, opened 2026-09-17
SIMD-0646: Disable legacy and v0 transaction formats

apfitzge:


gwalen:
It would be useful to know how many legacy/v0/v1 transactions land per epoch,
excluding vote transactions. Tracking these numbers over time would show v1
adoption, and could also serve as an activation criterion (e.g. v1 share of
non-vote transactions above some threshold) in addition to the fixed 180-day
date.

Many tutorials, examples and bootcamp materials still use the legacy format.
After activation that code will produce transactions that don't land, and new
developers will hit this at the start of their Solana journey.
Maybe the SIMD could include a migration plan: SDK support (web3.js 1.x, Kit,
Rust SDK), wallet and hardware wallet support, and updates to the docs and
courses SF maintains, with communication well ahead of activation.

Are we able to get some data on what supporting legacy/v0 costs validators,
and which part of the TPU pipeline carries most of it? If we could show the
latency/throughput gains of dropping these paths, it would help justify the
change and convince others.

How will historical data be handled after activation? Could there be nodes
that serve or replay old blocks and need to handle legacy and v0 transactions?
How would we support them?

PR #648, opened 2026-09-21
SIMD-0648: Unbound LoaderV3 Instruction Data

2501babe:
credit to https://github.com/anza-xyz/agave/issues/15356 for the suggestion


proposals/0648-loaderv3-unbounded-instruction-data.md
```diff
@@ -0,0 +1,115 @@
...
+feature: (fill in with feature key and github tracking issues once accepted)
+---
+
+## Summary
+
+LoaderV3 currently bounds instruction deserialization at 1232 bytes, the
+traditional maximum transaction size. This prevents using
+[Transaction V1](https://github.com/solana-foundation/solana-improvement-documents/pull/385)
+when writing to LoaderV3 buffers, which would allow program upgrades to use a
+fraction of the transactions now required.
```
buffalojoec: nit: it doesn't _prevent_ it, it just caps it, right?

proposals/0648-loaderv3-unbounded-instruction-data.md
```diff
@@ -0,0 +1,115 @@
...
+When deserializing LoaderV3 instructions, instead of aborting on instruction
+data larger than 1232 bytes, deserialize all the data.
+
+As an aside, System and Vote both enforce a 1232 byte limit on instruction
+deserialization as well. There is however no reason to change this, since
+neither program has any valid instruction longer than low hundreds of bytes.
+
+### Edge Cases
+
+N/A
```
buffalojoec: Since the second parameter for `Write` is a `Vec<u8>`, we have to be careful that a phony vector length input can't force a huge allocation.

buffalojoec (commented):
LGTM

proposals/0648-loaderv3-unbounded-instruction-data.md
2501babe: right, my wording below about "fully" using txv1 is more accurate, ill update this location

proposals/0648-loaderv3-unbounded-instruction-data.md
2501babe: i tested and it doesnt, but as a side effect of implementation details of bincode. ill add a note, this is easy to handle in code though. we shouldnt need to deserialize the buffer at all... we could just check the length and memcpy directly into the destination

buffalojoec (approved):

PR #649, opened 2026-09-21
SIMD-0649: Priority Ordering Within Entry Batches

MaxResnick:
Turns [Discussion #605](https://github.com/solana-foundation/solana-improvement-documents/discussions/605) into a proposal.

Makes non-increasing priority order within each entry batch a block validity rule.


proposals/0649-priority-ordering-within-entry-batches.md
```diff
@@ -0,0 +1,470 @@
...
+Scheduler experimentation has produced valuable performance work and exposed
+many transaction-pipeline bugs. Performance, however, does not require
+leaving the order of transactions within an entry batch to scheduler
+discretion. A scheduler can still decide what to include and how to form
+batches; it just has to write each batch down in priority order.
+
+Enforcing the order in replay makes the rule hold for every client and every
+scheduler, and the check is cheap: one integer comparison per transaction
+next to the signature verification that replay already performs.
+
```
umbnat92, line 58: I would clearly state here what is the expected outcome of this SIMD.

umbnat92 (commented):
Despite it's clear what this SIMD doesn't want to solve, it's not clear what is its goal.

Starting from system's properties and constraints, this SIMD does not solve any clear problem. However, it is adding extra burden, eventually representing a consensus breaking feature (especially in a multiclient world). Thus it's not clear if what we gain from the model is worth the extra risk.

> Enforcing the order in replay makes the rule hold for every client and every
scheduler

If this is the desiderata, the model can be easily circumvented. Priority is only relevant for conflicting txs when it comes to execution, so a scheduler can order by whatever priority it wants provided that it close the entry batch when it is convenient to violate the order. Thus, the proposed model is not really enforcing any priority order.

Finally, the `2-FEC-set minimum` choice is not researched enough. What is missing (and I'd like to see) is:

1. Current status
2. Division by scheduler and client
3. Dependency from market condition
4. Sensitive analysis on N-FEC-set

PR #650, opened 2026-09-21
SIMD-0650: bn254 pairing output

zz-sol:
this PR introduce a syscall that generates pairing results (the Fq12 element) for bn254 curve. current syscall only does pairing identity check; it does not produce the acutal result if the pairing result is not identity. note that bls12-381 syscall allows for generating pairing results already. so introduce this new syscall seems natural.

the motivation here is for cheaper Groth16 proof verification with solana program. for Groth16 verifier, because we cannot compute pairing result, we need to do 4 pairings and check if the result is identity; whereas if we were to introduce this syscall, the caller can do 3 pairings and compare the output with a pre-computed Fq12 element.


zz-sol:
@samkim-crypto please review

PR #651, opened 2026-09-21
SIMD-0558: Describe pointer validation & update CU cost

soundsonacid:
we would like to amend the cu cost here because "sol_get_leader is more expensive than sol_get_sysvar would have been, which does not make sense, considering it does less work" (richie quote)

we would like to include the pointer validation semantics per @buffalojoec request


soundsonacid:
cc: @buffalojoec @bw-solana @topointon-jump

bw-solana:
for paper trail, can we update the PR description to explain why we need to amend?

buffalojoec:
Is this replacing https://github.com/solana-foundation/solana-improvement-documents/pull/644?

soundsonacid:
> Is this replacing #644?

yes, i spoke with @ripatel-fd and agreed that this would replace 644

buffalojoec:
I don't really care either way, but if @bw-solana stamps #644 you can just merge that and rebase this. It already has mine and Tom's approval. If not, let's close it out.

buffalojoec (dismissed):

topointon-jump (approved):

proposals/0558-leader-info-syscall.md
bw-solana: can update this per the CU formula update above

bw-solana: can update this per the CU formula update above

bw-solana (commented):
couple minor nits

bw-solana (approved):

PR #661, opened 2026-09-22
sBPF SIMDs: clarify SBPF version scope (0166/0173/0174), fix 0166 drawbacks and 0460 stack table

Bruce039:
A few places where the sBPF SIMDs no longer describe which SBPF version a change applies to, plus two internal inconsistencies. Checked against `anza-xyz/sbpf` @ `471370a`.

**Version scope (0166, 0173, 0174)**

0173 and 0174 say their changes apply to "SBPF-version v2 or higher". In sbpf every one of them is gated on `== SBPFVersion::V2`: `disable_lddw`, `disable_le`, `move_memory_instruction_classes`, `callx_uses_src_reg` for 0173, and `enable_pqr`, `explicit_sign_extension_of_results`, `swap_sub_reg_imm_operands`, `disable_neg` for 0174 ([program.rs#L28-L93](https://github.com/anza-xyz/sbpf/blob/471370a1ecc304024f8338224011a65321d494f2/src/program.rs#L28-L93)). SBPFv3 (SIMD-0377) drops all of them. Dynamic stack frames (0166) are `manual_stack_frame_bump = V1 || V2`; v3 goes back to a fixed 4 KiB bump on `call` without gaps, which SIMD-0460 also notes ("this is already the case in SBPFv3").

I added a short note to each instead of rewording the existing "v2 or higher" / `0x01` sentences, because #562 is already editing those lines.

**SIMD-0166 Drawbacks**

The design says the VM keeps a shadow stack and restores the stack pointer on return, "so a function does not need to manually adjust the stack pointer to its original value with a second `add64`". The Drawbacks section then says most functions get "two extra instructions: one to increment the stack pointer and another one to decrement it". sbpf saves `r10` on call and restores it on exit ([interpreter.rs#L143](https://github.com/anza-xyz/sbpf/blob/471370a1ecc304024f8338224011a65321d494f2/src/interpreter.rs#L143), [#L599](https://github.com/anza-xyz/sbpf/blob/471370a1ecc304024f8338224011a65321d494f2/src/interpreter.rs#L599)), so it is one extra instruction.

**SIMD-0460 stack table**

The example has 4 KiB frames with 4 KiB gaps, but the last two rows were `0x200004000..0x200006000` and `0x200005000..0x200007000`, which overlap each other and don't match the 4 KiB step of the rows above. Corrected to `0x200004000..0x200005000` (mapped → unmapped) and `0x200005000..0x200006000` (gap).

Also "functionallity" in 0173. Lint passes, and this merges cleanly with #562, #638 and #652.


proposals/0166-dynamic-stack-frames.md
```diff
@@ -134,6 +134,9 @@ retrieving the caller’s frame pointer address to access those parameters.
 As per the description in SIMD-0161, programs compiled with dynamic stack 
 frames must contain the `0x01` flag on their ELF header `e_flags` field.
 
+Dynamic stack frames apply to SBPFv1 and SBPFv2. SBPFv3 returns to a fixed
+4 KiB frame bump on `call` / `callx`, without stack frame gaps.
```
buffalojoec, line 138: I'd just drop the last sentence. SBPFv3 SIMDs came after this one, and they can specify this info there. If they don't call out the reverting of SBPFv2 features clear enough, they should.

proposals/0173-sbpf-instruction-encoding-improvements.md
```diff
@@ -66,6 +66,10 @@ SBPF-version v2 or higher in its program header (see SIMD-0161). Some now
 unreachable verification and execution checks around `LDDW` can be safely
 removed (see motivation).
 
+These changes are specific to SBPFv2. SBPFv3 (SIMD-0377) does not carry them
+over, and the reference implementation (`anza-xyz/sbpf`) gates each of them on
+`SBPFVersion::V2` exactly.
```
buffalojoec, line 71: Same here.

proposals/0174-sbpf-arithmetics-improvements.md
```diff
@@ -51,6 +51,10 @@ None.
 The following must go into effect if and only if a program indicates the
 SBPF-version (v2) or higher in its program header (see SIMD-0161).
 
+These changes are specific to SBPFv2. SBPFv3 (SIMD-0377) does not carry them
+over, and the reference implementation (`anza-xyz/sbpf`) gates each of them on
+`SBPFVersion::V2` exactly.
```
buffalojoec, line 56: And here.

buffalojoec (approved):

PR #662, opened 2026-09-22
SIMD-0249: fix direction of the lifted commission-increase restriction

0xShadowX:
Fixes the direction of the commission rule in SIMD-0249 (`delay_commission_updates`, active on mainnet).

The `UpdateCommission` section said the program would "no longer restrict commission increases from happening during the **first** half of the epoch". The old rule was the other way round: increases were only allowed up to the epoch midpoint (`relative_slot * 2 <= slots_per_epoch`, [`is_commission_update_allowed`](https://github.com/anza-xyz/agave/blob/25f128d0fcb1735e8b1f31b5f2083fdd28c11df8/programs/vote/src/vote_state/mod.rs#L1002-L1014)), and the gate turns that check off ([vote_processor.rs#L192-L193](https://github.com/anza-xyz/agave/blob/25f128d0fcb1735e8b1f31b5f2083fdd28c11df8/programs/vote/src/vote_processor.rs#L192-L193), [mod.rs#L812-L820](https://github.com/anza-xyz/agave/blob/25f128d0fcb1735e8b1f31b5f2083fdd28c11df8/programs/vote/src/vote_state/mod.rs#L812-L820)). So the restriction being lifted is on the **second** half.

Also removes the leftover template placeholder "Why should we not do this?" under the optional Drawbacks heading.

No behaviour described by the SIMD changes. Lint passes, and there's no conflict with #638 (which only touches the frontmatter).


buffalojoec (commented):
LGTM but I don't know if we should be updating so many proposals in one amendment. I'll defer to @jacobcreech, especially given the fact that this is one of maybe 20 PRs that just came in that look like this.

0xShadowX:
Thanks @buffalojoec, fair point. I narrowed this PR to SIMD-0249 only: the direction of the lifted commission-increase restriction (first half → second half, with the agave references in the description) plus the leftover Drawbacks placeholder in the same file. The text fixes to 0133, 0185, 0291 and 0387 are dropped from this PR.

